PR #10, opened 2015-01-30
Correct spelling of "suppress"

alexhenrie:
"Suppress" is spelled with two P's, even in British English.


jzmaddock:
Merged to develop

PR #15, opened 2015-10-25
Add < and ==, making polynomial an ordered ring.

jeremy-murphy:
Hey, just wondering if there is interest in improving the `polynomial` class?

I have a few things in mind but ordered ring operators were just an easy thing to start off with.

Cheers.

### DIFF
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -14,6 +14,7 @@
 #include <boost/math/tools/rational.hpp>
 #include <boost/math/tools/real_cast.hpp>
 #include <boost/math/special_functions/binomial.hpp>
+#include <boost/operators.hpp>
 
 #include <vector>
 #include <ostream>
@@ -99,7 +100,8 @@ T evaluate_chebyshev(const Seq& a, const T& x)
 }
 
 template <class T>
-class polynomial
+class polynomial 
+    : ordered_ring_operators< polynomial<T> >
 {
 public:
    // typedefs:
@@ -150,6 +152,11 @@ public:
    std::vector<T> chebyshev()const
    {
       return polynomial_to_chebyshev(m_data);
+   }
+   
+   const std::vector<T>& data() const
+   {
+       return m_data;
    }
 
    // operators:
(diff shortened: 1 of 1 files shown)


include/boost/math/tools/polynomial.hpp
```diff
@@ -300,6 +307,18 @@ inline polynomial<T> operator * (const U& a, const polynomial<T>& b)
...
+bool operator == (polynomial<T> const &a, polynomial<T> const &b)
+{
+    return a.data() == b.data();
+}
+
+template <typename T>
+bool operator < (polynomial<T> const &a, polynomial<T> const &b)
+{
+    return a.data() < b.data();
+}
```
jeremy-murphy: Hmmm, hold on, I just realized that maybe polynomial order is not as straightforward as lexicographical order. I will sleep on it for now.

jzmaddock:
I don't think polynomials are really ordered?

jeremy-murphy:
You're right, they don't have a natural total ordering like integers do, it's just a default total ordering for practical convenience. Originally I was just going to add equality, which is necessary for GCD, but then I gave in to the temptation of adding more since it seemed harmless.

jeremy-murphy:
An argument in favour of granting a default total ordering is that it is required, strictly speaking, for a type to be *regular* by the definition in Elements of Programming, and thus interoperable with all the facilities of the STL. (E.g. it permits sort and binary search.)

jeremy-murphy:
However, total ordering is not necessary for GCD, so if it is too contentious I am happy to take it out. Equality is the essential thing.

jzmaddock:
I think making this play nice with std lib containers is a good argument in favour of these.

However,

* If this class is to get a major workover, then I'd rather make all the changes (or at least most of them) at once, and:
* I suspect these operators sort in the "wrong" order (so far as there is one) in that they sort on the x^0 coefficient first and the x^n coefficient last.

jeremy-murphy:
OK, sure, that makes sense. I was just going on the usual assumption that smaller pull requests are easier to review but in this case it's preferable to see the whole impact in one go. 

Ah, you're right, of course. I agree there is no right or wrong order, only a convention. So the correct definition of `operator<` would be to swap the arguments, `return b.data() < a.data()`, right? Anyway, I'm going to leave ordering out for now because it is distracting from the real purpose of GCD.

I'll close this PR and come back soon with a complete one for Euclidean GCD, but I'm inclined to leave Stein GCD to a separate PR. Is that OK?

jzmaddock:
On 31/10/2015 02:58, Jeremy W. Murphy wrote:
>
> OK, sure, that makes sense. I was just going on the usual assumption 
> that smaller pull requests are easier to review but in this case it's 
> preferable to see the whole impact in one go.
>
> Ah, you're right, of course. I agree there is no right or wrong order, 
> only a convention. So the correct definition of |operator<| would be 
> to swap the arguments, |return b.data() < a.data()|, right? Anyway, 
> I'm going to leave ordering out for now because it is distracting from 
> the real purpose of GCD.
>
> I'll close this PR and come back soon with a complete one for 
> Euclidean GCD, but I'm inclined to leave Stein GCD to a separate PR. 
> Is that OK?
>
OK that sounds fine, Best John.

PR #16, opened 2015-10-27
Fixing skewness formula (triangular distribution)

jleni:
I noticed a problem in the skewness of the triangular distribution.
The formula is now correct (very likely it was typo).

You can find more information here:
https://en.wikipedia.org/wiki/Triangular_distribution


jzmaddock:
Manually fixed in develop, many thanks for the report!

PR #17, opened 2015-11-01
Polynomial division

jeremy-murphy:
OK there are a lot of changes but they fall into two categories: 

1. quotient and remainder for polynomials over a field and over a unique factorization domain,
2. changes to existing operators to enforce the representation of zero.

The functionality is complete and there are unit tests.

Division algorithms are from Knuth, The Art of Computer Programming: Volume 2, Third edition, 1998:
specifically, 4.6.1D (division of polynomials over a field) and 4.6.1R (pseudo-division of polynomials).

By necessity, this also adds equality operators `==` and `!=` and for completeness a unary negate operator `-`.

### DIFF
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -108,11 +108,13 @@ public:
 
    // construct:
    polynomial(){}
+   
    template <class U>
    polynomial(const U* data, unsigned order)
       : m_data(data, data + order + 1)
    {
    }
+   
    template <class U>
    polynomial(const U& point)
    {
@@ -132,6 +134,12 @@ public:
       }
    }
 
+   template <typename I>
+   polynomial(I first, I last)
+      : m_data(first, last)
+   {
+   }
+   
    // access:
    size_type size()const { return m_data.size(); }
    size_type degree()const { return m_data.size() - 1; }
@@ -150,6 +158,11 @@ public:
    std::vector<T> chebyshev()const
    {
       return polynomial_to_chebyshev(m_data);
+   }
+   
+   std::vector<T> const& data() const
+   {
+       return m_data;
    }
 
    // operators:
@@ -227,6 +240,42 @@ public:
 private:
    std::vector<T> m_data;
 };
+
+
+/* Calculates a / b and a % b, returning the pair (quotient, remainder) together
+ * because the same amount of computation yields both.
+*/
+template <typename T>
+std::pair< polynomial<T>, polynomial<T> > quotient_remainder(const polynomial<T>& dividend, const polynomial<T>& divisor)
+{
+    if (divisor.degree() == 0 && divisor[0] == T(0))
+        throw std::domain_error("Divide by zero.");
+    std::vector<T> intermediate_result(dividend.data());
+
+    {
+        T const normalizer = divisor[0];
+        for (std::size_t i = 0; i < dividend.degree() - divisor.degree() + 1; i++)
+        {
+            if (intermediate_result[i] != T(0))
+            {
+                T const coefficient = intermediate_result[i] /= normalizer;
+                for (std::size_t j = 1; j < divisor.degree() + 1; j++)
+                {
+                    intermediate_result[i + j] += -divisor[j] * coefficient;
+                }
+            }
+        }
+    }
+    
+    {
+        typedef BOOST_DEDUCED_TYPENAME std::vector<T>::iterator iterator;
+        iterator f = intermediate_result.begin();
+        iterator m = f + dividend.degree() - divisor.degree() + 1;
+        iterator l = m + divisor.degree();
+        return std::make_pair(polynomial<T>(f, m), polynomial<T>(m, l));
+    }
+}
+
 
 template <class T>
 inline polynomial<T> operator + (const polynomial<T>& a, const polynomial<T>& b)
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -11,6 +11,7 @@
 #endif
 
 #include <boost/assert.hpp>
+#include <boost/config.hpp>
 #include <boost/math/tools/rational.hpp>
 #include <boost/math/tools/real_cast.hpp>
 #include <boost/math/special_functions/binomial.hpp>
@@ -109,35 +110,49 @@ template <typename T>
 std::pair< polynomial<T>, polynomial<T> >
 unchecked_synthetic_division(const polynomial<T>& dividend, const polynomial<T>& divisor)
 {
-    BOOST_ASSERT(dividend.degree() >= divisor.degree());
-    BOOST_ASSERT(divisor.degree() != 0 || divisor[0] != T(0));
+    BOOST_ASSERT(divisor.degree() <= dividend.degree());
     
     std::vector<T> intermediate_result(dividend.data());
+    if (divisor.degree() == T(0))
+        intermediate_result.insert(begin(intermediate_result), T(0));
     
     {
-        T const normalizer = divisor[0];
-        for (std::size_t i = 0; i < dividend.degree() - divisor.degree() + 1; i++)
+        typedef BOOST_DEDUCED_TYPENAME std::vector<T>::reverse_iterator reverse_iterator;
+        T const normalizer = divisor[divisor.size() - 1];
+        for (reverse_iterator i = intermediate_result.rbegin(); 
+             i != intermediate_result.rbegin() + dividend.degree() - divisor.degree() + 1; 
+             i++)
         {
-            if (intermediate_result[i] != T(0))
+            if (*i != T(0))
             {
-                T const coefficient = intermediate_result[i] /= normalizer;
+                T const coefficient = *i /= normalizer;
                 for (std::size_t j = 1; j < divisor.degree() + 1; j++)
-                {
-                    intermediate_result[i + j] += -divisor[j] * coefficient;
-                }
+                    *(i + j) += -*(divisor.data().rbegin() + j) * coefficient;
             }
         }
     }
     
     {
         typedef BOOST_DEDUCED_TYPENAME std::vector<T>::iterator iterator;
-        iterator const f = intermediate_result.begin();
-        iterator const m = f + dividend.degree() - divisor.degree() + 1;
-        iterator const l = m + (*m == T(0) ? 1 : divisor.degree());
+        iterator const f = intermediate_result.begin(); // remainder
+        iterator const m = f + std::max(divisor.degree(), 1ul); // quotient
+        iterator const l = intermediate_result.end();
         BOOST_ASSERT(m - f > 0);
         BOOST_ASSERT(l - m > 0);
-        return std::make_pair(polynomial<T>(f, m), polynomial<T>(m, l));
+        polynomial<T> const quotient(m, l);
+        polynomial<T> const remainder = *f != T(0) ? polynomial<T>(f, m) : polynomial<T>();
+        return std::make_pair(quotient, remainder);
     }
+}
+
+
+/**
+ * Returns the zero element for multiplication of polynomials.
+ */
+template <class T>
+polynomial<T> zero_element(std::multiplies< polynomial<T> >)
+{
+    return polynomial<T>();
 }
 
 /* Calculates a / b and a % b, returning the pair (quotient, remainder) together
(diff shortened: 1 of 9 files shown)


include/boost/math/tools/polynomial.hpp
```diff
@@ -134,7 +236,11 @@ class polynomial
 
    // access:
    size_type size()const { return m_data.size(); }
-   size_type degree()const { return m_data.size() - 1; }
+   size_type degree()const
+   { 
+       BOOST_ASSERT(size() != 0);
```
jzmaddock: If the size is zero, then the order is zero (with the order zero coefficient zero), so I think I'd prefer:

if(empty()) return 0;

rather than the assert.

include/boost/math/tools/polynomial.hpp
```diff
@@ -98,8 +111,86 @@ T evaluate_chebyshev(const Seq& a, const T& x)
...
+        reverse_iterator const last = intermediate_result.rbegin() + dividend.degree() - divisor.degree() + 1;
+        for (reverse_iterator i = intermediate_result.rbegin(); i != last; i++)
+        {
+            if (*i != T(0))
+            {
+                T const coefficient = *i /= normalizer;
+                // for (std::size_t j = 1; j < divisor.degree() + 1; j++)
+                    // *(i + j) -= *(divisor.data().rbegin() + j) * coefficient;
+                reverse_iterator const j = i + 1;
+                std::transform(j, j + divisor.degree(), divisor.data().rbegin() + 1, j, _1 - _2 * coefficient);
```
jzmaddock: Given that this will in effect be a new component, lets just go with C++11 here and use native lambda's and ditch the dependency on Boost.Lambda?

jzmaddock:
Other than the suggestions above, looks fine, thanks, John.

include/boost/math/tools/polynomial.hpp
jeremy-murphy: But that's the problem, the degree of the zero polynomial is not zero, it's undefined (http://mathworld.wolfram.com/ZeroPolynomial.html). By convention, some systems return -1 or -∞ but neither of those options are easy here. I thought an undefined value translated well to saying that this function is undefined when `size() == 0`. What about throwing an exception? Either way, I don't think users should expect `degree()` to return a useful value for the zero polynomial.

include/boost/math/tools/polynomial.hpp
jeremy-murphy: Wouldn't that suddenly break the class for users compiling to '03, even if they're not using division? I'm definitely in favour of using newer language standards but I'm not sure if it's worth breaking backward compatibility just over one little lambda.

include/boost/math/tools/polynomial.hpp
jzmaddock: > But that's the problem, the degree of the zero polynomial is not zero, 
> it's undefined (http://mathworld.wolfram.com/ZeroPolynomial.html). By 
> convention, some systems return -1 or -∞ but neither of those options 
> are easy here. I thought an undefined value translated well to saying 
> that this function is undefined when |size() == 0|. What about 
> throwing an exception? Either way, I don't think users should expect 
> |degree()| to return a useful value for the zero polynomial.
>

OK, an exception is better than an assert then - range_error seems to be 
the closest - or maybe just runtime_error?

include/boost/math/tools/polynomial.hpp
jzmaddock: > Wouldn't that suddenly break the class for users compiling to '03, 
> even if they're not using division? I'm definitely in favour of using 
> newer language standards but I'm not sure if it's worth breaking 
> backward compatibility just over one little lambda.
>

I guess I'm easy either way, there shouldn't really be many folks using 
the class at present though.

include/boost/math/tools/polynomial.hpp
jeremy-murphy: On thinking about it further, I would actually recommend against throwing an exception. If a user calls `degree()` on a zero polynomial, it's tantamount to calling `front()` on an empty `vector`: it's simply a logic bug. The user should be checking whether a polynomial is zero and, knowing the mathematics of polynomials, not asking what the degree of the zero polynomial is, just as we don't divide by zero and expect forgiveness. If it were to throw an exception, `logic_error` would be most appropriate, the user now has to check for that exception, which is probably less productive than if they properly checked for the zero polynomial value in the first place.

Now, to counter my own argument: `size()` actually provides exactly the right interface because it's defined for all polynomials including our problematic friend the zero polynomial. We don't really need `degree()`, but we have it, so we can make it a bit inefficient with the reassurance that we can always use `size()` for the most consistent and efficient interface.

So, I don't mind particularly, but I am inclined towards the first option because I think the second option fosters a lack of rigour in thinking about arithmetic. In the end, it's your call. :)

jeremy-murphy: I also realize that I'm being slightly hypocritical, because I throw a `domain_error` in `quotient_remainder` for divide by zero. The same harsh logic above should apply here, but somehow it seemed OK to be kind in that case. Ultimately, I would stick to my guns of "user beware" in favour of efficiency and optionally have a less efficient interface that throws an exception.

jeremy-murphy: A last thought before I go to bed: if we provide the most efficient implementation with the least safeguards, it can easily be wrapped in functions that offer more safeguards. Doing the opposite (removing safeguards for added efficiency) is non-trivial.

jeremy-murphy:
By the way, I did slip unary minus (negate) in here for the sake of completeness and testing arithmetic relations.

include/boost/math/tools/polynomial.hpp
jeremy-murphy: If it's OK with you I'd rather keep it simple and not broach the question of C++1y at the moment because it opens so many questions. If/when we come to proposing a Boost.Polynomial library in its own right then I would definitely be in favour of C++11 as a minimum or base.

jzmaddock: > If it's OK with you I'd rather keep it simple and not broach the 
> question of C++1y at the moment because it opens so many questions. 
> If/when we come to proposing a Boost.Polynomial library in its own 
> right then I would definitely be in favour of C++11 as a minimum or base.
>

That's fine, John.

jeremy-murphy:
OK, I've made some substantial changes to the way division is implemented and I think that's almost all of the work done. So it's probably worth giving the whole thing a once-over again.

It uses `enable_if`/`disable_if` to enable the correct division function depending on whether the coefficients are integer or not. This seemed to be the most natural boundary to separate on until we can really identify what is a *field* versus a *unique factorization domain*.

A couple of arithmetic relations unit tests fail for polynomials over integers. Not necessarily actual bugs, maybe I'm just trying to test something that we can't expect from integer polynomials. Certainly the case of `a * 0.5 == a / 2` is not something I would expect to work. Also, division of *reducible* polynomials seems not to work as expected; I need to look into this a bit more to understand why.

But yeah, apart from these curious unit test failures, I'm hoping that it's 99% done.

jeremy-murphy:
OK, I fixed the failing test for `a * 0.5 == a / 2`, that was a bug from a combination of a) not exposing the heterogeneous division operators and b) the single-argument constructor not being explicit.

Are you OK with that single-argument constructor being explicit now?

include/boost/math/tools/polynomial.hpp
```diff
@@ -98,8 +111,125 @@ T evaluate_chebyshev(const Seq& a, const T& x)
...
+* @tparam  T   Coefficient type, must be an integer.
+* 
+* Template-parameter T actually must be a unique factorization domain but we
+* don't currently have that subtlety of distinction.
+*/
+template <typename T, typename N>
+BOOST_DEDUCED_TYPENAME enable_if_c<std::numeric_limits<T>::is_integer, void >::type
+division_impl(polynomial<T> &q, polynomial<T> &u, const polynomial<T>& v, N n, N k)
+{
+    q[k] = u[n + k] * std::pow(v[n], k);
```
jeremy-murphy: I might find a better algorithm for integer power, as I assume this is not ideal.

jzmaddock:
> Are you OK with that single-argument constructor being explicit now?
>
Seems not unreasonable to me.  John.

include/boost/math/tools/polynomial.hpp
jzmaddock: > I might find a better algorithm for integer power, as I assume this is 
> not ideal.
>

Actually std::pow is generally pretty well optimized for that case - the 
"obvious" divide and conquer algorithm is often not much (or indeed any) 
faster.

jeremy-murphy: OK, thanks, I'll leave it be.

jeremy-murphy:
OK, fixed that bug. John, do you have any ad hoc regression tests you could run using the class to check I didn't break anything along the way?

Otherwise, apart from maybe some more minor documentation updates, it's done.

jeremy-murphy:
John, are there any more changes to the documentation or unit tests that you would like to see?

jzmaddock:
On 23/11/2015 21:53, Jeremy W. Murphy wrote:
>
> John, are there any more changes to the documentation or unit tests 
> that you would like to see?
>

Sorry I haven't had a chance to look yet :(  Will do soon.... honest!

jzmaddock:
This looks OK to me, however with MSVC I see:

```
test_polynomial.cpp(173): error C2666: 'boost::operator /': 2 overloads have similar conversions
M:\data\boost\boost\boost/operators.hpp(252): note: could be 'boost::math::tools::polynomial<double> boost::operator /(T,const U &)' [found using argument-dependent lookup]
        with
        [
            T=boost::math::tools::polynomial<double>,
            U=double
        ]
M:\data\boost\boost\boost/operators.hpp(252): note: or       'boost::math::tools::polynomial<double> boost::operator /(T,const T &)' [found using argument-dependent lookup]
        with
        [
            T=boost::math::tools::polynomial<double>
        ]
M:\data\boost\boost\boost/operators.hpp(252): note: or       'boost::math::tools::polynomial<int> boost::operator /(T,const U &)' [found using argument-dependent lookup]
        with
        [
            T=boost::math::tools::polynomial<int>,
            U=int
        ]
M:\data\boost\boost\boost/operators.hpp(252): note: or       'boost::math::tools::polynomial<int> boost::operator /(T,const T &)' [found using argument-dependent lookup]
        with
        [
            T=boost::math::tools::polynomial<int>
        ]
test_polynomial.cpp(173): note: while trying to match the argument list '(const boost::math::tools::polynomial<double>, int)'
test_polynomial.cpp(167): note: while compiling class template member function 'void test_arithmetic_relations<TestType>::test_method(void)'
        with
        [
            TestType=double
        ]
test_polynomial.cpp(166): note: see reference to function template instantiation 'void test_arithmetic_relations<TestType>::test_method(void)' being compiled
        with
        [
            TestType=double
        ]
test_polynomial.cpp(166): note: see reference to class template instantiation 'test_arithmetic_relations<TestType>' being compiled
        with
        [
            TestType=double
        ]
M:\data\boost\boost\boost/test/tree/test_case_template.hpp(63): note: see reference to function template instantiation 'void test_arithmetic_relations_invoker::run<TestType>(boost::type<TestType> *)' being compiled
        with
        [
            TestType=double
        ]
M:\data\boost\boost\boost/test/tree/test_case_template.hpp(63): note: while compiling class template member function 'void boost::unit_test::ut_detail::test_case_template_invoker<TestCaseTemplate,TestType>::operator ()(void)'
        with
        [
            TestCaseTemplate=test_arithmetic_relations_invoker,
            TestType=double
```

Any ideas?

jeremy-murphy:
From what I understand [here](https://msdn.microsoft.com/en-us/library/dyafzty4.aspx), the compiler can't resolve such ambiguities and I had missed explicitly casting that one integer! I hope the commit I just pushed is all that was needed to fix it, but I'll try to get access to MSVC somehow for testing.

jzmaddock:
On 27/11/2015 22:04, Jeremy W. Murphy wrote:
>
> From what I understand here 
> <https://msdn.microsoft.com/en-us/library/dyafzty4.aspx>, the compiler 
> can't resolve such ambiguities and I had missed explicitly casting 
> that one integer! I hope the commit I just pushed is all that was 
> needed to fix it, but I'll try to get access to MSVC somehow for testing.
>

Thanks, that gets it.

I want to merge current develop to master one more time before release 
(after the beta is out), so I'll defer merging to master until then.  
Please do nag me if I forget!

Thanks for this, John.

PR #21, opened 2016-01-16
Support Boost.Multiprecision types as polynomial coefficients.

jeremy-murphy:
Boost lambda was not automatically deducing the return type for multiprecision arithmetic and so required explicit specification.


jzmaddock:
Thanks!

PR #23, opened 2016-02-09
C++11 initializer_list constructor and assignment for polynomial class

jeremy-murphy:
A constructor and copy assignment operator from `initializer_list`.
Semantics are copied from `vector` in libstdc++.

### DIFF
--- a/doc/internals/polynomial.qbk
+++ b/doc/internals/polynomial.qbk
@@ -25,6 +25,7 @@
       polynomial(I first, I last);
       template <class U>
       explicit polynomial(const U& point);
+      polynomial(std::initializer_list<T> l); // C++11
 
       // access:
       size_type size()const;
--- a/example/polynomial_arithmetic.cpp
+++ b/example/polynomial_arithmetic.cpp
@@ -118,8 +118,8 @@ in descending order of degree, Boost.Math stores in [*ascending order of degree]
   boost::array<double, 4> const d3a = {{10, -6, -4, 3}};
   polynomial<double> const a(d3a.begin(), d3a.end());
 
-  boost::array<double, 2> const d1a = {{-2, 1}};
-  polynomial<double> const b(d1a.begin(), d1a.end());
+  // With C++11 and later, you can also use initializer_list construction.
+  polynomial<double> const b{{-2.0, 1.0}};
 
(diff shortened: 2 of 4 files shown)


jeremy-murphy:
I made a brief example of initializer_list construction, so I think this one is ready to go.

jzmaddock:
Thanks for this!

Can we please change the #if logic to use the appropriate Boost.Config macros?

jzmaddock:
Never mind, I merged it locally and made the changes.
Also changed the call to pow with integer arguments to use a trivial integer-power routine - fixes a build failure on VC10, but also loss of digits when T is a built in type with more digits than double (ie long long).

It's a pretty brain dead power function and there's no error handling - any exponent in double figures will likely cause numeric overflow :(  Note sure what we do about that...

jeremy-murphy:
Ah, sorry, I wasn't aware of those particular Boost macros. Next time.

I'm fairly sure that there are integer power algorithms in Knuth and at least for 32-bit integers in Hacker's Delight, so I should be able to come up with something. Begs the question of where such an algorithm should go in Boost in general.

jzmaddock:
> I'm fairly sure that there are integer power algorithms in Knuth and 
> at least for 32-bit integers in Hacker's Delight, so I should be able 
> to come up with something. Begs the question of where such an 
> algorithm should go in Boost in general.
>

Good question, I guess we haven't provided this before because:

* For floating point bases, use of std::pow is just as quick typically.
* For built-in integer types as bases, overflow occurs so easily, that 
it's questionable whether these are viable general purpose algorithms?

Either way, the version I posted could use having the tail recursion 
removed at the very least...

jeremy-murphy:
Right, yeah, it's a curious problem in that std::pow is satisfactory for most cases, but the fact that it's wrong for some cases does create the need at least for a correct algorithm. I didn't see anything in Knuth specific to integral bases but I only looked quickly. HD has this function, which is more or less an iterative version:
```C
int iexp(int x, unsigned n)
{
    int p = x, y = 1;
    while (true)
    {
        if (n & 1) y *= p;
        n >>= 1;
        if (n == 0) return y;
        p *= p;
    }
}
```
We can of course play with the types.

PR #24, opened 2016-02-09
Performance test for euclidean and binary gcd.

jeremy-murphy:
This currently includes two other pending pull requests, so the file of interest is `test_gcd.cpp`.

At first I was envisioning testing the cross product of different data types (unsigned, polynomial, etc) and functions (gcd_euclidean and gcd_binary), but the polynomial stuff isn't ready and it started to get very complicated. So this just tests the two gcd algorithms on unsigned.

The existing `exec_timed_test` didn't seem to quite fit for gcd so I made a simplified version here.

I don't quite understand what the latter parameters for `report_execution_time` are, so I have just provided some calculated values that vaguely seem to fit.

Do I need to include the resulting `performance_tables.qbk` file or is that generated automatically by some standard system for measuring performance?

### DIFF
--- a/reporting/performance/test_gcd.cpp
+++ b/reporting/performance/test_gcd.cpp
@@ -54,14 +54,13 @@ struct test_function_template
     test_function_template(pair<T, T> const &data) : data(data) {}
     
     template <typename Function>
-    void operator()(Function f) const
+    void operator()(pair<Function, string> const &f) const
     {
-        auto result = exec_timed_test_foo(bind(f, data.first, data.second));
-        cout << f.target_type().name() << ": " << result.first << " (" << result.second << ")" << endl;
+        auto result = exec_timed_test_foo(bind(f.first, data.first, data.second));
         report_execution_time(result.first, 
                             string("gcd method comparison with ") + compiler_name() + string(" on ") + platform_name(), 
-                            typeid(decltype(data.first)).name(), 
-                            typeid(f).name());
+                            f.second, 
+                            boost_name());
     }
 };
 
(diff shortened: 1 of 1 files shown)


jzmaddock:
I'll defer this until we have #23 merged.

performance_tables.qbk is regenerated when all the performance tests are run, so no need to include it in the commit just yet.  I'll double check the args to report_execution_time and the tables generated later.

Thanks for this!

pabristow:
Are there any changes to documentation from these changes?
Would users benefit from an example of use of initilizer list ?  Or just a reference to the test?

pabristow:
Sorry - fat thumb - didn't mean to close.

jeremy-murphy:
I think initializer_list usage is quite intuitive, so I'll put in something minimal into the documentation that I otherwise overlooked. :)

jeremy-murphy:
What kind of precision do you expect from a performance test? I see a 10-15% fluctuation in measurement, which does not inspire me with confidence. I guess should use a range of input values, too.

jzmaddock:
On 10/02/2016 01:37, Jeremy W. Murphy wrote:
>
> What kind of precision do you expect from a performance test? I see a 
> 10-15% fluctuation in measurement, which does not inspire me with 
> confidence. I guess should use a range of input values, too.
>

That doesn't surprise me.  However, note that:

* Compilers are often way too clever here - executing the same statement 
over and over often gets optimised away to just one call.
* Any result that measures a short duration typically has a lot of noise 
- you seem to require about a second of execution time to get consistent 
results.

John.

jeremy-murphy:
OK, I'm back from a trip overseas, so I hope to finish this up soon.

jeremy-murphy:
This might be ready to go now: any thoughts?

PR #25, opened 2016-03-02
Gcd unit tests

jeremy-murphy:
A few basic gcd unit tests.

### DIFF
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -958,6 +958,7 @@ run test_signed_zero.cpp ../../test/build//boost_unit_test_framework ;
 
 run complex_test.cpp ../../test/build//boost_unit_test_framework ;
 run test_polynomial.cpp ../../test/build//boost_unit_test_framework ;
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
Any thoughts on this PR? I think it is fine as is although minimal; I could add some more variety to the values tested just to increase confidence in correctness.

pabristow:
You have never have too many tests ;-)

 

(and examples of use).

 

Paul

jeremy-murphy:
Well, I think that's mostly true because most of the time people don't write enough tests.

But in the case of algorithms as simple as Euclidean gcd, I struggle to think of new examples that actually test something novel. Admittedly, 'binary' gcd is more complex, but there just aren't many code paths through Euclidean gcd, right? (And I intend to make it even simpler.)

I really value good tests, but I don't want to add more just for the sake of it (slightly contrary to what I said above). I'll keep thinking about it for a few days and I welcome suggestions.

jzmaddock:
Looks good to me, have you checked code coverage?

PR #26, opened 2016-03-02
Polynomial Stein gcd

jeremy-murphy:
This is the basic implementation required for using the "binary" gcd algorithm on a polynomial.

### DIFF
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -408,6 +408,14 @@ public:
        *this = quotient_remainder(*this, value).second;
        return *this;
    }
+   
+   template <typename U>
+   polynomial& operator >>=(U const &value)
+   {
+       BOOST_ASSERT(value <= m_data.size());
+       m_data.erase(m_data.begin(), m_data.begin() + value);
+       return *this;
+   }
 
    /** Remove zero coefficients 'from the top', that is for which there are no
     *        non-zero coefficients of higher degree. */
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -229,11 +229,10 @@ quotient_remainder(const polynomial<T>& dividend, const polynomial<T>& divisor)
 
 template <class T>
 class polynomial :
-    equality_comparable< polynomial<T>,
-    dividable< polynomial<T>,
-    dividable2< polynomial<T>, T,
-    modable< polynomial<T>,
-    modable2< polynomial<T>, T > > > > >
+    ordered_euclidean_ring_operators< polynomial<T>,
+    dividable< polynomial<T>, T,
+    modable< polynomial<T>, T,
+    right_shiftable< polynomial<T> > > > >
 {
 public:
    // typedefs:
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
Any thoughts here? I realize using `operator <` this way might be controversial.

jzmaddock:
If we're going to have an operator <, then if the limb counts are the same, we should compare limb-by-limb I guess, with higher order limbs taking priority.

Is there precedent for polynomial ordering?

jeremy-murphy:
Unfortunately, 99% of web results for polynomial ordering are about ordering the terms within a single expression and polynomial sorting is, well, about polynomial algorithms for sorting.  :)  I looked into this topic last year and my conclusion was that there was no precedent.

I believe `<` by itself is OK, it satisfies the concept of Strict Weak Ordering, my main concern is the interaction with `==`: elements that are equivalent according to the `<` relation are not necessarily equal according to `==`. Whereas they would be if `<` was a Total Ordering (I think -- I still get the strict/weak/partial/total terminology a bit mixed up). But is that terrible? My understanding is that the STL components only require a Strict Weak Ordering.

An argument in favour of this definition of `<` is that it is the most efficient for Stein gcd: a total ordering would also work but increases the complexity of ranking from constant to linear.

jeremy-murphy:
Gah, I accidentally called the branch "Steiner" when I meant "Stein", how embarrassing.

jzmaddock:
On 10/03/2016 01:58, Jeremy W. Murphy wrote:
>
> Gah, I accidentally called the branch "Steiner" when I meant "Stein", 
> how embarrassing.
>

I suspect he'll live with it ;)

jzmaddock:
On 10/03/2016 01:55, Jeremy W. Murphy wrote:
>
> Unfortunately, 99% of web results for polynomial ordering are about 
> ordering the terms within a single expression and polynomial sorting 
> is, well, about polynomial algorithms for sorting. :) I looked into 
> this topic last year and my conclusion was that there was no precedent.
>
> I believe |<| by itself is OK, it satisfies the concept of Strict Weak 
> Ordering, my main concern is the interaction with |==|: elements that 
> are equivalent according to the |<| relation are not necessarily equal 
> according to |==|. Whereas they would be if |<| was a Total Ordering 
> (I think -- I still get the strict/weak/partial/total terminology a 
> bit mixed up). But is that terrible? My understanding is that the STL 
> components only require a Strict Weak Ordering.
>
> An argument in favour of this definition of |<| is that it is the most 
> efficient for Stein gcd: a total ordering would also work but 
> increases the complexity of ranking from constant to linear.
>

Understood, I'm just uncomfortable with this definition as is.

Can we either pass a functor to gcd for operator <, or else use a traits 
class for that operation within gcd, and then we could leave operator< 
undefined for polynomials?

pabristow:
I trust you guys are saving this erudite discussion to record in the Boost.Math documentation ;-)

 

But I’ll do nothing until it has come to a conclusion…

 

Paul

jeremy-murphy:
I think a traits class sounds like a good idea. It may well be too much to expect that a single `operator<` will optimally perform all the tasks expected of it for a polynomial. In the end, given that equality is defined by `a.data() == b.data()`, it is probably consistent for less than to be `a.data() < b.data()`.

What about right shift? I guess we could have a complementary left shift that increases the order of all the terms and adds a zero constant to the end. That would make the behaviour equivalent to integers, I think.

jeremy-murphy:
I'll close this PR since I think the final result will look somewhat different.

PR #28, opened 2016-03-28
Polynomial: shift operators and odd/even predicates

jeremy-murphy:
Shifting bits is a natural enough operation on a binary digital computer. In the binary representation of integers, a right shift removes a factor of 2, or halves. A more abstract way of thinking about this is that it removes the smallest prime factor, which allows for translation of the shift operator into other domains such as polynomials. The smallest prime factor of a polynomial is _x_, so removal equates to decreasing the degree of all terms and thus effectively dropping the constant term. Thus, we can think of polynomials as being odd or even based on the value of the constant term: odd if it is non-zero and even otherwise. Each right shift reduces the degree of the polynomial, eventually reducing to the zero polynomial.

This PR adds left-shift (`<<=` and `<<`) and right-shift (`>>=` and `>>`) operators and `odd()`/`even()` predicates for the `polynomial` class. 

As well as being useful for general abstract algebraic and number theoretical fun, they are specifically useful for application of the Stein gcd algorithm.


jeremy-murphy:
I'll add documentation/examples once we agree on the interface and implementation.

jzmaddock:
That all looks reasonable to me.

jzmaddock:
Pushed to branch gcd.

jeremy-murphy:
Apart from some polishing of the example, I think this PR is about done. Any more thoughts on it?

jzmaddock:
It's fine, I just want to leave develop stable until after the release, hence merging to the gcd branch for now.

test/test_polynomial.cpp
```diff
@@ -227,20 +235,50 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_non_integral_arithmetic_relations, T, non_int
...
+    BOOST_CHECK_EQUAL(a, c);
+    polynomial<T> zero = zero_element(multiplies< polynomial<T> >());
+    // Multiplying zero by x should still be zero.
+    zero <<= 1u;
+    BOOST_CHECK_EQUAL(zero, zero_element(multiplies< polynomial<T> >()));
+}
 
-    a *= a;
 
-    BOOST_CHECK_EQUAL(a, b*b*b*b);
```
jeremy-murphy: I merged the test for self-multiplication into the general test of multiplication.

jzmaddock:
All merged to develop now - thanks!

PR #29, opened 2016-03-28
gcd: example of gcd_binary for polynomial (WIP)

jeremy-murphy:
**Don't merge this PR, it's just for demonstration.**

This includes the PR for polynomial shift, so if you ignore those changes and concentrate on `gcd_Stein` (nee binary) you'll see that I have abstracted out the concept of odd/even and the concept of less-than wrt to Stein gcd.

One could also make the shift operator more abstract, but I don't see the need at the moment.


jeremy-murphy:
The interesting parts of this PR are now merged via other PRs.

PR #30, opened 2016-03-30
gcd: Alternative implementation of Stein gcd.

jeremy-murphy:
@jzmaddock, this is an alternative implementation of Stein gcd for performance testing.

I don't believe it will actually make a difference, but it is worth ruling out.


jeremy-murphy:
This is now in the revamp branch so I'll close this one.

PR #31, opened 2016-04-04
gcd: revamp

jeremy-murphy:
This is a relatively quick rewrite of the gcd algorithms using code from From Mathematics to Generic Programming (more because it was conveniently on hand).

### DIFF
--- a/include/boost/math/common_factor_rt.hpp
+++ b/include/boost/math/common_factor_rt.hpp
@@ -9,10 +9,11 @@
 
 #include <boost/assert.hpp>
 #include <boost/core/enable_if.hpp>
-#include <boost/type_traits.hpp>
 #include <boost/mpl/and.hpp>
+#include <boost/type_traits.hpp>
 
 #include <algorithm>
+#include <limits>
 
 namespace boost {
 namespace math{
@@ -93,7 +94,7 @@ namespace detail
 
 
 template <typename T>
-BOOST_DEDUCED_TYPENAME enable_if_c<mpl::and_< has_right_shift_assign<T>, has_less<T> >::value, T>::type
+BOOST_DEDUCED_TYPENAME enable_if_c<mpl::and_< mpl::and_< has_right_shift_assign<T>, has_left_shift_assign<T> >, has_less<T> >::value, T>::type
 optimal_gcd(T const &a, T const &b)
 {
     return detail::Stein_gcd(a, b);
@@ -101,7 +102,7 @@ optimal_gcd(T const &a, T const &b)
 
 
 template <typename T>
-BOOST_DEDUCED_TYPENAME disable_if_c<mpl::and_< has_right_shift_assign<T>, has_less<T> >::value, T>::type
+BOOST_DEDUCED_TYPENAME disable_if_c<mpl::and_< mpl::and_< has_right_shift_assign<T>, has_left_shift_assign<T> >, has_less<T> >::value, T>::type
 optimal_gcd(T const &a, T const &b)
 {
     return detail::Euclid_gcd(a, b);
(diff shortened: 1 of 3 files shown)


jzmaddock:
Pushed to branch gcd, along with many other changes required to get the tests passing.

jzmaddock:
Performance test results for various compilers start here: https://htmlpreview.github.io/?https://github.com/boostorg/math/blob/gcd/reporting/performance/html/index.html#special_function_and_distributio.section_gcd_method_comparison_with_Clang_version_3_8_0_trunk_256686_on_linux

jeremy-murphy:
Hi John,

thanks for making such an extensive comparison! I regret that I have not
got into the habit of generating such results myself; is it fairly
straightforward? I imagine the simple part is running the performance
executable, but generating the HTML requires the usual docbook magic?

It's curious that the Stein algorithm from boost is better than the
textbook variant for multiprecision numbers, but for built-in numbers it's
the other way around. But it reminds me to ask, is it actually worth
measuring the generic gcd implementations on multiprecision types? I know I
expressed curiosity about the result, but I seem to recall from your
earlier comparison that the generic implementations were terrible by
comparison to the specific ones?

I'm not sure exactly which implementation is in use in each case: does
"boost" mean the extant implementation by Daryle Walker and Paul Moore? And
which textbook: TAOCP? What about the Stepanov implementation?

Looks like the mixed Euclidean/Stein algorithm is a clear winner except in
extreme cases. We could use that by default safely, still allowing for the
rare possibility of numeric types that are only in one but not the other
domain. Also, I'm thinking we should probably make a public interface for
the named algorithms so that users can call them explicitly. That is, "gcd"
would do the best thing for a numeric type, but a user could call
"Euclid_gcd" for example if that's what they explicitly want. What do you
think?

Cheers.

Jeremy

jzmaddock:
> thanks for making such an extensive comparison! I regret that I have not
> got into the habit of generating such results myself; is it fairly
> straightforward? I imagine the simple part is running the performance
> executable, but generating the HTML requires the usual docbook magic?
>
Once you have you're user-config.jam set up to build documentation (see 
https://svn.boost.org/trac/boost/wiki/BoostDocs/GettingStarted) then you 
just cd into libs/math/reporting/performance and invoke bjam.  That will 
run all the tests and build the single page summary that I linked to.  
Likewise in the reporting/accuracy directory.

> It's curious that the Stein algorithm from boost is better than the
> textbook variant for multiprecision numbers, but for built-in numbers it's
> the other way around. 
Other way around I think - but that's because the "textbook" algorithm 
has been messed around with (see below).

> But it reminds me to ask, is it actually worth
> measuring the generic gcd implementations on multiprecision types? I 
> know I
> expressed curiosity about the result, but I seem to recall from your
> earlier comparison that the generic implementations were terrible by
> comparison to the specific ones?

Not much point other than curiosity, multiprecision types need special 
handling in general.
>
> I'm not sure exactly which implementation is in use in each case: does
> "boost" mean the extant implementation by Daryle Walker and Paul 
> Moore? And
> which textbook: TAOCP? What about the Stepanov implementation?
>
Boost is your re-written code.

Textbook is basically some code I threw together in order to have 
something to experiment with, it's pretty much the "wikipedia" version 
since I couldn't be bothered to walk upstairs to fetch Knuth ;)

It's been messed about with though, for example whenever you see 
something like:

while(x&1) x >>=1;

which is common to all the binary algorithm variants, it's more 
efficient on most platforms to write:

x >>= lsb(x);

where lsb(x) finds the least-significant bit of x.  On many platforms 
this is a compiler intrinsic so this reduces to 2 opcodes if we ignore 
possible load and stores.  It's also hugely more efficient for 
multiprecision types.

> Looks like the mixed Euclidean/Stein algorithm is a clear winner except in
> extreme cases. We could use that by default safely, still allowing for the
> rare possibility of numeric types that are only in one but not the other
> domain. Also, I'm thinking we should probably make a public interface for
> the named algorithms so that users can call them explicitly. That is, 
> "gcd"
> would do the best thing for a numeric type, but a user could call
> "Euclid_gcd" for example if that's what they explicitly want. What do you
> think?

I suspect there's less between them than we think, but yes, the mixed 
algorithm looks best in general.

A public interface sounds OK to me, I think we also need a traits class 
for things like

* which algorithm to use.
* functions odd/even.
* function lsb and whether it's available.

etc.

I'll look into that shortly.

John.

jeremy-murphy:
Just a note that I clear forgot to implement lcm.

jzmaddock:
> Just a note that I clear forgot to implement lcm.
>
Yes, I saw that and put it back in for the gcd branch.

jzmaddock:
Just a note to say I've pushed some changes to the "gcd" branch to include:

* Mixed binary algorithm.
* Traits class support.
* Bitscanning support.

Can you see if this traits class provides what polynomial needs to be able to use gcd?

See https://github.com/boostorg/math/commit/1ef4bb64901f76084689b97f4546b3e294a84147

jeremy-murphy:
I'll try to get around to it on the weekend.

PR #32, opened 2016-04-09
FFT

jeremy-murphy:
**Work-in-progress: do not merge.**

Now I got a bit distracted from gcd and started to look at polynomial multiplication again.

So here are two textbook algorithms for FFT, the textbook in this case being Introduction to Algorithms (2009) by Cormen et al.

My plan is to use the GNU Scientific Library as a benchmark for comparison. From cursory observations it looks OK so far.

Maybe important thoughts at the moment are what kind of interface would be ideal?


jeremy-murphy:
[GSL](https://www.gnu.org/software/gsl/manual/html_node/Mathematical-Definitions.html#Mathematical-Definitions) briefly discusses whether to use +/- or -/+ exponent in the forward/inverse pair of FFT. Cormen and Numerical Recipes use +/-; GSL and FFTPACK use -/+.

jzmaddock:
Getting a correct interface is going to be tricky IMO, the main issue for me is that there is no check (or possibility of one) for correct output array size.  One option would be to make the output type a Container, rather than an iterator.  But I'm not sure if that would restrict use cases?

I'm also unsure at present, how this will be used with say integer types where std::complex is unavailable.

Best, John.

jeremy-murphy:
Another option is to operate in-place, as GSL does. Might seem weird when converting to/from reals but maybe not.

Hmm, might just be that the user has to convert their data to floating point?

jzmaddock:
On 11/04/2016 02:15, Jeremy W. Murphy wrote:
>
> Another option is to operate in-place, as GSL does. Might seem weird 
> when converting to/from reals but maybe not.
>
> Hmm, might just be that the user has to convert their data to floating 
> point?
>

If the goal is to use FFT acceleration of polynomial multiplication, 
then we can't use an inherently lossy format (floating point) to arrive 
at an exact integer result?

BTW std::complex is restricted to just float, double and long double, 
nothing "longer".

jeremy-murphy:
Right. I seem to recall something in Knuth about numeric precision in this
context but I'll have track it down.
On 11 Apr 2016 10:15 PM, "jzmaddock" <notifications@github.com> wrote:



On 11/04/2016 02:15, Jeremy W. Murphy wrote:
>
> Another option is to operate in-place, as GSL does. Might seem weird
> when converting to/from reals but maybe not.
>
> Hmm, might just be that the user has to convert their data to floating
> point?
>

If the goal is to use FFT acceleration of polynomial multiplication,
then we can't use an inherently lossy format (floating point) to arrive
at an exact integer result?

BTW std::complex is restricted to just float, double and long double,
nothing "longer".

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/boostorg/math/pull/32#issuecomment-208312464>

jeremy-murphy:
P. 307 in volume 2: "By a stroke of good luck, it turns out that everything
will work properly if we do the calculations with only a modest amount of
precision."
Of course there is more to it than good luck but the sentiment is
reassuring.

jeremy-murphy:
There are also areas of research on [integer FFT](http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=984749&queryText=integer%20Fast%20Fourier%20transform&refinements=4294944462) and of course a zillion articles on how to make it [fast](http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=6021384&queryText=speeding%20up%20Fast%20Fourier%20transform&newsearch=true).

jzmaddock:
On 11/04/2016 15:42, Jeremy W. Murphy wrote:
>
> There are also areas of research on integer FFT 
> <http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=984749&queryText=integer%20Fast%20Fourier%20transform&refinements=4294944462> 
> and of course a zillion articles on how to make it fast 
> <http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=6021384&queryText=speeding%20up%20Fast%20Fourier%20transform&newsearch=true>.
>

I confess this is an area I know very little about, I do notice GMP 
doesn't use floating point FFT's precisely because the last bit may be 
off: 
https://gmplib.org/manual/Other-Multiplication.html#Other-Multiplication

jzmaddock:
On 11/04/2016 15:42, Jeremy W. Murphy wrote:
>
> There are also areas of research on integer FFT 
> <http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=984749&queryText=integer%20Fast%20Fourier%20transform&refinements=4294944462> 
> and of course a zillion articles on how to make it fast 
> <http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=6021384&queryText=speeding%20up%20Fast%20Fourier%20transform&newsearch=true>.
>

Forgot to say: Karatsuba may actually be of more interest, since it's 
advantages kick in much sooner?

ckormanyos:
Sorry, I think I just might have over-written some comments by mistake. Sorry.

Some multiple precision libraries split 32-bit integer values into 16-bit integer values.
The FFT is carried out with double precision floating point. The 16-bit integer values
are converted back to 32-bit integer values when handling the carries after the
convolution and reverse transformation. This seems to preserve precision
for up to a few million points in the FFT.

I have recently written an extended complex class designed to handle
user-defined-types in addition to boult-in float, double and long double.
It has been through some basic tests with multiprecision and fixed-point.

Another potential Boost author was interested in getting an extended complex
class into boost a few years ago, but we never reached the review process.

Regards, Chris

jeremy-murphy:
Hi Chris,
Thanks for pointing that out. Yeah, Knuth goes into some detail on choosing
a decomposition of large numbers with sufficient precision to avoid
numerical error.
This implementation based on Cormen et al is probably too naive and I need
to take into account Knuth's formulas.

jeremy-murphy:
I'm going to close the branches that I'm not actively working on so that they don't distract me and others -- anyone is welcome to pick up where I left off of course.

PR #33, opened 2016-04-14
Algorithm 4.5.2C: gcd of n integers

jeremy-murphy:
A fairly trivial algorithm but necessary for calculating the gcd of polynomials over a unique factorization domain.

### DIFF
--- a/doc/gcd/math-gcd.qbk
+++ b/doc/gcd/math-gcd.qbk
@@ -25,6 +25,8 @@ programming problems.
 
    template < typename IntegerType >
       IntegerType  gcd( IntegerType const &a, IntegerType const &b );
+   template < typename ForwardIterator >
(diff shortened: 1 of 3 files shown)


jeremy-murphy:
What do you think of the name, `gcd_n`? I thought about calling it just `gcd` but it seemed like a lot of template magic to distinguish it from the other 2-arity `gcd` functions, and after all I guess it is a different algorithm.

jzmaddock:
Name is a good question, in C++11 land we could make gcd a variadic template that accepts n arguments, and I can see genuine uses for this, but your range-based version has good uses to.

Maybe ask on the mailing list for suggestions?

jeremy-murphy:
This is ready to go from my end, I'm happy with `gcd_range`.

jeremy-murphy:
OK, I've changed it to return the iterator position at termination too.

PR #34, opened 2016-04-20
The subresultant algorithm for gcd of polynomial over UFD

jeremy-murphy:
_This is done! Please review in detail._

**Algorithm 4.6.1C: Greatest common divisor over a unique factorization domain.**

The subresultant algorithm from Knuth is a missing piece of the polynomial gcd puzzle, but does not close the gap entirely. 

As I mentioned in conversation elsewhere on this PR, the algorithm needs to calculate intermediate values of size exponentially larger than the inputs, which rapidly leads to overflow of the intermediate variables for even trivially small input values. This turns out to be a common problem in integer algorithms and is called _intermediate expression swell_. It's quite likely that better algorithms than this one will not be entirely free of multi-precision integers, so it should not (yet) be taken as an intrinsic weakness of this algorithm.


pabristow:
I feel there must be some documentation changes from all this work ;-)

Tell me if you want me to do anything.

include/boost/math/tools/polynomial.hpp
```diff
@@ -608,6 +616,117 @@ polynomial<T> operator - (polynomial<T> a)
...
+    {
+        // Pseudo-division.
+        N const delta = u.degree() - v.degree();
+        r = u % v;
+        if (!r)
+            goto gcd_end;
+        if (r.degree() == 0)
+        {
+            v = polynomial<U>(U(1));
+            goto gcd_end;
```
jzmaddock: Is it possible to write this without a goto?

jzmaddock:
Paul, did you mean to close this?

jeremy-murphy:
For this specific PR yes we can remove the statement that gcd only works
for polynomials over a field but don't worry I can do that. :)

include/boost/math/tools/polynomial.hpp
jeremy-murphy: Yes although I left the goto in because I wasn't immediately sure how to
take it out without double-checking the termination condition. And I find
anything taboo strangely fascinating. :)

pabristow:
Oops- no of course not - sorry – speed of mouse deceived the brain.

include/boost/math/tools/polynomial.hpp
jeremy-murphy: Well, it's easy enough to just replace each goto with the return statement
but then you end up with a duplicate expression. I'll do that for now
because it's trivial and probably a better place to progress from.

jeremy-murphy:
I'm not sure how to reopen -- is it only the closer that can do that?

pabristow:
Sorry about my closing mistake.   Hope this does the trick.

jeremy-murphy:
Thanks, Paul.

jeremy-murphy:
The pseudo-division implicit in this algorithm results in intermediate values of enormous magnitude; I currently can't imagine anyone using it safely except on a arbitrary precision type.
For example, calculating the gcd of two polynomials of degree 7 with coefficients in the range 0-31 requires intermediate values that require at least 64-bits of storage.
So I'm inclined to restrict this to non-POD coefficient types for the time being, although that's hardly foolproof. Thoughts?

jeremy-murphy:
Actually, let me update that now that I got the multiprecision to work, it sometimes requires 157 bits of storage.

jzmaddock:
On 01/05/2016 14:13, Jeremy W. Murphy wrote:
>
> Actually, let me update that now that I got the multiprecision to 
> work, it sometimes requires 157 bits of storage.
>

Ouch!  Is this division in general of gcd in particular?

Can we use binary gcd and avoid division completely in that case?

John.

jeremy-murphy:
It's the pseudo-division inherent in quotient and remainder of polynomials
over a ufd, so it's not an issue with gcd per se.
But yeah, maybe we're just restricted to using Stein gcd on polynomials
over a ufd.
However, last time I tested that, the results were not great...

jeremy-murphy:
I finally got my hands on a book of computer algebra, and it turns out that there are better, more tailored algorithms for gcd of polynomials over a ufd, such as applying modular homomorphisms to reduce the problem to gcd of polynomials over a finite field, as actually mentioned on the [Wikipedia page](https://en.wikipedia.org/wiki/Polynomial_greatest_common_divisor#Modular_GCD_algorithm).

Now, I assume that although the aforementioned algorithm will become the default for single-precision coefficients, this algorithm might still have a role to play, as presumably there will be some polynomials, dense or sparse, with multi-precision coefficients, that are more easily calculated by this algorithm than the other. (Apparently the modular gcd is inefficient with sparse polynomials.)

So I suggest we keep it, but I think it should be limited to only work on multi-precision coefficients. Of course, curious and reckless users can always find a way to use it for single-precision coefficients, but I just think it would be negligent for us to let them do that _with ease_. What do you think?

jeremy-murphy:
Whoops.

jeremy-murphy:
Aha, I just read more of section 4.6.2 (factorization of polynomials) in which Knuth discusses but does not quite define the aforementioned modular gcd algorithm. He also says

> Perhaps the best general procedure would be to start with the computation of gcd(_u_(_x_), _v_(_x_)) modulo a fairly small prime _p_, not a divisor of both _l_(_u_) and _l_(_v_). If the result _q_(_x_) is 1, we're done; if it has high degree, we use Algorithm 4.6.1C; otherwise we use one of the methods above...

Meaning, one of the modular methods.

jzmaddock:
On 18/06/2016 13:33, Jeremy W. Murphy wrote:
>
> I finally got my hands on a book of computer algebra, and it turns out 
> that there are better, more tailored algorithms for gcd of polynomials 
> over a ufd, such as applying modular homomorphisms to reduce the 
> problem to gcd of polynomials over a finite field, as actually 
> mentioned on the Wikipedia page 
> <https://en.wikipedia.org/wiki/Polynomial_greatest_common_divisor#Modular_GCD_algorithm>.
>
> Now, I assume that although the aforementioned algorithm will become 
> the default for single-precision coefficients, this algorithm might 
> still have a role to play, as presumably there will be some 
> polynomials, dense or sparse, with multi-precision coefficients, that 
> are more easily calculated by this algorithm than the other. 
> (Apparently the modular gcd is inefficient with sparse polynomials.)
>
> So I suggest we keep it, but I think it should be limited to only work 
> on multi-precision coefficients. Of course, curious and reckless users 
> can always find a way to use it for single-precision coefficients, but 
> I just think it would be negligent for us to let them do that /with 
> ease/. What do you think?
>

I confess to know next to nothing about number theory.  Empirical 
scientist type here!

 From what I've read, the modular version is of greatest importance for 
say rational types where even just a few intermediate calculations can 
cause all kinds issues.  For all other types it probably comes down to a 
bit of empirical testing to determine which method is best for which 
number type, and then use some kind of traits class / heuristic to 
dispatch to the best (or approximately the best) algorithm?

jeremy-murphy:
Btw, I'm also thinking to move the gcd functions into a separate file `polynomial_gcd.hpp` in the `tools` directory, not for any especially logical reason, it just seems neater. Is that OK or do you prefer it all remains in the one file?

jzmaddock:
On 19/06/2016 13:24, Jeremy W. Murphy wrote:
>
> Btw, I'm also thinking to move the gcd functions into a separate file 
> |polynomial_gcd.hpp| in the |tools| directory, not for any especially 
> logical reason, it just seems neater. Is that OK or do you prefer it 
> all remains in the one file?
>

Seems logical/fine to me.

jeremy-murphy:
OK, the algorithm and unit tests are done, so please review them in detail at your leisure.

I introduced `leading_coefficient` for convenience with gcd algorithms, but maybe it should go in the base polynomial header file?

It's currently in the `boost::math::tools` namespace and, without really thinking about it, I also put an overload of `gcd` for integer polynomials in that namespace too. I think I'll take the `gcd` overload out though, because as I mentioned earlier, I don't think anyone wants to get surprised by this algorithm.

jeremy-murphy:
Oh and documentation of this algorithm should include some warning about intermediate value swell making it generally unsuitable for anything but arbitrary precision coefficients or trivially small inputs.

jeremy-murphy:
Hey, I'm still alive, and I would still like to finish this PR. Is there anything you think is missing? I have put a paragraph about the algorithm in the docs but there are no examples of usage for gcd on polynomials yet.

jzmaddock:
Well I'm certainly glad you're still alive ;)  My apologies I _have_ been slow to get to this.... right now I'd like to keep the develop branch clean in case we have any late SNAFU's that need fixing for the upcoming release.... so I've pulled your changes to branch "gcd_polynomial" and made a few small changes:

* Code now compiles with msvc (not really sure why it didn't to be honest).
* The new gcd in boost::math::tools is pulled into boost::math via a using declaration - that way we both overload boost::math::gcd and provide a version in the same namespace as polynomial for ADL lookup.
* polynomial.hpp includes the new gcd code by default, so there's no need for a separate header include - my feeling is that code behaviour shouldn't change depending which headers you include.
* I've updated the new gcd overload's disable_if conditions to reject things which aren't integers (to match the function it calls).

Does this look OK to you?

Over than that it looks fine, but more comments to follow shortly...

jeremy-murphy:
Hi John. I haven't looked at the changes yet but what you describe in your first message all sounds fine.
The only part I'm curious to understand better is the separation of functionality into different header files. I've noticed that some Boost libraries will put some non-member class functions in separate files, and I'm not entirely sure of the rationale. The best example is probably iostream insertion and extraction operators, which I understand is to avoid including the heavy <iostream>. But then for example Boost.Units separates the arithmetic operators from the quantity type, which seems weird because they're required for basic usage. Is that just about minute control of what is in out of the namespace at any given time?

jeremy-murphy:
I'll comment on the second list of items soon.

jzmaddock:
On 27/08/2016 11:08, Jeremy W. Murphy wrote:
>
> Hi John. I haven't looked at the changes yet but what you describe in 
> your first message all sounds fine.
> The only part I'm curious to understand better is the separation of 
> functionality into different header files. I've noticed that some 
> Boost libraries will put some non-member class functions in separate 
> files, and I'm not entirely sure of the rationale. The best example is 
> probably iostream insertion and extraction operators, which I 
> understand is to avoid including the heavy . But then for example 
> Boost.Units separates the arithmetic operators from the quantity type, 
> which seems weird because they're required for basic usage. Is that 
> just about minute control of what is in out of the namespace at any 
> given time?
>

Generally speaking we do tend to go for reasonably fine grained control 
and "only pay for what you need" - when it's appropriate to do so - in 
this case if we're overloading another function then my gut says it 
should always be overloaded, and not just if you include some other 
header.  ie no change in behaviour based on which headers you include 
and in which order.  Besides the additional header isn't that heavyweight?

PR #35, opened 2016-05-02
Fix indefinite hang with polynomial *= itself

kundor:
This code:
```C++
#include <boost/math/tools/polynomial.hpp>

int main() {
    double acoef[] = {1., 2.};
    boost::math::tools::polynomial<double> a(acoef, 1);
    a *= a;
    return 0;
}
```
will hang indefinitely (either with Boost 1.59.0, or with the current develop branch.)
This is because `polynomial::operator *=(const polynomial<U>& value)` goes through the elements of `value` in a loop, pushing back some new elements on `m_data`; when `value` is `*this`, this grows the list we are looping through, so it goes on forever.

A one-character fix is to delete the `&` in the function signature.

### DIFF
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -417,15 +417,12 @@ public:
           *this = zero;
           return *this;
       }
-      int i = value.size() - 1;
-      if (!i)
-          return multiplication(value[0]);
-      polynomial base(*this);
-      m_data.resize(size() + i);
-      for(; i >= 0; --i)
-         for(int j = base.size() - 1; j >= 0; --j)
-            m_data[i+j] += base[j] * value[i];
-      return subtraction(base);
+      std::vector<T> prod(size() + value.size() - 1);
+      for (size_type i = 0; i < value.size(); ++i)
+         for (size_type j = 0; j < size(); ++j)
+            prod[i+j] += m_data[j] * value[i];
+      m_data.swap(prod);
+      return *this;
    }
 
    template <typename U>
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -408,7 +408,7 @@ public:
        return *this;
    }
    template <class U>
-   polynomial& operator *=(const polynomial<U> value)
+   polynomial& operator *=(const polynomial<U>& value)
    {
       // TODO: FIXME: use O(N log(N)) algorithm!!!
       polynomial const zero = zero_element(std::multiplies<polynomial>());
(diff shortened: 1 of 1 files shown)


jeremy-murphy:
Thanks for catching this, Nick. I'm not sure we want to always copy the right-hand operand, and there are a couple of copies going on in there already with `base` and `t`, so I'd hope that we could safely handle self-multiplication by rearranging those copies without introducing a new one.

kundor:
A possibility without the copy is in this second commit 9c8e3f2. Essentially I just reversed the order of the loops, so that they go from the highest-order terms down. That way, if `value` is `*this`, we don't overwrite `value[i]` until after we're done using it.

I also eliminated the copying into `t`.

It is necessary for the loop index variables to be `int` instead of `size_type` because unsigned types always satisfy the `i >= 0` condition.
It's necessary to subtract `base` at the end to remove the initial values in `m_data` which were added to. Instead, `m_data` could be set to all 0s to start, but that would clobber `value` in the event that `value` is `*this`.

kundor:
Another possibility would be 
```C++
std::vector<T> prod(size() + value.size() - 1);
for (size_type i = 0; i < value.size(); ++i)
    for (size_type j = 0; j < size(); ++j)
        prod[i+j] += m_data[j] * value[i];
m_data.swap(prod);
```
This eliminates the copy `base` and the need for the subtraction.

kundor:
The more I thought about this third possibility, the more I liked it, so I committed it (627a4ce). With either the original code doing `push_back`, or the 9c8e3f2 version doing `resize`, there was bound to be a reallocation. This version has that same reallocation, but no others, and it avoids copying the previous members of `m_data` only to have to subtract them.

jeremy-murphy:
This looks great. Have you run the polynomial unit tests with it?

jzmaddock:
Just to say it fails the tests - I haven't investigated why yet.

kundor:
It occurred to me that the original code did not assume that `T` could be default-constructed with a zero value, so I removed that assumption here, too.

Can you tell me how I can run the polynomial unit tests? I'm not familiar with the build system.

jzmaddock:
Testing now.... looking good so far.

For the record the way to run the tests, assuming you have already bootstrapped bjam in the boost root directory, then:

cd libs/math/test
../../../b2 toolsetname -jN

where "toolsetname" is the name of the compiler to test - gcc msvc etc, and N is the number of tasks to run in parallel (and trust me you will want to run in parallel).

jeremy-murphy:
John, is it worth trying to get this bugfix into the 1.61 release?

jzmaddock:
On 04/05/2016 02:49, Jeremy W. Murphy wrote:
>
> John, is it worth trying to get this bugfix into the 1.61 release?
>

Much too late, the RC's are already out, if we wait for every bug report 
to be fixed we'll never release (we've been there and done that!)

John.

PR #36, opened 2016-05-02
Additional polynomial units tests & small normalization bug

kundor:
I added a unit test for the polynomial self-multiply-assignment problem fixed in pull request #35.

Also, I added a missing call to `normalize()` in `operator=(std::initializer_list<T> l)`, which allowed polynomials to exist with zero leading coefficients, and added tests to check that polynomials constructed and assigned with initalizer lists are normalized (which fail before commit 4302398, and succeed afterward.)


jzmaddock:
Thanks!

PR #39, opened 2016-05-11
operator bool() and set_zero(): test for zero and set to zero respectively.

jeremy-murphy:
Checking for zero or non-zero and setting to zero are common operations, so fast implementations are useful.

### DIFF
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -191,8 +191,8 @@ std::pair< polynomial<T>, polynomial<T> >
 division(polynomial<T> u, const polynomial<T>& v)
 {
     BOOST_ASSERT(v.size() <= u.size());
-    BOOST_ASSERT(v != zero_element(std::multiplies< polynomial<T> >()));
-    BOOST_ASSERT(u != zero_element(std::multiplies< polynomial<T> >()));
+    BOOST_ASSERT(v);
+    BOOST_ASSERT(u);
 
     typedef typename polynomial<T>::size_type N;
     
@@ -245,9 +245,9 @@ template <typename T>
 std::pair< polynomial<T>, polynomial<T> >
 quotient_remainder(const polynomial<T>& dividend, const polynomial<T>& divisor)
 {
-    BOOST_ASSERT(divisor != zero_element(std::multiplies< polynomial<T> >()));
+    BOOST_ASSERT(divisor);
     if (dividend.size() < divisor.size())
-        return std::make_pair(zero_element(std::multiplies< polynomial<T> >()), dividend);
+        return std::make_pair(polynomial<T>(), dividend);
     return detail::division(dividend, divisor);
 }
 
@@ -389,7 +389,7 @@ public:
    polynomial& operator %=(const U& /*value*/)
    {
        // We can always divide by a scalar, so there is no remainder:
-       *this = zero_element(std::multiplies<polynomial>());
+       this->clear();
        return *this;
    }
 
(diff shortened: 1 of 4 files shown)


include/boost/math/tools/polynomial.hpp
```diff
@@ -455,6 +454,18 @@ class polynomial :
        normalize();
        return *this;
    }
+   
+   // Useful for fast test of equality to zero.
+   operator bool() const
```
jzmaddock: This should be explicit when BOOST_NO_CXX11_EXPLICIT_CONVERSION_OPERATORS is not set, otherwise use an implicit conversion to an "unmentionable" type, for example multiprecision::number uses:

   typedef bool (self_type::*unmentionable_type)()const;

   BOOST_MP_FORCEINLINE operator unmentionable_type()const
   {
      return is_zero() ? 0 : &self_type::is_zero;
   }

Which prevents accidental conversion to an arithmetic type.

jeremy-murphy: Ahh, wow, that's a new one on me. What about naming a function that sets the polynomial to zero? I have used the verb `clear` but maybe `set_zero` is more appropriate? This is on the assumption that it is more efficient than assignment from zero, which I assume the compiler is not quite smart enough to optimize away for a class such as polynomial.

jeremy-murphy:
I still need to update the docs after we settle on the names.

jeremy-murphy:
And what about the horrible workaround for gcc 4.6.x?

include/boost/math/tools/polynomial.hpp
jzmaddock: On 12/05/2016 12:08, Jeremy W. Murphy wrote:
>
> In include/boost/math/tools/polynomial.hpp 
> <https://github.com/boostorg/math/pull/39#discussion_r63002315>:
>
> > @@ -455,6 +454,18 @@ class polynomial :
> >         normalize();
> >         return *this;
> >     }
> > +
> > +   // Useful for fast test of equality to zero.
> > +   operator bool() const
>
> Ahh, wow, that's a new one on me. What about naming a function that 
> sets the polynomial to zero? I have used the verb |clear| but maybe 
> |set_zero| is more appropriate? This is on the assumption that it is 
> more efficient than assignment from zero, which I assume the compiler 
> is not quite smart enough to optimize away for a class such as polynomial.
>

clear() works for me.

jzmaddock:
On 12/05/2016 12:10, Jeremy W. Murphy wrote:
>
> And what about the horrible workaround for gcc 4.6.x?
>

I don't believe that's required in this case - it was the presence of 
other operator() overloads that caused the issue (it called the bool 
operator rather than the correct one if memory serves).

jeremy-murphy:
John, if you're happy with my choice of `set_zero` for the modifying function, then this is all done from my end.

jzmaddock:
Can you mark operator bool in the docs as explicit to match the code?

Re the name "set_zero", I confess I'm not overly keen on set_ and get_ prefixes, so I would be inclined to go with either just "zero" or even "clear".  Does assigning from a literal zero produce the same result BTW?

Thanks for this, John.

jeremy-murphy:
I'm definitely not in favour of the object-oriented habit of 'getters and setters' on a per-member basis, but I thought it was OK in this case. I thought "clear" was too tied to the underlying implementation and is not meaningful to a polynomial per se (although that's not a huge sin). "zero" seems ambiguous as to whether it is setting or querying? I don't love "set_zero" but the alternatives all seem more problematic.

jzmaddock:
OK lets go with what we have and move on, merging to develop...

PR #40, opened 2016-05-11
pow function for polynomials

kundor:
I would like to propose a function
```C++
template <class T>
polynomial<T> pow(polynomial<T> base, int exp);
```
This can be found via argument-dependent lookup, so that polynomials can be used more generically.

This raises a domain exception (according to the default policy) if the exponent is negative. I thought of trying `pow(base[0], exp)` in the case that `base` is a constant (degree-0 polynomial) and `exp` is negative. This works fine when `T` is `double`, or when `T` is `int` and `base[0]` is ±1.
But it fails if `pow(T, int)` cannot be found, and gives a huge swath of errors if `T` is from Boost.Multiprecision, and perhaps isn't worth the trouble.

### DIFF
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -649,29 +649,13 @@ template <class T>
 polynomial<T> pow(polynomial<T> base, int exp)
 {
     if (exp < 0)
-    {
-        if (base.size() == 1)
-        {
-            using std::pow;
-            T res(pow(base[0], exp)); // Use ADL to find the right pow
-            if (res == T(0)) // integral
-                return policies::raise_domain_error(
-                        "boost::math::tools::pow<%1%>",
-                        "Negative powers not supported for integer constants.",
-                        base, policies::policy<>());
-            return polynomial<T>(res);
-        }
         return policies::raise_domain_error(
                 "boost::math::tools::pow<%1%>",
-                "Negative powers are not supported for non-constant polynomials.",
+                "Negative powers are not supported for polynomials.",
                 base, policies::policy<>());
         // if the policy is ignore_error or errno_on_error, raise_domain_error
         // will return std::numeric_limits<polynomial<T>>::quiet_NaN(), which
         // defaults to polynomial<T>(), which is the zero polynomial
-        /* ARG, but throw_on_error (default) requires boost::math::policies::digits<T,policy>
-         * which requires std::numeric_limits<T> to be specialized,
-         * just to try to format val into %1% in message, which isn't even there!! */
-    }
     polynomial<T> result(T(1));
     if (exp & 1)
         result = base;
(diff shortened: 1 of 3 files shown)


test/test_polynomial.cpp
```diff
@@ -298,3 +298,26 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_odd_even, T, all_test_types)
     BOOST_CHECK_EQUAL(odd(b), false);
     BOOST_CHECK_EQUAL(even(b), true);
 }
+
+BOOST_AUTO_TEST_CASE_TEMPLATE( test_pow, T, all_test_types )
+{
+    polynomial<T> a(d3a.begin(), d3a.end());
+    polynomial<T> const one = identity_element(std::multiplies< polynomial<T> >());
```
jeremy-murphy: Btw, I don't think we have to go to this length to specify one and zero after all unless it is inside a generic algorithm where the binary operation is parameterized. I think when you know the operation it's fine to just say
`polynomial<T> zero;
polynomial<T> one(T(1));`

jzmaddock: +1, I'd prefer to simplify that, though it doesn't make much difference in a test case I admit.

jzmaddock:
What was the issue with raising a multiprecision type to an integer power - that should work (via ADL)?

You would still have a problem when T is an integer though, as it would likely call pow(double, int) which would throw away digits when T has more than 53-bit precision.  So the existing code is probably just fine, I was just curious what the multiprecision issue was.

kundor:
I only tried using `pow` for the base type when `exp` is negative (otherwise, the normal approach works fine for degree-0 polynomials.) When `T` is an integer, if `base[0]` is ±1, it gives the right result, and otherwise `T(pow(base[0], exp))` should truncate the result to 0. I caught that and raised a `domain_error`.

I dug into the errors more and found that it's specifically for `cpp_rational`. G++ gives many pages of error messages boiling down to "no matching function for call to `pow`".

include/boost/math/tools/polynomial.hpp
```diff
@@ -644,6 +645,30 @@ bool even(polynomial<T> const &a)
...
+    if (exp & 1)
+        result = base;
+    /* "Exponentiation by squaring" */
+    while (exp >>= 1)
+    {
+        base *= base;
+        if (exp & 1)
+            result *= base;
+    }
+    return result;
```
jeremy-murphy: Apart from the constraint on `exp` this is after all a generic exponentiation algorithm, which would be preferable to see implemented separately from the context of the polynomial class.

Marshall actually implemented [one version](https://github.com/boostorg/algorithm/blob/develop/include/boost/algorithm/algorithm.hpp) (from Knuth) of this algorithm but it explicitly uses modulo to test for oddness, which is of course a performance drag and a reason why I think `odd` deserves to be a trait function for types.

kundor: For an int, I would be surprised if testing `exp & 1` vs `n % 2 == 1` were actually different performance-wise  (I'd expect the compiler to emit the same assembly.) But that version uses `y = x * y` and `x = x * x` instead of `*=`, creating lots of extra temporaries .

PR #41, opened 2016-05-15
Polynomial fixes: for mixed-type arithmetic etc.

kundor:
I was fixing some minor problems in the polynomial documentation, when I noticed that `operator / (polynomial<T>, U)` and `operator % (polynomial<T>, U)` were not actually templated on the type of the scalar `U`, as reported in the docs, but were provided by Boost.Operators and only accepted scalars of type `T`.

This isn't very noticeable, since either way, the code compiles only if `U` is implicitly convertible to `T`. However, it does cause different behavior. For instance, `polynomial<int>{2,4,6} / 1.5` is an identity operation (the `1.5` is truncated to `1`), while `polynomial<int>{2,4,6} * 0.6667` yields {1,2,4}.
More seriously, if `a` is a `polynomial<int>`, `a = a / 1.5` and `a /= 1.5` would give different results.

To fix this, I added templated implementations of these operators just like all the other polynomial-scalar operators.

For consistency, I used Boost.Operators to provide all the polynomial-polynomial operations (it was being used just for division, modulus, and operator !=), which are conveniently packed together in the class `ordered_euclidean_ring_operators< polynomial<T> >`.

The self-assign operators are all templated on `polynomial<U>`. However, `operator /=` and `%=` call out to `quotient_remainder`, which only takes two matching `polynomial<T>` arguments. So it's ambiguous what to convert when calling `quotient_remainder(polynomial<T>, polynomial<U>)`, causing an error.

To fix this, `polynomial<T>::operator /=(const polynomial<U>& q)` can call `quotient_remainder(*this, polynomial<T>(value))`.  But then you might as well not template the operator on type U, and just make it `polynomial::operator /=(const polynomial& q)`; you can still use it with any `polynomial<U>` because there is an implicit conversion to `polynomial<T>`.

Unfortunately, just removing the templating causes `polynomial<U>` arguments to choose the scalar version `operator /=(const U& value)` instead. So I had to guard that function with `enable_if< boost::is_convertible<U,T> >`.

This situation _does_ result in inconsistent behavior. For instance, multiplying a `polynomial<int>` by a `polynomial<double>` will do each multiplication with `double`s before truncating back to `int`, whereas dividing by a `polynomial<double>` will first truncate all the values to `int`.
Similarly, with `polynomial<int> a{3}`, `a /=  1.5` results in 2, but `a /= polynomial<double>{1.5}` results in 3.

Changing this, so that it promotes internal computations and then truncates, would mean changing the implementation of quotient_remainder.

I would be happy to fix this up if you want to merge #39 first (it will conflict a little.)


jeremy-murphy:
Thanks for picking up these defects. I remember now when I was implementing `operator >>`, etc that it did not work using the Boost.Operator facility because I needed a new template type for the right-hand parameter.

jeremy-murphy:
It would be great if you could include in this PR the unit tests that would otherwise fail without these changes.

jeremy-murphy:
Hmmm, I guess this begs the question as to whether non-PoD numbers (such as polynomial) should promote according to the same rules as PoDs. What do the multiprecision classes do, John? How do they handle arithmetic with a mixture of types?

kundor:
Speaking of which, is there really a need to template the shift operators on the right-hand parameter? Is there a case for calling it with anything but an `int` (or something convertible thereto)? I don't think this class is meant to support polynomials of degree 32,000 or more ;-) (a signed int goes up at least to 2^15-1, by the standard.) Taking an `int` and checking that it's positive is preferable to taking an unsigned number, in that otherwise a call `p <<= -1` will compile without complaint, and silently try to extend `p` by 4 million elements.

jzmaddock:
On 15/05/2016 16:46, Nick Matteo wrote:
>
> Speaking of which, is there really a need to template the shift 
> operators on the right-hand parameter? Is there a case for calling it 
> with anything but an |int| (or something convertible thereto)? I don't 
> think this class is meant to support polynomials of degree 32,000 or 
> more ;-) (a signed int goes up at least to 2^15-1, by the standard.) 
> Taking an |int| and checking that it's positive is preferable to 
> taking an unsigned number, in that otherwise a call |p <<= -1| will 
> compile without complaint, and silently try to extend |p| by 4 million 
> elements.
>

I wouldn't think so no.  Jeremy?

jzmaddock:
On 15/05/2016 12:53, Jeremy W. Murphy wrote:
>
> Hmmm, I guess this begs the question as to whether non-PoD numbers 
> (such as polynomial) should promote according to the same rules as 
> PoDs. What do the multiprecision classes do, John? How do they handle 
> arithmetic with a mixture of types?
>

Well first off they don't use Boost.Operators which is rather ill suited 
to this kind of thing, but basically given:

a op b

If there is an implicit (ie non-lossy) conversion from b to a, then b is 
converted to typeof(a), otherwise if a is implicitly convertible to 
typeof(b) then a gets promoted.

Of course internally, the value may be used directly where it is more 
efficient to do so, but the operators themselves enforce the above rule.

In this situation, for:

polynomial<T> op U

we would want to enforce that there is an implicit, non-lossy conversion 
from U to T, which means no multiplication of polynomial<int> by a 
double etc.

The logic used is somewhat multiprecision-specific, and in 
boost/multiprecision/traits/is_restricted_conversion.hpp

HTH, John.

kundor:
I rebased the branch, and added the tests Jeremy asked for.

jeremy-murphy:
With respect to the shift operators, no, there is no particular need for them to be templated, and off the cuff it only makes sense to call them with an integral type. How to handle negative values raises a curious thought, though: should `operator>>(-x)` be equivalent to `operator<<(x)`? I understand that for integers, shifting by a negative amount is undefined behaviour, so probably we should follow suit, but we're not obliged to.

jeremy-murphy:
Hmmm, so for `operator*=(U)` we should be restricting `U` to be implicitly convertible to `T`?

kundor:
Re: `operator*=(U)`.  We can restrict all the scalar operations to arguments that are implicitly convertible to `T`, by just removing the templating and only taking `T`. The question is whether we want
```C++
polynomial<int> p{3,2};
p *= 1.5;
```
to give us 3x + 4, analogous to what `int i=3; i *= 1.5;` would do,  or be an identity operation,  or be a compile-time error.

Personally, I would rather have it act like built-in `int`s, i.e. multiply by the double and then truncate, which is what it does now.

Making it an error would require a templated version of the function just to fail, using something like the `is_lossy_conversion` trait from multiprecision, which seems like overkill.

On the other hand, removing the templating from poly-poly operations (`operator *=(polynomial<U>)` etc.), would make their behavior more consistent (in that currently, `operator/=(polynomial)` must act that way.) I think currently I am in favor of removing arithmetic between polynomial<T> and polynomial<U> entirely, which means removing the template from each `operator op=(polynomial)` function, and making the conversion constructor `polynomial(const polynomial<U>& p)` `explicit`.

kundor:
Commit 910a62a removes arithmetic between polynomials with different base types, so you'll have to explicitly cast one operand or the other (e.g. if `p` is a `polynomial<int>` and `q` is a `polynomial<double>`, you can write `polynomial<double>(p) * q`, or `p *= polynomial<int>(q)`.

All the tests in test_polynomial.cpp still pass.

include/boost/math/tools/polynomial.hpp
```diff
@@ -527,64 +531,59 @@ class polynomial :
...
     }
 
+    template <class U>
+    polynomial& modulus(const U& value)
+    {
+       typedef typename std::vector<T>::iterator iter;
+       for (iter it = m_data.begin(); it != m_data.end(); ++it)
+           *it -= T(value * T(*it / value));
+       return *this;
+    }
```
jeremy-murphy: Should `modulus` still have a `U` template?

include/boost/math/tools/polynomial.hpp
```diff
@@ -387,30 +387,37 @@ class polynomial :
...
 
    template <class U>
-   polynomial& operator %=(const U& /*value*/)
+   polynomial& operator %=(const U& value)
    {
-       // We can always divide by a scalar, so there is no remainder:
-       *this = zero_element(std::multiplies<polynomial>());
+       // In the case that T is integral, this preserves the semantics
+       // p == r*(p/r) + (p % r), for polynomial<T> p and U r.
+       if (std::numeric_limits<T>::is_integer)
```
jeremy-murphy: Since this is a compile-time constant, can't we combine it with `enable_if` for static dispatch?

include/boost/math/tools/polynomial.hpp
kundor: Yes, `operator %=(U)`, and the other self-modifying operators for scalar arguments, are all still templated. We can change that if you want to be more strict.

include/boost/math/tools/polynomial.hpp
kundor: I seem to recall that using `enable_if` based only on `T`, which is not one of the template parameters for the function, doesn't work. Let me check.

kundor: Yes, I was right: GCC 6.1, GCC 5.3, and Clang 703.0.31 all complain vociferously. To separate them, it's necessary to add a dummy condition to the enable_if that depends on `U`, e.g. go back to `BOOST_DEDUCED_TYPENAME enable_if_c<boost::is_convertible<U,T>::value &&
std::numeric_limits<T>::is_integer, polynomial >::type& operator %=(const U& value)`.

I don't think this mess is worth it:
- the modulus operator on a polynomial doesn't seem performance-critical
- a conditional testing a compile-time boolean should only emit the chosen branch anyway

jeremy-murphy:
I probably agree with all of the changes in this PR in principle, but to be honest, I find it too big to think about all at once and be sure that every detail is right. I suggest that you split this into three PRs, one for each of the independent things being changed:

1. the bug fix for `*` versus `*=`
2. adding `operator<` leading to `ordered_euclidean_ring_operators`
3. changes to the shift operators.

I realize that's not fun but that's what I would do.

NAThompson:
Thanks, and sorry this sat for so long. Most of it has since been fixed or overtaken on develop. Scalar `/` has been templated on `U` for a while, so `polynomial<int>{2,4,6} / 1.5` now matches `* 0.6667`. The integer remainder bug you found was still there, though: `polynomial<int> % scalar` always returned zero. #1484 fixes that and credits you as co-author. We're leaving out the Boost.Operators rework, `operator<` and the removal of mixed-type arithmetic. Jeremy asked for those to be split out, and they'd be API changes today. Closing as superseded by #1484.

PR #46, opened 2016-07-18
Fix a lognormal mode formula in the documentation

DenisKolodin:
Mode of lognormal distribution has to be: mode
Wiki [source](https://en.wikipedia.org/wiki/Log-normal_distribution).


pabristow:
Merged - thanks for spotting this.

PR #48, opened 2016-08-20
More misspellings

inkstink:
I found and fixed a few more misspellings.


pabristow:
How did I miss those!

jzmaddock:
Thanks! Now fixed in develop.

PR #51, opened 2016-10-25
Fix issues in the inspection report

zerotypos-found:
- Remove Non-ASCII characters (i.e. replace dashes with hyphens).
- Protect min from macro expansion.


jzmaddock:
Thanks for that!

PR #52, opened 2016-11-05
Reimplement quaternion routines without valarray

matwey:
The major issue is that std::vallarray allocates heap memory at construction.
This is time-consuming operation and leads to unnecessary dramatic performance loss.

See discussion for reference: http://lists.boost.org/boost-users/2015/07/84711.php

### DIFF
--- a/include/boost/math/quaternion.hpp
+++ b/include/boost/math/quaternion.hpp
@@ -11,6 +11,7 @@
 #define BOOST_QUATERNION_HPP
 
 
+#include <algorithm>
 #include <complex>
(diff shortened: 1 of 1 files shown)


ZaMaZaN4iK:
Hello. Is it still required? If yes, can you review and merge it please?

jzmaddock:
Hi Alex, the basic issue is that this class could use a complete re-write to bring it into the 21st century.  I'll try to look into this PR some time soon.

matwey:
Why can't it be more faster 20th century version until it will be rewritten into 21st?

jzmaddock:
See https://github.com/boostorg/math/pull/90

PR #55, opened 2017-02-22
Remove Jacobi elliptic functions from TODO list

NAThompson:
. . . . as they have already been added to the library.


jzmaddock:
Quite right, however, it's the quickbook source that has to be changed - the HTML is periodically generated from that, we don't edit it directly.  Thanks!

PR #58, opened 2017-02-28
Addition of the barycentric rational interpolant.

NAThompson:
This commit can be regarded as a "friend" to the cubic b spline, filling out the feature set.
Whereas the cubic b spline interpolates uniformly spaced data, the barycentric rational interpolant can interpolate arbitrarily spaced data.
The price that is paid is that evaluation of the interpolant is O(N), rather than O(1) for the cubic b spline interpolant.

The routine supports essentially arbitrary approximation order d; with error of O(h^d+1).
It is cppcheck-clean, run under AddressSanitizer and UndefinedBehaviorSanitizer, and works with boost::multiprecision.

References:

1) Floater, M.S. & Hormann, K. Numer. Math. (2007) 107: 315. doi:10.1007/s00211-007-0093-y

2) Numerical Recipes, Section 3.4.1

### DIFF
--- a/doc/internals/barycentric_rational_interpolation.qbk
+++ b/doc/internals/barycentric_rational_interpolation.qbk
@@ -1,0 +1,45 @@
+[section:barycentric Barycentric Rational Interpolation]
+
+[h4 Synopsis]
+
+``
+#include <boost/math/tools/barycentric_rational_interpolation.hpp>
+``
+
+
+[h4 Description]
+
+Barycentric rational interpolation is a high-precision interpolation routine for non-uniformly spaced samples.
+It requires [bigo](N) time for construction, and [bigo](N) time for each evaluation.
+Linear time evaluation is not optimal; for instance the cubic b spline can be evaluated in constant time.
+However, using the cubic b spline requires uniformly spaced samples, which are not always available.
+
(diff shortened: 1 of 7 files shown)


jzmaddock:
This, along with the cubic spline code is now in develop (not sure why it isn't showing up as merged, but there we go).

I've made some minor changes to both to fix some compiler errors, missing #includes etc.  Also extended the examples, and moved the docs around a bit.  Can you take a look at https://www.dropbox.com/s/9uwe1tujc0p12b5/math.pdf?dl=0 and make sure the docs are OK? There are some minor formatting errors, but otherwise look OK to me now.  These two interpolators are in their own sub-section under "tools" BTW.

pabristow:
This looks good to me.  Very useful tools. Nice examples too.  Pretty obvious how they work I think (novice novices will use this code!).  But I did notice several typos and trivial not-niceties.  (Most seriously a dud character in a comment). We might even run to an umlaut for Prof Dr Schoedinger ;-) Could provide more links internally, for example to root finding tools, and externally, and additions to the indexes.  I suggest that at some point I go through this with my nitcomb and then commit and push.  But I don't want to collide with other changes, so tell me when would be convenient.

PR #59, opened 2017-03-06
Adaptive Trapezoidal Quadrature

NAThompson:
This routine estimates the definite integral of a function f.
Assuming that f is periodic, it can be shown that this routine converges exponentially fast.
In fact, the test cases given exhibit exponential convergence with decreasing stepsize.

A potential improvement is using the Bulirsch sequence rather than the Romberg sequence to schedule the refinements.
However, the convergence is so rapid for functions of the class specified above that there seems to be no need at present.

This code is cppcheck clean, and runs successfully under AddressSanitizer and UndefinedBehaviorSanitizer.


NAThompson:
This should be ready to go.

NAThompson:
Looks like I've broke the "test/compile_test/poison.hpp" test by adding 'std::isfinite' to the concept list. However, without adding this, the trapezoidal concept test doesn't pass, so I'm a bit flummoxed here.

jzmaddock:
I've moved this to https://github.com/boostorg/math/pull/70 so that we can trigger a CI build, that also contains a bunch of minor fixes for older compilers.

PR #62, opened 2017-04-12
Derivative of Legendre polynomials of the first kind. 

NAThompson:
Very simple implementation; not much there but still more convenient to have it written than rewrite it every time it's needed.

### DIFF
--- a/doc/sf/legendre.qbk
+++ b/doc/sf/legendre.qbk
@@ -7,36 +7,42 @@
 ``
 
    namespace boost{ namespace math{
-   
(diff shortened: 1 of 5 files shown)


jzmaddock:
That looks great, thanks.

Note that I am not checking anything in detail or merging to develop until 1.64 is out, at which point it seems I will have some catching up to do!

NAThompson:
No problem; I'm just getting the hang of how this project is put together, so 1.65 is a good merge target for me.

PR #65, opened 2017-05-03
Simple cleanup.

NAThompson:
Remove unused typedef. Fix dangling reference to policy in airy. Pipe unused variables in compile test to /dev/null to suppress compiler warnings. Remove comment within comment. Remove pragma about SSE2 as it is now a standard instruction set.


jzmaddock:
Moved to https://github.com/boostorg/math/pull/71 to trigger CI test.

PR #67, opened 2017-05-17
Chebyshev polynomials.

NAThompson:


include/boost/math/special_functions/chebyshev.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+template<class Real, bool second=false>
+inline Real chebyshev_imp(unsigned n, Real const & x)
+{
+    using std::cosh;
+    using std::acosh;
+    using std::pow;
+    Real T0 = 1;
+    Real T1;
+    if (second)
```
jzmaddock: To be honest, I think I'd rather have 2 separate functions than switch based on a template parameter - there's not much overlap between the two cases other than the recursion at the end?

include/boost/math/special_functions/chebyshev.hpp
```diff
@@ -0,0 +1,96 @@
...
+       std::swap(T0, T1);
+       T1 = boost::math::chebyshev_next(x, T0, T1);
+       ++l;
+    }
+    return T1;
+}
+} // namespace detail
+
+template<class Real>
+Real chebyshev_t(unsigned n, Real const & x)
```
jzmaddock: Paul, can I persuade you to make these forwarders follow the example here: http://www.boost.org/doc/libs/1_64_0/libs/math/doc/html/math_toolkit/special_tut/special_tut_impl.html so that the behaviour of these matches the rest of the lib?  Thanks!

doc/math.qbk
```diff
@@ -619,6 +620,7 @@ and as a CD ISBN 0-9504833-2-X  978-0-9504833-2-0, Classification 519.2-dc22.
 [include quaternion/math-quaternion.qbk]
 [include octonion/math-octonion.qbk]
 [include gcd/math-gcd.qbk]
+[include interpolators/cubic_b_spline.qbk]
```
jzmaddock: This seems to be a hangover from something else - that file should be included already (in a different location).

jzmaddock:
Thanks Nick, in addition to the line comments I added, can I get you to hook up testing of this in libs/math/test/compile_test/instantiate.hpp so that our concept checking is enabled for this code?

Thanks!

jzmaddock:
Since you so kindly volunteered to tackled this ;)  I have one other suggestion and a question:

Suggestion: it would also be useful to have access to the roots of the chebyshev polynomials (these are trivially calculated I think?)

Question: Now that we have bumped the polynomial class up to a first class citizen of Boost.Math, is there a use for

template <class T>
polynomial<T> chebyshev_t_polynomial(int n);  // etc

For accessing the actual coefficients?  Presumably polynomial arithmetic plus the recurrence relation would suffice for that unless there's an easier way?

Best, John.

include/boost/math/special_functions/chebyshev.hpp
NAThompson: That seems reasonable to me; will make the change.

NAThompson:
Yeah, I'll get the compile test and concept checks working. I'm still trying to figure out the convention for the zeros; should we use the "only present non-negative zeros" (which are natural for the Legendre polynomial zeros), or follow the notation used in <a href="https://www.amazon.com/Approximation-Theory-Practice-Applied-Mathematics/dp/1611972396/ref=sr_1_1?ie=UTF8&qid=1495384518&sr=8-1&keywords=approximation+theory+and+approximation+practice">Trefethen</a>.

I have a Chebyshev tranform in the works, which will allow "time-compression" of function evaluation, numerical differentiation, and numerical integration, as a class. This class should satisfy your `etc` in `chebyshev_t_polynomial(int n); // etc`.

NAThompson:
New PR elsewhere.

PR #70, opened 2017-06-16
New PR derived from #59

jzmaddock:
Created this PR so we can get CI testing, also contains many minor fixes.

### DIFF
--- a/doc/math.qbk
+++ b/doc/math.qbk
@@ -620,7 +620,7 @@ and as a CD ISBN 0-9504833-2-X  978-0-9504833-2-0, Classification 519.2-dc22.
 [include octonion/math-octonion.qbk]
 [include gcd/math-gcd.qbk]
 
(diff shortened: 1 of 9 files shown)


NAThompson:
Just so you don't have to fix this on all my other PRs, what is the benefit of using  ` boost::math::tools::epsilon<Real>()` over `std::numeric_limits<Real>::epsilon()`?

jzmaddock:
On 16/06/2017 19:26, Nick wrote:
>
> Just so you don't have to fix this on all my other PRs, what is the 
> benefit of using |boost::math::tools::epsilon<Real>()| over 
> |std::numeric_limits<Real>::epsilon()|?
>

It's designed to work with types that don't specialize numeric_limits - 
for example types with runtime variable precision. Testing with both 
real_concept and std_real_concept should catch any conceptual errors 
like this.

Best, John.

---
http://www.avg.com

PR #72, opened 2017-06-25
PR66: Moved to here for integration testing.

jzmaddock:


test/tanh_sinh_quadrature_test.cpp
```diff
@@ -0,0 +1,790 @@
...
+template<class Real>
+void test_detail()
+{
+    std::cout << "Testing tanh_sinh_detail on type " << boost::typeindex::type_id<Real>().pretty_name() << "\n";
+    Real tol = sqrt(boost::math::tools::epsilon<Real>());
+    tanh_sinh_detail<Real, boost::math::policies::policy<> > integrator(20, boost::math::tools::epsilon<Real>());
+    auto f = [](const Real& x, const Real&) { return x*x; };
+    Real err;
+    Real L1;
+    Real Q = integrator.integrate(f, &err, &L1, 0, boost::math::tools::epsilon<Real>(), boost::math::tools::epsilon<Real>(), tol);
```
jzmaddock: Ugh, not sure how that slipped through, will remove as suggested.

include/boost/math/quadrature/detail/tanh_sinh_detail.hpp
```diff
@@ -0,0 +1,689 @@
...
+
+#include <cmath>
+#include <vector>
+#include <boost/math/tools/atomic.hpp>
+#include <boost/detail/lightweight_mutex.hpp>
+#include <typeinfo>
+#include <boost/math/constants/constants.hpp>
+#include <boost/math/special_functions/next.hpp>
+
+
```
jzmaddock: I keep trying to find a way to make this work, with GCC-6.3 and:

#ifdef __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wliteral-range"
#pragma GCC diagnostic ignored "-Woverflow"
#endif

I still see:

In file included from ..\..\../boost/math/quadrature/tanh_sinh.hpp:35:0,
                 from tanh_sinh_quadrature_test.cpp:17:
..\..\../boost/math/quadrature/detail/tanh_sinh_detail.hpp:561:7: warning: floating constant truncated to zero [-Woverflow]

Yet strangely, -Wno-overflow does silence things when use don the command line :(

Using #pragama GCC system_header seems to work though - not ideal but it will have to do I guess.

test/tanh_sinh_quadrature_test.cpp
```diff
@@ -0,0 +1,790 @@
...
+      boost::math::tgamma(z), tol);
+   BOOST_CHECK_CLOSE_FRACTION(integrator.integrate([](const Real& t) {  using std::exp; return exp(-t*t); }, std::numeric_limits<Real>::has_infinity ? -std::numeric_limits<Real>::infinity() : -boost::math::tools::max_value<Real>(), std::numeric_limits<Real>::has_infinity ? std::numeric_limits<Real>::infinity() : boost::math::tools::max_value<Real>()),
+      boost::math::constants::root_pi<Real>(), tol);
+}
+
+template <class Real>
+void test_2_arg()
+{
+   BOOST_MATH_STD_USING
+   std::cout << "Testing 2 argument functiors on type " << boost::typeindex::type_id<Real>().pretty_name() << "\n";
```
jzmaddock: Oops, thanks for spotting that!

include/boost/math/quadrature/detail/tanh_sinh_detail.hpp
```diff
@@ -0,0 +1,689 @@
...
+   //
+   // First row is special:
+   //
+   Real h = m_t_max / m_inital_row_length;
+
+   std::vector<Real> temp(m_inital_row_length + 1, Real(0));
+   for (std::size_t i = 0; i < m_inital_row_length; ++i)
+   {
+      Real t = h * i;
+      if ((first_complement < 0) && (t < m_t_crossover))
```
jzmaddock: Ugh... that logic is all wrong.... the "problem" is that if there's a bug in the first row's values the error generated gradually gets erased by subsequent rows so I didn't spot it.  Will fix by using the same logic used by subsequent rows (which are thankfully correct!)

PR #79, opened 2017-08-14
enable move semantics for polynomial

lakshayg:
This patch is to enable move semantics in boost polynomial. I will be updating in small patches so that I can get feedback on how to move forward and stay consistent with boost standards. Any feedback/advice in appreciated.

### DIFF
--- a/include/boost/math/tools/polynomial.hpp
+++ b/include/boost/math/tools/polynomial.hpp
@@ -290,6 +290,12 @@ public:
           m_data.push_back(point);
    }
 
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
+   // move:
+   polynomial(polynomial&& p)
+      : m_data(std::move(p.m_data)) { }
+#endif
+
    // copy:
    polynomial(const polynomial& p)
       : m_data(p.m_data) { }
@@ -319,26 +325,26 @@ public:
 
 
    // access:
-   size_type size()const { return m_data.size(); }
-   size_type degree()const
+   size_type size() const { return m_data.size(); }
+   size_type degree() const
    {
        if (size() == 0)
            throw std::logic_error("degree() is undefined for the zero polynomial.");
        return m_data.size() - 1;
-    }
+   }
    value_type& operator[](size_type i)
    {
       return m_data[i];
    }
-   const value_type& operator[](size_type i)const
+   const value_type& operator[](size_type i) const
    {
       return m_data[i];
    }
-   T evaluate(T z)const
+   T evaluate(T z) const
    {
       return m_data.size() > 0 ? boost::math::tools::evaluate_polynomial(&m_data[0], z, m_data.size()) : 0;
    }
-   std::vector<T> chebyshev()const
+   std::vector<T> chebyshev() const
    {
       return polynomial_to_chebyshev(m_data);
    }
@@ -354,6 +360,19 @@ public:
    }
 
    // operators:
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
+   polynomial& operator =(polynomial&& p)
+   {
+       m_data = std::move(p.m_data);
+       return *this;
+   }
+#endif
+   polynomial& operator =(const polynomial& p)
+   {
+       m_data = p.m_data;
+       return *this;
+   }
+
    template <class U>
    polynomial& operator +=(const U& value)
    {
@@ -409,20 +428,25 @@ public:
        normalize();
        return *this;
    }
+
+   template <typename U, typename V>
+   void multiply(const polynomial<U>& a, const polynomial<V>& b) {
+       if (!a || !b)
+       {
+           this->set_zero();
+           return;
+       }
+       std::vector<T> prod(a.size() + b.size() - 1, T(0));
+       for (unsigned i = 0; i < a.size(); ++i)
+           for (unsigned j = 0; j < b.size(); ++j)
+               prod[i+j] += a.m_data[i] * b.m_data[j];
+       m_data.swap(prod);
+   }
+
    template <class U>
    polynomial& operator *=(const polynomial<U>& value)
    {
-      // TODO: FIXME: use O(N log(N)) algorithm!!!
-      if (!value)
-      {
-          this->set_zero();
-          return *this;
-      }
-      std::vector<T> prod(size() + value.size() - 1, T(0));
-      for (size_type i = 0; i < value.size(); ++i)
-         for (size_type j = 0; j < size(); ++j)
-            prod[i+j] += m_data[j] * value[i];
-      m_data.swap(prod);
+      this->multiply(*this, value);
       return *this;
    }
 
@@ -564,6 +588,26 @@ inline polynomial<T> operator + (const polynomial<T>& a, const polynomial<T>& b)
    result += b;
    return result;
 }
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
+template <class T>
+inline polynomial<T> operator + (polynomial<T>&& a, const polynomial<T>& b)
+{
+   a += b;
+   return a;
+}
+template <class T>
+inline polynomial<T> operator + (const polynomial<T>& a, polynomial<T>&& b)
+{
+   b += a;
+   return b;
+}
+template <class T>
+inline polynomial<T> operator + (polynomial<T>&& a, polynomial<T>&& b)
+{
+   a += b;
+   return a;
+}
+#endif
 
 template <class T>
 inline polynomial<T> operator - (const polynomial<T>& a, const polynomial<T>& b)
@@ -572,61 +616,75 @@ inline polynomial<T> operator - (const polynomial<T>& a, const polynomial<T>& b)
    result -= b;
    return result;
 }
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
+template <class T>
+inline polynomial<T> operator - (polynomial<T>&& a, const polynomial<T>& b)
+{
+   a -= b;
+   return a;
+}
+template <class T>
+inline polynomial<T> operator - (const polynomial<T>& a, polynomial<T>&& b)
+{
+   b -= a;
+   return -b;
+}
+template <class T>
+inline polynomial<T> operator - (polynomial<T>&& a, polynomial<T>&& b)
+{
+   a -= b;
+   return a;
+}
+#endif
 
 template <class T>
 inline polynomial<T> operator * (const polynomial<T>& a, const polynomial<T>& b)
 {
-   polynomial<T> result(a);
-   result *= b;
+   polynomial<T> result;
+   result.multiply(a, b);
    return result;
 }
 
 template <class T, class U>
-inline polynomial<T> operator + (const polynomial<T>& a, const U& b)
+inline polynomial<T> operator + (polynomial<T> a, const U& b)
 {
-   polynomial<T> result(a);
-   result += b;
-   return result;
+   a += b;
+   return a;
 }
 
 template <class T, class U>
-inline polynomial<T> operator - (const polynomial<T>& a, const U& b)
+inline polynomial<T> operator - (polynomial<T> a, const U& b)
 {
-   polynomial<T> result(a);
-   result -= b;
-   return result;
+   a -= b;
+   return a;
 }
 
 template <class T, class U>
-inline polynomial<T> operator * (const polynomial<T>& a, const U& b)
+inline polynomial<T> operator * (polynomial<T> a, const U& b)
 {
-   polynomial<T> result(a);
-   result *= b;
-   return result;
+   a *= b;
+   return a;
 }
 
 template <class U, class T>
-inline polynomial<T> operator + (const U& a, const polynomial<T>& b)
+inline polynomial<T> operator + (const U& a, polynomial<T> b)
 {
-   polynomial<T> result(b);
-   result += a;
-   return result;
+   b += a;
+   return b;
 }
 
 template <class U, class T>
-inline polynomial<T> operator - (const U& a, const polynomial<T>& b)
+inline polynomial<T> operator - (const U& a, polynomial<T> b)
 {
-   polynomial<T> result(a);
-   result -= b;
-   return result;
+   b -= a;
+   return -b;
 }
 
 template <class U, class T>
-inline polynomial<T> operator * (const U& a, const polynomial<T>& b)
+inline polynomial<T> operator * (const U& a, polynomial<T> b)
 {
-   polynomial<T> result(b);
-   result *= a;
-   return result;
+   b *= a;
+   return b;
 }
 
 template <class T>
@@ -636,19 +694,17 @@ bool operator == (const polynomial<T> &a, const polynomial<T> &b)
 }
 
 template <typename T, typename U>
-polynomial<T> operator >> (const polynomial<T>& a, const U& b)
+polynomial<T> operator >> (polynomial<T> a, const U& b)
 {
-    polynomial<T> result(a);
-    result >>= b;
-    return result;
+    a >>= b;
+    return a;
 }
 
 template <typename T, typename U>
-polynomial<T> operator << (const polynomial<T>& a, const U& b)
+polynomial<T> operator << (polynomial<T> a, const U& b)
 {
-    polynomial<T> result(a);
-    result <<= b;
-    return result;
+    a <<= b;
+    return a;
 }
 
 // Unary minus (negate).


include/boost/math/tools/polynomial.hpp
```diff
@@ -290,6 +290,10 @@ class polynomial :
           m_data.push_back(point);
    }
 
+   // move:
+   polynomial(polynomial&& p)
+      : m_data(std::move(p.m_data)) { }
+
```
lakshayg: Is there a way to detect if C++11 is supported and disable this constructor if it is not?

lakshayg: Just found out about BOOST_NO_RVALUE_REFERENCES in Boost.Config.

jzmaddock:
On 14/08/2017 11:13, Lakshay Garg wrote:
>
> *@lakshayg* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In include/boost/math/tools/polynomial.hpp 
> <https://github.com/boostorg/math/pull/79#discussion_r132916570>:
>
> > @@ -290,6 +290,10 @@ class polynomial :
>             m_data.push_back(point);
>      }
>   
> +   // move:
> +   polynomial(polynomial&& p)
> +      : m_data(std::move(p.m_data)) { }
> +
>
> Just found out about BOOST_NO_RVALUE_REFERENCES in Boost.Config.
>

Nod, I would prefer to keep C++03 support if possible, some more 
comments will follow soon....

---
http://www.avg.com

lakshayg:
@jzmaddock I amended the commit a few minutes ago. It now checks if Rvalue references are supported.

include/boost/math/tools/polynomial.hpp
```diff
@@ -558,59 +562,52 @@ class polynomial :
...
 
 template <class T>
-inline polynomial<T> operator + (const polynomial<T>& a, const polynomial<T>& b)
+inline polynomial<T> operator + (polynomial<T> a, const polynomial<T>& b)
 {
-   polynomial<T> result(a);
-   result += b;
-   return result;
+   a += b;
+   return a;
```
jzmaddock: I'm not completely certain about this, but I believe we need a:

return std::move(a);

Here to be sure that we don't make an accidental copy.  If the function is expanded inline it will make not one jot of difference of course, but IMO we can't rely on that.

To make it C++03 compatible we would need something like:

#ifndef BOOST_NO_RVALUE_REFERENCES
#  define BOOST_MATH_MOVE(x) std::move(x)
#else 
#  define BOOST_MATH_MOVE(x) x
#endif  

Then 

return BOOST_MATH_MOVE(a);

jzmaddock: Sorry about the messed up formatting, haven't figured this stupid interface out yet :(

include/boost/math/tools/polynomial.hpp
```diff
@@ -290,6 +290,12 @@ class polynomial :
           m_data.push_back(point);
    }
 
+#if !defined(BOOST_NO_RVALUE_REFERENCES)
```
glenfe: That is the deprecated macro; use BOOST_NO_CXX11_RVALUE_REFERENCES

include/boost/math/tools/polynomial.hpp
lakshayg: I benchmarked this code and this seems to be working fine. If you still insist then I will go ahead and change it to `return std::move(a);`

include/boost/math/tools/polynomial.hpp
lakshayg: Done! Thanks.

include/boost/math/tools/polynomial.hpp
```diff
@@ -558,59 +562,52 @@ class polynomial :
...
-   polynomial<T> result(a);
-   result -= b;
-   return result;
+   a -= b;
+   return a;
 }
 
 template <class T>
-inline polynomial<T> operator * (const polynomial<T>& a, const polynomial<T>& b)
+inline polynomial<T> operator * (polynomial<T> a, const polynomial<T>& b)
```
jzmaddock: I think we need to be careful about changing the parameter type to an object type rather than an lvalue-ref.

We have 4 possibilities here:

2 l-value refs: need a new temporary for the result.
left r-value ref, right l-value ref: reuse left argument for the result.
right r-value ref, left lvalue ref: reuse right argument for the result (except division!)
both rvalue-refs: reuse either for the result.

Your change handles the first two but not the last two, and in order to avoid overload ambiguities, I *think* the only way to do this right is to provide all 4 overloads.  I realise that sucks, but it does work well (or at least it did for Boost.Multiprecision when I did this there).

For multiplication, I think we should also rewrite the *= operator to carry out 2-arg multiplication since there isn't really an efficient in-place multiplication, something like:

`
void multiply(const Polynomial& a, const polynomial& b)
{
  // allocates temporary vector for result, then swaps it into *this.
}
`

The code would be much the same as now, but I think we can avoid at least one copy doing this when we're doing 2 arg multiplication and both in-place  and 2 arg multiplication can simply call the above routine?

include/boost/math/tools/polynomial.hpp
lakshayg: I think we don't need to `return std::move(...)`. Please see: https://youtu.be/AKtHxKJRwp4?t=38m3s and let me know if I interpreted something incorrectly.

include/boost/math/tools/polynomial.hpp
lakshayg: I just created the overloads

```
operator+(const polynomial<T>&, const polynomial<T>&)
operator+(const polynomial<T>&, polynomial<T>&&)
operator+(polynomial<T>&&, const polynomial<T>&)
operator+(polynomial<T>&&, polynomial<T>&&)
```
This causes the following code to use the overload with both const lvalue refs
```
polynomial<double> p{1, 2, 3};
polynomial<double> q{1, 2, 3};
auto sum = p + q + p;
```

Which in my opinion is undesirable. Please advise. @jzmaddock

lakshayg: I could not understand what you said about `operator *=`, could you please explain?

jzmaddock:
> I think we don't need to |return std::move(...)|. Please see: 
> https://youtu.be/AKtHxKJRwp4?t=38m3s and let me know if I interpreted 
> something incorrectly.
>

No you're quite right!

---
http://www.avg.com

lakshayg:
@jzmaddock Is this ready to be merged then? Or do you see any further improvements which can be made?

jzmaddock:
> I think we don't need to |return std::move(...)|. Please see: 
> https://youtu.be/AKtHxKJRwp4?t=38m3s and let me know if I interpreted 
> something incorrectly.
>

Forgot to say.... as long as we're returning a named value.  If we 
condense things down to:

T foo(T a, T b)
{
   return a += b;
}

Then we do need a move.

---
http://www.avg.com

lakshayg:
Yes right, but I prefer to keep it this way. It is one extra line but looks cleaner to the eye.

jzmaddock:
On 14/08/2017 13:18, Lakshay Garg wrote:
>
> *@lakshayg* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In include/boost/math/tools/polynomial.hpp 
> <https://github.com/boostorg/math/pull/79#discussion_r132936753>:
>
> >  }
>   
>   template <class T>
> -inline polynomial<T> operator * (const polynomial<T>& a, const polynomial<T>& b)
> +inline polynomial<T> operator * (polynomial<T> a, const polynomial<T>& b)
>
> I just created the overloads
>
> |operator+(const polynomial<T>&, const polynomial<T>&) operator+(const 
> polynomial<T>&, polynomial<T>&&) operator+(polynomial<T>&&, const 
> polynomial<T>&) operator+(polynomial<T>&&, polynomial<T>&&) |
>
> This causes the following code to use the overload with both const 
> lvalue refs
>
> |polynomial<double> p{1, 2, 3}; polynomial<double> q{1, 2, 3}; auto sum 
> = p + q + p; |
>
> Which in my opinion is undesirable. Please advise. @jzmaddock 
> <https://github.com/jzmaddock>
>

You should see the 2 lvalue ref overload called first - and that's 
correct, at some point a new temporary needs to be created - and then 
the second operator+ should call the rvalue+lvalue overload - and that 
one can just call operator += on the rvalue and return it.

---
http://www.avg.com

jzmaddock:
On 14/08/2017 13:29, Lakshay Garg wrote:
>
> *@lakshayg* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In include/boost/math/tools/polynomial.hpp 
> <https://github.com/boostorg/math/pull/79#discussion_r132938545>:
>
> >  }
>   
>   template <class T>
> -inline polynomial<T> operator * (const polynomial<T>& a, const polynomial<T>& b)
> +inline polynomial<T> operator * (polynomial<T> a, const polynomial<T>& b)
>
> I could not understand what you said about |operator *=|, could you 
> please explain?
>

I'm suggesting we reimplement like this:

inline polynomial<T> operator * (const polynomial<T>& a, const polynomial<T>& b)

{
   polynomial<T> t;  // no need to copy anything, cheap to construct 
zero order polynomial allocates nothing.
   t.multiply(a, b);
   return t;
}

Likewise operator *= changes to:

polynomial<T>& operator*=(const polynomial& a)
{
   multiply(*this, a);
   return *this;
}

And the new "multiply" member function is virtually the same as the 
current operator*=, but takes 2 arguments, rather than replying on *this 
for one of the input vectors.

Currently we have:

template <class T>
inline polynomial<T> operator * (const polynomial<T>& a, const 
polynomial<T>& b)
{
    polynomial<T> result(a);  // copies a, allocates memory.
    result *= b;   // reallocates more memory.
    return result;
}

Which does 2 allocations, where as I'm suggesting a little re-factoring 
would only do 1.

Best, John.

---
http://www.avg.com

include/boost/math/tools/polynomial.hpp
```diff
@@ -290,6 +290,12 @@ class polynomial :
           m_data.push_back(point);
```
lakshayg: Should we remove the `explicit` keyword from this constructor? I believe that it would provide us with opportunities to simplify the code and shave off some lines.

lakshayg:
Should we remove the `explicit` keyword from [the constructor][1]

```
    template <class U>
    explicit polynomial(const U& point)
    {
        if (point != U(0))
           m_data.push_back(point);
    } 
```
it should provide us with opportunities to simplify the code by removing some overloads like [this][2].

[1]: https://github.com/boostorg/math/pull/79/files/f0e7199034c0a208f3ef2ac20d4704dac201db9e#diff-46985bfe357deae42e1778baec521dfaR287
[2]: https://github.com/boostorg/math/pull/79/files/f0e7199034c0a208f3ef2ac20d4704dac201db9e#diff-46985bfe357deae42e1778baec521dfaR626

jzmaddock:
On 14/08/2017 21:20, Lakshay Garg wrote:
>
> Should we remove the |explicit| keyword from the constructor 
> <https://github.com/boostorg/math/pull/79/files/f0e7199034c0a208f3ef2ac20d4704dac201db9e#diff-46985bfe357deae42e1778baec521dfaR287>
>
> |template <class U> explicit polynomial(const U& point) { if (point != 
> U(0)) m_data.push_back(point); } |
>
> it should provide us with opportunities to simplify the code by 
> removing some overloads like this 
> <https://github.com/boostorg/math/pull/79/files/f0e7199034c0a208f3ef2ac20d4704dac201db9e#diff-46985bfe357deae42e1778baec521dfaR626>.
>

No: the explicit keyword is there to express semantics, we should only 
remove it if we *really* want to change the semantics, not to make 
coding easier.

Plus, if we rely on implicit conversion to polynomial<T> for the scalar 
operators these then become less efficient: it's far more efficient to 
multiply or add a scalar than to construct a polynomial temporary 
(memory allocation) and then go through an unnecessary loop in the 
operator implementation.

---
http://www.avg.com

include/boost/math/tools/polynomial.hpp
```diff
@@ -354,6 +360,12 @@ class polynomial :
    }
 
    // operators:
+   polynomial& operator =(polynomial p)
```
jzmaddock: Doesn't this create an unnecessary extra copy in C++03 land?

include/boost/math/tools/polynomial.hpp
```diff
@@ -572,45 +609,61 @@ inline polynomial<T> operator - (const polynomial<T>& a, const polynomial<T>& b)
...
-   polynomial<T> result(a);
-   result *= b;
+   polynomial<T> result;
+   result.multiply(a, b);
    return result;
 }
 
 template <class T, class U>
-inline polynomial<T> operator + (const polynomial<T>& a, const U& b)
+inline polynomial<T> operator + (polynomial<T> a, const U& b)
```
jzmaddock: As with the assignment operator, doesn't this create an extra full copy in C++03?

include/boost/math/tools/polynomial.hpp
lakshayg: We have two cases:
```cpp
polynomial<double> p{1, 2, 3, 4};

// case 1:
polynomial<double> q;
q = p; // this is handled optimally in both c++03 and c++11 since q has no memory allocated to it

// case 2:
polynomial<double> r{5, 6, 7, 8, 9, 1, 2, 3, 4};
r = p; // this is bad. We could have prevented a copy by overwriting r.m_data
```

Will fix it shortly. I'm afraid the code wil turn into a mess of `#if`s./`#ifndef`s. Can you suggest a way we can prevent this from happening?

include/boost/math/tools/polynomial.hpp
lakshayg: From what I understood, here U is some type other than polynomial (am I right?). In that case, don't we need to allocate memory for addition anyway?

jzmaddock:
On 15/08/2017 12:46, Lakshay Garg wrote:
>
> *@lakshayg* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In include/boost/math/tools/polynomial.hpp 
> <https://github.com/boostorg/math/pull/79#discussion_r133171075>:
>
> >     return result;
>   }
>   
>   template <class T, class U>
> -inline polynomial<T> operator + (const polynomial<T>& a, const U& b)
> +inline polynomial<T> operator + (polynomial<T> a, const U& b)
>
> From what I understood, here U is some type other than polynomial (am 
> I right?). In that case, don't we need to allocate memory for addition 
> anyway?
>

Nope, U is a scalar, so we just add b to the zeroth element in a. 
Likewise for multiplication we just multiply each element by the scalar.

There is a subsequent normalization step which may in some (not common) 
cases result in a change in size of the polynomial, but in most cases no 
memory allocation takes place.

---
http://www.avg.com

jzmaddock:
> Will fix it shortly. I'm afraid the code wil turn into a mess of 
> |#if|s./|#ifndef|s. Can you suggest a way we can prevent this from 
> happening?
>

Well you're not going to like it, but separate overloads for

operator=(const polynomial&)
operator=(polynomial&&)

would do the trick?

---
http://www.avg.com

include/boost/math/tools/polynomial.hpp
lakshayg: @jzmaddock I am a little confused by your comment. Here is what I think:

1. This function takes a polynomial and scalar, adds them and returns the result as a new polynomial
2. Suppose that we keep the signature I suggested:
  * If the input is a lvalue, a copy is made due to passing by value, scalar is added (does not allocate any memory) and we return the polynomial i.e. 1 memory allocation
  * If the input is a rvalue, the move constructor is invoked when passing and no copy is made. In the function body, we add using += and return the polynomial. This involves no copying.
3. I don't think we can do any better in terms of memory allocation by adding separate overloads for lvalue and rvalue.

Please let me know your views on this.

jzmaddock:
On 16/08/2017 12:37, Lakshay Garg wrote:
>
> *@lakshayg* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In include/boost/math/tools/polynomial.hpp 
> <https://github.com/boostorg/math/pull/79#discussion_r133424807>:
>
> >     return result;
>   }
>   
>   template <class T, class U>
> -inline polynomial<T> operator + (const polynomial<T>& a, const U& b)
> +inline polynomial<T> operator + (polynomial<T> a, const U& b)
>
> @jzmaddock <https://github.com/jzmaddock> I am a little confused by 
> your comment. Here is what I think:
>
>  1. This function takes a polynomial and scalar, adds them and returns
>     the result as a new polynomial
>  2. Suppose that we keep the signature I suggested:
>
>   * If the input is a lvalue, a copy is made due to passing by value,
>     scalar is added (does not allocate any memory) and we return the
>     polynomial i.e. 1 memory allocation
>   * If the input is a rvalue, the move constructor is invoked when
>     passing and no copy is made. In the function body, we add using +=
>     and return the polynomial. This involves no copying.
>
>  3. I don't think we can do any better in terms of memory allocation
>     by adding separate overloads for lvalue and rvalue.
>

Well I think this serves me right for trying to review code of an evening!

You're quite right of course, apologies for the noise!  John.

---
http://www.avg.com

lakshayg:
@jzmaddock all tests have passed. Do you see anything else that needs to be done?

jzmaddock:
On 17/08/2017 11:23, Lakshay Garg wrote:
>
> @jzmaddock <https://github.com/jzmaddock> all tests have passed. Do 
> you see anything else that needs to be done?
>

I don't think so, but let me look at this properly over the weekend: as 
you may have noticed, this class was originally an internal detail "just 
good enough to do what needed to be done", and it sort of leaked out due 
to public demand.  I'd like to find some time to sit down and read 
through the whole file properly and see what else if anything needs to 
be done, I should probably add to the tests as well.

Many thanks again for taking this on!

---
http://www.avg.com

pabristow:
I’ve viewed this interchange tangentially.

Would it be useful to briefly summarise the conclusions for a ‘implementation’ section in the documentation?

It seems complicated enough to me that other may find it helpful in future (and in implementing other functions too?).

Paul

lakshayg:
Would it be useful to briefly summarise the conclusions for a ‘implementation’ section in the documentation?
It seems complicated enough to me that other may find it helpful in future (and in implementing other functions too?).

​
Hey Paul. Many thanks for the suggestion. I'm not sure if I understand it completely.
​

​
​Are you suggesting that we should have a section called "implementation" in the documentation summarizing the implementation of this class? The suggestion sounds good but I cannot think of anyone who would benefit from such a documentation.

As far as a user of this library is considered, there is nothing that the user needs to know about this change. They can continue to use the library in the same manner and if they use C++11 and above, they should see improvement in the performance of some algorithms due to fewer memory allocations.

Talking about a library dev, move semantics are fairly common technique used in modern c++ code so it should be fairly easy for him/her to grasp them and start coding. Having a section for explaining the implementation will more-or-less be  like having a tutorial on move semantics.

Please let me know if you feel differently and would like to have an implementation section.

On a different note, I will see if there are any places which can be updated in the documentation and update them over this weekend.

Lakshay

pabristow:
I’ve only note that you and John seemed to have quite a long discussion about it ;-)

I didn’t follow in detail but what may be obvious to you may not be to future maintainers.
Perhaps a reference would suffice?

(I’ve even found that implementation notes have helped me years later when I have even forgotten that I wrote it!)

Paul

jzmaddock:
Interestingly when I benchmarked these changes, comparing new and old, it made very little difference with msvc-14.1.  My suspicion is that the compiler is so good at inlining the operators that the copies/moves just disappear anyway...  it'll help a lot in debug mode of course.

lakshayg:
On 20 August 2017 at 23:39, jzmaddock <notifications@github.com<mailto:notifications@github.com>> wrote:

Interestingly when I benchmarked these changes, comparing new and old, it made very little difference with msvc-14.1. My suspicion is that the compiler is so good at inlining the operators that the copies/moves just disappear anyway... it'll help a lot in debug mode of course.

​I benchmarked these changes on GCC and it indeed showed a significant speedup even after using -O3​
​.

PR #80, opened 2017-08-18
In the CDF/PDF for the t-distribution, check for df=+infinity, not x

sboukortt:
A few elements make me suspect that `x` was a typo:
- the comment says “+infinity”, but `x` could be negative; only `df` is necessarily positive;
- `detail::check_x` a few lines earlier prevents `x` from being infinite anyway.

However, I don’t think that it should affect the result, as the case where `df` is infinite is “caught” by `if (df > limit)`. In addition, if infinite values of `x` were allowed (it is not really clear to me why they are not?), then the return value would indeed be the same as for the normal distribution (0 or 1). The current code is merely suboptimal, not incorrect.

### DIFF
--- a/include/boost/math/distributions/detail/common_error_handling.hpp
+++ b/include/boost/math/distributions/detail/common_error_handling.hpp
@@ -113,6 +113,26 @@ inline bool check_x(
 } // bool check_x
 
 template <class RealType, class Policy>
+inline bool check_x_not_NaN(
+  const char* function,
+  RealType x,
+  RealType* result,
+  const Policy& pol)
+{
+  // Note that this test catches only NaN.
+  // Some distributions permit x to be infinite, leaving this test to catch any NaNs.
+  // See Normal, Logistic, Laplace and Cauchy for example.
+  if ((boost::math::isnan)(x))
+  {
+    *result = policies::raise_domain_error<RealType>(
+      function,
+      "Random variate x is %1%, but must be finite or + or - infinity!", x, pol);
+    return false;
+  }
+  return true;
+} // bool check_x_not_NaN
+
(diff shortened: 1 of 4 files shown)


sboukortt:
Ah, I see. Thanks for the clarification.

For the normal distribution, [support for infinity seems to be implemented like this](https://github.com/boostorg/math/blob/70e87cb407aba873498fe9db92c0c1ee24c04d32/include/boost/math/distributions/normal.hpp#L171). Is that code subject to the caveat that you mentioned, then?

pabristow:
I believe that normal does support infinite values OK (if the platform and real type support infinity of course).  Please tell me if you find it doesn’t ☹

(I lost the will to continue with all the many other distributions).

Paul

sboukortt:
Hi again,
Out of curiosity, does anything specific still need to be done before this can be merged?

pabristow:
No – I think it will make its way into master and then into the next release 1.66

Thanks for reporting this and providing a fix.

jzmaddock:
Apologies for jumping in late here, you're correct that this needs fixing, but IMO the fix isn't quite right:

* Later on, when df is greater than 1/eps the normal approximation is used anyway, so this check is entirely redundant since infinite degree of freedom is definitely > 1/eps.
* We could deal with infinite x here if we wished, and simply return 0 (PDF or CDF at -INF) or 1 (CDF at +INF).

pabristow:
OK – I’ll take a closer look at this.

but I’m also concerned that the test matrix isn’t doing the tests.

Paul

pabristow:
I've made a fuler revision of this allowing infinite argument and random variate, and updated docs to say so.
   1527667..9921c27  develop -> develop

PR #84, opened 2017-08-23
Typo removal.

NAThompson:


jzmaddock:
Thanks for that, I've simplified the tanh_sinh::integrate declarations in the docs a bit - the rationale is that the whole decltype overload/metaprogramming stuff is an implementation detail - it could (maybe should?) have been hidden inside the integrate method, it was just easier to add it at the top level.  Also reformatted a bit for shorter line length / easier reading.

PR #85, opened 2017-08-23
Readme

NAThompson:
This adds a README.rst so I can remember the list of steps to get code into boost/math.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,13 +62,11 @@ The following instructions will hopefully help the user get started.
         boost$ ./b2
 
 - To build the documentation, you must install some dependencies:
-
     .. code:: bash
 
         boost/libs/math/doc$ sudo apt install xsltproc docbook-xsl docbook-xml
         boost/libs/math/doc$ ../../../b2 release
-
-    This will build the `boost::math` documentation into `doc/index.html` and hence can be viewed with any browser. In addition, `doc/math.pdf` should contain the same information.
+This will build the `boost::math` documentation into `doc/index.html` and hence can be viewed with any browser. In addition, `doc/math.pdf` should contain the same information.
 
 - Write your feature. For the purposes of this contributing guide, we will test if a floating point number is an integer. In this case, place the file `is_float_int.hpp` in (say) the `include/boost/math/tools` directory.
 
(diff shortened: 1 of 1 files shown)


jzmaddock:
This is a good idea, however, how much does it duplicate: http://www.boost.org/doc/libs/1_64_0/libs/math/doc/html/math_toolkit/special_tut/special_tut_impl.html ?  If possible I'd like to keep all the "how to contribute" stuff in one place.

I think we should also add an extended description of what *is* in the library.

And finally, the existing (and rather old) TODO list here: http://www.boost.org/doc/libs/1_64_0/libs/math/doc/html/math_toolkit/issues.html  needs updating with your suggestions.

Sorry!

NAThompson:
I guess we have no clue which page would be read more, a README or the boost.org docs. I personally find the README to be more accessible, but you're right that we need a 'canonical point of authority' for each piece of information.

I could move this information to the qbk files, and simply display a CI badge with the readme, or we could move the TODO list in the qbk to the 'Contributing' section of the README. We might get more contributions and updates to the 'Contributing' and 'TODO' if it's in RST. . .

Do you have a preference?

jzmaddock:
On 24/08/2017 18:17, Nick wrote:
>
> I guess we have no clue which page would be read more, a README or the 
> boost.org docs. I personally find the README to be more accessible, 
> but you're right that we need a 'canonical point of authority' for 
> each piece of information.
>
> I could move this information to the qbk files, and simply display a 
> CI badge with the readme, or we could move the TODO list in the qbk to 
> the 'Contributing' section of the README. We might get more 
> contributions and updates to the 'Contributing' and 'TODO' if it's in 
> RST. . .
>
> Do you have a preference?
>

I think I'd prefer the readme to be a quick-ish overview with links to 
the actual docs on the website, otherwise it sort of duplicates what's 
already there.  The only thing I guess is that the readme could be more 
quickly updated, while the online docs refer to the last release.  I 
doubt it makes a huge difference though?

---
http://www.avg.com

jeking3:
Shouldn't this be named README.md so it shows up at the top level of a github view?  Also suggest adding a LICENSE file - you can copy one from the top level of https://github.com/jeking3/boost-ci to this repository so github recognizes this repository as a BSL-1.0 licensed project.

NAThompson:
> Shouldn't this be named README.md so it shows up at the top level of a github view?

github supports both .rst and .md format.

jeking3:
For me, the most important information in the README is the project status so that sounds reasonable.   If you look at the DateTime, Format, or Uuid projects on github that's the information I like to provide.

pabristow:
readme  https://github.com/boostorg/format,  for example, …   Look good to me.

I generally like the more info section,

but would links to the main boost.org site

and the boost general (not developer) lists for questions be good too?

And I’d like to see a direct link to the release docs index.html for the project, and perhaps another link to the develop branch?

“Note that by submitting patches you agree to license your modifications under the Boost Software License, Version 1.0<http://www.boost.org/LICENSE_1_0.txt>.”  is a good addition.

I’m less convinced by the re-iteration of the directories structure as this duplicates the normal github display?

License is good too, if redundant info, but makes it immediately clear.

PR #87, opened 2017-09-06
Merge Gauss Kronrod code into develop.

jzmaddock:


NAThompson:
The problem with the Chebyshev code is that we must have a discrete cosine transform to make it useful. The use case is to take a complicated function, and project it onto the Chebyshev polynomials, which have simple formulas for evaluation, integration, and differentiation, giving a numerical integrator, numerical differentiator, and interpolator. The projection requires the DCT to be efficient.

So I guess the question is: Does boost math need a DCT?

NAThompson:
Note: Trefethen's 'Chebfun' package is the canonical implementation of these algorithms.

PR #88, opened 2017-09-13
Chebyshev

NAThompson:
This is a first draft of the Chebyshev transform. Since the Chebyshev transform relies on the discrete cosine transform, I have put a dependency of FFTW3 in `boost/math/special_functions/chebyshev_transform.hpp`, and hence is must be linked with the flags `-lfftw3` if you use the double precision template instantiation, `-lfftw3f` for the float precision template instantiation, and `-lfftw3l` for the long double template instantiation. Ostensibly it could work in `__float128` precision (see [here](http://www.fftw.org/fftw3_doc/Precision.html) but I was unable to figure out how to make it compile if `__float128` isn't supported on the platform at all.

I used a C++17 `if constexpr` to simulate template metaprogramming on the `fftw_plan`. I think this is ugly and I think there's a better way, so if someone could take a look and see what can be done I'd appreciate it!

The numerical derivative works well, but does not have asymptotically optimal scaling, and I also need to be able expand the number of places where the function can be evaluated in case no coefficients are smaller than the tolerance. (Right now, they are only truncated.)

So, more work to do, as always, but I think this is enough to get the discussion going esp. with regard to the dependency on FFTW.


include/boost/math/special_functions/chebyshev_transform.hpp
```diff
@@ -0,0 +1,167 @@
...
+#include <boost/math/special_functions/chebyshev.hpp>
+
+namespace boost { namespace math {
+
+template<class Real>
+class chebyshev_transform
+{
+public:
+    template<class F>
+    chebyshev_transform(const F& f, Real a, Real b, Real tol=500*std::numeric_limits<Real>::epsilon()): m_a(a), m_b(b)
```
jzmaddock: Nick, I think the simplest thing to do here is to add a final enable_if argument:

`chebyshev_transform(const F& f, Real a, Real b, Real tol=500*std::numeric_limits<Real>::epsilon(), typename std::enable_if<std::is_same<double, Real>::value>::type* = nullptr)`

Then this constructor can only be called if type Real is a double, repeat for float and long double and you're done and the if constexpr can go.  It does mean some code duplication, but it's not a huge function so probably not an issue?

test/chebyshev_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+        BOOST_CHECK_CLOSE_FRACTION(chebyshev_t_prime(4, x), 16*x*(2*x*x - 1), tol);
+        // This one makes the tolerance have to grow too large; the Chebyshev recurrence is more stable than naive polynomial evaluation anyway.
+        //BOOST_CHECK_CLOSE_FRACTION(chebyshev_t_prime(5, x), 5*(4*x*x*(4*x*x - 3) + 1), tol);
+        x += 1/static_cast<Real>(1<<7);
+    }
+}
+
+
+template<class Real>
+void test_sin_chebyshev_transform()
```
jzmaddock: Nick, can we as with the header, split this into 2 tests: one that depends on fftw3 and one that doesn't?  Don't worry about the Jamfile, I can hook the one with the external dependency to only run when fftw3 is available.

Other than that, this looks pretty much ready to go to me!

NAThompson: Fixed!

include/boost/math/special_functions/chebyshev_transform.hpp
NAThompson: Yup, will try to get this done!

NAThompson:
Ok, now I just need to make the constructor adaptive so that if we don't pick a large enough number of Chebyshev points, we can expand it. Also should consider doing an indefinite integration routine, using the equation given [here](https://en.wikipedia.org/wiki/Chebyshev_polynomials#Differentiation_and_integration)

NAThompson:
I couldn't get the `enable_if` to work; kept on getting this error once I added the other constructors:

    /include/boost/math/special_functions/chebyshev_transform.hpp:24:49: 
    error: no type named 'type' in 'std::__1::enable_if<false, void>'; 'enable_if' cannot be used to disable this declaration
    typename std::enable_if<std::is_same<double, Real>::value>::type* = nullptr)

But presumably this is a small problem; other than this I think it's ready for review.

PR #96, opened 2017-12-26
Factor integer

NAThompson:
This is a first pass at integer factorization using the Pollard rho algorithm and trial division. These two algorithms should allow us to factor basically all 64 bit integers, which is more than sufficient for my purpose, which is to put the pieces in place to create an FFT.

Even libraries implementing the GNFS (such as [msieve](https://sourceforge.net/projects/msieve/)) first use trial division and Pollard rho on all inputs before unleashing the GNFS on them, so IMO there is no shame in not providing a sieve at this juncture.

Still a ways to go here, including replacing Floyd's cycle finding with Brent's cycle finding algorithm, using multiple iterates before computing the gcd, and increasing the usability by perhaps supporting cancellation. But this is a lot of fun, so look around and see if there's some nice improvements that can be made.


NAThompson:
I'm going to move this to boost integer; will close soon.

PR #109, opened 2018-02-10
Put explicit deprecation warning on boost/math/common_facto…

NAThompson:
…r.hpp


jzmaddock:
Good idea, however we've just added  BOOST_HEADER_DEPRECATED(A) for this which will give better error messages - can you change it over?  Usage is 

```
BOOST_HEADER_DEPRECATED("<boost/math/common_factor.hpp>")
```

Thanks!

PR #115, opened 2018-02-16
[ci skip] Remove gcd tests from boost.math

NAThompson:


NAThompson:
. . .  looks like this will take one more commit, my bad.

PR #117, opened 2018-03-07
Use constexpr when compiling with clang

CaseyCarter:
Clang compilers do not define `__clang`, they define `__clang__`. (https://godbolt.org/g/R7gX4h)


CaseyCarter:
Travis failure is unrelated to the change; it's a timeout.

PR #119, opened 2018-03-17
Remove pile of warnings so that the build log is easier to search thr…

NAThompson:
…ough.

### DIFF
--- a/test/compile_test/test_compile_result.hpp
+++ b/test/compile_test/test_compile_result.hpp
@@ -52,7 +52,10 @@ template <class T1, class T2>
 inline void check_result_imp(T1, T2)
 {
    // This is a static assertion that should always fail to compile...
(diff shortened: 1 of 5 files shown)


test/compile_test/test_compile_result.hpp
```diff
@@ -52,7 +52,10 @@ template <class T1, class T2>
 inline void check_result_imp(T1, T2)
 {
    // This is a static assertion that should always fail to compile...
+#pragma GCC diagnostic push
```
jzmaddock: Do we know which GCC version first started supporting this?

NAThompson: 4.8: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=33255

PR #121, opened 2018-04-13
Add derivative of barycentric rational interpolation.

NAThompson:
Quick feature roundout of barycentric rational interpolation, based on:

http://www.ams.org/journals/mcom/1986-47-175/S0025-5718-1986-0842136-8/S0025-5718-1986-0842136-8.pdf


NAThompson:
The only build error was a network problem, so I think this can be merged.

PR #124, opened 2018-05-01
Real->{Real, Complex} integrand support for Gauss-Kronrod

NAThompson:
Couldn't get it to work with `cpp_dec_float`, but everything else works.


jzmaddock:
This fails with expression templates enabled when the function being integrated returns an expression template, for example in our tests we have:

```    auto f = [](const Real& x)
    {
       return 5*x + 7;
    };
```

I guess it would be nice to retain this, will look at it later.

NAThompson:
I'm not sure it's a big priority. Plus the tests are now consistently using the `auto [](Real t)->Real` syntax.

PR #125, opened 2018-05-03
Complex valued integrand support for exp-sinh quadrature

NAThompson:


NAThompson:
Well, that bloodbath of a build is totally inexplicable to me. Error of 10^-25? Not consistent with a bad parse, not consistent with a stray long double. Not even close to the 128 bit epsilon of 10^-32.

jzmaddock:
The bug is in

```
template<class Real, class Policy>
void exp_sinh_detail<Real, Policy>::init(const mpl::int_<4>&)
```
everything works if I disable it.  But I don't see the error yet...

jzmaddock:
OK, this is nearly fixed: appveyor failures are due to serialization (I've filed a bug report), travis has one failure: https://travis-ci.org/boostorg/math/jobs/382315019

I don't have access to this platform - do you?   I'm not sure if it's a multiprecision or a type_traits issue, or equally likely a Mac-clang bug.  If you have reproduce, can you change number.hpp:642:

```
#if BOOST_WORKAROUND(BOOST_MSVC, < 1900)
   template <class T>
#else
   template <class T, class = typename boost::disable_if_c<boost::is_constructible<T, self_type const&>::value || !boost::is_default_constructible<T>::value, T>::type>
#endif
```

To

```
#if 1//BOOST_WORKAROUND(BOOST_MSVC, < 1900)
   template <class T>
#else
   template <class T, class = typename boost::disable_if_c<boost::is_constructible<T, self_type const&>::value || !boost::is_default_constructible<T>::value, T>::type>
#endif
```

and then also try:

```
#if BOOST_WORKAROUND(BOOST_MSVC, < 1900)
   template <class T>
#else
   template <class T, class = typename boost::disable_if_c<std::is_constructible<T, self_type const&>::value || !std::is_default_constructible<T>::value, T>::type>
#endif
```

Thanks!

NAThompson:
Built against boost/multiprecision latest commit (cbbfff8bfd036faef0f03bb2e472abe003a6c4d5), and of course the latest boost.math.

```
$ clang++ --version
Apple LLVM version 9.1.0 (clang-902.0.39.1)
Target: x86_64-apple-darwin17.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Command line:

```
$ make
clang++ --std=c++14 -O2 -Wfatal-errors -I./include -I../multiprecision/include -I../container_hash/include test/test_cubic_b_spline.cpp -o main.x
```

which is as close as I can get to the config from the failing job log.

On this system, it succeeded without any changes.

When applying the first suggestion:

```
#if 1//BOOST_WORKAROUND(BOOST_MSVC, < 1900)
```

it compiled and the test passed.

When applying the second suggestion:

```
#if BOOST_WORKAROUND(BOOST_MSVC, < 1900)
   template <class T>
#else
   template <class T, class = typename boost::disable_if_c<std::is_constructible<T, self_type const&>::value || !std::is_default_constructible<T>::value, T>::type>
#endif
```

again, the compile and test succeeded.

I also did these tests using `b2`, via the command line:

```
➜  test git:(complex_exp_sinh) ✗ ../../../b2 -j3 -d+0 -q toolset=clang --compiler=clang++ --cxxflags=--std=c++14 misc
```

with the same results.

I thought maybe the compiler would be "similar enough" to reproduce, but I guess not.

jzmaddock:
Thanks Nick, good to know it works with clang-9 anyway.  I'll try and test this another way.

NAThompson:
BTW, is Apple Clang 8 relevant? I get the compiler auto-updated, so anyone on apple clang 8 is forcefully rejecting updates.

PR #127, opened 2018-05-31
Changes required to use Boost.Math on CUDA devices.

jzmaddock:


NAThompson:
@jzmaddock : Should this be closed? Looks like @mborland got it working.

PR #128, opened 2018-06-12
Support for complex-valued integrands in sinh-sinh quadrature.

NAThompson:
Finally found a relevant example of a complex-valued function which could be evaluated by quadrature over the entire real line, the Dirichlet eta function.


jzmaddock:
Thanks for this, one odd failure is unconnected to this commit, merging...

PR #131, opened 2018-06-25
Use <boost::math::policies::real> instead of <real>

kuhlenough:
When compiling for VxWorks 7 
Dinkum STL declares an inline real() function in the global namespace that clang can not distinguish from policies::real otherwise.

### DIFF
--- a/example/Jamfile.v2
+++ b/example/Jamfile.v2
@@ -91,7 +91,7 @@ test-suite examples :
    [ run normal_misc_examples.cpp : : : <exception-handling>off:<build>no  ]
    [ run owens_t_example.cpp  ]
    [ run policy_eg_1.cpp  ]
-   [ run policy_eg_10.cpp  ]
+   [ run policy_eg_10.cpp : : : <target-os>vxworks:<build>no ] # VxWorks' complex.h has conflicting declaration of real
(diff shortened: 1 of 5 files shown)


NAThompson:
The code changes look good to me, and the build failure appears to be a problem on Travis's side.

jzmaddock:
Ah... question: do we want to have to fully qualify everything in the _examples_  remember that code code gets extracted and injected into the docs.  My gut says probably not - std::real should be hidden away inside the std namespace and not cause a conflict here?

kuhlenough:
With Dinkum there's a real() in the C global namespace that's causing the problem, not std::real. I realize not many platforms are using Dinkum anymore, so the fix is specific.  If keeping the examples clean is paramount, an alternative is:  

```
#if BOOST_DINKUMWARE_STDLIB >= 650
#define <real> <boost::math::policies::real>
#endif
```

jzmaddock:
MSVC uses the Dinkum std lib and there are no issues.  So it's restricted to VxWorks.  It's also not std conforming to do that, as in:

```
std::complex<double> cd;
// do something
double d = real(cd); // oops
```

Unless `::real` is an alias for `std::real` ?

Either way my inclination is to just let the examples fail to build on that one platform, and fix up just the tests.

PR #146, opened 2018-08-30
PR for integration testing of LambertW

jzmaddock:


NAThompson:
I didn't fix the linker error; I think it is due to test/Jamfile.v2 line 426, which should be removed.

jzmaddock:
>I didn't fix the linker error; I think it is due to test/Jamfile.v2 line 426, which should be removed.

There was some messed up preprocessor logic that I think was responsible.  Fingers crossed!

PR #147, opened 2018-09-03
Kummer-M

NAThompson:
This is my first special function commit, and frankly, it's not an area of mathematics in which I have a lot of expertise, so I'm hoping gauge if anyone is interested in going forward with this to help me out. I think using the confluent hypergeometric function will be much easier than the 2F1,  so it might be a nice starting point to move into the "higher analysis" of generalized hypergeometrics like Appell series.

### DIFF
--- a/doc/math.qbk
+++ b/doc/math.qbk
@@ -616,6 +616,10 @@ and as a CD ISBN 0-9504833-2-X  978-0-9504833-2-0, Classification 519.2-dc22.
 [/ Lambert W function]
 [include sf/lambert_w.qbk]
 
+[section:hypergeometrics Hypergeometric]
+[include sf/kummer_m.qbk]
+[endsect]
+
 [section:zetas Zeta Functions]
 [include sf/zeta.qbk]
 [endsect]
--- a/doc/sf/kummer_m.qbk
+++ b/doc/sf/kummer_m.qbk
@@ -1,0 +1,58 @@
+[section:kummer_m The Kummer-M function]
+
+[h4 Synopsis]
+
+``
+#include <boost/math/special_functions/kummer_m.hpp>
+``
+
+   namespace boost{ namespace math{
+
+   template <class T1, class T2, class T3>
+   ``__sf_result`` kummer_m(T1 a, T2 b, T3 x);
+
+   }} // namespaces
+
+The return type of these functions is computed using the __arg_promotion_rules:
+the return type is `double` if T is an integer type, and T otherwise.
+
+
+Returns the [@https://en.wikipedia.org/wiki/Confluent_hypergeometric_function Kummer-M] of /x/.
+
+[h4 Discussion]
+
+The Kummer /M/ function `M(a,b,x)` denotes the non-singular solution of Kummer's confluent hypergeometric equation
+
+   xy'' + (b-x)y' - ay = 0
+
+This equation has solution
+
+   M(a,b, x) = \u2211[sub n=0][super infin] /a/[super (n)] / /b/[super (n)] /x/[super n]/n!
+
+where `a[super (n)]` denotes the rising factorial.
+This expression shows that the Kummer-M function is a hypergeometric function, namely, `M(a,b,x) = [sub 1]F[sub 1](a; b; x)`.
+
+The Kummer-M function will arise from solution of any ordinary differential equation with linear coefficients, to wit, by appropriate change of variable, we can solve
+
+   (Ax+b)y'' + (Cx+d)y' + (ex+d)y = 0
+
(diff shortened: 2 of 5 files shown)


NAThompson:
Have you seen the asymptotics from Handbook of Exact Solutions to Ordinary Differential Equations, S.2.7-6, by Andrei Polyanin? I think these may expressions may be of use for the large parameters you speak of, and were going to be my next line of attack.

It looks like your implementation already has a lot of good stuff, so should I just close this PR and start working on your branch or do you see anything of value here (the doc writeup might be used)?

jzmaddock:
>Have you seen the asymptotics from Handbook of Exact Solutions to Ordinary Differential Equations, S.2.7-6, by Andrei Polyanin? I think these may expressions may be of use for the large parameters you speak of, and were going to be my next line of attack.

It's a while since I stopped on this, but I believe at least one of those is in use (the first one, recast in terms of 2F0 see Pearson below), not sure about the other 2 and if they offer anything different.  The main references I've been using are:

NUMERICAL METHODS FOR THE COMPUTATION OF THE CONFLUENT AND GAUSS HYPERGEOMETRIC FUNCTIONS, JOHN W. PEARSON †, SHEEHAN OLVER ‡, AND MASON A. PORTER

Pearson's thesis: Computation of Hypergeometric Functions.

Algorithms for Rational Approximations for A Confluent Hypergeouietric Function, Yudell L. Luke.

Hypergeometric Type Functions and Their Symmetries Jan Derezi´nski

The main problem area that had me stumped was a, b and z all large and negative (or equivalent values from symmetry relations).  If I could shift any one of the parameters either much larger or smaller via recurrence relations then everything would be fine - but the direction of stability of the recurrence relations is itself  unstable!  As you apply the relations repeatedly in any particular direction, you hit a point where they suddenly go crazy and the magnitude and/or sign of the result fluctuates violently and then the direction of stability reverses... even if the relations were consistently unstable at least we could use Millers algorithm, but we can't even rely on that :(

>It looks like your implementation already has a lot of good stuff, so should I just close this PR and start working on your branch or do you see anything of value here (the doc writeup might be used)?

It's pretty complete apart from the one problem area I think, the docs I'm sure could use some improvement.  So yes, by all means see if you can improve on what's in that branch!

In the even that the problem areas can't be "solved", my plan was to just write a checked summation of the series that throws an evaluation_error if it becomes too unstable.

Best, John.

NAThompson:
Moving discussion to #148

ckormanyos:
This would be a good one to get deeper into in the fall evenings. I looked into Olver's "Asymptotics and Special Functinos", but there is no detailed information about hypergeometric_1f1. I've seen some CAS's do amazing things with hypergeometrics (complex, real argument and parameters, etc.), but I have no idea how they do it. With kind regards, Chris
   On Monday, September 3, 2018, 7:16:40 PM GMT+2, jzmaddock <notifications@github.com> wrote:  
 
 

Have you seen the asymptotics from Handbook of Exact Solutions to Ordinary Differential Equations, S.2.7-6, by Andrei Polyanin? I think these may expressions may be of use for the large parameters you speak of, and were going to be my next line of attack.


It's a while since I stopped on this, but I believe at least one of those is in use (the first one, recast in terms of 2F0 see Pearson below), not sure about the other 2 and if they offer anything different. The main references I've been using are:

NUMERICAL METHODS FOR THE COMPUTATION OF THE CONFLUENT AND GAUSS HYPERGEOMETRIC FUNCTIONS, JOHN W. PEARSON †, SHEEHAN OLVER ‡, AND MASON A. PORTER

Pearson's thesis: Computation of Hypergeometric Functions.

Algorithms for Rational Approximations for A Confluent Hypergeouietric Function, Yudell L. Luke.

Hypergeometric Type Functions and Their Symmetries Jan Derezi´nski

The main problem area that had me stumped was a, b and z all large and negative (or equivalent values from symmetry relations). If I could shift any one of the parameters either much larger or smaller via recurrence relations then everything would be fine - but the direction of stability of the recurrence relations is itself unstable! As you apply the relations repeatedly in any particular direction, you hit a point where they suddenly go crazy and the magnitude and/or sign of the result fluctuates violently and then the direction of stability reverses... even if the relations were consistently unstable at least we could use Millers algorithm, but we can't even rely on that :(


It looks like your implementation already has a lot of good stuff, so should I just close this PR and start working on your branch or do you see anything of value here (the doc writeup might be used)?


It's pretty complete apart from the one problem area I think, the docs I'm sure could use some improvement. So yes, by all means see if you can improve on what's in that branch!

In the even that the problem areas can't be "solved", my plan was to just write a checked summation of the series that throws an evaluation_error if it becomes too unstable.

Best, John.

PR #150, opened 2018-09-12
Remove deprecation warning from the BOOST_BIG_ENDIAN and BOOST_LITTLE…

NAThompson:
…_ENDIAN macros.

### DIFF
--- a/include/boost/math/special_functions/detail/fp_traits.hpp
+++ b/include/boost/math/special_functions/detail/fp_traits.hpp
@@ -216,9 +216,9 @@ template<> struct fp_traits_non_native<double, double_precision>
 
 private:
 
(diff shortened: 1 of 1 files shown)


pdimov:
This patch doesn't look correct, even though Travis is passing with flying colors; `BOOST_ENDIAN_BIG_BYTE` and `BOOST_ENDIAN_LITTLE_BYTE` are always defined, to 0 or 1, so the proper way to check them is with `#if`, not `#ifdef`. See f.ex. https://github.com/boostorg/predef/blob/develop/include/boost/predef/detail/endian_compat.h

NAThompson:
@pdimov : Nice catch, thanks!

PR #153, opened 2018-10-05
Add a fake boost_math target to boost-install

pdimov:
This adds a target `boost_math` that does nothing to the `boost-install` line for dependency resolution purposes. Some libraries such as `random` depend on `math`, but there's no `math`, so the fake target helps with that.


pdimov:
Comments, reservations?

jzmaddock:
Just not got to it yet :(

Merging now.

PR #165, opened 2018-11-25
Complex Newton's method

NAThompson:


NAThompson:
This is, I think, as good as it can reasonably get without subjecting it to users. [Here's](https://codereview.stackexchange.com/questions/209224/complex-newtons-method) a code review which said it looked pretty good.

So I think this is ready to merge provided everyone likes the feature and thinks it's a good fit.

jzmaddock:
Looks good to me too, merging...

PR #172, opened 2018-12-28
Do not include the whole Predef for a single macro

Kojoley:


jzmaddock:
The CI failures are new, but unrelated to this, will fix separately.

PR #174, opened 2019-01-05
return static_cast<result_type>(0) in round().

pulver:


NAThompson:
This "compiler is out of heap space" breaks every build. Looks like we're not the only ones with [this problem](https://github.com/appveyor/ci/issues/742) either.

jzmaddock:
Test failure is spurious, merging.

PR #179, opened 2019-01-27
Add move constructor for barycentric rational interpolation.

NAThompson:


NAThompson:
Build failure from lack of disk space; not worth another build to fix.

PR #195, opened 2019-04-13
Vectorize mean calculation.

NAThompson:


include/boost/math/tools/univariate_statistics.hpp
```diff
@@ -30,12 +30,47 @@ auto mean(ForwardIterator first, ForwardIterator last)
         }
         return mu;
     }
-    else
+    else if constexpr (std::is_same_v<typename std::iterator_traits<ForwardIterator>::iterator_category, std::random_access_iterator_tag>)
```
jzmaddock: Is this otherwise dependent upon C++17?  If yes, please go ahead and merge, otherwise can we add some configuration logic to fall back to the old code for non-C++ 17 compilers?

NAThompson: Yeah, the whole file uses C++17.

PR #209, opened 2019-05-21
Use std::size if __cplusplus >= 201703L

### LINKED ISSUE #208: (Interpolators) Catmull-Rom requires C++17 std::size
There are several uses of `std::size` in `include/boost/math/interpolators/catmull_rom.hpp`. This is only available in C++17 and prevents the use of this file in projects using older language standards. Would it be appropriate to use .size() or boost::size() be used instead?  See:
https://www.boost.org/doc/libs/1_70_0/libs/range/doc/html/range/concepts/random_access_range.html

.size() has the disadvantage of not working for C arrays, while boost::size() introduces a dependency on another part of boost that may not be desirable.

hazelnusse:
Fixes #208.

### DIFF
--- a/include/boost/math/interpolators/catmull_rom.hpp
+++ b/include/boost/math/interpolators/catmull_rom.hpp
@@ -22,7 +22,9 @@ namespace boost{ namespace math{
         typename Point::value_type alpha_distance(Point const & p1, Point const & p2, typename Point::value_type alpha)
         {
             using std::pow;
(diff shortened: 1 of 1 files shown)


jzmaddock:
That change will break msvc which never sets __cplusplus to the correct version.  I don't really see how this fixes things for you either since unqualified calls to size() will find nothing?

Whatever, closing in favour of https://github.com/boostorg/math/pull/210.

PR #219, opened 2019-06-20
Whittaker shannon

NAThompson:
This was part of the Daubechies wavelets PR, but that one was getting just huge, and hence I figured I'd split it into more digestible chunks.


NAThompson:
@pulver : Do you mind taking a look at the `.prime()` function on the Whittaker-Shannon derivative? I think if you wrote the automatic differentiation formula for evaluation of the derivative it would be way more accurate than the braindead evaluation that I'm using here.

pulver:
> @pulver : Do you mind taking a look at the `.prime()` function on the Whittaker-Shannon derivative? I think if you wrote the automatic differentiation formula for evaluation of the derivative it would be way more accurate than the braindead evaluation that I'm using here.

Is there a reason why `operator()` can't be written as a template function, so that it can be called with a parameter of type `autodiff_fvar<double,1>`? The return value `y` will then have the derivative via `y.derivative(1)`.

The alternative of writting a general autodiff overload, as was done with `lambert_w0`, must calculate all derivatives up to any given order, which would be a bit of an overkill if you're only interested in the first.

pabristow:
If you chaps are thinking about a change here, you might consider putting a warning in the code and the docs that this is subject to further work-in-progress.  So anyone using it is forewarned that they might have to revisit their code later (and it will be a breaking change). Just a thought.  (Either seem reasonable but avoiding unwanted computation is never bad).

NAThompson:
I got it working; basically I just following the autodiff algorithm through by hand.

NAThompson:
CI failure was a travis problem; shouldn't show up again.

PR #220, opened 2019-06-26
Cardinal quadratic B-spline interpolation.

NAThompson:
Again another PR to split the wavelets PR into digestible chunks.


pulver:
Fyi we're looking into the autodiff-related build errors
```
unknown location(0): fatal error: in "test_autodiff_7/gamma_hpp<float>": Throw location unknown (consider using BOOST_THROW_EXCEPTION)
```
Not sure why these weren't showing up on the previous PR #218.

CC: @kedarbhat

NAThompson:
@pulver : No problem; the CI system should serve us, not the other way around. So take a look at it as time permits.

PR #222, opened 2019-06-28
Disable gamma_hpp autodiff test template.

pulver:


NAThompson:
Why comment it out instead of just deleting it?

mattpulver:
> Why comment it out instead of just deleting it?

Will do.

NAThompson:
I don't quite understand what happened here. It looks like the `autodiff` type was being passed into the `float_distance`, which is clearly undefined, but why was it being passed into that function?

kedarbhat:
Hi Nick, `float_distance` works for `autodiff` types (this is part of the next_hpp tests in test_autodiff_8.cpp). 

The errors in the gamma function tests were due to infinite values being passed into `float_distance` (or rather, non-finite valued instances of `autodiff_fvar` were being passed into `float_distance`), at least that's the way I'm reading the  error below (please correct me if you're looking at a different error).

I've submitted a PR to Matthew (in the `pulver/autodiff` repo) to catch exceptions and print the inputs which caused them, which will give more useful diagnostics.

_unknown location(0): fatal error: in "test_autodiff_7/gamma_hpp<float>": Throw location unknown (consider using BOOST_THROW_EXCEPTION)
Dynamic exception type: boost::wrapexcept<std::domain_error>
std::exception::what: Error in function float_distance<N5boost4math15differentiation11autodiff_v16detail4fvarIfLm5EEE>(N5boost4math15differentiation11autodiff_v16detail4fvarIfLm5EEE, N5boost4math15differentiation11autodiff_v16detail4fvarIfLm5EEE): Argument a must be finite, but got depth(1)(inf,0,0,0,0,0)_

NAThompson:
> I've submitted a PR to Matthew (in the pulver/autodiff repo) for possible inclusion after the Boost 1.71 release, so that thrown exceptions are caught and the inputs which caused the exception are printed

That's just in the tests, correct?

kedarbhat:
Correct, the library itself remains untouched.

PR #227, opened 2019-07-08
Theis interpolation. [CI SKIP]

NAThompson:


NAThompson:
This interpolator is horrible.

PR #228, opened 2019-07-15
Cardinal b splines

NAThompson:


jzmaddock:
That looks fine to me, two quick questions:

1) Do we have changes in develop unmerged to master (ie waiting for the closed-for-beta period to end)?
2) Should the basis be provided as separate arguments, ie as:

```
template<unsigned n, typename Real>
   Real cardinal_b_spline(Real x, Real basis_start = -1, Real basis_end = 1);
```

Rather than providing a separate function for the [0,1] case?

I don't have a strong opinion on this, just asking ;)

NAThompson:
> Do we have changes in develop unmerged to master (ie waiting for the closed-for-beta period to end)?

No, develop got merged to master last week and I'm aware of no problems with it. If there's some bugs discovered post-beta we'll have to cherry pick.

>     Should the basis be provided as separate arguments, ie as:

template<unsigned n, typename Real>
   Real cardinal_b_spline(Real x, Real basis_start = -1, Real basis_end = 1);

Rather than providing a separate function for the [0,1] case?

This is actually closer in spirit to what the Cox paper does, namely he creates B-splines on arbitrary knot sets and so the basis functions can have arbitrary support. That loses some symmetry relative to what I'm doing here, but it's something I would like to deploy in the future. But as for now, I would say no.

pabristow:
https://www.boost.org/doc/libs/1_70_0/libs/math/doc/html/math_toolkit/ellint/ellint_intro.html

"This implementation uses k throughout: this matches the requirements of the Technical Report on C++ Library Extensions. However, you should be extra careful when using these functions! "

And k as the first parameter is to match all the other elliptic functions.

So although Boost does not match other implementations, there is some consistency and logic to it.

PR #233, opened 2019-08-04
Gegenbauer polynomials

NAThompson:
I think this will be easier than the Jacobi polynomials; in any case, my use case only requires that I have some continuous parameter that I can optimize over, and the Gegenbauer polynomials give me one.

### DIFF
--- a/include/boost/math/special_functions/gegenbauer.hpp
+++ b/include/boost/math/special_functions/gegenbauer.hpp
@@ -1,0 +1,41 @@
+//  (C) Copyright Nick Thompson 2019.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_MATH_SPECIAL_CARDINAL_B_SPLINE_HPP
+#define BOOST_MATH_SPECIAL_CARDINAL_B_SPLINE_HPP
+
+#include <array>
+#include <cmath>
+#include <boost/math/tools/recurrence.hpp>
+
+namespace boost { namespace math {
+
+
+template<typename Real1, typename Real2>
+Real2 gegenbauer(unsigned n, Real1 alpha, Real2 x)
+{
+    if (x < 0) {
+        if (n&1) {
+            return -gegenbauer(n, alpha, -x);
+        }
+        return gegenbauer(n, alpha, -x);
+    }
+
+    if (x > 1) {
+        throw std::domain_error("Recurrence implementations for the Gegenbauer polynomials have unknown stability for x > 1.");
+    }
+
+    // (n+1)C_{n+1}(alpha, x) - (2x(n+alpha))C_{n}(alpha, x) + (n-1+2alpha)C_{n-1}(alpha, x) = 0
+    auto Recurrence = [&](unsigned j) { return std::make_tuple<Real2, Real2, Real2>( n-1+2*alpha, -2*x*(n+alpha), n+1); };
+
+    Real2 factor = 1;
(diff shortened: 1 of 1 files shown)


include/boost/math/special_functions/gegenbauer.hpp
```diff
@@ -0,0 +1,41 @@
+//  (C) Copyright Nick Thompson 2019.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_MATH_SPECIAL_CARDINAL_B_SPLINE_HPP
```
jzmaddock: The include guard name needs changing - cut and paste error?

include/boost/math/special_functions/gegenbauer.hpp
```diff
@@ -0,0 +1,41 @@
...
+    if (x > 1) {
+        throw std::domain_error("Recurrence implementations for the Gegenbauer polynomials have unknown stability for x > 1.");
+    }
+
+    // (n+1)C_{n+1}(alpha, x) - (2x(n+alpha))C_{n}(alpha, x) + (n-1+2alpha)C_{n-1}(alpha, x) = 0
+    auto Recurrence = [&](unsigned j) { return std::make_tuple<Real2, Real2, Real2>( n-1+2*alpha, -2*x*(n+alpha), n+1); };
+
+    Real2 factor = 1;
+    boost::uintmax_t max_iter = n + 1;
+    Real2 Calphax = boost::math::tools::function_ratio_from_backwards_recurrence(Recurrence, factor, max_iter);
```
jzmaddock: Do you really want the ratio here?  Surely you want to apply recurrences up to n - or else recurse down to n = 0 and normalise (Millers Algorithm)?

include/boost/math/special_functions/gegenbauer.hpp
NAThompson: Yup, just got that one.

include/boost/math/special_functions/gegenbauer.hpp
NAThompson: I hadn't yet worked out the ratio; actually I just PR'd this to move it from one computer to another.

NAThompson:
If I've still got John online:

- I have recursed upwards, which is exactly what the GSL does. But OMG, something very weird is happening. Check out the ULP plot:

Gegenbaurer6ULP

There appears to be 3 distinct regions where the upwards recurrence is unstable. I have no explanation. In addition, the regions have very small measure; almost everywhere else it is perfectly stable.

jzmaddock:
That's weird for sure, but the errors are still relatively small?  Is this near the zeros of the polynomial maybe?  I doubt the recursing down would help if so - normalisation would be by comparison to the zeroth order polynomial which is known.

jzmaddock:
Actually looking at the recurrence relations It's not at all obvious to me which direction is the stable one.  My suspicion might be that both directions are meta-stable to some extent.

NAThompson:
Uh, whoops! Looks like you are precisely correct; these are the roots.

NAThompson:
I just did a literature review on this problem: Basically, there are sophisticated mechanisms to sum a series of Gegenbauer polynomials (going beyond the standard Clenshaw recurrence) but it doesn't appear that there is much study of how to evaluate the roots to high accuracy; moreover this problem is ill-conditioned so I'm not sure I should agonize over this too much. The fact that it appears in the quadratic means that there's very little scope for improvement.

In any case, I'm removed the draft label; presumably the only thing to try is continued fractions or Miller's algorithm. The continued fractions might devastate the performance; so Miller's algorithm is the only thing left to try.

PR #234, opened 2019-08-05
Jacobi polynomials

NAThompson:


NAThompson:
@jzmaddock : Did you have a nice tool for gamma function ratios such as are seen in the [Jacobi polynomial derivatives](https://en.wikipedia.org/wiki/Jacobi_polynomials)? For some reason I thought you had such a thing . . .

jzmaddock:
>@jzmaddock : Did you have a nice tool for gamma function ratios such as are seen in the Jacobi polynomial derivatives? For some reason I thought you had such a thing . . .

Yep, several such tools.  Assuming k is an integer, then that's a rising factorial (aka pochhammer), see https://www.boost.org/doc/libs/1_70_0/libs/math/doc/html/math_toolkit/factorials/sf_rising_factorial.html.

In the general case, we have https://www.boost.org/doc/libs/1_70_0/libs/math/doc/html/math_toolkit/sf_gamma/gamma_ratios.html

NAThompson:
> Assuming k is an integer, then that's a rising factorial (aka pochhammer)

The division by 2^k means I get some overflow resistance by doing the rising factorial by hand. Hopefully there aren't any cool tricks for stable evaluations of Pochhammers that I lost . . .

PR #249, opened 2019-09-06
roots: Use C versions of fma and copysign

neheb:
fma and copysign are C++11 functions that do not get included in std::
by uClibc-ng. Switch to the C version.

### DIFF
--- a/include/boost/math/tools/roots.hpp
+++ b/include/boost/math/tools/roots.hpp
@@ -800,8 +800,8 @@ namespace detail
     inline T discriminant(T const & a, T const & b, T const & c)
     {
         T w = 4*a*c;
-        T e = std::fma(-c, 4*a, w);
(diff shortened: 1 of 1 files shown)


jzmaddock:
That would break long double support.

Am I correct in thinking that you're using gcc + libstdc++  ?  If so can you tell me what happens if you compile:

```
#include <cmath>

#ifdef _GLIBCXX_USE_C99_MATH_TR1
#error 1
#else
#error 2
#endif
```
?

neheb:
error 2.

Local Ubuntu compile gives error 1.

neheb:
This is the solution in a different project: https://github.com/facebook/folly/blob/82a534256b216e9c12852a09e3d934bd76ec1d18/folly/portability/Math.h#L32

My C++ is not that great though.

PR #265, opened 2019-10-20
Linear regression

NAThompson:


pabristow:
For some types of data, I have found least squares decidedly un-'robust', but there is a need for the basic algorithm, and this looks like it.

For other purposes, we can add other various more 'robust' methods later?

NAThompson:
I definitely will be adding the robust linear regression soon, but I'm trying to decide *which* robust linear regression should be added. I actually like NRs, but I haven't seen anything about it's pitfalls. The [Theil-Sen](https://en.wikipedia.org/wiki/Theil%E2%80%93Sen_estimator) estimator looks really good, but naively it's quadratic time, and there are too many different options for making it efficient which I can't evaluate right now.

PR #266, opened 2019-10-24
Remove fake boost_math library target; no longer required

pdimov:


NAThompson:
Just read through the linked issue; seems like this is relatively low risk.

@pdimov : Is it important that this go into 1.72?

pdimov:
It's not critical, but we gain nothing if it doesn't go into 1.72. In general we only discover installation/downstream issues when a release goes out, so delaying this for 1.73 won't help anyone.

This basically reverts my earlier PR https://github.com/boostorg/math/pull/153.

NAThompson:
Ok cool. @jzmaddock : Cool with this getting merged to both develop and master as soon as the CI cycles?

jzmaddock:
>Ok cool. @jzmaddock : Cool with this getting merged to both develop and master as soon as the CI cycles?

Absolutely.

PR #283, opened 2019-12-20
roots: Fix fma_workaround

neheb:
fma takes three parameters, not one.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


neheb:
A second problem showed up. Because fma_workaround was not marked inline, it was generating multiple functions, per the template and causing linker errors. Fixed that as well.

neheb:
Seems buildbot is stuck.

jzmaddock:
Many thanks for the patch - appveyor (msvc tests) is irrelevant for this change, so merging.

PR #285, opened 2019-12-20
roots: Use boost::math::copysign

neheb:
std::copysign is not available with uClibc-ng.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


neheb:
Updated patch to use an if directive to select boost::math::copysign when std version is not available.

neheb:
This succeeded compilation. Seems the buildbots have problems.

NAThompson:
@neheb : Just took a look at the build and the diffs; looks good to me. AppVeyor takes forever to run; presumably this is not time critical?

neheb:
I'd like for this to make the next release of boost, but otherwise no.

PR #286, opened 2019-12-20
Fix more typos, via latest version of typochecker tool

bwignall:
With current version of typochecker utility (https://github.com/bwignall/typochecker), which has an updated set of likely typos, this "should" eliminate a lot of false positives:

`python PATH/TO/corrector.py -d path/to/math -w significand -w Mathematica -w cont -w comm`


NAThompson:
Just read through all these diffs; looks good thanks!

pabristow:
From reading these, and a forensic examination of the source code, I conclude that this typochecker program now includes a metaphysical AI algorithm that in Phase 1 cunningly detects places where a typos might have existed and inserts a suitable one from its databse.  In Phase 2 these typos are rediscovered and corrected.  (If one turns the sound volume up to ear and speaker damaging levels one can hear a 'whoop' when these events occur during both Phase 1 and Phase 2).

They will be merged after the conclusion of Solstice Celebrations...

PR #290, opened 2019-12-23
Entropy of distributions.

NAThompson:
There are still quite a few more distributions to do, but this is a good start.

Should it be called  `differential_entropy` or is `entropy` just fine?

Table from Wikipedia:

https://en.wikipedia.org/wiki/Differential_entropy


NAThompson:
@pabristow , @jzmaddock : Gimme your thoughts on this; looks like once the build cycles it'll basically be ready to go.

NAThompson:
It seems like this is pretty uncontroversial, and I want to get working on the Kullback-Liebler divergence without merge commit issues, so I'm merging. Can revert if anyone objects.

jeremy-murphy:
> There are still quite a few more distributions to do, but this is a good start.
> 
> Should it be called `differential_entropy` or is `entropy` just fine?
> 
> Table from Wikipedia:
> 
> https://en.wikipedia.org/wiki/Differential_entropy

This should be differential entropy, because the random variable (or more specifically, the PDF from which it comes) is continuous.
The formula for (discrete) entropy and differential entropy are very similar, summation in one case, integration in the other, but a key difference in the result is that differential entropy can be negative.

I'm not sure if Boost.Math already has a dependency on Boost.Units, but whether it does or not, I would consider giving the result the unit type "bits", so as to distinguish from the other units that (differential) entropy sometimes uses: nats,

Cheers.

include/boost/math/distributions/cauchy.hpp
```diff
@@ -347,6 +347,13 @@ inline RealType kurtosis_excess(const cauchy_distribution<RealType, Policy>& /*d
       std::numeric_limits<RealType>::quiet_NaN(), Policy());
 }
 
+template <class RealType, class Policy>
+inline RealType entropy(const cauchy_distribution<RealType, Policy> & dist)
+{
+   using std::log;
+   return log(2*constants::two_pi<RealType>()*dist.scale());
+}
```
jeremy-murphy: I think integer `2` will implicitly convert to `double`, which could cause extra rounding when `RealType` is not `double`.
I would be strict and write it as `RealType(2)`.

jzmaddock: Not in any normal situation I can think of, it's actually a conceptual requirement of Boost.Math that all real-valued types interoperate with integer literals, and we make plenty of use of that throughout the library.  Besides I can think of no situation where `2.0` (double precision) looses bits compared to `2`.  The only situation I can think of is if RealType isn't constructible or multipliable by `int` and has a lossy _implicit_ conversion operator to double.
BTW in the multiprecision case, multiplying by an integer is likely to be significantly more efficient than constructing a temporary `RealType` and multiplying.

jeremy-murphy: > Not in any normal situation I can think of, it's actually a conceptual requirement of Boost.Math that all real-valued types interoperate with integer literals, and we make plenty of use of that throughout the library. Besides I can think of no situation where `2.0` (double precision) looses bits compared to `2`. The only situation I can think of is if RealType isn't constructible or multipliable by `int` and has a lossy _implicit_ conversion operator to double.

What I'm thinking of is the so-called ["double rounding"](https://www.exploringbinary.com/double-rounding-errors-in-floating-point-conversions/) error, when `RealType` is `float`, then the extra conversion from `double` back down to `float` incurs an extra rounding.

> BTW in the multiprecision case, multiplying by an integer is likely to be significantly more efficient than constructing a temporary `RealType` and multiplying.

Yes, I understand that Boost.Math probably has a policy of being performance-friendly to multi-precision types, but even at the cost of accuracy to other floating point types? I'm not expecting you to flip a long-standing policy right now, and I'm sure that C++ being the language it is could solve all situations to everyone's liking (but with a lot more work).

Surely this sort of small literal value performance penalty must arise frequently for multi-precision types? Could it not be solved in general with `constexpr` constructors or a small-object optimization?

NAThompson: Doesn't look like the double rounding problem happens on gcc:

https://godbolt.org/z/m4q-vz

jeremy-murphy: Yes, it does look like clang and GCC optimize the integer literal away to a compile-time effect. Even integer non-literal values appeared to not have double rounding, so that's good.
I assumed that the literal `int` would promote to `double` first.

It's a different story when the literal is `2.0` and so actually has the `double` type. See for yourself, the `float` function now looks very inefficient. But that's not the actual situation here, so it's case closed for now. Something to think about though.  :)

PR #292, opened 2019-12-26
Fix typos, after fixing bug in typochecker

bwignall:
I thought I was done for awhile, but I found a subtle bug in the typochecker tonight, so I re-ran it on this codebase.


bwignall:
Sorry for the extra PR; I only suspected/found/fixed the bug after my last one.

NAThompson:
Just read through these diffs; looks good!

ckormanyos:
Very cool. I've read and re-read some of those lines numerous times. Never noticed. Really good catches those are...

PR #291, opened 2019-12-25
Fix typos

bwignall:
Another iteration of applying typochecker (with an improved dataset), similar to https://github.com/boostorg/math/pull/286 .

PR #301, opened 2020-01-12
Fix typos

bwignall:
Another iteration of applying typochecker (with an improved dataset), similar to #291 .


NAThompson:
@bwignall : The `.html` files are generated from the `.qbk`; can you run your tool and exclude `.html`?

bwignall:
@NAThompson : done.

NAThompson:
@bwignall : Thanks again; I just looked through all these; they look good. Will merge once the build cycles.

Issue #310, opened 2020-01-24
New dependency on chrono in 1.72 broke our build.


barcharcraz:
See https://github.com/microsoft/STL/pull/428 for more details. Basically chrono has (optionally) a separately compiled component, and uses a prefix and suffix header to enforce abi, this breaks our build by triggering warnings.

We can avoid this by defining the macro to use chrono header only, but we _really_ don't want to depend on boost chrono (at least not by accident) 

The commit in question for boost is f4e5abfbb1f2e3f1c07cb144a895620c8004e712, and it appears that it adds support for a wall-clock timeout on "hypergeometric_pFq".

Is this intentional?

ckormanyos:
> appears that it adds support for a wall-clock timeout on "hypergeometric_pFq"

I had also noticed this, but remained silent earlier. I can investigate this phenomenon. it might seem unusual to need hard time in hypergeometric series calculations. Either it is intentional and needed or not.

i will try to find out.

jzmaddock:
It was intentional to have a timeout as otherwise pFq can just go on forever trying to reach the requested precision.

However, this is new code, so I'm wondering how you're pulling it in?  Ah... I guess you're including `<boost/math/special_functions.hpp>` rather than the individual headers maybe?

It does occur to me, that since use of chrono is an internal detail, and that header is C++11 anyway (initializer_lists), that we should just switch to `std::chrono`.

barcharcraz:
yep we were pulling in all of special_functions.hpp, I've switched our code to include just what we need, however it's still strange to pull in chrono. It looks like the hypergeometric series calculations already have support for number of iterations based timeout (as I would expect), however they _also_ support a wall clock timeout which is somewhat strange.

jzmaddock:
>It looks like the hypergeometric series calculations already have support for number of iterations based timeout (as I would expect), however they also support a wall clock timeout which is somewhat strange.

There are 2 different interfaces here for 2 different use cases: the regular fixed precision case uses an iteration limit as with all the other special functions, but there is also a variable precision interface which keeps upping the working precision until the required accuracy has been achieved - and that one uses a wall clock timeout (and is expected to take some considerable time to complete in some cases).  It could be debatable whether the latter interface belongs with this library or Multiprecision, and whether it's a header API or a tool, but it has proved to be exceptionally useful as a means of double checking other results, so I'm assuming others will find it useful also.

barcharcraz:
[...]

PR #311, opened 2020-01-24
Switch pFq over to using <chrono> since it's all C++11 code anyway.

jzmaddock:
Removes dependency on Boost.Chrono which breaks Microsoft's std lib build.
Fixes https://github.com/boostorg/math/issues/310.


jzmaddock:
@NAThompson : The CI tests are all failing for autodiff, looks like some disconnected change somewhere else has exposed a latent bug?  Should the use of `decay_t` refer to `std::decay_t` or `boost::decay_t` ?  I assume the former, but wanted to check.

NAThompson:
Didn't ever notice this before. It appears that `std::decay_t` would be fine.

@pulver: Did you intend on `std::decay_t` or `boost::decay_t`?

pulver:
> Did you intend on `std::decay_t` or `boost::decay_t`?

I believe I would have intended `std::decay_t`. I am puzzled why it wasn't originally that way. I'll submit a fix this weekend, or feel free to make the change directly.

PR #312, opened 2020-01-25
autodiff: replace decay_t<T> with typename std::decay<T>::type.

pulver:
This is a fix for an issue brought up in #311. For compatibility with C++11, we will use the more verbose `typename std::decay<T>::type` rather than the more succinct C++14 version `std::decay_t<T>`.


jzmaddock:
Haha, looks like we've both been at this, there's an alternative in https://github.com/boostorg/math/pull/313 which also removes the #include of special_functions.hpp in favour of just those we really need, and also adds a #include test case.

pulver:
@jzmaddock Ok great once #313 passes then I'm happy to go with that over this PR. Cheers.

pulver:
Use #313 instead.

PR #318, opened 2020-02-27
adding dirichlet distribution

mrityunjay-tripathi:
I am adding Dirichlet distribution in ```include/boost/math/distributions/```. The Dirichlet distribution is a multivariate distribution function, so vector type inputs are needed, but the tests in ```test/compile_test/test_compile_result.hpp``` are for ```int```, ```double```, ```float```, ```long double```, ```unsigned```. How can I add tests for Dirichlet distribution?
TODO:
- [ ] Decision on ```VectorType``` and it's implementation.
- [ ] Cross check ```cdf``` function. (I could not find it, so tried to derive it.)
- [ ] Implementation of ```find_concentration``` function using ```x``` and ```probability``` as arguments.
- [ ] Implementation of ```quantiles``` function
- [ ] Implementation of ```skewness``` function
- [ ] Implementation of ```kurtosis``` function.

Also, none of the distribution functions generate random samples of the corresponding distribution. That means no sampling function implemented. In PyTorch, we can do something like,
```
>>> import torch
>>> import torch.distributions as dist
>>> d = dist.dirichlet.Dirichlet(torch.tensor([0.3, 0.3, 0.3]))
>>> d.sample()
tensor([0.7736, 0.1908, 0.0356])
>>> d.rsample(torch.tensor([5]))
tensor([[2.3072e-02, 9.4317e-01, 3.3756e-02],
        [1.8339e-01, 1.6803e-03, 8.1493e-01],
        [7.3136e-01, 1.9131e-01, 7.7323e-02],
        [2.1086e-01, 9.5684e-03, 7.7958e-01],
        [9.6672e-06, 4.6919e-02, 9.5307e-01]])
>>> 
```
Can we add such feature or its implemented as such?

### DIFF
--- a/include/boost/math/distributions/dirichlet.hpp
+++ b/include/boost/math/distributions/dirichlet.hpp
@@ -1,0 +1,405 @@
+// boost/math/distributions/dirichlet.hpp
+
+// Copyright Mrityunjay Tripathi 2020.
+
+// Use, modification and distribution are subject to the
+// Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// https://en.wikipedia.org/wiki/Dirichlet_distribution
+// https://mast.queensu.ca/~communications/Papers/msc-jiayu-lin.pdf
+
+// The Dirichlet distribution is a family of continuous multivariate probability
+// distributions parameterized by a vector 'alpha' of positive reals.
+// It is a multivariate generalization of the dirichlet distribution, hence its
+// alternative name of multivariate dirichlet distribution (MBD).
+// Dirichlet distributions are commonly used as prior distributions in
+// Bayesian statistics, and in fact the Dirichlet distribution is the
+// conjugate prior of the categorical distribution and multinomial distribution.
+
+#ifndef BOOST_MATH_DIST_DIRICHLET_HPP
+#define BOOST_MATH_DIST_DIRICHLET_HPP
+
+#include <boost/math/distributions/fwd.hpp>
+#include <boost/math/special_functions/gamma.hpp>
+#include <boost/math/special_functions/digamma.hpp>
+#include <boost/math/distributions/complement.hpp>
+#include <boost/math/special_functions/fpclassify.hpp>
+#include <boost/math/distributions/detail/common_error_handling.hpp>
+
+#if defined(BOOST_MSVC)
+#pragma warning(push)
+#pragma warning(disable : 4702) // unreachable code
+// in domain_error_imp in error_handling
+#endif
+
+#include <utility>
+
+namespace boost
+{
+namespace math
+{
+namespace dirichlet_detail
+{
+// Common error checking routines for dirichlet distribution function:
+template <class VectorType, class RealType, class Policy>
+inline bool check_concentration(const char *function,
+                                const VectorType &concentration,
+                                RealType *result,
+                                const Policy &pol)
+{
+  for (size_t i = 0; i < concentration.size(); ++i)
+  {
+    if (!(boost::math::isfinite)(i) || (i <= 0))
+    {
+      *result = policies::raise_domain_error<RealType>(
+          function,
+          "Concentration Parameter is %1%, but must be > 0 !", concentration, pol);
+      return false;
+    }
+  }
+  return true;
+} // bool check_concentration
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_prob(const char *function,
+                       const VectorType &p,
+                       RealType *result,
+                       const Policy &pol)
+{
+  for (size_t i = 0; i < p.size(); ++i)
+  {
+    if ((i < 0) || (i > 1) || !(boost::math::isfinite)(i))
+    {
+      *result = policies::raise_domain_error<RealType>(
+          function,
+          "Probability argument is %1%, but must be >= 0 and <= 1 !", i, pol);
+      return false;
+    }
+  }
+  return true;
+} // bool check_prob
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_x(const char *function,
+                    const VectorType &x,
+                    RealType *result,
+                    const Policy &pol)
+{
+  for (size_t i = 0; i < x.size(); ++i)
+  {
+    if (!(boost::math::isfinite)(x) || (x < 0) || (x > 1))
+    {
+      *result = policies::raise_domain_error<RealType>(
+          function,
+          "x argument is %1%, but must be >= 0 and <= 1 !", x, pol);
+      return false;
+    }
+  }
+  return true;
+} // bool check_x
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_dist(const char *function,
+                       const VectorType &concentration,
+                       RealType *result,
+                       const Policy &pol)
+{
+  return check_concentration(function, concentration, result, pol);
+} // bool check_dist
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_dist_and_x(const char *function,
+                             const VectorType &concentration,
+                             const VectorType &x,
+                             RealType *result,
+                             const Policy &pol)
+{
+  return check_dist(function, concentration, result, pol) && check_x(function, x, result, pol);
+} // bool check_dist_and_x
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_dist_and_prob(const char *function,
+                                const VectorType &concentration,
+                                const VectorType &p,
+                                RealType *result,
+                                const Policy &pol)
+{
+  return check_dist(function, concentration, result, pol) && check_prob(function, p, result, pol);
+} // bool check_dist_and_prob
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_mean(const char *function,
+                       const VectorType &mean,
+                       RealType *result,
+                       const Policy &pol)
+{
+  for (size_t i = 0; i < mean.size(); ++i)
+  {
+    if (!(boost::math::isfinite)(i) || (i <= 0))
+    {
+      *result = policies::raise_domain_error<RealType>(
+          function,
+          "mean argument is %1%, but must be > 0 !", i, pol);
+      return false;
+    }
+  }
+  return true;
+} // bool check_mean
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_variance(const char *function,
+                           const VectorType &variance,
+                           RealType *result,
+                           const Policy &pol)
+{
+  for (size_t i = 0; i < variance.size(); ++i)
+  {
+    if (!(boost::math::isfinite)(i) || (i <= 0))
+    {
+      *result = policies::raise_domain_error<RealType>(
+          function,
+          "variance argument is %1%, but must be > 0 !", i, pol);
+      return false;
+    }
+  }
+  return true;
+} // bool check_variance
+} // namespace dirichlet_detail
+
+template <class VectorType = std::vector<double>, class RealType = double, class Policy = policies::policy<>>
+class dirichlet_distribution
+{
+public:
+  dirichlet_distribution(VectorType concentration) : concentration(concentration)
+  {
+    RealType result;
+    dirichlet_detail::check_dist(
+        "boost::math::dirichlet_distribution<%1%>::dirichlet_distribution",
+        concentration,
+        &result, Policy());
+    sum_concentration = accumulate(concentration.begin(), concentration.end(), 0);
+  } // dirichlet_distribution constructor.
+
+  // Accessor functions:
+  VectorType Concentration() const
+  {
+    return concentration;
+  }
+
+  size_t Order() const
+  {
+    return concentration.size();
+  }
+
+  static VectorType find_concentration(
+      VectorType mean,     // Expected value of mean.
+      VectorType variance) // Expected value of variance.
+  {
+    assert(("Dimensions of mean and variance must be same!", mean.size() == variance.size()));
+    static const char *function = "boost::math::dirichlet_distribution<%1%>::find_concentration";
+    RealType result = 0; // of error checks.
+    if (!(dirichlet_detail::check_mean(function, mean, &result, Policy()) && dirichlet_detail::check_variance(function, variance, &result, Policy())))
+    {
+      return result;
+    }
+    VectorType c;
+    for (size_t i = 0; i < mean.size(); ++i)
+    {
+      c.push_back(mean[i] * (((mean[i] * (1 - mean[i])) / variance[i]) - 1));
+    }
+    return c;
+  } // RealType find_concentration
+
+  // TODO
+  // static VectorType find_concentration(
+  //     VectorType x,           //  x.
+  //     VectorType probability) // cdf
+  // {
+  //   assert(("", x.size() == probability.size()));
+  //   static const char *function = "boost::math::dirichlet_distribution<%1%>::find_conentration";
+  //   RealType result = 0; // of error checks.
+  //   if (!(dirichlet_detail::check_prob(function, probability, &result, Policy()) && dirichlet_detail::check_x(function, x, &result, Policy())))
+  //   {
+  //     return result;
+  //   }
+  //   return ;
+  // } // RealType find_concentration(x, probability)
+
+private:
+  VectorType concentration; // https://en.wikipedia.org/wiki/Concentration_parameter.
+  RealType sum_concentration;
+}; // template <class RealType, class Policy> class dirichlet_distribution
+
+
+template <class VectorType, class RealType, class Policy>
+inline const std::pair<RealType, RealType> range(const dirichlet_distribution<VectorType, RealType, Policy> & /* dist */)
+{ // Range of permissible values for random variable x.
+  using boost::math::tools::max_value;
+  return std::pair<RealType, RealType>(static_cast<RealType>(0), static_cast<RealType>(1));
+}
+
+
+template <class VectorType, class RealType, class Policy>
+inline const std::pair<RealType, RealType> support(const dirichlet_distribution<VectorType, RealType, Policy> & /* dist */)
+{ // Range of supported values for random variable x.
+  // This is range where cdf rises from 0 to 1, and outside it, the pdf is zero.
+  return std::pair<RealType, RealType>(static_cast<RealType>(0), static_cast<RealType>(1));
+}
+
+
+template <class VectorType, class RealType, class Policy>
+inline VectorType mean(const dirichlet_distribution<VectorType, RealType, Policy> &dist)
+{ // Mean of dirichlet distribution = c[i]/sum(c).
+  VectorType m;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    m.push_back(dist.concentration[i] / dist.sum_concentration);
+  }
+  return m;
+
+} // mean
+
+
+template <class VectorType, class RealType, class Policy>
+inline VectorType variance(const dirichlet_distribution<VectorType, RealType, Policy> &dist)
+{
+  VectorType v;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    v.push_back(dist.concentration[i] / dist.sum_concentration * (1 - dist.concentration[i] / dist.sum_concentration) / (1 + dist.sum_concentration));
+  }
+  return v;
+} // variance
+
+
+template <class VectorType, class RealType, class Policy>
+inline VectorType mode(const dirichlet_distribution<VectorType, RealType, Policy> &dist)
+{
+  static const char *function = "boost::math::mode(dirichlet_distribution<%1%> const&)";
+  VectorType m;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    if ((dist.concentration[i] <= 1))
+    {
+      result = policies::raise_domain_error<RealType>(
+          function,
+          "mode undefined for alpha = %1%, must be > 1!", dist.alpha(), Policy());
+      return result;
+    }
+    else
+    {
+      m.push_back((dist.concentration[i] - 1) / (dist.sum_concentration - dist.Order()));
+    }
+  }
+  return m;
+} // mode
+
+
+template <class VectorType, class RealType, class Policy>
+inline RealType entropy(const dirichlet_distribution<VectorType, RealType, Policy> &dist)
+{
+  RealType t1 = 1;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    t1 *= tgamma(dist.concentration[i]);
+  }
+  t1 = std::log(t1 / tgamma(dist.sum_concentration));
+  RealType t2 = (dist.sum_concentration - dist.Order()) * digamma(dist.sum_concentration);
+  RealType t3 = 0;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    t3 += (dist.concentration[i] - 1) * digamma(dist.concentration[i]);
+  }
+  return t1 + t2 - t3;
+}
+
+
+template <class VectorType, class RealType, class Policy>
+inline RealType pdf(const dirichlet_distribution<VectorType, RealType, Policy> &dist, const VectorType &x)
+{ // Probability Density/Mass Function.
+  BOOST_FPU_EXCEPTION_GUARD
+
+  static const char *function = "boost::math::pdf(dirichlet_distribution<%1%> const&, %1%)";
+
+  BOOST_MATH_STD_USING // for ADL of std functions
+
+  // Argument checks:
+  RealType result = 0;
+  if (!dirichlet_detail::check_dist_and_x(function, x, &result, Policy()))
+  {
+    return result;
+  }
+  using boost::math::tgamma;
+  RealType f = 1;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    f *= std::pow(x[i], dist.concentration[i] - 1);
+  }
+  f /= dist.normalizing_factor;
+  return f;
+} // pdf
+
+
+template <class VectorType, class RealType, class Policy>
+inline RealType cdf(const dirichlet_distribution<VectorType, RealType, Policy> &dist, const VectorType &x)
+{ // Cumulative Distribution Function dirichlet.
+  BOOST_MATH_STD_USING // for ADL of std functions
+
+      static const char *function = "boost::math::cdf(dirichlet_distribution<%1%> const&, %1%)";
+
+  // Argument checks:
+  RealType result = 0;
+  if (!dirichlet_detail::check_dist_and_x(function, dist.concentration, x, &result, Policy()))
+  {
+    return result;
+  }
+  RealType c = 1;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    c *= std::pow(x[i], dist.concentration[i]) / tgamma(dist.concentration[i]) / dist.concentration[i];
+  }
+  c *= tgamma(dist.sum_concentration);
+  return c;
+} // dirichlet cdf
+
+template <class VectorType, class RealType, class Policy>
+inline RealType cdf(const complemented2_type<dirichlet_distribution<VectorType, RealType, Policy>, RealType> &c)
+{ // Complemented Cumulative Distribution Function dirichlet.
+
+  BOOST_MATH_STD_USING // for ADL of std functions
+
+  static const char *function = "boost::math::cdf(dirichlet_distribution<%1%> const&, %1%)";
+
+  RealType const &x = c.param;
+  dirichlet_distribution<RealType, Policy> const &dist = c.dist;
+
+  // Argument checks:
+  RealType result = 0;
+  if (!dirichlet_detail::check_dist_and_x(function, x, &result, Policy()))
+  {
+    return result;
+  }
+  RealType cumm = 1;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    cumm *= std::pow(x[i], dist.concentration[i]) / tgamma(dist.concentration[i]) / dist.concentration[i];
+  }
+  cumm *= tgamma(dist.sum_concentration);
+  return cumm;
+} // dirichlet cdf
+
+} // namespace math
+} // namespace boost
+
+// This include must be at the end, *after* the accessors
+// for this distribution have been defined, in order to
+// keep compilers that support two-phase lookup happy.
+#include <boost/math/distributions/detail/derived_accessors.hpp>
+
+#if defined(BOOST_MSVC)
+#pragma warning(pop)
+#endif
+
+#endif // BOOST_MATH_DIST_dirichlet_HPP
--- a/test/compile_test/dist_dirichlet_incl_test.cpp
+++ b/test/compile_test/dist_dirichlet_incl_test.cpp
@@ -1,0 +1,25 @@
+//  Copyright Mrityunjay Tripathi 2020.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Basic sanity check that header <boost/math/distributions/dirichlet.hpp>
+// #includes all the files that it needs to.
+//
+#include <boost/math/distributions/dirichlet.hpp>
+//
+// Note this header includes no other headers, this is
+// important if this test is to be meaningful:
+//
+#include "test_compile_result.hpp"
+
+void compile_and_link_test()
+{
+    TEST_DIST_FUNC(dirichlet)
+}
+
+template class boost::math::dirichlet_distribution<std::vector<double>, double, boost::math::policies::policy<>>;
+template class boost::math::dirichlet_distribution<std::vector<float>, float, boost::math::policies::policy<>>;
+#ifndef BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
+template class boost::math::dirichlet_distribution<std::vector<long double>, long double, boost::math::policies::policy<>>;
+#endif


mrityunjay-tripathi:
> Also, none of the distribution functions generate random samples of the corresponding distribution. That means no sampling function implemented. In PyTorch, we can do something like,

Ok, got it. It is implemented in ```boostorg/random``` repository.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+// The Dirichlet distribution is a family of continuous multivariate probability
+// distributions parameterized by a vector 'alpha' of positive reals.
+// It is a multivariate generalization of the dirichlet distribution, hence its
+// alternative name of multivariate dirichlet distribution (MBD).
+// Dirichlet distributions are commonly used as prior distributions in
+// Bayesian statistics, and in fact the Dirichlet distribution is the
+// conjugate prior of the categorical distribution and multinomial distribution.
+
+#ifndef BOOST_MATH_DIST_DIRICHLET_HPP
+#define BOOST_MATH_DIST_DIRICHLET_HPP
```
NAThompson: Could you make this BOOST_MATH_DISTRIBUTIONS_DIRICHLET_HPP? I recognize that the other files have gotten out of sync with this pattern over time, but that's what we want to move towards.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+#include <utility>
+
+namespace boost
+{
+namespace math
+{
+namespace dirichlet_detail
+{
+// Common error checking routines for dirichlet distribution function:
+template <class VectorType, class RealType, class Policy>
```
NAThompson: Generally, for classes that accept vectors, we do:

```
template<class RandomAccessContainer>
foo (RandomAccessContainer const & v, typename RandomAccessContainer::value_type x);
```
and use `using Real = typename RandomAccessContainer::value_type`, which forces the correct deduction.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+inline bool check_concentration(const char *function,
+                                const VectorType &concentration,
+                                RealType *result,
+                                const Policy &pol)
+{
+  for (size_t i = 0; i < concentration.size(); ++i)
+  {
+    if (!(boost::math::isfinite)(i) || (i <= 0))
```
NAThompson: `i <=0`? But `i` is a `size_t`. . . 

This will cause warning when used with `Eigen` or any other container that has signed indexes. Use `decltype(concentration.size()) i = 0; i < concentration.size()`. . .

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+
+template <class VectorType, class RealType, class Policy>
+inline bool check_prob(const char *function,
+                       const VectorType &p,
+                       RealType *result,
+                       const Policy &pol)
+{
+  for (size_t i = 0; i < p.size(); ++i)
+  {
+    if ((i < 0) || (i > 1) || !(boost::math::isfinite)(i))
```
NAThompson: Again, weird to check for `i < 0` here. Also make `i` the same type as `p.size()`.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+  }
+  return true;
+} // bool check_variance
+} // namespace dirichlet_detail
+
+template <class VectorType = std::vector<double>, class RealType = double, class Policy = policies::policy<>>
+class dirichlet_distribution
+{
+public:
+  dirichlet_distribution(VectorType concentration) : concentration(concentration)
```
NAThompson: This makes a copy; generally we don't like to copy containers. I'd make this a move constructor: `VectorType && concentration`; also, it's nice to use syntax that shows `concentration` is a member variable: `concentration_` or `m_concentration`.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+          function,
+          "variance argument is %1%, but must be > 0 !", i, pol);
+      return false;
+    }
+  }
+  return true;
+} // bool check_variance
+} // namespace dirichlet_detail
+
+template <class VectorType = std::vector<double>, class RealType = double, class Policy = policies::policy<>>
```
NAThompson: I wouldn't bother with the defaults here. Use `template<class RandomAccessContainer>`, and then use `using Real = typename RandomAccessContainer::value_type`.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+    RealType result;
+    dirichlet_detail::check_dist(
+        "boost::math::dirichlet_distribution<%1%>::dirichlet_distribution",
+        concentration,
+        &result, Policy());
+    sum_concentration = accumulate(concentration.begin(), concentration.end(), 0);
+  } // dirichlet_distribution constructor.
+
+  // Accessor functions:
+  VectorType Concentration() const
```
NAThompson: Isn't this accessor a big memcopy?

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+  {
+    return concentration;
+  }
+
+  size_t Order() const
```
NAThompson: Generally we want to stick with the snake case conventions.

pabristow:
Looks a really promising start.

It would be really nice if the Real defaults could be vector<double> and all other parameters double.
This would make it appear to work quite like C, Python, R etc, out of the box.

For testing, \boost\libs\math\test\test_beta_dist.cpp is a slightly useful template.  However it is probably over-complicated, at least to get started on testing.  I found it easier to start with testing just double values and then refactor later with other types and higher precision.   This will at least get some sanity checks.

 (Although the usefulness of high values for this distribution is unclear, our policy is to aim for tests at least the precision of 128-bit (~36 decimal digits)).

There is also the important question of test values.  Python could be used to generate some test values for a start.  https://cran.r-project.org/web/packages/DirichletReg/DirichletReg.pdf might also be slightly useful but is ominously marked as 'orphaned'  - but is  quite new.  Our favorite Wolfram doesn't seem to be able to help.

The beta distribution tests should be applicable when they should be identical, so this might be the first thing to try?

We also like to have some example(s).  Being able to show how to combine with Boost.Random to replicate the Wikipedia/Python example(s) would be good.

There are lots of tests of the that check for 'dodgy' input like negative, NaN or inf.  These check throwing (if the policy is default).   (and don't forget any examples should use thow'n'catch blocks so that the error messages are displayed). You can probably leave me to tidy these up later.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+  //   return ;
+  // } // RealType find_concentration(x, probability)
+
+private:
+  VectorType concentration; // https://en.wikipedia.org/wiki/Concentration_parameter.
+  RealType sum_concentration;
+}; // template <class RealType, class Policy> class dirichlet_distribution
+
+
+template <class VectorType, class RealType, class Policy>
```
NAThompson: Again, `template<class RandomAccessContainer>` and then `using Real = typename RandomAccessContainer::value_type`.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+}
+
+
+template <class VectorType, class RealType, class Policy>
+inline VectorType mean(const dirichlet_distribution<VectorType, RealType, Policy> &dist)
+{ // Mean of dirichlet distribution = c[i]/sum(c).
+  VectorType m;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    m.push_back(dist.concentration[i] / dist.sum_concentration);
```
NAThompson: Repeatedly calling `push_back` is bad; get it to the correct size in line 255 and then just fill it up.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+} // mean
+
+
+template <class VectorType, class RealType, class Policy>
+inline VectorType variance(const dirichlet_distribution<VectorType, RealType, Policy> &dist)
+{
+  VectorType v;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    v.push_back(dist.concentration[i] / dist.sum_concentration * (1 - dist.concentration[i] / dist.sum_concentration) / (1 + dist.sum_concentration));
```
NAThompson: Again, get rid of the push_back for performance reasons.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,405 @@
...
+  // Argument checks:
+  RealType result = 0;
+  if (!dirichlet_detail::check_dist_and_x(function, x, &result, Policy()))
+  {
+    return result;
+  }
+  RealType cumm = 1;
+  for (size_t i = 0; i < dist.Order(); ++i)
+  {
+    cumm *= std::pow(x[i], dist.concentration[i]) / tgamma(dist.concentration[i]) / dist.concentration[i];
```
NAThompson: `std::pow` won't compile in `float128` precision. Use: `using std::pow` and then call `pow`. Also have a unit test showing compatibility with `#include <boost/multiprecision/float128.hpp>`.

include/boost/math/distributions/dirichlet.hpp
NAThompson: BTW, in your unit tests, try to use both `std::vector` and an `Eigen` vector as the template type.

NAThompson:
Looks like a great start. I'll go through your other items on your checklist later.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: That makes sense.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: It's really silly, it should be ```concentration[i]```.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: Here, it should be```p[i]```.

mrityunjay-tripathi: seems like I need more sleep :D

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: Thanks, this solves my first checklist.

mrityunjay-tripathi:
Hi @pabristow @NAThompson Can we use something like ```validate_args``` to see if the user wants to check for argument validity. Currently, we are validating the arguments when an instance of the object is created which is inefficient.

pabristow:
I have a vague recollection that we had a complaint similar to this many years ago.  In the end we decided to keep the apparently redundant check because there were circumstances when it would allow a 'bad' value to sneak in.  Checking seems quite cheap, as is construction generally?

mrityunjay-tripathi:
Right. If some 'invalid' value comes in for reasons unknown to the user, it would be a disaster.

NAThompson:
@mrityunjay-tripathi Could you write a google benchmark and show how fast it is with and without the argument validation? This is a really hard question to answer without hard data. An example is [here](https://github.com/boostorg/math/blob/dc52a787c8eacd9d45d9d674a0d479e6f577ba1f/example/daubechies_wavelets/bench.cpp).

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -12,19 +12,20 @@
...
 // Bayesian statistics, and in fact the Dirichlet distribution is the
 // conjugate prior of the categorical distribution and multinomial distribution.
 
-#ifndef BOOST_MATH_DIST_DIRICHLET_HPP
-#define BOOST_MATH_DIST_DIRICHLET_HPP
+#ifndef BOOST_MATH_DISTRIBUTIONS_DIRICHLET_HPP
+#define BOOST_MATH_DISTRIBUTIONS_DIRICHLET_HPP
 
 #include <boost/math/distributions/fwd.hpp>
+#include <boost/multiprecision/float128.hpp>
```
NAThompson: Don't include `boost/multiprecision/float128.hpp` in this file. Anyone who wants to use the `float128` type will #include it in their code. For instance, it should be in your unit tests, but not here, because it'll slow compilation.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -43,350 +44,418 @@ namespace math
...
+  RandomAccessContainer &alpha() const { return m_alpha; }
+  // Set the concentration parameters.
+  RandomAccessContainer &alpha() { return m_alpha; }
 
-  size_t Order() const
-  {
-    return concentration.size();
-  }
+  // Get the order of concentration parameters.
+  decltype(m_alpha.size()) &order() const { return m_alpha.size(); }
```
NAThompson: This shouldn't be a reference.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -43,350 +44,418 @@ namespace math
...
     }
   }
   return m;
 } // mode
 
+template <class RandomAccessContainer, class Policy>
+inline typename RandomAccessContainer::value_type entropy(const dirichlet_distribution<RandomAccessContainer, Policy> &dist)
+{
+  using RealType = typename RandomAccessContainer::value_type;
+  RealType ent = std::log(dirichlet_detail::mvar_beta(dist.m_alpha, 0)) + (dirichlet_detail::alpha0(dist.m_alpha) - dist.order()) * digamma(dirichlet_detail::alpha0(dist.m_alpha));
```
NAThompson: `using std::log; log(dirichet_detail::...`

NAThompson:
@jzmaddock , @pabristow : For scalar distribution, using free functions to compute kurtosis, range, so on, is just fine. But for vector types, following this pattern is getting a bit awkward. Should each of these functions go from free functions acting on the class  to member functions?

NAThompson:
@mrityunjay-tripathi : We have a file "boost/libs/math/test/math_unit_test.hpp", which you can use to write unit tests. An example is "test/whittaker_shannon_test.cpp".

mrityunjay-tripathi:
>@mrityunjay-tripathi : We have a file "boost/libs/math/test/math_unit_test.hpp", which you can use to write unit tests. An example is "test/whittaker_shannon_test.cpp".

This is clear now. I will soon add the tests.

include/boost/math/distributions/dirichlet.hpp
pabristow: We don't test routinely float128 (because it is slow and GCC only).  When the whole package is working OK for the Big Three MSVC, GCC and Clang, we might run a one-off check.  Boost.Multipreciison cpp_bin_float_quad is entirely portable (if slower) but too slow for routine CI testing.  Lets make sure that we can walk first ;-)

mrityunjay-tripathi:
> @mrityunjay-tripathi Could you write a google benchmark and show how fast it is with and without the argument validation? This is a really hard question to answer without hard data. An example is [here](https://github.com/boostorg/math/blob/dc52a787c8eacd9d45d9d674a0d479e6f577ba1f/example/daubechies_wavelets/bench.cpp).

I am not familiar with 'google benchmark', but ready to delve into it. As soon as I add the tests, I will try to write the benchmark as well.

pabristow:
> @jzmaddock , @pabristow : For scalar distribution, using free functions to compute kurtosis, range, so on, is just fine. But for vector types, following this pattern is getting a bit awkward. Should each of these functions go from free functions acting on the class to member functions?

@jzmaddock devised this scheme - with good reasons - so he is best qualified to comment on this change to multi-thingys.

jzmaddock:
> For scalar distribution, using free functions to compute kurtosis, range, so on, is just fine. But for vector types, following this pattern is getting a bit awkward. Should each of these functions go from free functions acting on the class to member functions?

For better or worse I'd prefer to keep to the current pattern if possible.  Implementation and use wise, I think it makes very little practical difference.

I confess to a certain amount of hesitation in adding something that's not entirely mainstream as our first multivariate distribution.  I had always assume that the multivariate normal would come first, but no matter I guess.

With regard to the CDF, I note that multivariate normal has 2 different possible definitions, and I assume that is the case here also.  I think we should be careful about treading on shaky ground here.  Likewise, I assume there is no real notion of a quantile, not least because there is unlikely to be a unique vector of x-values corresponding to any given probability.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,474 @@
...
+  {
+    RealType result;
+    dirichlet_detail::check_dist(
+        "boost::math::dirichlet_distribution<%1%>::dirichlet_distribution",
+        alpha,
+        &result, Policy());
+  } // dirichlet_distribution constructor.
+
+  // Get the concentration parameters.
+  RandomAccessContainer &alpha() const { return m_alpha; }
```
jzmaddock: Should return a const-reference from a const member function here.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -0,0 +1,474 @@
...
+  // Set the concentration parameters.
+  RandomAccessContainer &alpha() { return m_alpha; }
+
+  // Get the order of concentration parameters.
+  decltype(m_alpha.size()) &order() const { return m_alpha.size(); }
```
jzmaddock: This returns a reference to a temporary - should return a value.

NAThompson:
> I am not familiar with 'google benchmark', but ready to delve into it. As soon as I add the tests, I will try to write the benchmark as well.

You will definitely enjoy adding google benchmark to your toolkit. It's amazing.

NAThompson:
> I confess to a certain amount of hesitation in adding something that's not entirely mainstream as our first multivariate distribution. I had always assume that the multivariate normal would come first, but no matter I guess.

I think the problem is that the multivariate normal factors into products so that there is little demand for such a thing. The useful definition of the cdf the translates into an ergonomic API is going to be a problem though . .

mrityunjay-tripathi:
> Likewise, I assume there is no real notion of a quantile, not least because there is unlikely to be a unique vector of x-values corresponding to any given probability.

@jzmaddock I thought of using some random values so that they satisfy the condition. But it seems there is no practical use for this. I was putting it for the sense of completeness of the implementation.

mrityunjay-tripathi:
>Writing down a useful definition of the cdf that translates into an ergonomic API is going to be a problem though . .

@NAThompson: I've used this deduction to calculate the CDF.

cdf
cdf1

Talking about efficiency, this distribution obviously is computationally expensive.

NAThompson:
```
./include/boost/math/distributions/dirichlet.hpp:53:20: error: need 'typename' before 'RandomAccessContainer::value_type' because 'RandomAccessContainer' is a dependent scope
   53 |   using RealType = RandomAccessContainer::value_type;
      |                    ^~~~~~~~~~~~~~~~~~~~~
      |                    typename
compilation terminated due to -Wfatal-errors.
```

mrityunjay-tripathi:
I was trying this [sample program from scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.dirichlet.html) but the values I am getting is not in sync with that of scipy :(
Not even the 'mean' is correct! What might be the reason?
```
#include <bits/stdc++.h>
#include <boost/math/distributions/dirichlet.hpp>
using namespace std;

void print(std::vector<long double> v)
{
  for (size_t i = 0; i < v.size(); ++i)
  {
    cout<<v[i]<<" ";
  }
  cout<<endl;
}
int main()
{
  std::vector<long double> alpha = {0.4, 5.0, 15.0};
  std::vector<long double> x = {0.2, 0.2, 0.6};
  
  using VectorType = typename std::vector<long double>;
  
  boost::math::dirichlet_distribution<VectorType> diri(std::move(alpha));
  long double p = pdf(diri, x);
  long double c = cdf(diri, x);
  std::vector<long double> m = mean(diri);
  std::vector<long double> v = variance(diri);
  std::vector<long double> std = standard_deviation(diri);
  long double en = entropy(diri);
  cout<<"pdf: "<<p<<endl;
  cout<<"cdf: "<<c<<endl;
  cout<<"entropy: "<<en<<endl;
  
  cout<<"mean: ";
  print(m);
  cout<<"variance: ";
  print(v);
  cout<<"statndard deviation: ";
  print(std);
}
```
The output:
```
pdf: 0.0203131
cdf: 6.90545e-05
entropy: -nan
mean: 0.02 0.25 0.75 
variance: 0.000933333 0.00892857 0.00892857 
statndard deviation: 0.0305505 0.0944911 0.0944911 
```

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -223,20 +199,20 @@ class dirichlet_distribution
   } // dirichlet_distribution constructor.
 
   // Get the concentration parameters.
-  RandomAccessContainer &alpha() const { return m_alpha; }
+  const RandomAccessContainer &get_alpha() const { return m_alpha; }
   // Set the concentration parameters.
-  RandomAccessContainer &alpha() { return m_alpha; }
+  RandomAccessContainer &set_alpha() { return m_alpha; }
```
NAThompson: What is the use case for `set_alpha()`? Isn't `alpha` set in the constructor?

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -223,20 +199,20 @@ class dirichlet_distribution
...
 
   // Get the order of concentration parameters.
-  decltype(m_alpha.size()) &order() const { return m_alpha.size(); }
+  size_t order() const { return m_alpha.size(); }
```
NAThompson: I would still use `decltype(m_alpha.size())` just to get rid of annoying sign compare warnings for RandomAccessContainers with signed indexes.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -248,9 +224,27 @@ class dirichlet_distribution
...
+    {
+      mb *= tgamma(m_alpha[i] + b);
+    }
+    mb /= tgamma(alpha_sum);
+    return mb;
+  } // normalizing_constant
+
+  RealType sum_alpha() const
+  {
+    return accumulate(m_alpha.begin(), m_alpha.end(), 0);
```
NAThompson: Should this be `std::accumulate`? `std::accumulate` works on any type with `+` defined; it doesn't need to use ADL for multiprecision types.

NAThompson:
Get your unit tests in, make them as simple and as meaningful as possible, and these errors will become obvious to you.

This is really coming together nicely; good work.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: It's not that important but provides a little flexibility. This helps in case ```m_alpha``` (concentration parameter) needs to be changed (or updated) on the go.

NAThompson: If you don't have an applications that needs this, I would definitely remove it. In fact, `set_alpha` is more expensive than generating a new object since instantiation is done by move.

mrityunjay-tripathi:
Ohh. Got it!
>  RealType sum_alpha() const
  {
    return accumulate(m_alpha.begin(), m_alpha.end(), 0);

The initializing value must be ```RealType```. Here I made the mistake of keeping it ```int```.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: No worries. I will remove it.

pabristow:
Something has gone wrong ;-)  (It often does...) 

Special cases have always proved good, especially if the result should be 'exact'.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.dirichlet.html

looks recent (Last updated on Dec 19, 2019). so likely to be correct?

After a little light googling, I also found a more recent bit of code in R and some tests that might be helpful?

https://github.com/bertcarnell/dirichlet/blob/7954c574e2bb1359883e8d4b810d266acc6c5a35/tests/testthat/test-dirichlet.r

  it("A two parameter Dirichlet is equivalent to a beta", {
    expect_equal(ddirichlet(c(0.3, 0.7), c(2, 3)), dbeta(0.3, 2, 3))

http://ugrad.stat.ubc.ca/R/library/MCMCpack/html/dirichlet.html  might be useful too?  Old but might be correct nonetheless?

HTH

mrityunjay-tripathi:
>After a little light googling, I also found a more recent bit of code in R and some tests that might be helpful?
https://github.com/bertcarnell/dirichlet/blob/7954c574e2bb1359883e8d4b810d266acc6c5a35/tests/testthat/test-dirichlet.r

@pabristow: This seems helpful. Thank you.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: that was throwing error (not in scope).

pabristow: I trust that your demo/test program has try'n'catch blocks so that throw error messages are visible, rather than unhelpfully bombing with abort

NAThompson: @pabristow : That only happens on Windows command line; in other shells the programs don't abort and do give helpful error messages.

NAThompson: @mrityunjay-tripathi: Sounds good.

NAThompson:
BTW, your code is a little verbose:

```cpp
  std::vector<long double> alpha = {0.4, 5.0, 15.0};
  std::vector<long double> x = {0.2, 0.2, 0.6};
  
  using VectorType = typename std::vector<long double>;
  
  boost::math::dirichlet_distribution<VectorType> diri(std::move(alpha));
```

Instead:

```
  std::vector<double> alpha = {0.4, 5.0, 15.0};
  std::vector<double> x = {0.2, 0.2, 0.6};
  auto diri = boost::math::dirichlet_distribution(std::move(alpha));
```

Note how long double doesn't help since your floating point literals are parsed as doubles.

include/boost/math/distributions/dirichlet.hpp
pabristow: We have had many complaints about not seeing the error messages, so not knowing what platform was being used, I thought it was worth repeating.  I'm sure Microsoft would say it was a feature, not a bug, which I suppose it is...

NAThompson: How about?

```cpp
auto order() const { return m_alpha.size(); }
```

mrityunjay-tripathi: That's nice.

mrityunjay-tripathi:
>Note how long double doesn't help since your floating point literals are parsed as doubles, and that the template argument can be deduced by the compiler.

How about using 'L' as suffix for those floating point literals?

NAThompson:
> How about using 'L' as suffix for those floating point literals?

Yes, that would fix the issue.

mrityunjay-tripathi:
```
#include <bits/stdc++.h>
#include <boost/math/distributions/dirichlet.hpp>
using namespace std;

template <typename T>
void print(std::vector<T> v)
{
  for (size_t i = 0; i < v.size(); ++i)
  {
    cout<<fixed<<setprecision(20)<<v[i]<<", ";
  }
  cout<<endl;
}
int main()
{
  using RealType = long double;
  using VectorType = typename std::vector<RealType>;
  
  VectorType alpha = {3.462347646231L, 5.2136450991334L, 15.341L};
  VectorType x = {0.03536252L, 0.28787621346L, 0.676761265L};
  
  boost::math::dirichlet_distribution<VectorType> diri(std::move(alpha));
  RealType p = pdf(diri, x);
  RealType c = cdf(diri, x);
  VectorType m = mean(diri);
  VectorType v = variance(diri);
  VectorType std = standard_deviation(diri);
  RealType en = entropy(diri);
  
  cout<<"alpha: ";
  print(diri.get_alpha());
  cout<<"x: ";
  print(x);
  
  cout<<"pdf: "<<p<<endl;
  cout<<"cdf: "<<c<<endl;
  cout<<"entropy: "<<en<<endl;
  
  cout<<"mean: ";
  print(m);
  cout<<"variance: ";
  print(v);
  cout<<"standard deviation: ";
  print(std);
  cout<<"Normalizing Constant, B: "<<diri.normalizing_constant()<<endl;
  cout<<"Sum of Concentration Parameters, a0: "<<diri.sum_alpha()<<endl;
  cout<<"Sum of quantiles: "<<std::accumulate(x.begin(), x.end(), 0.0)<<endl;

}  
```

The Output:
```
alpha: 3.46234764623100000009, 5.21364509913340000002, 15.34100000000000000016, 
x: 0.03536252000000000000, 0.28787621346000000000, 0.67676126499999999999, 
pdf: 6.12940449892651902141
cdf: 0.00015248845250954929
entropy: 93.39123000056522383439
mean: 0.14416241379342879160, 0.21708151201151955451, 0.63875607419505165389, 
variance: 0.00493183227490614376, 0.00679366744373401038, 0.00922360070303550949, 
standard deviation: 0.07022700531067905191, 0.08242370680656148689, 0.09603957883620434480, 
Normalizing Constant, B: 0.00000000084752204199
Sum of Concentration Parameters, a0: 24.01699274536439999983
Sum of quantiles: 0.99999999845999998360

```
I also cross-checked it with scipy. Its giving correct results now.

NAThompson:
What is the header `#include <bits/stdc++.h>` doing?

mrityunjay-tripathi:
>What is the header #include <bits/stdc++.h> doing?

Nothing xD. I have just the habit of putting it whenever writing a sketch code. (maybe bad practice)

NAThompson:
Cool. I would get rid of the header and start cleaning up the code you wrote there for an example (add it to (say) `examples/dirichlet_distribution.cpp`).

I'd also pass the vector by `const &`, and get rid of the `using namespace std;`.

pabristow:
I'm just checking this out using

Paul@hetpD MINGW64 /i/boost/libs/math (mrityunjay-tripathi/dirichlet)
$ git log
commit a73ff6b70ba66911ad09d2305a9e7818d55a348c (HEAD -> mrityunjay-tripathi/dirichlet)
Author: Mrityunjay Tripathi <mrityunjay2668@gmail.com>
Date:   Mon Mar 2 16:06:31 2020 +0530
    slight errors removed

I needed to remove

`using namespace std; `

 as this is not Boost good-practice on grounds of the risk of name clash.

(It might be OK in .cpp examples, but it is much better to avoid completely.  It always leads to trouble eventually, and can be very puzzling to trace).

and I needed to add

#include <iostream>
#include <numeric>

and use std::cout and std::endl; 

but it fails to compile

\boost\boost\math\distributions\dirichlet.hpp(235): error C4700: uninitialized local variable 'mb' used

I tried setting mb to zero (assuming it might be zeroed by default), but that still doesn't give anything like the results you show.  (MSVC doesn't do 80-but long doubles, only 64-bit but it should be nearly the same?).

 alpha: 3.46235 5.21365 15.341
  x: 0.0353625 0.287876 0.676761
  pdf: inf
  cdf: 6.28359e-06
  entropy: -inf
  mean: 0.150537 0.22668 0.667
  variance: 0.00532815 0.00730401 0.00925463
  standard deviation: 0.0729942 0.0854635 0.096201
  Normalizing Constant, B: 0
  Sum of Concentration Parameters, a0: 23
  Sum of quantiles: 1

What am I doing wrong?  (Old version of dirichlet.hpp).

mrityunjay-tripathi:
@pabristow: Yes, that one is old version of dirichlet.hpp
I removed errors locally and yet to push it.

NAThompson:
@mrityunjay-tripathi : If you haven't added any unit tests, tag your PR commit messages with `[CI SKIP]`.

include/boost/math/distributions/dirichlet.hpp
pabristow: But there are users  (Poor Peasants?) still in the Dark Ages of C++03.  Do we need to cut them off gratuitously just for an auto?  (or perhaps they are they cut off already?)

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -200,16 +200,14 @@ class dirichlet_distribution
...
 
   // Get the order of concentration parameters.
   size_t order() const { return m_alpha.size(); }
 
   // Get alpha from mean and variance.
   void find_alpha(
-      RandomAccessContainer &&mean,     // Expected value of mean.
-      RandomAccessContainer &&variance) // Expected value of variance.
+      RandomAccessContainer &mean,     // Expected value of mean.
+      RandomAccessContainer &variance) const// Expected value of variance.
```
mrityunjay-tripathi: ```const``` here needs to be removed. don't know what i was experimenting.

NAThompson:
I have slightly reworked your example:

```cpp
#include <iostream>
#include <numeric>
#include <boost/math/distributions/dirichlet.hpp>

template <typename RandomAccessContainer>
void print(RandomAccessContainer const & v)
{
   using Real = typename RandomAccessContainer::value_type;
   std::cout << std::fixed << std::setprecision(std::numeric_limits<Real>::digits10 + 2);
   std::cout << "{";
   for (size_t i = 0; i < v.size() - 1; ++i)
   {
      std::cout << v[i] << ", ";
   }
   std::cout << v.back() << "}\n";
}

int main()
{
   using Real = long double;
   
   std::vector<Real> alpha = {3.462347646231L, 5.2136450991334L, 15.341L};
   std::vector<Real> x = {0.03536252L, 0.28787621346L, 0.676761265L};
   
   auto diri = boost::math::dirichlet_distribution(std::move(alpha));
   Real c = cdf(diri, x);
   auto m = mean(diri);
   auto v = variance(diri);
   auto std = standard_deviation(diri);
   Real en = entropy(diri);
   
   std::cout << "alpha = ";
   print(diri.get_alpha());

   std::cout<< "x = ";
   print(x);
   
   std::cout << "pdf = " << pdf(diri, x) << "\n";

   std::cout << "cdf = " << c << "\n";
   std::cout << "entropy: " << en << "\n";
   
   std::cout << "mean: ";
   print(m);
   std::cout << "variance: ";
   print(v);
   std::cout << "standard deviation: ";
   print(std);
   std::cout << "Normalizing Constant, B: " << diri.normalizing_constant() << "\n";
   std::cout << "Sum of Concentration Parameters, a0: "<<diri.sum_alpha() << "\n";
   std::cout << "Sum of quantiles: " << std::accumulate(x.begin(), x.end(), 0.0L) << "\n";
}  
```

I have run this with `-fsanitize=address -fsanitize=undefined`, which is a very strong test, so things are looking good.

mrityunjay-tripathi:
>auto diri = boost::math::dirichlet_distribution(std::move(alpha));

I think this is what you intended?
```auto diri = boost::math::dirichlet_distribution<std::vector<Real>>(std::move(alpha));```

This is required since the default type was ```std::vector<double>``` and here we are using ```Real``` which is ```long double```.

**Comment by pabristow** [comment]
`std::setprecision(std::numeric_limits<Real>::max_digits10);`

is more correct and clearer than

`std::setprecision(std::numeric_limits<Real>::digits10 + 2);`

And I like to use it thus

```
//! Show value of z to the full possibly-significant max_digits10 precision of type T.
//! Restores io state on exit.
template<typename T>
void show_value(T z)
{
  std::streamsize precision = std::cout.precision(std::numeric_limits<T>::max_digits10);  // Save.
  std::cout.precision(std::numeric_limits<T>::max_digits10); // Show all posssibly significant digits.
  std::ios::fmtflags flags(std::cout.flags());
  std::cout.setf(std::ios_base::showpoint); // Include any trailing zeros.
  std::cout << z;
  // Restore:
  std::cout.precision(precision);
  std::cout.flags(flags);
} // template<typename T> void show_value(T z)

```
to show the implied precision with no zero-information trailing junk, only those that are noisy but possibly-significant, but including trailing zeros.

Pedants will rule the world, eventually!

**Comment by NAThompson** [comment]
> I think this is what you intended?
auto diri = boost::math::dirichlet_distribution<std::vector<Real>>(std::move(alpha));

This is required since the default type was std::vector<double> and here we are using Real which is long double.

It should compile (in C++17 mode at least, I think in C++14) without the explicit template argument, so removing the default type `std::vector<double>` will make the API more ergonomic.

@pabristow: Good catch on `max_digits10`.

File: include/boost/math/distributions/dirichlet.hpp (hunk 21 above)
Comment (NAThompson) [reply include/boost/math/distributions/dirichlet.hpp]: Why does `const` need to be removed here?

**Comment by pabristow** [comment]
Using MSVC latest c++14, I've now got 

(after adding #include <numeric> to dirichlet.hpp to find std::accumulate)
```

  dirichlet_basic_examples.cpp
  Generating code
  Previous IPDB not found, fall back to full compilation.
  All 421 functions were compiled because no usable IPDB/IOBJ from previous compilation was found.
  Finished generating code
  dirichlet_examples.vcxproj -> I:\Cpp\math\x64\Release\dirichlet_examples.exe
  Autorun "I:\Cpp\math\x64\Release\dirichlet_examples.exe"
  alpha: 3.4623476462310001 5.2136450991333998 15.340999999999999 
  x: 0.035362520000000001 0.28787621345999997 0.67676126500000000 
  pdf: 6.1294
  cdf: 0.000152488
  entropy: 93.3912
  mean: 0.14416241379342881 0.21708151201151957 0.63875607419505165 
  variance: 0.0049318322749061453 0.0067936674437340111 0.0092236007030355102 
  standard deviation: 0.070227005310679061 0.082423706806561498 0.096039578836204342 
  Normalizing Constant, B: 8.47522e-10
  Sum of Concentration Parameters, a0: 24.017
  Sum of quantiles: 1
```

Looks promising? 

(Will work on some Boost.test using these values - but we need more tests of course!)

**Comment by mrityunjay-tripathi** [comment]
>It should compile (in C++17 mode at least, I think in C++14) without the explicit template argument, so removing the default type std::vector<double> will make the API more ergonomic.

@NAThompson: Maybe in C++17. I am using C++14 and it throws this error. 
```
dirichlet_test.cpp:29:51: error: missing template arguments before ‘(’ token
    auto diri = boost::math::dirichlet_distribution(std::move(alpha));
```

File: include/boost/math/distributions/dirichlet.hpp (hunk 21 above)
Comment (mrityunjay-tripathi) [reply include/boost/math/distributions/dirichlet.hpp]: Because m_alpha is being assigned in this function but ```const``` forces it to be read-only.

**Comment by mrityunjay-tripathi** [comment]
Hi @pabristow @NAThompson: I've added tests (still more sample tests to be added, I was busy with my assignments). But I wanted to check if [this error](https://docs.google.com/document/d/1fDOplqMiwvP5XsMksxVuhGOAvQc4VPLS5uNMUwqzvzw/edit?usp=sharing) also comes during CI. I am worried if this turns out to be something stupid :(

**Comment by pabristow** [comment]
I've also been working up a basic example and checking that it works with Boost.Multiprecision.

with these patches

```
 #include <utility>
+#include <numeric>   <<<<<<<<<<<<<<< Needed to find std::accumulate
```
(The example.cpp happens to use accumulate, but this isn't always so.  Boost has uncovered countless examples of missing includes.  Always best to be explicit and not rely on includes in other files).

```
@@ -100,7 +101,7 @@ inline bool check_x(const char *function,
       return false;
     }
   }
-  return std::accumulate(x.begin(), x.end(), 0.0) <= 1.0;
+  return std::accumulate(x.begin(), x.end(), static_cast<RealType>(0.0)) <= 1.0;
 } // bool check_x

 template <class RandomAccessContainer, class Policy>
@@ -222,10 +223,10 @@ public:
     }
   } // void find_alpha

-  RealType normalizing_constant(RealType b = 0.0) const
+  RealType normalizing_constant(RealType b = static_cast<RealType>(0.0)) const
   {
     // B(a1,a2,...ak) = (tgamma(a1)*tgamma(a2)...*tgamma(ak)/tgamma(a1+a2+...+ak)
-    RealType mb = 1.0;
+    RealType mb = static_cast<RealType>(1);
     RealType alpha_sum = accumulate(m_alpha.begin(), m_alpha.end(), b * m_alpha.size());
     for (decltype(m_alpha.size()) i = 0; i < m_alpha.size(); ++i)
```

These static_casts cause a lot of clutter, but are essential for UDTs.

I can compute and pretty-print your SciPy sample values for the fundamental types, float, double ... and as well, float128 (on GCC only) cpp_bin_float_quad and cpp_dec_float_50.

Although I can't yet see a use for extravagant precisions like these, it is useful to confirm that the library can be use with UserDefinedTypes for which there are a typical example.

We don't routinely test (too slow) but a macro BOOST_MATH_TEST_MULTIPRECISION can be used to control, and/or they can be placed in a different in the B2 jamfile.

I'll look at your error messages separately.

**Comment by pabristow** [comment]
I’ve fired up your test using MSVC and I see immediately that you do not have A check around 

#include <boost/multiprecision/float128.hpp>

FLOAT128 is only possible on GCC, but not Clang not MSVC.

A way of dealing with this is to check 

#ifdef BOOST_HAS_FLOAT128

I’ve put

```
#ifdef BOOST_MATH_TEST_FLOAT128
#ifdef BOOST_HAS_FLOAT128
// Including this header below without float128 triggers:
// fatal error C1189: #error:  "Sorry compiler is neither GCC, not Intel, don't know how to configure this header."
#include <boost/multiprecision/float128.hpp>
using boost::multiprecision::float128;
#endif // ifdef BOOST_HAS_FLOAT128
#endif // #ifdef #ifdef BOOST_MATH_TEST_FLOAT128
```
A test of BOOST_MATH_TEST_FLOAT128   is also useful because of the time that float128 tests may take, it’s not worth testing routinely.  

2  As far as I remember, the macro defines for Boost.Test must come *before* the #include in order to have the desired effect.

// Defines must come before includes.
#define BOOST_TEST_MAIN
#define BOOST_MATH_CHECK_THROW BOOST_CHECK_THROW
#define BOOST_LIB_DIAGNOSTIC "on" // Report library file details.

#include <boost/test/unit_test.hpp>                       // for test_main

(aside Although it is not enforced, I believe that Boost best practice is to place the std:: includes after all the Boost ones, so that Boost takes precedence.)

#define BOOST_MATH_CHECK_THROW BOOST_CHECK_THROW

Is needed to get the real check command to work.  (Some platforms don’t have throws).

3  There are zillions of warnings from constants – very tiresomely, they all need casting to RealType to keep the compiler quiet.  Sigh!

4    in mode, the compiler didn’t like the return result for reasons I have yet to understand.  But it should not normally be executed, so I’ve commented it out to get going for the rest of the tests.

5  Below is using VS , I'll try using the b2 jamfile next...

`  if (dist.get_alpha()[i] <= 1)
    {
      result = policies::raise_domain_error<RealType>(
          function,
          "mode undefined for alpha = %1%, must be > 1!", dist.get_alpha()[i], Policy());
     // return result;
    }`


```
  test_dirichlet_dist.vcxproj -> I:\Cpp\math\x64\Release\test_dirichlet_dist.exe
  Autorun "I:\Cpp\math\x64\Release\test_dirichlet_dist.exe"
  Running 1 test case...
  Platform: Win32
  Compiler: Microsoft Visual C++ version 14.2
  STL     : Dinkumware standard library version 650
  Boost   : 1.73.0
I:/boost/libs/math/test/test_dirchlet_dist.cpp(126): error : in "test_main": exception std::domain_error expected but not raised
I:/boost/libs/math/test/test_dirchlet_dist.cpp(172): error : in "test_main": exception std::domain_error expected but not raised
I:/boost/libs/math/test/test_dirchlet_dist.cpp(75): error : in "test_main": difference{3.03688e-08} between boost::math::pdf(diri, x){0.0586615382185217} and pdf{0.058661539999999998} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(76): error : in "test_main": difference{6.21687e-06} between boost::math::cdf(diri, x){0.00071693445706326182} and cdf{0.00071692999999999995} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(77): error : in "test_main": difference{1.68492e-06} between boost::math::entropy(diri){17.674729780402942} and entropy{17.674700000000001} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(81): error : in "test_main": difference{5.74049e-09} between calc_mean[i]{0.69312878397890099} and mean[i]{0.69312878} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(82): error : in "test_main": difference{1.90732e-07} between calc_variance[i]{0.022781795654775592} and var[i]{0.022781800000000001} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(84): error : in "test_main": difference{9.79326e-09} between calc_skewness[i]{-0.4951551348491805} and skewness[i]{-0.49515513} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(81): error : in "test_main": difference{1.2966e-08} between calc_mean[i]{0.3068712160210989} and mean[i]{0.30687122} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(82): error : in "test_main": difference{1.90732e-07} between calc_variance[i]{0.022781795654775585} and var[i]{0.022781800000000001} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(84): error : in "test_main": difference{9.79322e-09} between calc_skewness[i]{0.49515513484916279} and skewness[i]{0.49515513} exceeds 2.2204460492503131e-09
I:/boost/libs/math/test/test_dirchlet_dist.cpp(126): error : in "test_main": exception std::domain_error expected but not raised
  
  *** 24 failures are detected in the test module "Master Test Suite"
I:/boost/libs/math/test/test_dirchlet_dist.cpp(172): error : in "test_main": exception std::domain_error expected but not raised
...
```
Don’t panic – most of these are due to too tight a tolerance.

Looking very promising despite these reports.

PS I don’t understand your compile errors, but try to sort out the above first?

File: include/boost/math/distributions/dirichlet.hpp [hunk 22]
```diff
@@ -50,6 +50,12 @@ inline bool check_alpha(const char *function,
                         const Policy &pol)
 {
   using RealType = typename RandomAccessContainer::value_type;
+  using std::invalid_argument;
+  if (alpha.size() < 1)
+  {
+    throw invalid_argument("Size of 'concentration parameters' must be greater than 0.");
+    return false;
```
NAThompson: Do we need both a throw and a `return false`? I'd just do `throw std::invalid_argument`.

include/boost/math/distributions/dirichlet.hpp
```diff
@@ -100,45 +92,33 @@ inline bool check_x(const char *function,
       return false;
     }
   }
-  return std::accumulate(x.begin(), x.end(), 0.0) <= 1.0;
+  return accumulate(x.begin(), x.end(), 0.0) <= 1.0;
```
NAThompson: `0.0` parses as a double precision value, no matter what your `RealType` is. Use `RealType(0)`. Also, use `std::accumulate`, rather than `accumulate`, since you don't need argument-dependent lookup here.

test/test_dirichlet_dist.cpp
```diff
@@ -0,0 +1,258 @@
...
+  x[0] = 0.5, x[1] = 0.5;
+  BOOST_CHECK_EQUAL(boost::math::pdf(dirichlet_distribution<V>(std::move(alpha)), x), static_cast<RealType>(1.0));
+
+  alpha[0] = 2.0, alpha[1] = 2.0;
+  x[0] = 0.5, x[1] = 0.5;
+  BOOST_CHECK_EQUAL(boost::math::pdf(dirichlet_distribution<V>(std::move(alpha)), x), static_cast<RealType>(1.5));
+
+  // Checking precalculated values on scipy.
+  // https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.dirichlet.html
+  alpha[0] = static_cast<RealType>(5.778238829L); alpha[1] = static_cast<RealType>(2.55821892973L);
```
NAThompson: Have these values been computed up to long double precision? If so it doesn't look like there's enough digits there. In fact, I don't think there are enough digits even for double precision.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: hmmm...

test/test_dirichlet_dist.cpp
mrityunjay-tripathi: Scipy's gamma function doesn't support some (high precision) data types. Yes, these are not long double or double.

include/boost/math/distributions/dirichlet.hpp
mrityunjay-tripathi: Oh yes.

pabristow:
I agree that the differences 1e-8ish look like using float precision?

We definitely need some long double precision (output to numeric_limits<long double>max_digits10 = 18 decimal digits) 

But before you change them all, I suspect we should also be using this

#include <boost/math/tools/test_value.hpp>  // for create_test_value and macro BOOST_MATH_TEST_VALUE.

to create the right type for testing with UDT like multiprecision.

For example

  BOOST_CHECK_CLOSE_FRACTION(  // Check -exp(-1) ~= -0.367879450 == -1
    lambert_w0(BOOST_MATH_TEST_VALUE(RealType, -0.36787944117144232159552377016146086744581113103176783450783680169746149574489980335714727434591964374662732527)),
    BOOST_MATH_TEST_VALUE(RealType, -1.),
    tolerance);

It's unfortunate that we can't use Wolfram to create spot values at really high precision, but they haven't this distribution implemented yet :-(

@jzmaddock Should we use BOOST_MATH_TEST_VALUE ?

pabristow:
If I increase the tolerance to float, the tests all pass, but there are two tests that don't throw when expected.

  alpha[0] = 1.26;
  alpha[1] = 2.99;
  x[0] = 0.5;
  x[1] = 0.75; // sum(x) > 1.0
  BOOST_MATH_CHECK_THROW(boost::math::pdf(dirichlet_distribution<V>(std::move(alpha)), x), std::domain_error);

and 

  alpha[0] = 1.56;
  alpha[1] = 4.00;
  x[0] = 0.31;
  x[1] = 0.71; // sum(x) > 1.
  BOOST_MATH_CHECK_THROW(boost::math::cdf(dirichlet_distribution<V>(std::move(alpha)), x), std::domain_error);

So almost all looking right, but imprecise tests.  Are there some parameter values that are expected to give exact results?  Beta_distribution?

pabristow:
I also wonder if we can dispense with passing a zero of the right type (required in the Dark Days of C++98) `test_spots(std::vector<float>(0.0F));`  and instead pass nothing, and be explicit 
by calling ` test_spots<std::vector<float> >();` ...   (This works OK)

(Since you have already upped the C++ standard version to 14, if not 17!)

Incidentally, is using c++14/17 really helpful - I'd much prefer to limit it to c++11 if possible?

@jzmaddock Boost.Math policy decision?  Your important views?

jzmaddock:
>Incidentally, is using c++14/17 really helpful - I'd much prefer to limit it to c++11 if possible?

It depends.... I think we need to mark C++03 as deprecated in the next release - that way we can start removing it from old code next year sometime (!).

For new code, I'm reasonably relaxed about the compiler requirements, although it would be nice to not require stuff gratuitously.  I guess recent GCC's are C++14 by default, as is MSVC so that seems like a reasonable target unless there's something in C++17 that essential?

jzmaddock:
>Should we use BOOST_MATH_TEST_VALUE ?

It's probably overkill unless you have extended precision values which you're certain are correct.

test/test_dirichlet_dist.cpp
NAThompson: Ok, I'd try to get some more digits here, maybe via Mathematica?

NAThompson: I'd also try to find some exact values.

pabristow:
Boost has to cope with systems without ‘throwability’ (eg embedded systems, real-time, IOT…).

Perhaps a vector containing a NaN?  But still thinking about this.

pabristow:
Sadly Dirichlet distribution in Wolfram says "Contribute this entry."  :-(

pabristow:
Correction of my bad - max_digits10 should be 21 for 80-bit long double, not 17 as I said before (that's only MSVC who can't be bothered to implement 80-bit long double).  But I have yet to discover who to make SciPy output all those digits. Still digits10 = 18 would be better than only 6 decimal digit output!

mrityunjay-tripathi:
>alpha[0] = 1.26;
alpha[1] = 2.99;
x[0] = 0.5;
x[1] = 0.75; // sum(x) > 1.0
BOOST_MATH_CHECK_THROW(boost::math::pdf(dirichlet_distribution(std::move(alpha)), x), std::domain_error);
and
alpha[0] = 1.56;
alpha[1] = 4.00;
x[0] = 0.31;
x[1] = 0.71; // sum(x) > 1.
BOOST_MATH_CHECK_THROW(boost::math::cdf(dirichlet_distribution(std::move(alpha)), x), std::domain_error);

@pabristow: I got it why this is the case. Actually the check for sum of x just returns false if the condition is not satisfied but doesn't throw the error.
```return accumulate(x.begin(), x.end(), 0.0) <= 1.0;``` (line 95).

>Boost has to cope with systems without ‘throwability’ (eg embedded systems, real-time, IOT…).

Never thought about that!
and Thank You for your constant support :)

**Comment by mrityunjay-tripathi** [comment]
Why do these tests pass even if there are several errors? Am I missing something?

**Comment by pabristow** [comment]
I don't think that the jamfile.v2 includes dirichlet tests or examples yet .

I will do this and check it out locally using the b2 build system with a variety of gcc clang and msvc versions and c++std variants too.  Job for next week.

(I think that it is premature to be using the CI suite yet.  I'm only testing locally.  When pushing, include the string [CI SKIP] in the commit message will avoid triggering the long-suffering and very tired and overworked CI systems.)

**Comment by pabristow** [comment]
I'm also trying to use beta distribution (which works with Boost.Multiprecision) for some really high precision tests.

https://github.com/bertcarnell/dirichlet/blob/7954c574e2bb1359883e8d4b810d266acc6c5a35/tests/testthat/test-dirichlet.r

says
  it("A two parameter Dirichlet is equivalent to a beta", {
    expect_equal(ddirichlet(c(0.3, 0.7), c(2, 3)), dbeta(0.3, 2, 3))

and this seems to work for PDF, but so far I get a different value for CDF.

I've been playing with

    //   dirichlet outputs             cdf: 0.061739999999999982  ??????????????
    // https://www.wolframalpha.com/input/?i=betaregularized%280.3%2C+0.3%2C+0.7%29
    // I0.3(0.3,0.7) = 0.6122328397117941119545947029213687501316859461148673
    // alpha 0.3, beta 0.7
    // (beta(0.3, alpha, beta) gamma(alpha + beta))/(gamma(alpha) gamma(beta))
    // betadistribution(0.3,0.7) 
    // mean  0.3, mode = 0.7, sd = 0.324037, variance = 0.105, skew = 0.822951 

    // N[PDF[betadistribution(0.3,0.7), 0.3], 55] =  0.6657228773303705085295919960559028800560346898849140226
    // N[PDF[betadistribution(2, 3), 0.3], 55] 1.764000000000000000000000000000000000000000000000000000
    // N[CDF[betadistribution(2, 3), 0.3], 55] 0.3483000000000000000000000000000000000000000000000000000

    // N[mode[betadistribution(2, 3), 0.3], 55] shows 
    // mean 0.4, mode = 0.3333, sd = 1/5 = 0.2
    // variance = 1/25 = 0.04, skewness = 2/7 = 0.285714
    // PDF is as expected, but not CDF whose plot shows about 0.35

I'm not sure if this is feasible and I suspect my ignorance and inexpertise.  Views from mathy experts?

I also would hope that one can do the equivalent of this for didchlet

    // Check beta distribution round-tripping:
    double cq = quantile(b, cdf(b, 0.3)); // 
    show_value(cq, "quantile(b, cdf(b, 0.3) = ", "\n"); // quantile(b, cdf(b, 0.3) = 0.29999999999999999
    BOOST_ASSERT_MSG(0.3000000000000000 == cq, "round_trip failed!");


    double qc = cdf(b, quantile(b, 0.3)); // 
    show_value(qc, "cdf(b, quantile(b, 0.3)) = ", "\n"); // cdf(b, quantile(b, 0.3)) = 0.30000000000000004
    BOOST_ASSERT_MSG(0.3000000000000000 == qc, "round_trip failed!");

How to do this for a multi distribution (in hand-waving mode ;-) )?

**Comment by mrityunjay-tripathi** [comment]
>and this seems to work for PDF, but so far I get a different value for CDF.

The deduction for CDF is definitely wrong and now it really seems complicated to deduce CDF of Dirichlet distribution though I will try to work on it.

**Comment by mrityunjay-tripathi** [comment]
I got [this answer on researchgate](https://www.researchgate.net/post/What_is_the_CDF_of_Dirichlet_Distribution) about CDF of Dirichlet distribution. Since the application for which I am using Dirichlet distribution is concerned about generating random distribution mostly. Can we omit CDF for now and wait if something (approximation or efficient deductions) comes up later.

**Comment by NAThompson** [comment]
@mrityunjay-tripathi : Why don't you try Cross-Validated and see if you get any better answers?

https://stats.stackexchange.com/

**Comment by pabristow** [comment]
For the record here is the helpful if negative reply.

"3rd Mar, 2020
Joachim Domsta
The State University of Applied Sciences in Elbląg,
Dear Mrityunjay Tripathi
do you mean the multivariate continuous Dirichlet distribution (extended beta)?
If yes, then do not be wondering, since then for the cpdf the formula is a very complicated combination of gamma multidimensional integrals. Moreover this is a pd concentrated on the simplex formed by points with positive coordinates summing up to 1In the case of 3D, this is concentrated on the triangle
x+y+z=1, where x,y,z >0.
What would you do with such function F(a,b,c) which is zero for all triples a,b,c, which sum up to value less than 1?
"

Sounds tricky, as my rusty maths quickly concluded.

File: test/test_dirichlet_dist.cpp [hunk 25]
```diff
@@ -28,24 +32,17 @@
...
     RandomAccessContainer &&mean,                       // mean
-    RandomAccessContainer &&mode,                       // mode
     RandomAccessContainer &&var,                        // variance
-    RandomAccessContainer &&skewness,                   // skewness
-    RandomAccessContainer &&kurtosis,                   // kurtosis
     typename RandomAccessContainer::value_type entropy, // entropy
     typename RandomAccessContainer::value_type pdf,     // pdf
-    typename RandomAccessContainer::value_type cdf,     // cdf
     typename RandomAccessContainer::value_type tol)     // Test tolerance.
 {
```
mrityunjay-tripathi: I removed ```mode```, ```skewness```, ```kurtosis```, ```cdf``` because I don't have predefined values for them. I thought of adding tests first for those that we have predefined values.

test/test_dirichlet_dist.cpp
```diff
@@ -54,183 +51,293 @@ void test_spot(
...
+  alpha[0] = static_cast<RealType>(5.310948003052013);
+  alpha[1] = static_cast<RealType>(8.003963132298916);
+  x[0] = static_cast<RealType>(0.35042614416132284);
+  x[1] = static_cast<RealType>(0.64957385583867716);
+  mean[0] = static_cast<RealType>(0.398872207937724);
+  mean[1] = static_cast<RealType>(0.601127792062276);
+  var[0] = static_cast<RealType>(0.016749888798155716);
+  var[1] = static_cast<RealType>(0.016749888798155716);
+  pdf = static_cast<RealType>(2.870121181949622);
+  entropy = static_cast<RealType>(-0.6347509574442718);
```
mrityunjay-tripathi: I couldn't get higher precision than this (float64 in NumPy), because gamma function in SciPy doesn't support float128.

pabristow:
I could generate some cpp_bin_float_50 values for this, but that is somewhat circular in that it assumes that your code is correct ;-)  
These values would also least be a check on regressions from ill-advised edits in future.

Cunning-er tests like Nick's are a Good Idea.  (Past experience is that it is all to easy to code these equations wrong :-( so cross checks are Really Useful)
We should also be able to compare the beta_distribution results ('Know Good' from comparison with Wolfram) with Dirichlet when k= 2?

I've asked again about the Dirichlet CDF
[https://stats.stackexchange.com/questions/57262/implementation-of-dirichlet-cdf/453526#453526](url)

pabristow:
We also like to have some worked examples of using functions and distributions, (often lifted from somewhere else so we can compare results).

Have you any examples yet?  Perhaps using with Boost.Random?

I am starting on documentation.

mrityunjay-tripathi:
Sorry, I couldn't get done much today because of all-day celebration of 'Holi'. 
>I could generate some cpp_bin_float_50 values for this, but that is somewhat circular in that it assumes that your code is correct ;-)

@pabristow: The ```pdf```, ```mean```, ```variance```, ```mode```, ```entropy``` are producing correct results now. I have to work on ```cdf``` and recheck ```skewness``` and ```kurtosis```. :)
>I've asked again about the Dirichlet CDF

I got the [equation for CDF](https://www.researchgate.net/publication/220461721_On_numerical_calculation_of_probabilities_according_to_Dirichlet_distribution) but it's toooo big. I am trying to implement it as soon as possible.
>We also like to have some worked examples of using functions and distributions, (often lifted from somewhere else so we can compare results).
Have you any examples yet? Perhaps using with Boost.Random?

Yes, I have but I am yet to get familiar with Boost.Random

>I am starting on documentation.

That's nice.

mrityunjay-tripathi:
>We should also be able to compare the beta_distribution results ('Know Good' from comparison with Wolfram) with Dirichlet when k= 2?

It seems unlikely that we will get the same results because the equation for cdf of dirichlet is much complex and has to be dealt using approximation (as far as I could get it).

pabristow:
https://www.researchgate.net/publication/220461721_On_numerical_calculation_of_probabilities_according_to_Dirichlet_distribution

looks intimidating to be an exercise for the student ;-)

mrityunjay-tripathi:
Sorry for the delay in the implementation of the CDF function, I've almost completed it but need to change the API accordingly. I got busy with the project proposal for GSoC. The approximation for Lauricella function mentioned in [this paper](https://www.researchgate.net/publication/220461721_On_numerical_calculation_of_probabilities_according_to_Dirichlet_distribution) is elaborated [here](https://link.springer.com/article/10.1007/s10444-014-9397-5). 
Sorry that I might take another few days  :(

pabristow:
Ah - that sounds extremely interesting and promising.  Approximations like these are really useful.
I have been trying to get my head around the Gouda and Szantai paper, but I have yet to get access to the Butler and Wood paper.  Comparing with beta distribution will give some good sanity checks on your code.

mrityunjay-tripathi:
I've removed std::invalid_argument almost everywhere, I will see if it is still used somewhere.

Also, my Institute is closed and I am on my way to home due to this whole Corona thing. I too can't access them right now, though I downloaded some of the papers (if it's that what you want) and can send you if you say.

pabristow:
I've searched better and found the article, but not even partially digested it.

I'm still struggling to re-understand how the error_handling works - don't worry about it for now.  Getting the cdf to work would be excellent.

I've download lauricella.zip from http://faculty.smu.edu/rbutler
Ronald Butler  C.F. Frensley Professor of Mathematical Sciences Department of Statistical Science
Southern Methodist University Dallas

contains a pdf for Laplace Approximation of Lauricella Functions  FA, FD, and Φ2

lauricella_NM2.pdf contains a section on 3.1.1 explaining how the there is an error in the Exton paper (equation 10)

We may be able to get help from Andrew Wood at Nottingham, UK.  This is above my 'pay grade' ;-)

http://www.smu.edu/statistics/faculty/butler.html is is 404, but the download above has some .m programs that might be useful?

HTH

Good luck avoiding the virus - at least you are not in the Death Age Zone!

evanmiller:
For test values, [Selected Tables in Mathematical Statistics, Volume IV (1977)](https://books.google.com/books?id=SIoaNt7WoGQC&pg=PP1&lpg=PP1&dq=selected+tables+in+mathematical+statistics+volume+4&source=bl&ots=iptkABzEes&sig=ACfU3U3lv6kHZ2L-cr31YrCmxr_0lEJscg&hl=en&sa=X&ved=2ahUKEwiGgpXa4YzqAhWDTN8KHZqtCJ4Q6AEwBXoECAcQAQ#v=onepage&q&f=false) has 200 pages of numerical tables dedicated to the Dirichlet distribution's CDF, inverse, mean, and variance. The book also includes a lengthy exposition on how to compute them.

pabristow:
Thanks for this.  Covid-19 and other things have pushed the Dirichlet onto the back burner, but this looks a very useful reference.  Library loans are not functionally at present, but I hope to return to this in due course.

PR #320, opened 2020-03-02
Fix typo in lambert w diagnostic output constant

### LINKED ISSUE #319: Typo in lambert w diagnostic output constant
There is a small typo in a constant used for diagnostic output in `special_functions/lambert_w.hpp` 
```C++
#ifdef BOOST_MATH_INSTRUMENT_LAMBERT_W_TERMS
  {
    int terms = 20; // Default to using all terms.
    if (absp < 0.001150)
    { // Very near singularity.
      terms = 6;
    }
```
Should read `absp < 0.01159` instead

Gattag:
Fixes #319

### DIFF
--- a/include/boost/math/special_functions/lambert_w.hpp
+++ b/include/boost/math/special_functions/lambert_w.hpp
@@ -363,7 +363,7 @@ T lambert_w_singularity_series(const T p)
 #ifdef BOOST_MATH_INSTRUMENT_LAMBERT_W_TERMS
   {
     int terms = 20; // Default to using all terms.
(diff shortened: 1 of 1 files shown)


NAThompson:
Thanks! I'll let @pabristow weigh in before merging, but this looks correct to me.

Gattag:
@pabristow Yeah I really shouldn't have found this, but I wanted to see an implementation of Fukushima's algorithm of approximation and found this code base with it and just luckily noticed that the constants were not the same when reading through it

pabristow:
You need to get out more? ;-)

PR #324, opened 2020-03-10
Bugfix/exp sinh quardature complex non standard range error

CGJackson:
I found a bug where the exp_sinh integrator failed to compile when used to integrate function returning a complex number over a non-native range (that is not from 0 to infinity).

This turned out to be due to the functions which mapped the integral back onto the native range requiring the function being integrated to return the same type that was used for the integration range, rather than the type inferred from the integrand. 

I have fixed the bug and added some tests to cover the case of a complex valued function being integrated over a non-native range.


NAThompson:
@CGJackson : Huge thanks for this!

PR #326, opened 2020-03-21
Accept b<= a in quadrature.

### LINKED ISSUE #325: Accept b = a in quadrature, return zero
If b=a in quadrature routines, then:

```
libc++abi.dylib: terminating with uncaught exception of type boost::wrapexcept<std::domain_error>: Error in function tanh_sinh<double>::integrate: Arguments to integrate are in wrong order; integration over [a,b] must have b > a.
```

For iterating over quadratures of different domains, it might be easier to allow b = a and just return 0.

Also, if b < a, maybe return -integral from b to a.

NAThompson:
Fixes #325

### DIFF
--- a/include/boost/math/quadrature/gauss.hpp
+++ b/include/boost/math/quadrature/gauss.hpp
@@ -1258,9 +1258,13 @@ public:
 
          if ((boost::math::isfinite)(a) && (boost::math::isfinite)(b))
          {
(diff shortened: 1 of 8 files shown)


NAThompson:
The CI was due to a Travis problem; will fix directly on develop if something else arises.

Issue #296, opened 2020-01-06
Circular distributions


ghost:
May math knowledge is quite limited. But in the past, I more than once had to deal with circular data. I usually use a Von Mises distribution to approximate my data. Having support in Boost.Math would have been very helpful back then.

NAThompson:
Not an unreasonable request. Looks like all the building blocks are already in Boost.Math. Got a number of other things to do at the moment, but I'll try to get this added as time permits.

pcjmuenster:
Is there anything I can do to assist you?

NAThompson:
@pcjmuenster : If you want to submit a PR I could help with the details. I'd probably just `cp normal.hpp von_mises.hpp` and just make the necessary adjustments to make it a Von Mises distribution.

NAThompson:
Looking at the [Von Mises distribution wiki](https://en.wikipedia.org/wiki/Von_Mises_distribution), the mean, variance, mode, pdf and entropy are straightforward. The only one that would require significant testing is the CDF.

Also a decision would have to be made about the `support()`, since `[-pi,pi]` is just as good as (say) `[0, 2pi]`.

pcjmuenster:
@NAThompson Thanks, I will try to look into it this weekend.

pcjmuenster:
Notes in case of my unforeseen death: 
For large values of kappa, the naive computation of the PDF fails since both exp(kappa) and bessel_i0(kappa) become quite large. For instance with kappa=100, the values exceed the range of single precision floating point numbers. However the maximum of such a distribution is still a fairly small number (cf. [Wolfram Alpha](https://www.wolframalpha.com/input/?i=von+mises+distribution+for+mu%3D0%2Ckappa%3D100%2Cx%3D0)). 
In a first attempt, I tried to approximate the logarithm of bessel_i0(kappa) instead and exponentiate the difference. I used the [Hankel Expansions](https://dlmf.nist.gov/10.40) for this. Boost currently uses different sets of polynomials to approximate the Bessel functions but doesn't yet include functions to directly retrieve their logarithms. For large kappa these could be trivially adapted from the polynomial approximations.
Suppose we use a similar a approach to calculate the variance. The quotient of bessel_i1 and bessel_i0 converges fast to 1. This could lead to a loss of precision when we substract the result from 1. I have yet to figure out how severe that is. (A simple test showed that under the assumption we could compute the logarithms of the Bessel functions exactly – which we can not – the error is 0.05% from the true variance.)

[...]

PR #330, opened 2020-03-27
Von Mises Distribution

pcjmuenster:
My trial to implement a circular distribution #296. Basic moments, PDF and CDF work for small values of κ, i. e. the concentration parameter. For larger values the used methods are too imprecise.


NAThompson:
@pcjmuenster : My apologies for not being able to look at this. We've got some work to do before the release; will probably not be able to look at this until after.

include/boost/math/distributions/von_mises.hpp
```diff
@@ -0,0 +1,842 @@
+//  Copyright John Maddock 2006, 2007.
+//  Copyright Paul A. Bristow 2006, 2007.
+
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_STATS_VON_MISES_HPP
+#define BOOST_STATS_VON_MISES_HPP
```
NAThompson, line 9: I think you've picked up an artefact of an old directory structure from other files. I'd change this to `BOOST_MATH_DISTRIBUTIONS_VON_MISES_HPP`.

include/boost/math/distributions/von_mises.hpp
```diff
@@ -0,0 +1,842 @@
+//  Copyright John Maddock 2006, 2007.
+//  Copyright Paul A. Bristow 2006, 2007.
```
NAThompson, line 2: Your name should be on the copyright.

include/boost/math/distributions/von_mises.hpp
```diff
@@ -0,0 +1,842 @@
...
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_STATS_VON_MISES_HPP
+#define BOOST_STATS_VON_MISES_HPP
+
+// https://en.wikipedia.org/wiki/Von_Mises_distribution
+// From MathWorld--A Wolfram Web Resource.
+// http://mathworld.wolfram.com/VonMisesDistribution.html
+
+#include <boost/math/distributions/fwd.hpp>
```
NAThompson, line 15: Can we remove the forward header? @jzmaddock  or @pabristow might have a better idea that I about why this is necessary but in general I find it causes problems.

include/boost/math/distributions/von_mises.hpp
```diff
@@ -0,0 +1,842 @@
...
+  //
+  RealType m_mean;  // distribution mean or location.
+  RealType m_concentration;    // distribution standard deviation or scale.
+}; // class von_mises_distribution
+
+typedef von_mises_distribution<double> von_mises;
+
+#ifdef BOOST_MSVC
+#pragma warning(push)
+#pragma warning(disable:4127)
```
NAThompson, line 75: Could you add a comment on why this warning should be disabled?

include/boost/math/distributions/von_mises.hpp
```diff
@@ -0,0 +1,842 @@
...
+  if(conc < 7.75)
+  {
+    RealType bessel_i0 = cyl_bessel_i(0, conc, Policy());
+    return exp(conc * cos(x - mean)) / bessel_i0 / boost::math::constants::two_pi<RealType>();
+  }
+  else if(conc < 50)
+  {
+    // Polynomial coefficients from boost/math/special_functions/detail/bessel_i0.hpp
+    // > Max error in interpolated form: 5.195e-08
+    // > Max Error found at float precision = Poly: 8.502534e-08
```
NAThompson, line 124: You are using `float` precision to do the interpolation, but allow result to be an arbitrary precision. Is there a better way?

include/boost/math/distributions/von_mises.hpp
```diff
@@ -0,0 +1,842 @@
...
+// float version of pdf_impl
+template <class RealType, class Policy>
+inline RealType pdf_impl(const von_mises_distribution<RealType, Policy>& dist, RealType x,
+                         const boost::integral_constant<int, 24>&)
+{
+  RealType mean = dist.mean();
+  RealType conc = dist.concentration();
+
+  BOOST_MATH_STD_USING
+  if(conc < 7.75)
```
NAThompson, line 115: I guess `conc < 7.75` is a cutoff of an asymptotic expansion? Could you add a comment if so?

include/boost/math/distributions/von_mises.hpp
```diff
@@ -0,0 +1,842 @@
...
+  // "Algorithm 518: Incomplete Bessel Function I0. The Von Mises Distribution", 1977, ACM
+  // DOI: 10.1145/355744.355753
+  RealType result = 0;
+  
+  int digits = std::numeric_limits<RealType>::max_digits10 - 1;
+  RealType ck = ((0.1611*digits - 2.8778)*digits + 18.45)*digits - 35.4;
+  if (conc > ck) {
+    RealType c = 24.0 * conc;
+    RealType v = c - 56;
+    RealType r = sqrt((54.0 / (347.0 / v + 26.0 - c) - 6.0 + c) / 12.0);
```
NAThompson, line 363: Lots of double precision constants here. Is arbitrary precision not possible?

test/test_von_mises.cpp
```diff
@@ -0,0 +1,379 @@
+// Copyright Paul A. Bristow 2010.
```
NAThompson, line 1: Again, your name should be here.

test/test_von_mises.cpp
```diff
@@ -0,0 +1,379 @@
...
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// test_von_mises.cpp
+
+// https://en.wikipedia.org/wiki/Von_Mises_distribution
+// From MathWorld--A Wolfram Web Resource.
+// http://mathworld.wolfram.com/VonMisesDistribution.html
+
+#include <pch.hpp> // include directory /libs/math/src/tr1/ is needed.
```
NAThompson, line 15: Do you need the entire `pch.hpp`? If not I'd just include what you need.

include/boost/math/distributions/von_mises.hpp
pabristow: Agree.

include/boost/math/distributions/von_mises.hpp
pabristow: definitely

include/boost/math/distributions/von_mises.hpp
pabristow: MSVC used to warn on anything that was constant.

// C 4127 "conditional expression is constant"

They have recently stopped doing this, at last, thank goodness!

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127?view=vs-2019

But we still test antediluvian versions of MSVC, so still worth keeping to reduce clutter?

include/boost/math/distributions/von_mises.hpp
pcjmuenster: As for the CDF, the mentioned paper gives constants depending on the precision for the values of `ck`, `a1`–`a4` and `v`. However, they are only for upto 12 decimal digits. I tried to extrapolate these values but that wasn't quite successful. I have looked around for implementations in other libraries and they all seem to use this old paper and the values for 8 digits (single precision).

pcjmuenster:
> Could you write a google benchmark for each, run it under AddressSanitizer and UBSan, and then do a couple plots?

What kind of benchmarks do you have in mind?

NAThompson:
@pcjmuenster : [Here's](https://github.com/boostorg/multiprecision/pull/203) a google benchmark for the GCD in multiprecision. I'd do a float and double test for the pdf and cdf.

include/boost/math/distributions/von_mises.hpp
pabristow: That's a little disappointing (though probably good enough for most users).

https://www.rdocumentation.org/packages/circular/versions/0.4-93 R documents mention this but I haven't found/downloaded the source code yet, but perhaps you already have?

These fitting functions are a Black Art but I know the Chief Wizard and perhaps he can advise @jzmaddock .

include/boost/math/distributions/von_mises.hpp
pcjmuenster: I don't understand what you are referring to. `pdf_impl` is overloaded depending on the precision.

include/boost/math/distributions/von_mises.hpp
jzmaddock: I do wonder if we can do better here:

1) I notice you've pulled out and duplicated the contents of I0 - why not just call that function directly?
2) Looking at the original paper, the series to evaluate for the CDF is given on the start of page 280 and everything else is just "faff" to try to evaluate that.  I confess it's going to be a nasty one to code up because the Bessel function recurrences are stable in the backwards not forwards direction.  But it would be true arbitrary precision.

I faced something similar with some of the 1F1 series expansions, and the way I coded it was to maintain a cache of Bessel function values, when filling the cache you start with an arbitrary small value, then iterate to fill the cache and then normalise all the values afterwards either using the last value from the previous cache or a single Bessel function call.  Which is not easy to explain and may or may not make sense!  

The code in question is `hypergeometric_1F1_AS_13_3_7_tricomi_series::refill_cache` in https://github.com/boostorg/math/blob/8eb40afd6b2cdc3ffa786d4494bfc4164794e8d4/include/boost/math/special_functions/detail/hypergeometric_1F1_bessel.hpp#L153.  Note that this is way over-complex for your case because it deals with both Bessel I and J, and positive and negative orders, neither of which are needed here.

pcjmuenster: The reason to duplicate the code was that I0(kappa) grows very fast just like e^kappa, dividing these two however leads to numbers < 100. So, in order to make the code work for large values of kappa, I calculate the quotient directly via polynomial approximation.

test/test_von_mises.cpp
pcjmuenster: Not necessary at all.

pcjmuenster:
> 
> 
> I'd do a float and double test for the pdf and cdf.

Until now, I manually typed in the formula into Wolfram|Alpha to get a correct value. In order to make a plot, I'd have to retrieve hundreds or thousands of these. WA has an API to automate these kind of things but unfortunately caching the results is against the ToS. Do you another way to get the exact values? (The papers I read are so old, they only have 4 digits of precision)

NAThompson:
> Until now, I manually typed in the formula into Wolfram|Alpha to get a correct value. In order to make a plot, I'd have to retrieve hundreds or thousands of these. 

No don't do that.

I'd do the following:

- Test the evenness of the PDF f(-x) = f(x) for various kappa.

- Test the oddness of the CDF about y = 0.5 for various kappa.

- Test quadratures of the PDF and see that it agrees with the CDF. I think `tanh_sinh` should work well here, though over the entire domain use `trapezoidal`.

- Test moments of the distribution; it looks like they can be evaluated as quadratures and then tested via modified Bessel functions.

include/boost/math/distributions/von_mises.hpp
NAThompson: > The reason to duplicate the code was that I0(kappa) grows very fast just like e^kappa, dividing these two however leads to numbers < 100. So, in order to make the code work for large values of kappa, I calculate the quotient directly via polynomial approximation.

Can you just test for values of kappa that cause overflow and then use an asymptotic there?

pcjmuenster:
> 
> 
> > Until now, I manually typed in the formula into Wolfram|Alpha to get a correct value. In order to make a plot, I'd have to retrieve hundreds or thousands of these.
> 
> No don't do that.
> 
> I'd do the following:
> 
>     * Test the evenness of the PDF f(-x) = f(x) for various kappa.
> 
>     * Test the oddness of the CDF about y = 0.5 for various kappa.
> 
>     * Test quadratures of the PDF and see that it agrees with the CDF. I think `tanh_sinh` should work well here, though over the entire domain use `trapezoidal`.
> 
>     * Test moments of the distribution; it looks like they can be evaluated as quadratures and then tested via modified Bessel functions.

So shall I plot the difference between say pdf(mu - x) and pdf(mu + x) and not between the calculated value and some yet to be retrieved exact value?

pabristow:
In the past we have used a modest number of spot values using Mathematica or some other 'known good' source.  Plot are pretty, but in the end they may not tell you much beyond the obvious over nearly all the range.  

But we have tried to include some spot values that are 'edge and corner' cases too.   These may reveal where a better algorithm would be useful (or has been used).

NAThompson:
> So shall I plot the difference between say pdf(mu - x) and pdf(mu + x) and not between the calculated value and some yet to be retrieved exact value?

Don't do it as a plot, do:

```
#include "math_unit_test.hpp"

template<typename Real>
void test_even()
{
     Real mu = ...;
     Real x = 0;
     Real dx = ...;
     while (mu + x < pi) {
             CHECK_ULP_CLOSE(pdf(mu+x), pdf(mu-x), 4);
              x+= dx;
     }
}
```

pcjmuenster:
> ```
> #include "math_unit_test.hpp"
> 
> template<typename Real>
> void test_even()
> {
>      Real mu = ...;
>      Real x = 0;
>      Real dx = ...;
>      while (mu + x < pi) {
>              CHECK_ULP_CLOSE(pdf(mu+x), pdf(mu-x), 4);
>               x+= dx;
>      }
> }
> ```

Switching from BOOST_CHECK_CLOSE to CHECK_ULP_CLOSE revealed that the symmetry doesn't increase with the precision. Everything's fine for `float` but for `double` the ULPs off are about 1e+6.

NAThompson:
> Everything's fine for float but for double the ULPs off are about 1e+6.

Yeah ULPs tests are pretty unforgiving! Try creating an [ulps plot](https://github.com/boostorg/math/blob/develop/include/boost/math/tools/ulps_plot.hpp) in float precision and you might be able to see where things are going wrong.

pabristow:
I'm currently playing with Nick's ulps toy (looks really helpful) and I'll put your distribution on my todo-next list.

pcjmuenster:
Appearantly the problem was the imprecision of the calculation of `mu - x` and `mu + x`. After switching to inputs that have a finite binary representation, PDF is now down to 2 ULPs and CDF to 32.

test/test_von_mises.cpp
```diff
@@ -330,11 +331,29 @@ void test_spots(RealType)
...
   // Error tests:
   check_out_of_range<boost::math::von_mises_distribution<RealType> >(0, 1); // (All) valid constructor parameter values.
-    
+
   BOOST_MATH_CHECK_THROW(quantile(von_mises_distribution<RealType>(0, 1), -1), std::domain_error);
   BOOST_MATH_CHECK_THROW(quantile(von_mises_distribution<RealType>(0, 1), 2), std::domain_error);
 } // template <class RealType>void test_spots(RealType)
 
+template <typename RealType>
+void test_symmetry(RealType)
```
NAThompson, line 340: This doesn't need an argument does it?

test/test_von_mises.cpp
```diff
@@ -330,11 +331,29 @@ void test_spots(RealType)
...
+{
+	RealType const pi = boost::math::constants::pi<RealType>();
+	RealType delta = 1.0 / (1 << 4);
+  for (RealType mean = 0; mean < pi; mean += delta) {
+    for (RealType conc = 0; conc < 100; conc = (conc + 1) * 1.5 - 1) {
+      von_mises_distribution<RealType> dist(mean, conc);
+      for (RealType x = 0; x < pi; x += delta) {
+        CHECK_ULP_CLOSE(pdf(dist, mean + x),
+                        pdf(dist, mean - x), 2);
```
NAThompson, line 349: Do you get 2 ULP accuracy in long double as well?

test/test_von_mises.cpp
```diff
@@ -330,11 +331,29 @@ void test_spots(RealType)
...
+        CHECK_ULP_CLOSE(cdf(dist, mean + x) - static_cast<RealType>(0.5),
+                        static_cast<RealType>(0.5) - cdf(dist, mean - x), 32);
```
NAThompson, line 351: This is a great test.

test/test_von_mises.cpp
```diff
@@ -351,8 +370,13 @@ BOOST_AUTO_TEST_CASE( test_main )
   // (Parameter value, arbitrarily zero, only communicates the floating point type).
   test_spots(0.0F);   // Test float. OK at decdigits = 0 tolerance = 0.0001 %
   test_spots(0.0);    // Test double. OK at decdigits 7, tolerance = 1e07 %
+  // Check symmetry of PDF and CDF
+  test_symmetry(0.0F);
+  test_symmetry(0.0);
+
 #ifndef BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
```
NAThompson, line 377: I think we can now assume long double support; this macro was for older systems.

test/test_von_mises.cpp
```diff
@@ -351,8 +370,13 @@ BOOST_AUTO_TEST_CASE( test_main )
...
-  //test_spots(0.0L); // Test long double.
+  test_spots(0.0L); // Test long double.
+  //test_symmetry(0.0L);
 #if !BOOST_WORKAROUND(__BORLANDC__, BOOST_TESTED_AT(0x0582))
   //test_spots(boost::math::concepts::real_concept(0.)); // Test real concept.
```
NAThompson, line 381: These lines can be deleted.

test/test_von_mises.cpp
```diff
@@ -363,7 +387,7 @@ BOOST_AUTO_TEST_CASE( test_main )
       "to pass.</note>" << std::endl;
 #endif
 
-   
+
```
NAThompson, line 390: Could you add a multiprecision test? Template it on float128 should be sufficient.

NAThompson:
Could you drop

- an ulps plot of the pdf and cdf into `doc/graphs`?
- a google benchmark into `reporting/performance`?

test/test_von_mises.cpp
pcjmuenster: I just used the same style that also `test_spots` uses.

pabristow:
I think I remember that passing zero as in test_symmetry(0.0); is a left-over from the dark days only just in the current millennium when templates didn't work properly on all compilers.

(But some of the testers may still using these old compilers/platforms, so we haven't been through to rationalise it.  I suspect the those that are marked as requiring C++11 may not try anyway?

"If it ain't broke, don't fix it" applies?  )

test/test_von_mises.cpp
pcjmuenster: Appearently necessary for some architectures …

test/test_von_mises.cpp
pcjmuenster: Yes.

pcjmuenster:
> 
> 
> Could you drop
> 
>     * an ulps plot of the pdf and cdf into `doc/graphs`?
> 
>     * a google benchmark into `reporting/performance`?

I wasn't sure how to use the ULP plot function in Boost.Math to plot a function with two variables.
So I made a color-coded one instead. What do you think? The red areas have an error of 260 ULPs.

ulp_plot_von_mises_pdf_float

NAThompson:
@pcjmuenster : Looks pretty good. For our own ability to reproduce this later though, could you fix the concentration parameter to (say) 64 and then use the univariate ulps plots in `boost/math/tools`? Put the file you write in `reporting/accuracy/von_mises_ulps.cpp`.

NAThompson:
> The ULP plots for the PDF look okay-ish, but the ones for the CDF clearly indicate that the cancelling out of significant digits at the end due to the formula causes severe imprecisions.

Can you post a 1D ULPs plot in the "bad region" you are finding? The condition number envelope should tell us if those errors are acceptable or not.

PR #331, opened 2020-03-28
Improvements for math_unit_test

NAThompson:


NAThompson:
I'm going to vote for no since I don't want to step on boost.test toes. In addition it makes the macro pretty verbose.

NAThompson:
@jzmaddock : Is this a post-release merge or is it low enough risk to go in now?

pabristow:
I don't think that this treads on Boost.Test toes at all - it has BOOST_MATH_ prefix. There is a long standing convention that all macro names used in boost are BOOST_ prefixed, and usually with a library specific like BOOST_MATH_  (Having been bitten by past experience, worse of all MAX and MIN by Microsoft) But I agree it is rather long.  Can we think of a shorter name? 

I like the concept a lot., even if we might refine it in the light of experience.  I'm doing some experiments with it.

PR #333, opened 2020-03-28
Better argument promotion for float_next

### LINKED ISSUE #332: float_distance with arg promotion has surprising behavior
Allowing two different template arguments in float_distance gives very surprising behavior, since it's not clear which floating point type the user intends to calculate the distance between.

If one type is integral, then this is sensible: Promote it to the floating point type. But if both are floating what is the intended behavior? The actual behavior is that it computes the more precise floating point distance.

NAThompson:
For some reason I couldn't get the unit tests for this to compile even without my diffs, so I'm flying a bit blind here.

When finished, should fix #332.


NAThompson:
@jzmaddock : What's the opinion on two integral types?

jzmaddock:
>What's the opinion on two integral types? Error or double promotion?

IMO it doesn't make any sense, I would say error.

NAThompson:
@jzmaddock : Looks like it's failing due to use of C++11 features. Since C++03 is deprecated, should we remove those build jobs?

NAThompson:
Just wanted to see if everything would go green; we can revert that commit if need be.

jzmaddock:
>Looks like it's failing due to use of C++11 features. Since C++03 is deprecated, should we remove those build jobs?

Sadly not yet: deprecated means "we will remove support in future" not that we've removed it right now.  So we're committed to continuing C++03 support till next year :(

NAThompson:
@jzmaddock : I'm somewhat worried that this change is a de facto removal of C++03 support. `next.hpp` is used pervasively; see [this build](https://travis-ci.org/github/boostorg/math/jobs/671225659).

jzmaddock:
> I'm somewhat worried that this change is a de facto removal of C++03 support. next.hpp is used pervasively;

I don't think we should do that - not till this time next year anyway - give me a moment and I'll see if I can back-port to C++03.

NAThompson:
Will this fix go into the release?

jzmaddock:
It's too late for the beta, and since it's not really a bug fix, I think it's too late really?

NAThompson:
I guess it's a question of definition as to whether its a bugfix. I say it is, since my mental model of how it should work has been broken a few times, and I've gotten incorrect answers.

At very least, it's an abstraction leak.

Issue #321, opened 2020-03-06
Invalid output for incomplete elliptic integral of the second kind with k = 1


shatalinra:
Directory: boost/math/special_functions/
Files: ellint_2.hpp

Inside function ellint_e_imp in file ellint_2.hpp there is a special branch at 74 line for case when k = 1. It seems the original idea was that in this case the expression in the integral becomes sqrt(1 - sin(phi)^2). Using trigonometry, it simplifies to sqrt(cos(phi)^2) which is always non-negative. However, it is wrong to substitute that expression with just cos(phi) which can be negative. Therefore it is a mistake to just replace integral with sine function in that case and intergal suddenly loses its monotonically increasing property.

NAThompson:
@shatalinra : Do you have a main program which can demonstrate what you expect to happen?

shatalinra:
Here is a minimal reproducible example showing that at k = 1 function suddenly stop producing monotonically increasing values if angle is close to pi.
```
#include <boost/math/special_functions/ellint_2.hpp>

int main()
{
	float a = boost::math::ellint_2(0.9f, 0.9f * boost::math::constants::pi<float>());
	float b = boost::math::ellint_2(0.9f, 1.1f * boost::math::constants::pi<float>());
	float c = boost::math::ellint_2(1.0f, 0.9f * boost::math::constants::pi<float>());
	float d = boost::math::ellint_2(1.0f, 1.1f * boost::math::constants::pi<float>());

	BOOST_ASSERT(a < b);
	BOOST_ASSERT(c < d);

	return 0;
}
```

ckormanyos:
> ... the integral becomes sqrt(1 - sin(phi)^2). Using trigonometry, it simplifies to sqrt(cos(phi)^2) which is always non-negative...

I can take a look into this.

Is it sufficient to simply use an absolute value?

Or is it necessary to do anything more complicated for small  arguments near pi/2 to do a Taylor series or any additional algorithmic operations such as that?

Kind regards, Chris

shatalinra:
@ckormanyos : could you elaborate what you meant by using an absolute value? If it is about integrand sqrt(cos(phi)^2), then yes it could be replaced with abs(cos(phi)). But I don't think integrating abs is the easiest way of solving this.

Actually I already implemented workaround in my code but sadly don't have time to properly contribute it. I took idea of using integrand's periodicity from 84 line at ellint_2.hpp file and just took some shortcuts due to k = 1:
```
bool invert = false;
[...]

PR #340, opened 2020-04-08
Correct behaviour of ellint_2 when k = 1.

jzmaddock:
Fixes https://github.com/boostorg/math/issues/321.


NAThompson:
@jzmaddock : Did you not want this going in the release? Seems like a pretty low-risk fix with good coverage.

jzmaddock:
It's too late the RC is already out.

PR #345, opened 2020-04-18
Changes for Embarcadero C++ clang-based compilers, targeting Boost 1.74. Change __BORLANDC__ to BOOST_BORLANDC and __CODEGEARC__ to BOOST_CODE…

eldiener:
…GEARC, which is defined in Boost config for the Embarcadero non-clang-based compilers.


eldiener:
Can this please be merged ?

PR #358, opened 2020-05-28
AGM

NAThompson:


ckormanyos:
This is cool.
Should we give it a try in some of the GSoC 2020 work, where we are using AGM for pi and log?
Or in a while? Or soon?

NAThompson:
The file `reporting/accuracy/test_agm.cpp` demonstrates ~2ULP accuracy over the interval [1,10000]. So I think it can be used whenever you like!

NAThompson:
Ooh the [mpmath](http://mpmath.org/doc/current/functions/powers.html) agm docs are quite good.

NAThompson:
Looks like `mpfr` has an AGM as well: [mpfr_agm](https://tspiteri.gitlab.io/gmp-mpfr-sys/mpfr/MPFR-Interface.html).

Should this be wrapped in multiprecision?

NAThompson:
I realize now that the 'vanilla AGM' is not all that useful; the variations on the theme are really powerful. In any case, this is still should still be a good starting point for other ideas.

PR #363, opened 2020-06-05
Gauss's constant G:=1/agm(1,sqrt(2))

NAThompson:


ckormanyos:
Cool idea to include G.

NAThompson:
Yeah it's the absolute simplest thing I could do with the AGM, so it's basically "free".

Issue #371, opened 2020-06-14
[Not an Issue] Fast nth fibonacci numbers


madhur4127:
A while back, I wrote a routine for [fast nth fibonacci number](https://github.com/CodeISM/ccptl/blob/master/Number%20Theory/fibonacci.h) calculation.

Is boost/math interested in adding this to the library?

I could add the functionality if the library's interested.

NAThompson:
Should it go in boost.integer?

madhur4127:
> Should it go in boost.integer?

The code that is linked only works it for *standard* C++ integers but I think it can be extended to work with arbitrary precision.

NAThompson:
I personally think it's a cool addition.

The Fibonacci numbers grow so fast it might be nice to allow for arbitrary sized integers.

madhur4127:
> Also, is there a fast way to detect if it's going to overflow? It might be nice to (say) return -1 if we know it'll overflow.

Binet formula can be used for a rough approximation, so the problem will reduce to whether the given number fits in the limit.

But this will be a problem if someone tries to use a modular integer wrapper (considering that it takes the "integer" type as a template parameter) and only needs the modulus of the Fibonacci number. In this case, it will never overflow.

NAThompson:
Can you do compile-time detection of where it's a modular type?

madhur4127:
> Can you do compile-time detection of whether it's a modular type?

I am out of ideas for that :(

jzmaddock:
>Can you do compile-time detection of whether it's a modular type?

That's tough: we don't have modular type yet in multiprecision, so we don't have our own trait, and numeric_limits doesn't really give us the option of distinguishing between fixed precision integers that handle overflow as 2^n modular arithmetic, and "deliberately" modular types that are modulo some number that perhaps isn't a 2^n.

NAThompson:
Ok, cool. Just to clarify, I think it's a cool idea whether or not we can take care of this inconvenience without heroics.

PR #376, opened 2020-06-24
Fibonacci numbers

madhur4127:
Based on #371 

Currently, the code checks for the sufficient bits using `std::numeric_limits<T>::digits` and throws if the result overflows. This check uses `std::log` and is *fast*. 94th Fibonacci number overflows the precision of `uint64_t`.

I would appreciate feedback on the overflow behavior. My desired approach would be to leave the overflow handling to the user and provide a fast bare-bones version for users to build abstractions. This approach also works in the case of custom modular arithmetic integer wrapper as the template parameter. This functionality can also be made available by providing a non-throwing overload. 

The time complexity is O(log2(N) * Multiply(B)) where F_n denotes the nth fibonacci number, B is the number of bits in F_n.


include/boost/math/special_functions/fibonacci.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+// Use, modification and distribution are subject to the
+// Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_MATH_SPECIAL_FIBO_HPP
+#define BOOST_MATH_SPECIAL_FIBO_HPP
+
+#include "boost/math/constants/constants.hpp"
```
NAThompson, line 11: Did you want this one in quotes?

include/boost/math/special_functions/fibonacci.hpp
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+#endif
+
+namespace boost {
+namespace math {
+
+template <typename T, class Policy>
+T fibonacci(unsigned long long n, const Policy &pol) {
+    using ull = unsigned long long;
+    BOOST_ASSERT(n >= 0);
```
NAThompson, line 27: If `n` is unsigned, isn't this assert superfluous?

include/boost/math/special_functions/fibonacci.hpp
```diff
@@ -0,0 +1,49 @@
...
+    if (n <= 3) return (n + 1) >> 1;
+    if (std::ceil(n * std::log2(boost::math::constants::phi<long double>()) - std::log2(5) / 2.0) > std::numeric_limits<T>::digits)
```
NAThompson, line 29: I might do a bit of analysis to see if you could get away with `float` rather than `long double` here. I worry the check will take longer than the computation!

include/boost/math/special_functions/fibonacci.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#ifdef _MSC_VER
+#pragma once
+#endif
+
+namespace boost {
+namespace math {
+
+template <typename T, class Policy>
+T fibonacci(unsigned long long n, const Policy &pol) {
```
NAThompson, line 25: `inline`?

NAThompson, line 25: Should there be a static assert that `T` is an integral type?

madhur4127:
Thanks, Nick for the review! Sorry for the delayed response!

include/boost/math/special_functions/fibonacci.hpp
madhur4127: I think a static assert for `is_arithmetic` should be better instead as there's no harm in using floating-point types.

I only require +,-,* operations on the type. 

C++20 style would be to use concepts and requires expression here, I guess.

include/boost/math/special_functions/fibonacci.hpp
madhur4127: I took the constants out from the `if` check. In future, I'm hoping if `std::log2` becomes `constexpr` then it would be very efficient. To deal with the imprecision issues with floating-point, I guess a difference of `std::numeric_limits<T>::epsilon` must be relaxed.

There are two extremes now, a very precise check or no check (in unchecked version).

NAThompson:
@madhur4127 : Could you also put a benchmark in `reporting/performance` as well as some unit tests?

madhur4127:
I am stuck on trying to identify floating point types and integral types in general.

Standard says that specialising type_traits for is_arithmetic, is_floating_point is UB.

This is because I think it would be fast to call O(1) closed expression in case of floating point type but it will have precision issues with integers.

Or should I add a new function and let user call the necessary?

pabristow:
Is boost::multiprecision::number_category any help to you?

template<class FPType>
typename std::enable_if<
  // (std::is_floating_point<FPType>::value == true), // Only enables fundamental types.
  ((boost::multiprecision::number_category<FPType>::value
    == boost::multiprecision::number_kind_floating_point) == true),
  void>::type
your_foo() ...

also boost::multiprecision::number_kind_integer and rational and complex.

jzmaddock:
>Is boost::multiprecision::number_category any help to you?

Yes but it adds a dependency to multiprecision.

numeric_limits is your friend here IMO: integer types have is_integer set to true, floating point types have min_exponent != max_exponent.  For types where is_specialized is false, then it's more likely to be a floating point type (reasoning is that variable precision floating point types can't easily specialize numeric_limits, butvariable precision integer types can).

HTH, John.

madhur4127:
I added the test file and performance stats (as shown below) in reporting/performance.

As the `n` increases cost of recursion is only minimal thus iterative has no performance advantage. I used GMP as the backend for both the functions. `recursive_slow` is not _supposed_ to be slow because I wrote it to outperform naive implementations.

Both of them work in O(log n) time and O(1) space (without taking into account big integer's space and time complexity for less verbosity).

```
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
recursive_slow/8             3525 ns         3521 ns       153697
recursive_slow/64            6014 ns         6003 ns       118048
recursive_slow/512           8802 ns         8784 ns        78288
recursive_slow/4096         13661 ns        13660 ns        52241
recursive_slow/32768        70636 ns        70632 ns         9678
recursive_slow/262144     1217721 ns      1217522 ns          570
recursive_slow/2097152   20469874 ns     20468924 ns           35
recursive_slow/4194304   43150902 ns     43142089 ns           17
iterative_fast/8             1419 ns         1419 ns       495748
iterative_fast/64            2389 ns         2389 ns       285226
iterative_fast/512           4168 ns         4167 ns       169008
iterative_fast/4096          8288 ns         8288 ns        92749
iterative_fast/32768        64556 ns        64555 ns        11057
iterative_fast/262144     1200102 ns      1200053 ns          576
iterative_fast/2097152   19781647 ns     19781013 ns           35
iterative_fast/4194304   42263450 ns     42195817 ns           17
```

madhur4127:
1. Because of imprecisions, the calculation for the exact overflow limit for a type is off by 1 (up to 100) and increases as the `n` increases. But on a bright side, when the overflow exception is thrown then it hints that overflow would _surely_ occur. So overflow can be missed but not falsely reported.

2. I couldn't convince myself about limiting the Fibonacci to types like integral only. Any type for which +, -, * is defined should be enough. This could easily be checked with C++20's requires clause, but till Boost.Math reaches that stage compiler errors would be the only help.

NAThompson:
> I couldn't convince myself about limiting the Fibonacci to types like integral only.

Yes, I now agree on this point; a use case might be computation of the [reciprocal Fibonacci constant](https://mathworld.wolfram.com/ReciprocalFibonacciConstant.html).

reporting/performance/test_fibonacci.cpp
```diff
@@ -0,0 +1,65 @@
...
+    T d = p.first * p.first + p.second * p.second;
+    return (n & 1) ? std::make_pair(d, c + d) : std::make_pair(c, d);
+}
+
+static void recursive_slow(benchmark::State &state) {
+    for (auto _ : state)
+        benchmark::DoNotOptimize(fib_rec(state.range(0)).first);
+}
+constexpr int bm_start = 1 << 3, bm_end = 1 << 22;
+BENCHMARK(recursive_slow)->Range(bm_start, bm_end);
```
NAThompson, line 26: Could you add `Complexity()` here?

```cpp
BENCHMARK(recursive_slow)->Range(bm_start, bm_end)->Complexity();
```

reporting/performance/test_fibonacci.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+static void iterative_fast(benchmark::State &state) {
+    for (auto _ : state)
+        benchmark::DoNotOptimize(boost::math::fibonacci<T>(state.range(0)));
+}
+BENCHMARK(iterative_fast)->Range(bm_start, bm_end);
```
NAThompson, line 32: Same thing here with `Complexity()`.

reporting/performance/test_fibonacci.cpp
madhur4127: `Complexity(benchmark::oLogN);` results in:

```
recursive_slow_BigO          -nan lgN        -nan lgN  
recursive_slow_RMS           -nan %          -nan %    
...
iterative_fast_BigO          -nan lgN        -nan lgN  
iterative_fast_RMS           -nan %          -nan %  
```

NAThompson: Oh, yeah you need to do `state.SetComplexityN(state.range(0))` after the for loop completes. Also, don't add the `benchmark::oLogN` key or else you're assuming the desired result.

include/boost/math/special_functions/fibonacci.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <limits>
+
+#ifdef _MSC_VER
+#pragma once
+#endif
+
+namespace boost {
+namespace math {
+namespace detail {
+// this should be constexpr in future
```
NAThompson, line 23: Have you checked godbolt to see if these get compiled into constants? If not, we might want to put:

```
fib_bits_phi = 0.6942419136306174; // = ln(phi)/ln(2)
```

include/boost/math/special_functions/fibonacci.hpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+template <typename T, class Policy>
+T inline fibonacci(unsigned long long n, const Policy &pol) {
+    // check for overflow using approximation to binet's formula: F_n ~ phi^n / sqrt(5)
+    if (n > 20 && n * detail::fib_bits_phi - detail::fib_bits_deno > std::numeric_limits<T>::digits)
+        return policies::raise_overflow_error<T>("boost::math::fibonacci<%1%>(unsigned long long)", "Possible overflow detected.", pol);
+    return fibonacci_unchecked<T>(n);
+}
+
```
NAThompson, line 52: It occurs to me we might want to add `fibonacci_next` generator so that we can call it in a loop; do you agree?

include/boost/math/special_functions/fibonacci.hpp
madhur4127: [Yes, they are reduced into constants](https://compiler-explorer.com/z/4a6dGP)

test/Jamfile.v2
```diff
@@ -490,6 +490,7 @@ test-suite special_fun :
    [ run test_trig.cpp test_instances//test_instances pch_light ../../test/build//boost_unit_test_framework  ]
    [ run test_zeta.cpp ../../test/build//boost_unit_test_framework test_instances//test_instances pch_light  ]
    [ run test_sinc.cpp ../../test/build//boost_unit_test_framework pch_light ]
+   [ run test_fibonacci.cpp ../../test/build//boost_unit_test_framework : : : <linkflags>-lgmp [ requires cpp_lambdas ] ]
```
madhur4127, line 493: I don't really require lamdba, but it's a hack for C++11 check required by `<cstdint>`

include/boost/math/special_functions/fibonacci.hpp
madhur4127: I don't understand what do you mean by a generator here. Can you please share some examples or elaborate more.

I think you meant like generating next Fibonacci number continuously but that can be done trivially by a user.
```c++
int a = 0, b = 1;
while (n--) 
    swap(a,b), b += a;
```
Fibonacci numbers need previous two values to compute the next term, so generator in that case would take the arguments and simply return the sum?

madhur4127: Or like Python's `yield`? Where calling `next()` would return next fibonacci number?

madhur4127:
Sorry for a newbie type question. But is there a way to add gmp to Appveyor build? Some tests are failing because of the `gmp.h` not found error.

My current solution is to add a check for gmp but that'll make the tests avoid builds on msvc which is undesirable.

```
// Jamfile.v2
[ run test_fibonacci.cpp ../../test/build//boost_unit_test_framework : : : <linkflags>-lgmp 
    [ check-target-builds ../config//has_gmp : : <build>no ] [ requires cpp_lambdas ] ]
```

test/Jamfile.v2
NAThompson: @madhur4127: I personally don't mind if this is written in C++17. I'm doing all my new features in C++17 so that if I need to make changes that require C++17 I won't break anything.

Hopefully that's not being obtuse.

include/boost/math/special_functions/fibonacci.hpp
NAThompson: @madhur4127: Yes, that's what I mean. For an example, [here's](https://github.com/boostorg/multiprecision/blob/aee74ade95f6ebb458b63b9796b3e0923e214783/example/representations/cfrac_patterns.cpp) one that generates a particular continued fraction; you then pass it to the continued fraction evaluator which calls it's call operator.

Could you write a quick code that evaluates the [reciprocal Fibonacci constant](https://en.wikipedia.org/wiki/Reciprocal_Fibonacci_constant) and add it to `examples`? I think you'll see how natural a generator is for that use; but if you find it can be done easily without a generator feel free to ignore this request.

NAThompson:
@madhur4127 : Please tag your commits with `[CI SKIP]` until you are confident that it's ready to go; we're getting backed up on the CI system.

madhur4127:
@NAThompson Sure, I am confident about passing CI tests this time. I would wish to make the CI run for this once successfully so that I could be sure about the working of the Fibonacci code without generator.

madhur4127:
@NAThompson, I think the PR is in good shape.

1. Changes till Fibonacci numbers passed the CI tests.
2. Fibonacci generator is added with test and an example which hopefully will pass the CI too (so no tagging of CI SKIP).

NAThompson:
@madhur4127 : Ok, looking good. Could you add documentation to `doc/sf`?

madhur4127:
Sorry guys, I will add the docs soon. Sorry for the loooong delay!

madhur4127:
Hi All,

Sorry for the looooong delay. I had added the documentation into `number_series.qbk` as it is similar in spirit to prime_numbers, bernoulli_numbers. I am open to relocating it 😄 

I also added comments and made the unchecked function constexpr. I couldn't make all of them constexpr because of `std::log` not being constexpr. Although compilers are smart in precomputing the results as seen in the godbolt link in the review comments.

Lastly the feature will require atleast C++14 because of the constexpr changes, but previously required C++11. But I plan to upgrade it as and when the constexpr `cmath` is available. I have added C++17 as the requirement in the test as @NAThompson pointed out.

The test failures are in `distribution_tests/test_kolmogorov_smirnov.cpp:59:44: error: no matching function` whereas my changes are in `special_functions` so I guess that's okay :)

NAThompson:
@madhur4127 : Are you happy with this? If so could you rebase on develop?

madhur4127:
Yes @NAThompson, I think it is complete with the documentation + tests.

jzmaddock:
I have only 2 comments:

1) The existing "unchecked" functions are called "unchecked_factorial" and "unchecked_bernoulli_b2n", right or wrong I think we should probably stick with that convention and use "unchecked_fibonacci" here.
2) The example could usefully go in the docs at the end: use quickbooks code import feature to directly import markup from the example rather than cut and pasting, see: https://www.boost.org/doc/libs/1_75_0/doc/html/quickbook/syntax/block.html#quickbook.syntax.block.import

madhur4127:
Sorry Nick (@NAThompson) and John (@jzmaddock),

This PR somehow got flushed out from my memory. Terrific 2020 eh.

1. Agreed, for consistency, I'll renamed it to `unchecked_fibonacci`
2. But there's a problem, the documentation, I wrote is kinda snippet based. For example:
```
   boost::math::fibonacci_generator<int> gen;
   int x = gen(); // x is 0
   int y = gen(); // y is 1
   for(int i = 0; i < 5; ++i) // this loop is same as gen.set(7);
       gen();
   int z = gen(); // z is 13boost::math::fibonacci_generator<int> gen;
 ```
This should really be a new file instead of a function in the same file of the documentation.

I like that way of documentation, for example, Boost.Asio is a nice example of that but this will be a overkill, I think.

I'm open to ideas :)

NAThompson:
> This should really be a new file instead of a function in the same file of the documentation.

Well, it could also be a separate PR just to get this one past the finish line.

madhur4127:
This seems a clever solution. I'll remove the documentation and finish this.

PR #385, opened 2020-06-29
Fix reciprocal of min for mpfr/cpp_bin_float in continued fractions.

NAThompson:
The issue is identified [here](https://github.com/boostorg/multiprecision/issues/253).

The issue can be reproduced by calling `continued_fraction_b` with b0 = 0 and a mpfr result_type.

### DIFF
--- a/include/boost/math/tools/fraction.hpp
+++ b/include/boost/math/tools/fraction.hpp
@@ -73,7 +73,7 @@ namespace detail
    struct tiny_value
    {
       static T get() {
-         return tools::min_value<T>(); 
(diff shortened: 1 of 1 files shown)


NAThompson:
@jzmaddock : Could you review?

jzmaddock:
That looks reasonable to me: can you just add a comment to the code so we know why the *16 was added?

NAThompson:
@jzmaddock : Good catch; done.

NAThompson:
Build failure from serialization and a timeout.

PR #387, opened 2020-07-02
Soldner's constant.

NAThompson:
@jzmaddock : Could you review? I have some worry that I might create header cycles this way . . .


NAThompson:
It'll create header cycles; sad!

Issue #354, opened 2020-05-18
boost::math::pow is not constexpr


pabristow:
`constexpr  int i = boost::math::pow<8>(2);`

MSVC 2019 c++std=17 reports

error C2131: expression did not evaluate to a constant

 ` const  int i = boost::math::pow<8>(2);` //  gives the expected 256.

Is there any reason why this could not be constexpr?

NAThompson:
Sorry I'm gonna have to punt on this one; I have no clue what the `constexpr` rules are . . .

cosurgi:
Would be cool if pow was constexpr.

PR #388, opened 2020-07-03
boost::math::pow is not constexpr #354

mborland:
Made pow constexpr as requested in issue #354. Forward declaring constexpr functions requires the language standard to be at least C++14. The if statements were left as is due to making them constexpr as well would raise the language standard to 17. The only change that would have is determining which of the return statements would compile. If you compile with -std=c++11 you will receive 6 [-Wc++14-extension] warnings but the results will remain as expected. Please let me know if you have any questions.

### DIFF
--- a/include/boost/math/special_functions/pow.hpp
+++ b/include/boost/math/special_functions/pow.hpp
@@ -35,7 +35,7 @@ template <int N, int M = N%2>
 struct positive_power
 {
     template <typename T>
-    static constexpr T result(T base)
+    static BOOST_CXX14_CONSTEXPR T result(T base)
     {
         T power = positive_power<N/2>::result(base);
         return power * power;
@@ -46,7 +46,7 @@ template <int N>
 struct positive_power<N, 1>
 {
     template <typename T>
-    static constexpr T result(T base)
+    static BOOST_CXX14_CONSTEXPR T result(T base)
     {
         T power = positive_power<N/2>::result(base);
         return base * power * power;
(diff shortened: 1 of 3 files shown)


NAThompson:
@mborland : Starting to look good. Could you throw a `static_assert(boost::math::pow<8>(2)==256, "Pow is not constexpr");` into `test/pow_test.cpp`? Then we'll be able to work out how many platforms we'll be able to support.

mborland:
@pabristow I did some testing locally to make sure that it worked on my machine first. I was going to rely on CI to check all platforms.

@NAThompson Added as requested.

include/boost/math/special_functions/pow.hpp
```diff
@@ -46,7 +46,7 @@ template <int N>
 struct positive_power<N, 1>
 {
     template <typename T>
-    static T result(T base)
+    static constexpr T result(T base)
```
jzmaddock, line 49: These require C++14 constexpr anyway?

So I would change `constexpr` to `BOOST_CXX14_CONSTEXPR` and make the constexpr test conditional on BOOST_NO_CXX14_CONSTEXPR not being defined.  While it might be possible to rewrite these for C++11 constexpr it risks wrecking runtime performance.

mborland: @jzmaddock Most functions in this header requires C++14 either due to variable declaration in the function body, or multiple return statements. I will proceed with the recommendation to change to BOOST_CXX_14_CONSTEXPR rather than bringing everything down to C++11 standard using obfuscated code.

mborland:
@NAThompson It looks like the two tests that ran and failed are the same two that failed in your PR #387. Is this the expected result?

NAThompson:
@mborland : Yup, those are the known serialization issues.

NAThompson:
@mborland : Thanks for the diffs; hope you had fun and learned a bit!

PR #389, opened 2020-07-03
Tanh-sinh with MPFR test.

NAThompson:


jzmaddock:
There's something strange going on in the original test case - I don't see why the checks in itrunc are failing?  So this might be an issue with mpfr_float....

NAThompson:
@jzmaddock : I noticed that as well, but concluded it was a separate issue. In any case, I realized that `itrunc` really needed to be a `lltrunc` in order to match the size of the lhs . . .

jzmaddock:
Nah, my bad.  It's overflow in `sqrt((2 - x) / x` that's the issue.

NAThompson:
The build failures come from the known issue in serialization.

PR #390, opened 2020-07-04
Implemented mode and associated tests

mborland:
A generic implementation of mode that allows for multi-modal distributions. Please let me know if you have any questions.


include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +513,58 @@ inline auto interquartile_range(RandomAccessContainer & v)
     return interquartile_range(v.begin(), v.end());
 }
 
+template<class RandomAccessIterator>
+auto mode(RandomAccessIterator first, RandomAccessIterator last)
+{
+    using Real = typename std::iterator_traits<RandomAccessIterator>::value_type;
```
NAThompson, line 519: Is `Real` the right language here? The probability that a set of real numbers has a mode is zero. I would use `Z`.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +513,58 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+    using Size = typename std::iterator_traits<RandomAccessIterator>::difference_type;
+
+    std::vector<Real> sorted(first, last);
+    std::sort(sorted.begin(), sorted.end());
+
+    std::vector<Real> modes {};
+    Size max_counter {};
```
NAThompson, line 526: Are you sure `max_counter` will be initialized to zero? There might be a custom type that doesn't do that.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +513,58 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+    }
+
+    for(const Real i : sorted)
+    {
+        if(i == previous_search)
+        {
+            continue;
+        }
+        
+        Size current_count {std::count(sorted.begin(), sorted.end(), i)};
```
NAThompson, line 542: The complexity of `std::count` is linear, so isn't this loop quadratic?

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +513,58 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+        if(current_count > max_counter)
+        {
+            modes.clear();
```
NAThompson, line 545: Isn't this call to clear pretty expensive?

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +513,58 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+
+        else if(current_count == max_counter)
+        {
+            modes.push_back(i);
+        }
+
+        previous_search = i;
+    }
+
+    BOOST_ASSERT_MSG(max_counter > 1, "Mode is undefined");
```
NAThompson, line 558: I would probably prefer to return an empty vector rather than die with an assert.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +513,58 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+
+    return modes;
+}
+
+template<class RandomAccessContainer>
+inline auto mode(RandomAccessContainer & v)
```
NAThompson, line 564: The other implementations return `auto` because they might return `double` for integer inputs, or `Real` for `Real` inputs, `Real` for complex inputs so on. But it appears this one always returns `std::vector<Real>`, so it might be nice to write that explicitly.

include/boost/math/statistics/univariate_statistics.hpp
NAThompson: I might even `static_assert` that `Z` is an integer type.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -11,6 +11,7 @@
 #include <tuple>
 #include <cmath>
 #include <boost/assert.hpp>
+#include <boost/math/tools/precision.hpp>
```
NAThompson, line 14: I'd probably just use `std::numeric_limits<Z>::max()` so that we don't have to bring in this header.

NAThompson:
@mborland : Could you put a google benchmark file into `reporting/performance`? You could just `cp rsqrt_performance.cpp` for a template.

include/boost/math/statistics/univariate_statistics.hpp
mborland: I will make the change to Z. I went with Real for consistency with the rest of the file.

include/boost/math/statistics/univariate_statistics.hpp
mborland: It would be. I swapped the range based for loop out with an iterator so it can replace begin() in the count. That will reduce the range it is iterating over.

include/boost/math/statistics/univariate_statistics.hpp
mborland: At worst it is linear, and should never be very large. Libstdc++ has it optimized to constant complexity for POD.

mborland:
@NAThompson The google benchmark file has been added as requested. I do not see asymptotic complexity in the results.

include/boost/math/statistics/univariate_statistics.hpp
jzmaddock: +1, asserts are for testing the logic of our code, not for user-errors or anything which is expected.  So either an exception or an empty vector as Nick suggests.

include/boost/math/statistics/univariate_statistics.hpp
jzmaddock: +1.  Clear should be as simple as setting the size of the vector to zero for POD types.  For non POD types the destructors would all be called.  I guess a more streamlined version in that case would be to resize the vector to 1, and then assign the new value?

reporting/performance/test_mode.cpp
```diff
@@ -0,0 +1,39 @@
...
+    std::vector<Z> v(100);
+    std::generate(v.begin(), v.end(), gen);
+
+    for (auto _ : state)
+    {
+        benchmark::DoNotOptimize(mode(v));
+    }
+}
+
+BENCHMARK_TEMPLATE(test_mode, float);
```
NAThompson, line 33: I feel fairly confident that this should not compile with any floating point type. Otherwise we need to have a semantics for when two floating point numbers are approximately equal.

As it stands, the mode is not robust against perturbations.

reporting/performance/test_mode.cpp
```diff
@@ -0,0 +1,39 @@
...
+BENCHMARK_TEMPLATE(test_mode, double);
+BENCHMARK_TEMPLATE(test_mode, long double);
+BENCHMARK_TEMPLATE(test_mode, int);
```
NAThompson, line 36: I would try int32_t, int64_t, uint64_t, uint32_t, and maybe __int128_t.

Then we need to measure the complexity. My apologies for recommending `rsqrt_performance.cpp` as a template, [chebyshev_clenshaw.cpp](https://github.com/boostorg/math/blob/develop/reporting/performance/chebyshev_clenshaw.cpp) is a better example for this.

Could you also add a benchmark for the vector {0,1,2,3,4,5,6,...}  and {1,1,2,2,3,3,4,4, ...}? I want to see if there's some way to break the complexity of the algorithm.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,74 @@ inline auto interquartile_range(RandomAccessContainer & v)
     return interquartile_range(v.begin(), v.end());
 }
 
+template<class RandomAccessIterator>
+auto mode(RandomAccessIterator first, RandomAccessIterator last) -> std::vector<typename std::iterator_traits<RandomAccessIterator>::value_type>
+{
+    using Z = typename std::iterator_traits<RandomAccessIterator>::value_type;
+    using Size = typename std::iterator_traits<RandomAccessIterator>::difference_type;
+
+    std::vector<Z> sorted(first, last);
```
NAThompson, line 521: There is an idiom in the standard library that sorting operations do not copy their input, they mutate it. See `std::nth_element`.

There might be a case to add a `sorted_mode` call which takes sorted data by `const &`, and this can forward to that call. I've found in statistical workflows, there are lots of things that require sorted data, so it's not a bad assumption that sorting is part of a larger workflow.

reporting/performance/test_mode.cpp
mborland: I added the assertion for integral values with the message that floating point values have not been implemented. I see what you are getting at with the floating point comparison problem.

reporting/performance/test_mode.cpp
mborland: Added those additional tests. Attached is a synopsis of the results.

[Results.txt](https://github.com/boostorg/math/files/4873188/Results.txt)

reporting/performance/test_mode.cpp
```diff
@@ -21,19 +18,113 @@ void test_mode(benchmark::State& state)
...
+
+    state.SetComplexityN(state.range(0));
 }
 
-BENCHMARK_TEMPLATE(test_mode, float);
-BENCHMARK_TEMPLATE(test_mode, double);
-BENCHMARK_TEMPLATE(test_mode, long double);
-BENCHMARK_TEMPLATE(test_mode, int);
-BENCHMARK_TEMPLATE(test_mode, cpp_bin_float_50);
+BENCHMARK_TEMPLATE(test_mode, int)->RangeMultiplier(2)->Range(1<<1, 1<<22)->Complexity(benchmark::oN);
```
NAThompson, line 113: Don't put the `benchmark::oN` key in these. First, because the expected complexity is `nlog(n)`, and second, because we're looking for performance bugs. If you specify the complexity class, then you'll never see that the best fit is (say) quadratic.

Just do:

```
BENCHMARK_TEMPLATE(test_mode, int)->RangeMultiplier(2)->Range(1<<1, 1<<22)->Complexity();
```

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -516,10 +516,15 @@ template<class RandomAccessIterator>
 auto mode(RandomAccessIterator first, RandomAccessIterator last) -> std::vector<typename std::iterator_traits<RandomAccessIterator>::value_type>
 {
     using Z = typename std::iterator_traits<RandomAccessIterator>::value_type;
+    static_assert(std::is_integral<Z>::value, "Floating point values have not yet been implemented.");
     using Size = typename std::iterator_traits<RandomAccessIterator>::difference_type;
 
     std::vector<Z> sorted(first, last);
-    std::sort(sorted.begin(), sorted.end());
+    
+    if(!std::is_sorted(sorted.begin(), sorted.end()))
```
NAThompson, line 524: I'm not convinced this is buying us much. `std::is_sorted` takes O(N) time if the vector is sorted, and `std::sort` also takes O(N) time if the vector is sorted.

I think having a `sorted_mode` would also help us with the benchmarks, because your results are currently dominated by the time for sorting.

reporting/performance/test_mode.cpp
mborland: Go it. The results are more sensical now.

Worst values
test_mode: 13.06 nlog(n)
sequential: 2.48 nlog(n)
sequential_pairs: 1.87 nlog(n)
sequential_multiple: 1.55 nlog(n)

include/boost/math/statistics/univariate_statistics.hpp
mborland: With a separate sorted mode function the results validate the addition.

Worst values
test_mode: Unchanged
sequential: 2.48 -> 2.01 nlog(n)
sequential_pairs: 1.87 -> 1.40 nlog(n)
sequential_multiples: 1.55 -> 1.08 nlog(n)

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -580,6 +574,22 @@ auto mode(RandomAccessIterator first, RandomAccessIterator last) -> std::vector<
...
+{
+    return sorted_mode(v.begin(), v.end());
+}
+
+template<class RandomAccessIterator>
+auto mode(RandomAccessIterator first, RandomAccessIterator last)
+{
+    using Z = typename std::iterator_traits<RandomAccessIterator>::value_type;
+    std::vector<Z> sorted(first, last);
+    std::sort(sorted.begin(), sorted.end());
```
NAThompson, line 588: I still recommend against copying, and instead recommend documenting that `mode` mutates its inputs.

include/boost/math/statistics/univariate_statistics.hpp
NAThompson: Wait, shouldn't this be O(N) for the sorted input?

NAThompson:
@mborland : Just ran the benchmark:

```
$ g++-9 -g --std=gnu++17 -ffast-math -fno-finite-math-only -march=native -Wfatal-errors -MMD -fvisibility=hidden -O3 -I../../ -I/usr/local/include -o reporting/performance/test_mode.x reporting/performance/test_mode.cpp -pthread -lbenchmark -lbenchmark_main -L/usr/local/lib
$ ./reporting/performance/test_mode.x
2020-07-04 15:05:05
Running ./reporting/performance/test_mode.x
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32K (x2)
  L1 Instruction 32K (x2)
  L2 Unified 262K (x2)
  L3 Unified 3145K (x1)
Load Average: 2.90, 3.60, 3.28
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
test_mode<int>/2                                        576 ns          574 ns      1256890
test_mode<int>/4                                        670 ns          662 ns      1067155
test_mode<int>/8                                        691 ns          687 ns      1000000
test_mode<int>/16                                       754 ns          752 ns      1084565
test_mode<int>/32                                       779 ns          772 ns      1155592
test_mode<int>/64                                       808 ns          804 ns       858548
test_mode<int>/128                                     1251 ns         1247 ns       564903
test_mode<int>/256                                     2294 ns         2283 ns       324230
test_mode<int>/512                                     4274 ns         4259 ns       175606
test_mode<int>/1024                                   10340 ns        10286 ns        61062
test_mode<int>/2048                                   31495 ns        31390 ns        23763
test_mode<int>/4096                                   77137 ns        76821 ns         9569
test_mode<int>/8192                                  172714 ns       170798 ns         3998
[1]    46061 segmentation fault  ./reporting/performance/test_mode.x
```

Recompiled with `-fsanitize=address -fsanitize=undefined`:

```
 ./reporting/performance/test_mode.x
2020-07-04 15:06:44
Running ./reporting/performance/test_mode.x
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32K (x2)
  L1 Instruction 32K (x2)
  L2 Unified 262K (x2)
  L3 Unified 3145K (x1)
Load Average: 2.26, 3.20, 3.16
=================================================================
==46572==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000005ef8 at pc 0x000106455186 bp 0x7ffee97b1ba0 sp 0x7ffee97b1b98
READ of size 4 at 0x602000005ef8 thread T0
    #0 0x106455185 in std::vector<std::iterator_traits<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >::value_type, std::allocator<std::iterator_traits<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >::value_type> > boost::math::statistics::sorted_mode<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >(__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > >, __gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > >) univariate_statistics.hpp:542

0x602000005ef8 is located 0 bytes to the right of 8-byte region [0x602000005ef0,0x602000005ef8)
allocated by thread T0 here:
    #0 0x106deec0f in wrap__Znwm (libasan.5.dylib:x86_64+0x7dc0f)
    #1 0x106453bb1 in std::vector<std::iterator_traits<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >::value_type, std::allocator<std::iterator_traits<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >::value_type> > boost::math::statistics::sorted_mode<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >(__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > >, __gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > >) <Invalid path>

SUMMARY: AddressSanitizer: heap-buffer-overflow univariate_statistics.hpp:542 in std::vector<std::iterator_traits<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >::value_type, std::allocator<std::iterator_traits<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >::value_type> > boost::math::statistics::sorted_mode<__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > > >(__gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > >, __gnu_cxx::__normal_iterator<int*, std::vector<int, std::allocator<int> > >)
Shadow bytes around the buggy address:
  0x1c0400000b80: fa fa 00 fa fa fa 00 fa fa fa 00 fa fa fa 00 fa
  0x1c0400000b90: fa fa 00 fa fa fa 00 fa fa fa 00 fa fa fa 00 fa
  0x1c0400000ba0: fa fa 00 fa fa fa 00 fa fa fa 00 fa fa fa fd fd
  0x1c0400000bb0: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fd
  0x1c0400000bc0: fa fa fd fd fa fa fd fa fa fa 00 00 fa fa 00 00
=>0x1c0400000bd0: fa fa 00 fa fa fa 00 fa fa fa 00 fa fa fa 00[fa]
  0x1c0400000be0: fa fa 04 fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x1c0400000bf0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x1c0400000c00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x1c0400000c10: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x1c0400000c20: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==46572==ABORTING
[1]    46572 abort      ./reporting/performance/test_mode.x
```

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,89 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+    if(max_counter == 1) // Return an empty vector if mode is undefined
+    {
+        modes.clear();
+    }
+
+    return modes;
+}
+
+template<class RandomAccessContainer>
+inline auto sorted_mode(RandomAccessContainer & v)
```
NAThompson, line 578: This can be marked `const`: `RandomAccessContainer const & v`.

BTW does this require `RandomAccess` or can it be a `std::list`?

reporting/performance/test_mode.cpp
```diff
@@ -0,0 +1,130 @@
+//  (C) Copyright Nick Thompson 2020.
```
NAThompson, line 1: Put your name in the copyright.

reporting/performance/test_mode.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    for (auto _ : state)
+    {
+        benchmark::DoNotOptimize(sorted_mode(v));
+    }
+
+    state.SetComplexityN(state.range(0));
+}
+
+BENCHMARK_TEMPLATE(test_mode, int)->RangeMultiplier(2)->Range(1<<1, 1<<22)->Complexity();
```
NAThompson, line 113: On most systems `int` is the same as `int32_t`, so this benchmark is superfluous.

test/univariate_statistics_test.cpp
```diff
@@ -833,6 +833,39 @@ void test_interquartile_range()
...
+    // Does it work with const qualification?
+    test = boost::math::statistics::mode(v.cbegin(), v.cend());
+    BOOST_TEST(ref == test[0]);
+
+    // Does it work with std::array?
+    std::array<Real, 6> u {1, 2, 2, 3, 4, 5};
+    test = boost::math::statistics::mode(u);
+    BOOST_TEST(ref == test[0]);
+
+    // Does it work with a bi-modal distribuition?
```
NAThompson, line 863: Nitpick: Spelling of distribution.

test/univariate_statistics_test.cpp
```diff
@@ -902,5 +935,11 @@ int main()
 
     test_interquartile_range<double>();
     test_interquartile_range<cpp_bin_float_50>();
+
+    test_mode<int>();
+    test_mode<int32_t>();
+    test_mode<int64_t>();
+    test_mode<u_int32_t>();
```
NAThompson, line 942: Is `u_int32_t` standard? I thought it was `uint32_t`.

include/boost/math/statistics/univariate_statistics.hpp
mborland: Now that it is not sorting it can take a forward iterator instead of random access iterator

mborland:
@NAThompson I am unable to replicate your seg fault. Ran with the same compiler flags on g++9, g++10, clang++10. I am not sure what the discrepancy is.

NAThompson:
@mborland : Did you compile with `-fsanitize=address -fsanitize=undefined`?

mborland:
I had not because I could not replicate the seg fault, but should have. You could de-reference last which is UB. Everything now shows a benchmark between 0.67 nlog(n) and 0.92n

NAThompson:
@mborland : With the new changes, I do not see any UB.

reporting/performance/test_mode.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include <benchmark/benchmark.h>
+
+template <class Z>
+void test_mode(benchmark::State& state)
+{
+    using boost::math::statistics::mode;
+    
+    std::random_device rd;
+    std::mt19937_64 mt(rd());
+    std::uniform_real_distribution<double> dist {1, 10};
```
NAThompson, line 17: This should be `std::uniform_int_distribution`.

reporting/performance/test_mode.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    auto gen = [&dist, &mt](){return dist(mt);};
+
+    std::vector<Z> v(state.range(0));
+    std::generate(v.begin(), v.end(), gen);
+
+    for (auto _ : state)
+    {
+        benchmark::DoNotOptimize(mode(v));
```
NAThompson, line 26: I would only  benchmark `sorted_mode`, since the performance of `std::sort` is understood.

test/univariate_statistics_test.cpp
```diff
@@ -833,6 +833,39 @@ void test_interquartile_range()
     BOOST_TEST_EQ(iqr, 6);
 }
 
+template<class Real>
```
NAThompson, line 836: Change `Real` to `Z`.

test/univariate_statistics_test.cpp
```diff
@@ -833,6 +833,39 @@ void test_interquartile_range()
...
+void test_mode()
+{
+    std::vector<Real> v {1, 2, 2, 3, 4, 5};
```
NAThompson, line 839: Could you add a zero length vector and a length one vector to the tests?

NAThompson:
@mborland : Might also want to add a test of `std::list` as well.

reporting/performance/test_mode.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    for (auto _ : state)
+    {
+        benchmark::DoNotOptimize(sorted_mode(v));
+    }
+
+    state.SetComplexityN(state.range(0));
+}
+
```
NAThompson, line 112: Looks like you forgot to measure `test_mode` here, also `test_mode` is calling `mode` rather than `sorted_mode`.

test/univariate_statistics_test.cpp
```diff
@@ -833,6 +834,59 @@ void test_interquartile_range()
...
+    BOOST_TEST(ref == test[0]);
+
+    // Does it work with const qualification? Only if pre-sorted
+    test = boost::math::statistics::sorted_mode(v.cbegin(), v.cend());
+    BOOST_TEST(ref == test[0]);
+
+    // Does it work with std::array?
+    std::array<Z, 6> u {1, 2, 2, 3, 4, 5};
+    test = boost::math::statistics::mode(u);
+    BOOST_TEST(ref == test[0]);
```
NAThompson, line 862: I would use `BOOST_TEST_EQ` here rather than `BOOST_TEST`; it'll give a nicer message if we ever break this.

test/univariate_statistics_test.cpp
NAThompson: I think we should do `uint32_t` here.

NAThompson:
@mborland : Could you update [univariate_statistics.qbk](https://github.com/boostorg/math/blob/develop/doc/statistics/univariate_statistics.qbk)?

NAThompson:
Ok, ran `cppcheck` over the diffs; it's clean. Ran the benchmarks, got O(N) scaling except for `test_mode`, which oddly looks more like NLog(n). Could you do a `vector::reserve` with (say) 16 elements and see if that increases the performance?

mborland:
@NAThompson I am seeing O(N) scaling for test_mode. Where would you add the vector::reserve? It does not play nicely with std::generate.

reporting/performance/test_mode.cpp
```diff
@@ -10,20 +10,21 @@
...
     
     std::random_device rd;
     std::mt19937_64 mt(rd());
     std::uniform_int_distribution<> dist {1, 10};
 
     auto gen = [&dist, &mt](){return dist(mt);};
 
-    std::vector<Z> v(state.range(0));
+    std::vector<Z> v;
+    v.reserve(state.range(0));
```
NAThompson, line 22: I meant call `v.reserve` in `sorted_mode`; I believe this is UB.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,78 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
 }
 
+template<class ForwardIterator>
+auto sorted_mode(const ForwardIterator first, const ForwardIterator last) -> std::vector<typename std::iterator_traits<ForwardIterator>::value_type>
+{
+    using Z = typename std::iterator_traits<ForwardIterator>::value_type;
+    static_assert(std::is_integral<Z>::value, "Floating point values have not yet been implemented.");
+    using Size = typename std::iterator_traits<ForwardIterator>::difference_type;
+
+    std::vector<Z> modes {};
```
NAThompson, line 522: modes.reserve(16);

NAThompson:
@mborland : Did that change make it faster?

mborland:
@NAThompson It's a hair faster. Dropped from avg 2.8N to 2.75N for test_mode.

NAThompson:
Ok, looks good. Will merge once the CI cycles.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,79 @@ inline auto interquartile_range(RandomAccessContainer & v)
     return interquartile_range(v.begin(), v.end());
 }
 
+template<class ForwardIterator>
+auto sorted_mode(const ForwardIterator first, const ForwardIterator last) -> std::vector<typename std::iterator_traits<ForwardIterator>::value_type>
```
jeremy-murphy, line 516: If you really want it to be generic, don't return a `std::vector`. Do as all the standard algorithms do: take an output iterator.

NAThompson: @jeremy-murphy : Good suggestion, but looking it up I only see it in [C++20](https://en.cppreference.com/w/cpp/iterator/output_iterator).

mborland: @NAThompson That is pretty idiomatic STL. Looking at the test_num benchmark it is nearly equivalent to returning the vector in terms of complexity. The one caveat is I can not use `benchmark::DoNotOptimize` because sorted_modes would no longer have a return value.

jeremy-murphy: @mborland , looks closely at the STL algorithms that take output iterators, they _do_ have a return value.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,79 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+    {
+        return modes;
+    }
+    
+    for(auto it {first}; it != last; ++it)
+    {
+        if(*it == previous_search)
+        {
+            continue;
+        }
```
jeremy-murphy, line 537: This will essentially do `find` on the input sequence twice. Yes, it's already in the cache, but it's not logical and not required. See my next comment down below about `previous_search`.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,79 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+            modes[0] = *it;
+            max_counter = current_count;
+        }
+
+        else if(current_count == max_counter)
+        {
+            modes.emplace_back(*it);
+        }
+
+        previous_search = *it;
```
jeremy-murphy, line 559: Instead of using `previous_search`, do `it = end_it`. But actually, take it a step further: change the loop to `while (first != last)` and keep moving `first` forward with `first = end_it`.

include/boost/math/statistics/univariate_statistics.hpp
mborland: @jeremy-murphy Found them. It had been a bit since I cracked open the C++ Programming Language.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,74 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+auto sorted_mode(ForwardIterator first, ForwardIterator last, OutputIterator output) -> decltype(output)
+{
+    using Z = typename std::iterator_traits<ForwardIterator>::value_type;
+    static_assert(std::is_integral<Z>::value, "Floating point values have not yet been implemented.");
+    using Size = typename std::iterator_traits<ForwardIterator>::difference_type;
+
+    std::vector<Z> modes {};
+    modes.reserve(16);
+    Size max_counter {0};
+    auto it {first};
```
jeremy-murphy, line 525: It's perfectly OK to use `first` as the variable that gets modified, you don't have to make a copy of it. That's generally how STL algorithms are implemented. `while (first != last)` was Stepanov's mantra.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,74 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+
+        else if(current_count == max_counter)
+        {
+            modes.emplace_back(*it);
+        }
+
+        it = end_it;
+    }
+
+    if(max_counter == 1) // Return an empty vector if mode is undefined
```
jeremy-murphy, line 557: This is not undefined, it's just GIGO. If the input is `{ 99 }` then the mode is `99`, that's it. 

And once you remove this edge case for size 1, then you can _also_ remove the special case for `first == last` because the `std::move` will be a no-op in that case.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,74 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
+    
+    while(it != last)
+    {
+        Size current_count {0};
+        auto end_it {it};
+        while(end_it != last && *end_it == *it)
+        {
+            ++current_count;
+            ++end_it;
+        }
```
jeremy-murphy, line 540: You don't have to increment `current_count` on every iteration of this loop. You can just loop until the condition is met and then `current_count += std::distance(it, end_it)`.

And then, you'll see that this loop is just `find_if` with a predicate for 'is not equal to `*it`'.

jeremy-murphy: Ahh, actually, if the input sequence is not random access, then `std::distance` will be linear. Hmm. Maybe best to leave it as is after all. :)

jeremy-murphy:
Here is a deeper issue for you: what are the concept requirements on the input elements for statistical mode? I would say they have to have an equivalence relationship (comparable by `==`), but not much else. Specifically, objects without ordering (comparable by `<`) can still have a mode calculated for them.

So, there is an uneasy mixing of sorting and ordering with mode here. It might be suitable for _special cases_ of input (i.e., integers), but it doesn't work in the general case. In the general case, the input needs to be partitioned into equivalence classes, perhaps like this: http://doi.org/10.1007/978-3-319-21840-3_36  

I know, this probably looked a simple algorithm on the surface.  ;)

jeremy-murphy:
Ah, that previous paper was actually in aid of an algorithm to find mode using equality comparisons: https://link.springer.com/chapter/10.1007/978-3-319-30139-6_28

jeremy-murphy:
And a slightly newer one: https://www.sciencedirect.com/science/article/pii/S030439751730662X

jeremy-murphy:
Anyway, if you don't have access to, or interest in, these papers on equivalence-class sorting that's perfectly OK. It shouldn't stop this code going through, which is useful for ordered types. It's just worth being aware of the whole landscape of mode algorithms.

mborland:
@jeremy-murphy Unfortunately those papers are behind a paywall, and I do not have enterprise access. I appreciate the insight.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -512,6 +512,63 @@ inline auto interquartile_range(RandomAccessContainer & v)
...
 
+template<class ForwardIterator, class OutputIterator>
+auto sorted_mode(ForwardIterator first, ForwardIterator last, OutputIterator output) -> decltype(output)
+{
+    using Z = typename std::iterator_traits<ForwardIterator>::value_type;
+    static_assert(std::is_integral<Z>::value, "Floating point values have not yet been implemented.");
+    using Size = typename std::iterator_traits<ForwardIterator>::difference_type;
+
+    std::vector<Z> modes {};
+    modes.reserve(16);
```
jeremy-murphy, line 523: A performance idea, and honestly you're under no obligation to pursue it. You can just ignore me. But did you consider trying `boost::container::small_vector` here? Because I assume _most_ of the time, `modes` is not going to grow large at all. Or it might not make a difference. Entirely up to you.

NAThompson: @jeremy-murphy : It would couple `boost::container` into `boost::math`.

mborland: @jeremy-murphy By calling `std::vector::reserve()` and using `std::vector::emplace_back()` dynamic memory allocation is avoided in most cases, which from the docs looks like what `boost::container::small_vector` accomplishes.

jeremy-murphy: OK, good, just a thought.  :)

jeremy-murphy:
Btw, did you consider the performance optimization of breaking the loop early if `max_count` > `last - it`?
In other words, if you can't find an equally large (or larger) mode in the remaining input data, then you don't need to process it.

And the algorithm wouldn't have to check on every iteration, it could just move `last` backwards each time it finds a new value for `max_count`. Of course that doesn't work for `ForwardIterator`, but then that's the disadvantage of such iterators.

mborland:
@jeremy-murphy I just compared the bench mark of the current configuration to an edited version that shifts `last` backwards as you suggested. Generally they are comparable but in the case of `sequential_multiple_test_mode` the performance dropped from O(N) to O(NlgN).

jeremy-murphy:
Ah, that's a pity about `sequential_multiple_test_mode`. Presumably it would have the most positive effect when that sequence is reversed, with the largest sequence at the front. Thanks for testing it!

doc/statistics/univariate_statistics.qbk
```diff
@@ -252,6 +263,36 @@ You should have /very/ good cause to pass negative values to the Gini coefficien
...
+    std::deque<int> d_modes;
+    std::array<int, 7> w {2, 2, 3, 1, 5, 4, 4};
+    boost::math::statistics::mode(w, std::back_inserter(d_modes));
+    // Modes are 2 and 4, d_modes.size() == 2
+
+/Nota bene/: The input data is altered: in particular, it is sorted. Makes a call to `std::sort`, and as such requires random access iterators.
+
+If your data is sorted, the following function can be used instead:
+
+    std::vector<int> v {1, 3, 2, 2, 5, 4};
```
NAThompson, line 283: Note that this data is not sorted.

doc/statistics/univariate_statistics.qbk
```diff
@@ -252,6 +263,36 @@ You should have /very/ good cause to pass negative values to the Gini coefficien
...
+    std::vector<int> modes; 
+    boost::math::statistics::sorted_mode(v, std::back_inserter(modes));
+    // Mode is 2, modes.size() == 1
+    std::deque<int> d_modes;
+    std::array<int, 7> w {2, 2, 3, 1, 5, 4, 4};
```
NAThompson, line 288: This data is not sorted.

doc/statistics/univariate_statistics.qbk
NAThompson: Might want to note that passing an unsorted list to `sorted_mode` is a bug.

NAThompson:
The two build errors are known and are from boost::serialization. Modulo the two trivial issues I found in the docs, I think this looks good.

@jeremy-murphy : Your suggestions were invaluable; I think this now is a very high quality implementation.

@mborland : Nice work, will merge once the doc diffs are in.

doc/statistics/univariate_statistics.qbk
```diff
@@ -252,6 +263,38 @@ You should have /very/ good cause to pass negative values to the Gini coefficien
...
+    boost::math::statistics::sorted_mode(v, std::back_inserter(modes));
+    // Mode is 2, modes.size() == 1
+    std::deque<int> d_modes;
+    std::array<int, 7> w {1, 2, 2, 3, 4, 4, 5};
+    boost::math::statistics::sorted_mode(w, std::back_inserter(d_modes));
+    // Modes are 2 and 4, d_modes.size() == 2  
+
+/Nota bene/: The requirements for sorted_mode are reduced to forward iterators because there is no call to `std::sort`.
+
+/Nota bene/: Passing unsorted data to sorted_mode is a bug.
```
jeremy-murphy, line 294: That's not strictly true, right? It doesn't have to be sorted, it just has to have equal values be consecutive, which is the same requirement as for `std::unique`. Sorting is just the simplest way to achieve it.

mborland: @jeremy-murphy You are correct. Consecutive equal values would work here.

NAThompson: I would probably describe this in a literary sense as a "distracting caveat"; not a true statements about the code need to be stated.

But I think having use cases would clarify whether this needs to be explicitly stated; do we have any? I personally have never needed to compute a mode.

mborland: @NAThompson I think of a case where data being handled by sorted_mode would be configured differently than what you would get from `std::sort`

PR #392, opened 2020-07-05
Fix up data generators for C++03 etc.

jzmaddock:

Issue #373, opened 2020-06-19
Jacobi Theta functions


evanmiller:
Hello! I have begun an implementation of the Jacobi Theta functions ([DLMF](https://dlmf.nist.gov/20)). My personal aspiration is to use them to implement a Kolmorogov-Smirnov distribution, but they could also be used as a complex basis for Boost's Jacobi elliptical functions. I have a few questions before I get any deeper into the implementation.

* Is there any interest in these functions?

* I haven't seen much complex-number code, nor do I have much C++ experience, but here is my proposed interface: (`template` keywords omitted)

```C++
// First function (complex)
std::complex<T> jacobi_theta1(std::complex<T> z, std::complex<T> q, const Policy& pol);
std::complex<T> jacobi_theta1(std::complex<T> z, std::complex<T> q);
// First function (real-valued)
T jacobi_theta1(T z, T q, const Policy& pol);
T jacobi_theta1(T z, T q);

// Second function (complex)
std::complex<T> jacobi_theta2(std::complex<T> z, std::complex<T> q, const Policy& pol);
std::complex<T> jacobi_theta2(std::complex<T> z, std::complex<T> q);
// Second function (real-valued)
T jacobi_theta2(T z, T q, const Policy& pol);
T jacobi_theta2(T z, T q);

// Third function (complex)
std::complex<T> jacobi_theta3(std::complex<T> z, std::complex<T> q, const Policy& pol);
std::complex<T> jacobi_theta3(std::complex<T> z, std::complex<T> q);
// Third function (real-valued)
T jacobi_theta3(T z, T q, const Policy& pol);
T jacobi_theta3(T z, T q);

// Third function minus 1, for accuracy (complex)
std::complex<T> jacobi_theta3m1(std::complex<T> z, std::complex<T> q, const Policy& pol);
std::complex<T> jacobi_theta3m1(std::complex<T> z, std::complex<T> q);
// Third function minus 1 (real-valued)
T jacobi_theta3m1(T z, T q, const Policy& pol);
T jacobi_theta3m1(T z, T q);

// Fourth function (complex)
std::complex<T> jacobi_theta4(std::complex<T> z, std::complex<T> q, const Policy& pol);
std::complex<T> jacobi_theta4(std::complex<T> z, std::complex<T> q);
// Fourth function (real-valued)
T jacobi_theta4(T z, T q, const Policy& pol);
T jacobi_theta4(T z, T q);

// Fourth function minus 1, for accuracy (complex)
[...]

PR #394, opened 2020-07-07
Jacobi Theta functions

evanmiller:
Implementations, tests, and ULP plotting programs are provided for the four Jacobi Theta functions per #373. Twenty-four public C++ functions are provided in all, covering various precision-preserving scenarios.

Documentation for collaborators is provided in the code comments. Proper documentation for end users will be provided when the implementation and APIs are finalized.

Some tests are failing; this implementation is meant to start a conversation. The core dilemma faced by the author was that large values of |q| resulted in slow convergence, and sometimes wildly inaccurate results. Following the implementation note in [DLMF 20.14](https://dlmf.nist.gov/20.14), I added code to switch over to the imaginary versions of the theta functions when |q| > 0.85.  This restored accuracy such that all of the identity tests passed for a loose-enough epsilon, but then lost precision to the point that the Wolfram Alpha spot checks failed. It is the author's hope that someone with floating-point experience can tame the exponential dragons and squeeze the ULPs back down to a reasonable range when |q| is large.

When #392 is merged I will add more thorough value tests, although I fully expect them to fail until the underlying precision issues are resolved.

As a final note, the precision issues do not affect the `z=0` case - the ULP plots indicate these return values within 2 ULP across all valid |q|. So that's a start.

Any feedback appreciated.


jzmaddock:
That looks really good!

You can save me some time by providing a couple of concrete examples which currently have low accuracy?

evanmiller:
@jzmaddock I can provide specific examples later today but the failing spot tests (`../../../b2 test_jacobi_theta`) should point you in the right direction. Also see the ULP plotting programs added to `reporting/accuracy`. Try large q / small tau (say q = 0.99 or tau = 0.001) with non-zero z (say 1.5) to trigger the `_IMAGINARY` code paths on any of the public functions.

To be clear: the real-valued code paths are inaccurate for large q. The `_IMAGINARY` code paths are accurate but imprecise. 

Just so you know: the `_IMAGINARY` functions are rescaled and rearranged a bit - they assume imaginary z, so the sines and cosines became hyperbolic. I then tried to be clever by consolidating the exponentials. Because the identity tests pass I believe the expressions are mathematically correct, but they're losing precision somehow. I suspect either the exp(-z*z) or sqrt(tau) but I am honestly out of my depth here.

Will provide actual examples tonight if I haven't heard from you.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,623 @@
...
+// transform representations above).
+//
+// Finally, the twelve functions above are each available with a third Policy
+// argument, which can be used to define a custom epsilon value. These Policy
+// versions bring the total number of functions provided by jacobi_theta.hpp
+// to twenty-four.
+//
+// See:
+// https://mathworld.wolfram.com/JacobiThetaFunctions.html
+// https://dlmf.nist.gov/20
```
NAThompson, line 86: This documentation is excellent.

jzmaddock: +1.

jzmaddock:
OK I have a test case, thanks.

Minor nits while I think of them:

* test_jacobi_theta.cpp needs to include boost/math/concepts/real_concept.hpp to build.
* The tests use some constants like M_PI which are Unix'isms: please use boost::math::constants::pi<double>() for portability.
* There are a bunch of places where you're accidentally calling ::log(double) (for example) rather than std::log(T) which will quite adversely effect the long double precisions.  Adding the using declarations or BOOST_MATH_STD_USING will fix these.  Instantiating the functions with boost::math::concepts::std_real_concept will help find all the locations.

After that the error rates don't look bad (200eps or so), but.... as an example jacobi_theta_2(3, 0.9921875) is so totally trivial to calculate (done in one iteration, and I don't see any obvious things that would loose precision) that I would like to investigate some more.

One possible improvement is in:


```
inline RealType
jacobi_theta2(RealType z, RealType q, const Policy& pol) {
    if (q <= 0.0 || q >= 1.0) {
        return policies::raise_domain_error<RealType>("boost::math::jacobi_theta2<%1%>(%1%)",
                "q must be greater than 0 and less than 1 but got %1%.", q, pol);
    }
    return jacobi_theta2tau(z, -log(q)/constants::pi<RealType>(), pol);
}
```

where log(q) *might* be better expressed as log1p(q-1) when q is ultra close to 1, but makes no difference in the case above.  And I'm not completely convinced it ever would make a difference, I need to think about that ;)

jzmaddock:
One more nit: if you instantiate with for example `number<debug_adaptor<cpp_bin_float_50::backend_type>>` ie a type that uses expression templates, then there are quite a few build errors where expressions are passed as function arguments.

The error in the case above come entirely from the line

```
delta1 = exp(tau*(z*(RealType(2*n) - z/constants::pi<RealType>())
                -constants::pi<RealType>() * RealType(n)*RealType(n)));
```

The issue being that z(2n-z/pi) happens to evaluate to very close to pi in this case.  I don't see any algebraic simplification though?

evanmiller:
> * test_jacobi_theta.cpp needs to include boost/math/concepts/real_concept.hpp to build.

Thanks

> * The tests use some constants like M_PI which are Unix'isms: please use boost::math::constants::pi() for portability.

Will do

> * There are a bunch of places where you're accidentally calling ::log(double) (for example) rather than std::log(T) which will quite adversely effect the long double precisions.  Adding the using declarations or BOOST_MATH_STD_USING will fix these.  Instantiating the functions with boost::math::concepts::std_real_concept will help find all the locations.

Aha! I never would have thought of that. It likely affects the use of `fmod` as well - this would basically reduce long doubles to doubles (for the `z` argument) on the _IMAGINARY path. Since my ULP plots use long doubles for the high-precision case, this would basically wreck all the plots.

I'll make these changes and report back.

NAThompson:
@evanmiller : It'll probably a few iterations on this, so make sure to tag your commits with `[CI SKIP]` so we don't DDOS the CI system.

evanmiller:
I've made the changes suggested and added a few more tests covering special values (did you know the third theta function is related to the gamma function?). I don't see any perceptible change to the test results though.

@NAThompson I will look into adding a performance benchmark later tonight.

evanmiller:
Good news! I think I've just about cracked the precision issues. I completed the square on the exponentials, and rewrote them as double-sided summations following [DLMF 20.13.4 and 20.13.5](https://dlmf.nist.gov/20.13).

The mean errors on the MathWorld data are now down to <10 (eps? ULPs? whatever `test_hetero` reports). They were 50-200 before. The ULP plots show the precision getting steadily worse as z grows - so there's still room for improvement. I should have a better picture on the overall accuracy once the 1000-bit code gets plugged in.

BTW What are we benchmarking against exactly? I don't see Jacobi theta functions in either GSL or RMath.

Commit incoming.

NAThompson:
> BTW What are we benchmarking against exactly? 

Ourselves! Run the benchmark under `perf record` and examine the asm; you'll enjoy it.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,646 @@
+// Jacobi theta functions
+// Copyright Evan Miller 2020
```
NAThompson, line 2: Make sure to add the Boost Licence under the copyright statement.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,646 @@
...
+
+        if (n%2) {
+            delta = delta2 - delta1;
+        } else {
+            delta = delta1 - delta2;
+        }
+
+        result += delta;
+        n++;
+    } while (abs(delta * sqrt(tau)) > eps || (abs(result * sqrt(tau)) > eps * eps && abs(delta/result) > eps));
```
NAThompson, line 207: Could you add a comment about the error analysis that leads to this termination criteria?

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,646 @@
...
+        delta2 = exp(tau*(z*(-RealType(2*n+1) - z/constants::pi<RealType>())
+                    -constants::pi<RealType>() * RealType(n + 0.5)*RealType(n + 0.5)));
+
+        delta = delta1 + delta2;
+
+        result += delta;
+        n++;
+    } while (abs(delta * sqrt(tau)) > eps || (abs(result * sqrt(tau)) > eps * eps && abs(delta/result) > eps));
+    
+    if (abs(result) < eps * eps)
```
NAThompson, line 236: I don't get this one; could you add another comment here?

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,646 @@
...
+_IMAGINARY_jacobi_theta3tau(RealType z, RealType tau, const Policy& pol) {
+    BOOST_MATH_STD_USING
+    unsigned n = 1;
+    RealType eps = policies::get_epsilon<RealType, Policy>();
+    RealType /* q_n, */ delta, delta1, delta2, result = exp(-z*z*tau/constants::pi<RealType>());
+
+    do {
+        delta1 = exp(tau*(z*(RealType(2*n) - z/constants::pi<RealType>())
+                    -constants::pi<RealType>() * RealType(n)*RealType(n)));
+        // q_n = exp(-tau * constants::pi<RealType>() * RealType(n + 0.5)*RealType(n + 0.5) );
```
NAThompson: This is commented out in every implementation.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,646 @@
...
+        return RealType(0);
+
+    return result * sqrt(tau);
+}
+
+template <class RealType, class Policy>
+inline RealType
+_IMAGINARY_jacobi_theta4tau(RealType z, RealType tau, const Policy& pol) {
+    BOOST_MATH_STD_USING
+    unsigned n = 1;
```
NAThompson, line 292: I'd go with a 64 bit counter; I'll probably try to get a few million digits out of this and it's best if it doesn't wrap. Also, I'd venture the benchmarks will show there is no performance penalty.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,672 @@
...
+            z += constants::two_pi<RealType>();
+        }
+
+        return _IMAGINARY_jacobi_theta4tau(z, 1/tau, pol);
+    }
+
+    do {
+        q_n = exp(-tau * constants::pi<RealType>() * RealType(n + 0.5)*RealType(n + 0.5));
+        delta = q_n * cos(RealType(2*n+1)*z);
+        result += RealType(2) * delta;
```
NAThompson, line 437: I worry that this cast of `2` to `RealType(2)` will prevent generation of an exponent shift in the exponent and for a mul. I'll see what godbolt says in a bit . .

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: The `sqrt(tau)` stuff was removed in the latest commit, but I can clarify what is left.

reporting/accuracy/plot_jacobi_theta1.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include <iostream>
+#include <boost/math/tools/ulps_plot.hpp>
+#include <boost/core/demangle.hpp>
+#include <boost/math/special_functions/jacobi_theta.hpp>
+
+using boost::math::tools::ulps_plot;
+
+int main() {
+    using PreciseReal = long double; // boost::multiprecision::float128;
+    using CoarseReal = double;
```
NAThompson, line 11: In this case, there is no problem, but in general I like to use `float128` to test `double`, or use `long double` to test `float`. Obviously that isn't sufficient to guarantee accuracy of the plot, but I have had problems where `long double` wasn't sufficient to get a good ulps plot for `double`.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Removed in latest commit.

reporting/accuracy/plot_jacobi_theta1.cpp
```diff
@@ -0,0 +1,31 @@
...
+    };
+
+    std::string filename = "jacobi_theta1_" + boost::core::demangle(typeid(CoarseReal).name()) + ".svg";
+    int samples = 2500;
+    int width = 1100;
+    PreciseReal clip = 100;
+    auto plot = ulps_plot<decltype(jacobi_theta1_precise), PreciseReal, CoarseReal>(jacobi_theta1_precise, -10.0, 10.0, samples);
+    plot.clip(clip).width(width);
+    std::string title = "jacobi_theta1 ULP plot at " + boost::core::demangle(typeid(CoarseReal).name()) + " precision";
+    //plot.title(title);
```
NAThompson, line 27: Is the title string unused?

reporting/accuracy/plot_jacobi_theta1.cpp
evanmiller: `boost::multiprecision::float128` does not compile on my machine - that's why I switched it to `long double`.

reporting/accuracy/plot_jacobi_theta1.cpp
evanmiller: I just copy-pasted this from `test_agm.cpp`, which has the same line commented out.

NAThompson: Note that if you don't need a title that's fine; but you'll need to label it in the docs.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Will do.

test/test_jacobi_theta.hpp
```diff
@@ -0,0 +1,517 @@
...
+            jacobi_theta2tau(z + z, tau + tau) * A,
+            jacobi_theta1tau(constants::quarter_pi<RealType>() - z, tau) * jacobi_theta1tau(constants::quarter_pi<RealType>() + z, tau),
+            eps);
+
+    _check_close( // DLMF 20.7.18
+            jacobi_theta3tau(z + z, tau + tau) * A,
+            jacobi_theta3tau(constants::quarter_pi<RealType>() - z, tau) * jacobi_theta3tau(constants::quarter_pi<RealType>() + z, tau),
+            eps);
+
+    _check_close( // DLMF 20.7.19
```
NAThompson, line 450: If you want, you can use my "math_unit_test.hpp" and use CHECK_ULP_CLOSE if you think it'll be a bit more ergonomic. It also compiles much faster that Boost.Test.

reporting/accuracy/plot_jacobi_theta1.cpp
NAThompson: Ok, then I'd use `using CoarseReal = float`; it should only depend on the difference in precisions.

reporting/accuracy/plot_jacobi_theta1.cpp
NAThompson: Bah then I need to fix that in `test_agm.cpp`. . .

NAThompson:
Screen Shot 2020-07-07 at 10 23 36 PM

It looks like your implementation is now correct; all values are basically in the condition number envelope.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Well - `n` enters into the equation as `exp(-n*n)`. By the time `n` hits 1000 you'll have roughly 500,000 digits of accuracy.

For machine precision, `n` is lucky to hit 5!

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Easy enough to change!

NAThompson:
Screen Shot 2020-07-07 at 10 32 27 PM

It's jacobi3 that confuses me. The error is pretty good, but it's uncorrelated with the condition number envelope, which feels very strange.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: I had some tests choking on identities that were supposed to evaluate to zero (e.g. passing in multiples of Pi). I can eliminate this check if you think it's best - it was designed to iron out vanishingly small return values ("epsilon of an epsilon") that the iterator treated as zero anyway.

jzmaddock: Looks wrong to me as well - very small return values should presumably be possible?

If there's something that should be exactly zero then I'd rather we look at those as special cases.  Also test values that are close to but not exactly zero.

include/boost/math/special_functions/jacobi_theta.hpp
jzmaddock: +1, also in the multiprecision case, multiplying by a literal integer is *much* more efficient than the cast.  BTW we assume in our conceptual requirements that all number types will interact correctly with integer types, so multiplying/dividing by literals (mostly 2!) is pretty commonplace in the library.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: If I remove this check, I get some disagreement with Wolfram Alpha (which returns 0 instead of a Very Small Number) for e.g. `jacobi_theta4(0, 0.9921875)`. I will leave the check out for now, and hope that the 1000-bit implementation will clear things up.

Regarding exactly zero, this only happens in a couple of cases, as far as I know: `jacobi_theta1(n*pi, q)` and `jacobi_theta2((n+0.5)*pi, q)`. The special case  `jacobi_theta1(0, q)=0` is already handled, but any other case will slip through the floating-point cracks, unless we re-parameterize z to be a phase (à la `sin_pi`).

Studying the equations for `theta3` and `theta4`, these should never truly return zero. Perhaps a strongly worded letter to Wolfram Alpha is in order on this topic.

reporting/accuracy/plot_jacobi_theta2.cpp
```diff
@@ -7,15 +7,14 @@
...
+    using CoarseReal = float;
 
-    // TODO Small q is fine but larger q has pretty bad ULPs
     auto jacobi_theta_coarse = [](CoarseReal z) {
-        return boost::math::jacobi_theta2<CoarseReal>(z, 0.05);
+        return boost::math::jacobi_theta2<CoarseReal>(z, 0.9);
     };
     auto jacobi_theta_precise = [](PreciseReal z) {
-        return boost::math::jacobi_theta2<PreciseReal>(z, 0.05);
+        return boost::math::jacobi_theta2<PreciseReal>(z, 0.9);
```
NAThompson, line 17: 0.9 is somewhat of a bad choice, since it's parsed as a double and then promoted to long double or parsed as double and cast back to float. That adds a source of inaccuracy that is uninteresting in this context. I'd use a representable here, like 15/16.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,665 @@
...
+inline RealType jacobi_theta4m1(RealType z, RealType q, const Policy& pol);
+template <class RealType, class Policy>
+inline RealType jacobi_theta3m1tau(RealType z, RealType tau, const Policy& pol);
+template <class RealType, class Policy>
+inline RealType jacobi_theta4m1tau(RealType z, RealType tau, const Policy& pol);
+
+// Internal convergence criterion:
+// If the delta is more than an absolute epsilon, keep going.
+// If the delta is more than an relative epsilon, and the denominator
+// is non-zero, keep going.
```
NAThompson, line 168: I'm not sure this comment is very informative; I was hoping for something like "expected error is cond(f; x)eps/2, so when delta <..." and then some perturbative analysis.

evanmiller: The comment represents the extent of my thinking and, well, education. I'm happy to incorporate a more formal convergence analysis but it's simply outside my ken.

reporting/accuracy/plot_jacobi_theta2.cpp
evanmiller: Good catch, will fix.

include/boost/math/special_functions/jacobi_theta.hpp
NAThompson: Agreement with Wolfram Alpha is not a requirement; there are quite a few bugs in Mathematica. Personally I'd rather see unit tests of properties and quadratures to check accuracy.

include/boost/math/special_functions/jacobi_theta.hpp
NAThompson: lol, alright np. I'll see if I can get a minute to work out some error analysis for these functions.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Re: quadrature, the first 3 equations here ought to verify the whole `tau` domain for z=0: https://dlmf.nist.gov/20.10

I previously tried checking relationships with elliptic integrals (e.g. [DLMF 20.9.4](https://dlmf.nist.gov/20.9)), but I ran into some accuracy issues and didn't know which function was to blame. So the existing property checks mostly use simple functions.

reporting/accuracy/plot_jacobi_theta1.cpp
evanmiller: Uncommented these lines, resolving

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Fixed in latest commit. I opted for `delta + delta` which I imagine compiles to the same thing.

include/boost/math/special_functions/jacobi_theta.hpp
jzmaddock: IMO there's no need to require `delta < eps`, normally we would just use `abs(result * eps) > abs(delta)` as if that's true, then adding `delta` can have no effect on `result`.  The only exception would be if `delta` can grow as well as shrink.

evanmiller: The sign of `delta` can alternate, which could theoretically put `result` at 0 after two iterations. So I wanted to handle that somehow, and distinguish the situation from an underflow.

reporting/accuracy/plot_jacobi_theta2.cpp
evanmiller: Fixed in latest commit.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Added quadrature tests and significantly tightened test tolerances in the latest commit. Quadrature is such an elegant way to verify a function, by the way!

Will revisit the elliptic integral relations tonight.

evanmiller: Added tests against Jacobi elliptic functions (`sn`, `cd`, etc) in latest commit, as well as a couple of tests against `ellint_rf`.

NAThompson:
I don't know exactly what you did, but the jacobi_theta3 ulps plot looks a lot better:

Screen Shot 2020-07-08 at 9 50 04 PM

There is still room for improvement where the condition number envelope -> 0, but now the error is correlated with the condition number.

evanmiller:
> There is still room for improvement where the condition number envelope -> 0, but now the error is correlated with the condition number.

Yes that does look better - might be the tweak I made to the `m1` functions today.
> Other than around q = 1; I expect the error to be much higher there; but it's identically zero. Do you understand this?

I am guessing `exp` is underflowing here and these functions are returning 0 at all precisions (cf. Mathematica discussion above).

evanmiller:
The test suite now includes quadrature for the 4 main functions and basically every identity I could get my hands on. I've tightened the tolerances on the existing identity equations to 100EPS or so. Based on the ULP plots, there's room for improvement, but I am basically satisfied with the function implementations at this point. As a bonus, the new tests now serve as a cross-check for related elliptic functions and integrals (including the Riemann zeta function, which I didn't expect to see here!).

When #392 gets merged in, I'll go about generating a grid of high-precision point-test data and finalizing the tolerances. If that all looks good, and assuming everyone is satisfied with the proposed API, then we can start discussing user-facing documentation.

I should warn the audience that I don't have any particular interest in benchmarks or performance optimization with these functions - my preference on that front would be to put it off until later. I do enjoy performance engineering, but I like to know I'm working on the frontier of an actual performance bottleneck before investing any time into it.

NAThompson:
> I should warn the audience that I don't have any particular interest in benchmarks or performance optimization with these functions - my preference on that front would be to put it off until later. I do enjoy performance engineering, but I like to know I'm working on the frontier of an actual performance bottleneck before investing any time into it.

There is no need to agonize, but we at least need to know what the performance is. Part of the fun is learning new skills! Drop a benchmark in real quick and run it under perf; you'll have a blast.

evanmiller:
(Following up from @jzmaddock's comment yesterday)

> One possible improvement is in:
> 
> ```
> template <class RealType, class Policy>
> inline RealType
> jacobi_theta2(RealType z, RealType q, const Policy& pol) {
>     if (q <= 0.0 || q >= 1.0) {
>         return policies::raise_domain_error<RealType>("boost::math::jacobi_theta2<%1%>(%1%)",
>                 "q must be greater than 0 and less than 1 but got %1%.", q, pol);
>     }
>     return jacobi_theta2tau(z, -log(q)/constants::pi<RealType>(), pol);
> }
> ```
> 
> where log(q) _might_ be better expressed as log1p(q-1) when q is ultra close to 1, but makes no difference in the case above. And I'm not completely convinced it ever would make a difference, I need to think about that ;)

`log1p` only helps if you have a precise value of `x`. If we `q` is close to 1, then subtracting 1 from it will yield an imprecise result and defeat its purpose. (Think of it this way - if it help with precision, then `log` would check for values close to 1 and call `log1p` itself.)

It would only make sense to call `log1p` if the user were passing in (for example) the complement of q. But if they have a high-precision complement of q, then the user can always call `log1p` themselves and pass the result (divided by pi) to the `tau` parameterization.

evanmiller:
> There is no need to agonize, but we at least need to know what the performance is. Part of the fun is learning new skills! Drop a benchmark in real quick and run it under perf; you'll have a blast.

My intention was not to antagonize - I'm just trying to prioritize my time.

jzmaddock:
>log1p only helps if you have a precise value of x

I think that's right.

evanmiller:
I've added a test data generator to the code base, and put some of its output under test coverage. Right now it's just testing random data in z-tau space - next I'd like to hone in on the problematic area of |q|>0.99.

evanmiller:
I have added user documentation and finalized the test tolerances. A couple of questions:

* Is there a "house style" for equations? I made some equation PNGs/SVGs using a LaTeX wrapper on my machine, and added them to the `equations` folder. I see there is a [script](https://github.com/boostorg/math/blob/develop/doc/equations/generate.sh) to make these files from MML, but the script looks not-especially-portable.

* How do I go about generating a multi-platform Accuracy table for inclusion in the docs?

NAThompson:
@evanmiller : I use [this](https://viereck.ch/latex-to-svg/) to make the equations, then put the result in the `equations` folder. It's not optimal, but it works ok.

> How do I go about generating a multi-platform Accuracy table for inclusion in the docs?

I personally have never bothered with this; if you have an ULPs plot this goes a long way towards understanding the behavior. Also, don't compiler flags have a bigger effect than platform?

evanmiller:
@NAThompson I've added some ULPs plots to the docs. I'm not thrilled with these two:

jacobi_theta1
jacobi_theta4

But the others look OK. I believe this PR is otherwise ready for final evaluation.

NAThompson:
@evanmiller : Those are good! The green line is the condition number of function evaluation; you really can't expect to do better than than.

Locally could you drive up the number of samples and screenshot it? Then we can see how good the q->1 evaluation is.

evanmiller:
OK! I was hoping the errors would center around zero, but if it's good enough for you then it's good enough for me.

Here is a graph zoomed in on q > 0.9 with 25,000 samples:

jacobi_theta1

I'm guessing the discontinuities have something to do with the number of iterations / convergence criteria.

NAThompson:
> I was hoping the errors would center around zero

Yeah, that almost never happens; weird structure always arises. There's a section on this in Higham's Accuracy and Stability of Numerical Algorithms, section 1.17 "Rounding Errors are Now Random".

evanmiller:
@NAThompson Thanks. Since the errors are acceptable, I will go ahead and pull off the WIP sticker here. Let me know if I've overlooked anything.

jzmaddock:
>Is there a "house style" for equations? I made some equation PNGs/SVGs using a LaTeX wrapper on my machine, and added them to the equations folder. I see there is a script to make these files from MML, but the script looks not-especially-portable.

Anything broadly similar to what we have already is fine - don't worry about .png's as we don't use them any more (HTML output uses SVG's as does PDF output).  If you could add the latex (if you still have it) as a comment in the .qbk source next to the equation include that would be useful just in case we later spot any typos ;)

evanmiller:
> If you could add the latex (if you still have it) as a comment in the .qbk source next to the equation include that would be useful just in case we later spot any typos ;)

Will do.

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+[section:jacobi_theta Jacobi Theta Functions]
+
+[section:jacobi_theta_overview Overview of the Jacobi Theta Functions]
+
+The Jacobi Theta functions are a set of four inter-related periodic functions of /x/ which are expressed in terms of a parameter /q/ (also called the nome), or a closely related value, [tau]
+[footnote [@https://en.wikipedia.org/wiki/Theta_function Wikipedia: Theta function]]
+[footnote [@https://mathworld.wolfram.com/JacobiThetaFunctions.html Weisstein, Eric W. "Jacobi Theta Functions." From MathWorld - A Wolfram Web Resource.]]
+[footnote [@https://dlmf.nist.gov/20 Digital Library of Mathematical Functions: Theta Functions, Reinhardt, W. P.,  Walker, P. L.]].
+
+Mathematically, the functions are described by:
```
NAThompson, line 17: "Mathematically, the functions are described by:" -> "They are" (no semicolon)

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+
+[equation jacobi_theta1] [/ \theta_1(x, q) = 2 \sum_{n=0}^\infty (-1)^n q^{(n+\frac{1}{2})^2} \sin((2n+1)x) ]
+[equation jacobi_theta2] [/ \theta_2(x, q) = 2 \sum_{n=0}^\infty q^{(n+\frac{1}{2})^2} \cos((2n+1)x) ]
```
NAThompson, line 20: I personally like `:=` for definitions.

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+[equation jacobi_theta3] [/ \theta_3(x, q) = 1 + 2 \sum_{n=1}^\infty q^{n^2} \cos(2nx) ]
+[equation jacobi_theta4] [/ \theta_4(x, q) = 1 + 2 \sum_{n=1}^\infty (-1)^n q^{n^2} \cos(2nx) ]
+
+Graphically, setting /q/=0.15:
```
NAThompson, line 24: I'd put this after the graph, as a caption; then "Plots of the four theta functions for /q/=0.15.

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+Appropriately multiplied and divided, these four theta functions can be used
+to implement the [link math_toolkit.jacobi.jac_over Jacobi elliptic functions]; but this is not really
+recommended, as the existing Boost implementations are likely faster and
+more accurate.
+
+Most applications will want to use the /q/ parameterization of the functions: `__jacobi_theta1`, `__jacobi_theta2`, `__jacobi_theta3`, and `__jacobi_theta4`, where /q/ is restricted to the domain (0, 1). However, a second [tau] parameterization is provided for all four functions, where
+
+[equation jacobi_theta_nome] [/ q = \exp(i\pi\tau) ]
+
+Note that Boost assumes that [tau] is a purely imaginary number. This assumption is not required by the mathematics, but it does cover the most common application domains.
```
NAThompson, line 37: I like to make every sentence on its own line just to make future diffs more legible.

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+
+[heading Accuracy considerations]
+
+The purpose of the [tau] parameterization is to provide increased accuracy either when /q/ is expressible as an exponential or is very close to unity. For example, instead of:
```
NAThompson, line 41: Don't put semicolons before equations; the mathematics is part of the sentence.

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+
+  jacobi_theta1(x, exp(-a));
+
+A more accurate computation will take advantage of [tau]:
+
+  jacobi_theta1tau(x, a / boost::math::constants::pi<T>());
+
+Internally, Boost implements the /q/ parameterization by taking the logarithm of /q/ and passing it to the [tau] parameterization; as such, using the [tau] parameterization directly will generally yield greater precision. As another example, if the complement of /q/ is known with great accuracy, then instead of:
```
NAThompson, line 49: Newline for each sentence.

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+It is more accurate to use `__log1p` and pass in the result to the [tau] version:
+
+  jacobi_theta1tau(x, -boost::math::log1p(-q_complement) / boost::math::constants::pi<T>());
+
+Additional "minus 1" versions of the third and fourth theta functions are provided. Similar in spirit to `__expm1`, these functions return one less than the evaluated function, and yield increased accuracy when /q/ is small.
+
+[heading Testing]
+
+Results of the theta functions are tested against Wolfram Alpha data, as well as random values computed at high precision. In addition, the tests verify the majority of the identities described in [@https://dlmf.nist.gov/20.7 DLMF Chapter 20.7].
+
```
NAThompson, line 62: Could you include the ulps plots you produced?

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+[optional_policy]
+
+[heading Accuracy]
+
+The following [link math_toolkit.ulps_plots ULPs plot] is representative, fixing /q/=0.5 and varying /x/ from 0 to 2[pi]:
+
+[graph jacobi_theta1_float]
+
+The envelope represents the function's [link math_toolkit.cond condition number]. Note that relative accuracy degenerates periodically near [theta][sub 1]=0.
```
NAThompson, line 109: A better link would be:

https://en.wikipedia.org/wiki/Condition_number

(I couldn't get the shard to the one variable section, but that would be even better.

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+
+[graph jacobi_theta1]
+
+[optional_policy]
+
+[heading Accuracy]
+
+The following [link math_toolkit.ulps_plots ULPs plot] is representative, fixing /q/=0.5 and varying /x/ from 0 to 2[pi]:
+
+[graph jacobi_theta1_float]
```
NAThompson, line 107: Looking at this graph, it appears that the x->0 accuracy is not limited by the condition number of function evaluation and hence it could be implemented better. Is this a case to recommend use of `jacobi_theta1tau`?

doc/sf/jacobi_theta.qbk
```diff
@@ -0,0 +1,349 @@
...
+      ``__sf_result`` jacobi_theta2tau(T x, T tau, const Policy&);
+  }} // namespaces
+
+[heading Description]
+
+The functions calculate the value of second [link math_toolkit.jacobi_theta.jacobi_theta_overview Jacobi Theta function], parameterized either in terms of the nome /q/:
+
+[equation jacobi_theta2] [/ \theta_2(x, q) = 2 \sum_{n=0}^\infty q^{(n+\frac{1}{2})^2} \cos((2n+1)x) ]
+
+Or in terms of an imaginary [tau]:
```
NAThompson, line 157: I'm confused. You say you enforce that tau is purely imaginary, yet your function signature says `tau` is the same type as `x`. . .

tools/jacobi_theta_data.cpp
```diff
@@ -0,0 +1,66 @@
+
+#include "mp_t.hpp"
+#include <boost/math/tools/test_data.hpp>
+#include <boost/test/included/prg_exec_monitor.hpp>
+#include <boost/math/special_functions/jacobi_theta.hpp>
+#include <fstream>
```
NAThompson, line 6: I don't understand what this file is doing. It appears you are spitting out values of the function from your own code; but isn't that tautologous?

NAThompson:
@evanmiller : Could you drop this file into `reporting/performance/jacobi_theta_performance.cpp`?

```cpp
//  (C) Copyright Nick Thompson 2020.
//  Use, modification and distribution are subject to the

#include <random>
#include <benchmark/benchmark.h>
#include <boost/math/special_functions/jacobi_theta.hpp>
#include <boost/multiprecision/float128.hpp>
#include <boost/multiprecision/mpfr.hpp>
#include <boost/multiprecision/cpp_bin_float.hpp>

using boost::multiprecision::number;
using boost::multiprecision::mpfr_float_backend;
using boost::multiprecision::float128;
using boost::multiprecision::cpp_bin_float_50;
using boost::multiprecision::cpp_bin_float_100;
using boost::math::jacobi_theta1;
using boost::math::jacobi_theta1tau;

template<class Real>
void JacobiTheta1(benchmark::State& state)
{
    std::random_device rd;
    std::mt19937_64 mt(rd());
    std::uniform_real_distribution<long double> unif(0,0.01);

    Real x = static_cast<Real>(unif(mt));
    Real q = static_cast<Real>(unif(mt));
    for (auto _ : state)
    {
        benchmark::DoNotOptimize(jacobi_theta1(x, q));
        x += std::numeric_limits<Real>::epsilon();
    }
}

BENCHMARK_TEMPLATE(JacobiTheta1, float);
BENCHMARK_TEMPLATE(JacobiTheta1, double);
BENCHMARK_TEMPLATE(JacobiTheta1, long double);
BENCHMARK_TEMPLATE(JacobiTheta1, float128);
BENCHMARK_TEMPLATE(JacobiTheta1, number<mpfr_float_backend<100>>);
BENCHMARK_TEMPLATE(JacobiTheta1, number<mpfr_float_backend<200>>);
BENCHMARK_TEMPLATE(JacobiTheta1, number<mpfr_float_backend<300>>);
BENCHMARK_TEMPLATE(JacobiTheta1, number<mpfr_float_backend<400>>);
BENCHMARK_TEMPLATE(JacobiTheta1, number<mpfr_float_backend<1000>>);
BENCHMARK_TEMPLATE(JacobiTheta1, cpp_bin_float_50);
BENCHMARK_TEMPLATE(JacobiTheta1, cpp_bin_float_100);

template<class Real>
void JacobiTheta1Tau(benchmark::State& state)
{
    std::random_device rd;
    std::mt19937_64 mt(rd());
    std::uniform_real_distribution<long double> unif(0,0.01);

    Real x = static_cast<Real>(unif(mt));
    Real q = static_cast<Real>(unif(mt));
    for (auto _ : state)
    {
        benchmark::DoNotOptimize(jacobi_theta1tau(x, q));
        x += std::numeric_limits<Real>::epsilon();
    }
}

BENCHMARK_TEMPLATE(JacobiTheta1Tau, float);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, double);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, long double);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, float128);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, number<mpfr_float_backend<100>>);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, number<mpfr_float_backend<200>>);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, number<mpfr_float_backend<300>>);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, number<mpfr_float_backend<400>>);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, number<mpfr_float_backend<1000>>);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, cpp_bin_float_50);
BENCHMARK_TEMPLATE(JacobiTheta1Tau, cpp_bin_float_100);

BENCHMARK_MAIN();
```

Also, running this file, I find that something has gone wrong with the ADL for `mpfr` types, there are some warnings from unused policies from `g++-10`.

doc/sf/jacobi_theta.qbk
```diff
@@ -32,7 +32,7 @@ more accurate.
 
 Most applications will want to use the /q/ parameterization of the functions: `__jacobi_theta1`, `__jacobi_theta2`, `__jacobi_theta3`, and `__jacobi_theta4`, where /q/ is restricted to the domain (0, 1). However, a second [tau] parameterization is provided for all four functions, where
```
NAThompson, line 33: Is there equivalent Mathematica syntax that has the same nome/tau/scaling conventions? It might be nice to put that in here  . . .

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,616 @@
...
+
+    do {
+        q_n = exp(-tau * constants::pi<RealType>() * RealType(n)*RealType(n));
+        delta = q_n * cos(RealType(2*n)*z);
+        result += delta + delta;
+        n++;
+    } while (!_jacobi_theta_converged(result, q_n, eps));
+
+    if (abs(result) < eps * eps)
+        return RealType(0);
```
NAThompson, line 436: I don't understand why this needs to be done. There are plenty of floating point numbers less than eps*eps; they are perfectly valid outputs aren't they?

include/boost/math/special_functions/jacobi_theta.hpp
NAThompson: The other point that you are doing a slow division instead of a faster multiplication still stands though. There is still a bit of thinking to do on this termination condition . . .

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: You are correct; I removed similar checks after previous discussion, but forgot to take this one out as well.

tools/jacobi_theta_data.cpp
evanmiller: > @evanmiller : Could you drop this file into `reporting/performance/jacobi_theta_performance.cpp`?

Will do!

evanmiller: This is what the other `tools/*_data.cpp` files do – use a high-precision implementation to compare against the low-precision implementations.

evanmiller:
> @evanmiller : Could you drop this file into `reporting/performance/jacobi_theta_performance.cpp`?

Will do!

doc/sf/jacobi_theta.qbk
evanmiller: I don't think Mathematica provides a tau parameterization. I'll add a link in the docs to EllipticTheta, though.

doc/sf/jacobi_theta.qbk
evanmiller: `tau` the C++ argument is real-typed. The provided number is implicitly multiplied by *i*. This seemed simpler than trying to provide a complex-valued API.

doc/sf/jacobi_theta.qbk
evanmiller: `tau` can help with precision issues on `q` but not `x`.

doc/sf/jacobi_theta.qbk
evanmiller: OK

doc/sf/jacobi_theta.qbk
evanmiller: OK

doc/sf/jacobi_theta.qbk
evanmiller: OK

doc/sf/jacobi_theta.qbk
evanmiller: OK

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,616 @@
...
+    return abs(delta) < eps && (abs(result) == 0.0 || abs(delta/result) < eps);
+}
+
+template <class RealType>
+inline RealType
+_jacobi_theta_sum(RealType tau, RealType z_n, RealType z_increment, RealType eps) {
+    RealType delta, partial_result = 0;
+
+    do {
+        delta = exp(-tau*z_n*z_n/constants::pi<RealType>());
```
NAThompson, line 189: Maybe precompute `tau/constants::pi<RealType>()` so that there isn't a division on each iteration.

People don't realize that elementary special function evaluations are in the same complexity class as division!

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,616 @@
...
+        while (z < -constants::pi<RealType>()) {
+            z += constants::two_pi<RealType>();
+        }
+
+        return _IMAGINARY_jacobi_theta1tau(z, 1/tau, pol);
+    }
+
+    do {
+        q_n = exp(-tau * constants::pi<RealType>() * RealType(n + 0.5)*RealType(n + 0.5) );
+        delta = q_n * sin(RealType(2*n+1)*z);
```
NAThompson, line 316: Calling `sin` on each iteration seems expensive; can't it be done via a Clenshaw recurrence?

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,616 @@
...
+        while (z < -constants::pi<RealType>()) {
+            z += constants::two_pi<RealType>();
+        }
+
+        return _IMAGINARY_jacobi_theta4tau(z, 1/tau, pol);
+    }
+
+    do {
+        q_n = exp(-tau * constants::pi<RealType>() * RealType(n + 0.5)*RealType(n + 0.5));
+        delta = q_n * cos(RealType(2*n+1)*z);
```
NAThompson, line 380: Clenshaw recurrence instead of `cos` on each iteration?

doc/sf/jacobi_theta.qbk
NAThompson: I'm getting more confused about the `tau` parametrization. It now looks to me like you are trading a multiplication on each iteration for a logarithm and a call to `std::exp` on each iteration.  . .

tools/jacobi_theta_data.cpp
jzmaddock: Sort of, here's how I approach this:

Let's say we have a principal method, plus a bunch of special case handling (asymptotic expansions at one end of the range say as an example).  The main things that can go wrong:

1) The principal method has an outright bug (eg terms in series not correctly calculated).
2) The principal method looses precision due to either cancellation or an incorrect termination condition.
3) There are bugs in the special cases, or the selection logic between them.

We can deal with (1) by comparison to a "known good" - so job done on that I think.

For (2) and (3) what I try and do is calculate spot values at very high precision using the principal method alone, the important thing is to remove all the special case handling as otherwise as @NAThompson says you're not really testing much.  In an ideal world you would use a completely different method, but those are really hard to find.  When I remember, what I should have done (but didn't always) is archive a "naive" version of the principal method in the test value generator file and call that rather than the actual method - I suspect I may have been sloppy and just quickly commented stuff out in the header occasionally though :(

However, with regard to different methods: do the defining series for these functions satisfy that?  They don't have to be fast for test value generation as they're only run the once.  In extreme cases (1F1) I've resorted to interval arithmetic using mpfi_float to handle series with lots of cancellation when generating spot tests.

doc/sf/jacobi_theta.qbk
evanmiller: Implementing purely in `q`, without a `log`, would require a `pow` in place of the `exp`. The private `IMAGINARY` code paths have to do the `exp` anyway since they are implementing hyperbolic sine and cosine.

Exposing `tau` has the added benefit of letting the user provide a pre-logged argument for the sake of precision, e.g. with `q` close to 1 (so the user can use `log1p`) or if the argument provided would have been an exponential such as `exp(-a)`.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: The iteration count is typically between 2 and 5 so I'm not sure a recurrence formula will buy much (also because there's already an `exp` in each iteration).

doc/sf/jacobi_theta.qbk
NAThompson: You don't need a call to `pow`, since q^(n+1)^2 = q^n^2*qn*qn*q. Here's some pseudocode:

```cpp
template<typename Real>
Real new_theta1(Real x, Real q)
{
    using std::sqrt;
    using std::sin;
    using std::abs;
    Real scale = 2*sqrt(sqrt(q));
    Real term = sin(x);
    Real sum = term;
    size_t n = 1;
    Real qn = q;
    Real qnsq = q;
    Real prefactor = qnsq*qn;
    while (prefactor > abs(sum)*std::numeric_limits<Real>::epsilon())
    {
        term = prefactor*sin((2*n+1)*x);
        if (n & 1) {
            sum -= term;
        } else {
            sum += term;
        }
        qn *= q;
        qnsq = qnsq*q*qn*qn;
        prefactor = qnsq*qn;
        ++n;
    }
    return scale*sum;
}
```

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: If we're going to squeeze out divisions like that, it might make sense to divide `tau` by `pi` further up the call stack.

include/boost/math/special_functions/jacobi_theta.hpp
NAThompson: Ok, so most Clenshaw recurrence discussions descend rather than ascend, so we instead can use

sin((2(n+1)+1)x) = sin((2n+1)x)*cos(2x) + cos((2n+1)x)*sin(2x)

NAThompson: Combine this with the recurrence for q^n^2 and you can do the whole thing with a single call to std::cos and std::sin and then only muls and adds after that.

NAThompson:
I'm a bit concerned about this; it seems like it might not be terminating on the right timescale:

Screen Shot 2020-08-02 at 2 10 57 PM

tools/jacobi_theta_data.cpp
evanmiller: That makes sense. We have essentially two implementations of each function, the `IMAGINARY` path (for `tau` < 1) and the regular path. The code chooses the one that is expected to converge more quickly. So one possibility would be to invert the branch logic so that the test data always chooses the slower path when generating test data.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,616 @@
...
+    return result * sqrt(tau);
+}
+
+// Begin public API
+
+// First Jacobi theta function (Parameterized by tau - assumed imaginary)
+// = 2 * Σ (-1)^n * exp(iπτ*(n+1/2)^2) * sin((2n+1)z)
+template <class RealType, class Policy>
+inline RealType
+jacobi_theta1tau(RealType z, RealType tau, const Policy& pol)
```
jzmaddock, line 288: Point of pedantry: I'd prefer it if the actual entry points followed the logic in https://www.boost.org/doc/libs/1_73_0/libs/math/doc/html/math_toolkit/special_tut/special_tut_impl.html, what we gain is the ability to handle arguments of different types (including integers), plus expression templates when used with multiprecision types.

On a related note, I'd like to see the entry points added to this test: https://github.com/boostorg/math/blob/develop/test/compile_test/instantiate.hpp,   

Also note that each special function appears in multiple sections in that file - you will very likely get test failures when doing this, but it does catch a lot of issues - including now, not forwarding policies correctly ;)

evanmiller: To be clear - should I be using `promote_args` and friends in the public API? (I was advised against this previously.)

I've added the entry points (privately) to instantiate.hpp - lots of failures - will keep me busy tomorrow.

doc/sf/jacobi_theta.qbk
evanmiller: Very cool! Happy to explore performance improvements after everything else is settled.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: How's it look now?

jzmaddock: >To be clear - should I be using promote_args and friends in the public API? (I was advised against this previously.)

Yeah, my bad, you had a complex number interface before and promote_args doesn't work (easily) with that.  Now that it's all real-valued, might as well be consistent with everything else.

aside 1: promote_args is mis-named, should really be "make_common_floating_point_type" or something.
aside 2: I guess it would help if I updated it to function with complex number types as well!

>I've added the entry points (privately) to instantiate.hpp - lots of failures - will keep me busy tomorrow.

Likely most will disappear once the argument type handling is in place.

>How's it look now?

You're way ahead of me, will catch up with this later!

include/boost/math/special_functions/jacobi_theta.hpp
NAThompson: Ok, I think I've figured this out. The reason that the sign of `delta` can alternate is because you're including the oscillatory part in it (sin((2n+1)x) or cos((2n+1)x)). I think you should only be testing the magnitude of q^n^2, because the magnitude of complex phase of the original complex valued Jacobi function is always unity. As an aside, the auxiliary functions look much more difficult to implement individually than just using the original Jacobi function in the complex plane. . .

evanmiller: The convergence tests operate on `q_n` and do not involve the oscillatory part. My original intention was to operate in the complex plane, but the API got too complicated when trying to support arbitrary numeric types.

NAThompson: Wait then aren't all these calls to `abs` superfluous?

evanmiller: Sorry, I should have said: the numerator is `q_n` (always positive) but the denominator is `result` (which may oscillate). It would probably be better to change the denominator to `previous_q_n`.

NAThompson: I personally might track \sum q^{n^2} since it provides a nice L1 bound on the sum.

evanmiller: I'm not sure the sum will work because of the alternating signs. In addition to the oscillating component there is the (-1)^n in two of the formulas. With q close to 1, the straight sum could be large relative to the final result, and so an error that is small relative to the sum may be large relative to the result.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -0,0 +1,830 @@
...
+inline typename tools::promote_args<T, U>::type jacobi_theta4m1tau(T z, U tau, const Policy& pol);
+
+// Internal convergence criterion:
+// If the delta is more than an absolute epsilon, keep going.
+// If the delta is more than an relative epsilon, and the denominator
+// is non-zero, keep going.
+template <class RealType>
+inline bool
+_jacobi_theta_converged(RealType result, RealType delta, RealType eps) {
+    return abs(delta) < eps && (abs(result) == 0.0 || abs(delta/result) < eps);
```
jzmaddock, line 180: You need a using std::abs here.

evanmiller: I will plan to eliminate `abs` when I revamp this function per other discussions.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: I have updated the convergence criterion to compare `q_n` to the previous one. Everything including ULP plots look pretty much the same. Let me know what you think.

jzmaddock:
This looks like it's ready to go?  @NAThompson ?

NAThompson:
I'm happy with this; the only potential point of improvement is the trigonometric recurrence and the q^{n^2} recurrence. But perfect is the enemy of the good.

evanmiller:
@NAThompson's ideas may also yield accuracy improvements – but those can always take the form of a future PR. So you have my blessing to merge as well.

pulver:
Are the build errors on https://travis-ci.org/github/boostorg/math/builds/718417184 related to this PR?

NAThompson:
My bad; I told @evanmiller to `[CI SKIP]` until he was ready to merge, and then I just merged it. So looks like this one is my fault.

jzmaddock:
Testing the "obvious" fixes now....

jzmaddock:
Tentative fixes pushed: just looks like the usual concept-failures.  Glad someone was paying attention to develop results as I missed that!

Issue #398, opened 2020-07-08
cyl_bessel_j_zero does not respect policies::promote_double<false>


evanmiller:
I'm auditing a code base to eliminate uses of `long double` on x86 (i.e. eliminate all x87 instructions). After compiling, I dump the disassembly and look for forbidden instructions and function calls.

I found that the following small test program - which I am led to believe by Boost documentation will not promote doubles to long doubles - includes calls to a number of `long double` functions when compiled.

```
#include <boost/math/special_functions/bessel.hpp>

int main() {
    typedef boost::math::policies::policy<
        boost::math::policies::promote_double<false>
        > numeric_policy;

    boost::math::cyl_bessel_j_zero((double)0.5, 1, numeric_policy());
}
```

A sampling of disassembly:

```
$  otool -t -V a.out | grep 'long double' | head -n 5
00000001000039a0	callq	__ZN5boost4math7lanczos19lanczos_initializerINS1_12lanczos17m64EeE4initC2Ev ## boost::math::lanczos::lanczos_initializer<boost::math::lanczos::lanczos17m64, long double>::init::init()
000000010000db3e	callq	__ZL4fabse ## fabs(long double)
000000010000db67	callq	__ZL3cose ## cos(long double)
000000010000db98	callq	__ZL5floore ## floor(long double)
000000010000dc5d	callq	__ZL3sine ## sin(long double)
```

Is this expected behavior, or a bug?

I realize that I can disable x87 with a compiler flag - but my compiler is actually crashing when I try that. So I am hoping to resolve the issue within Boost.

NAThompson:
The Lanczos one I recognize: It's required. There is an ill-conditioned filter computation that really does need that extra precision.

evanmiller:
I am porting code to a platform without native 80-bit or 128-bit (i.e. 'long double' is equivalent to 'double').

Will the Lanczos code fail to converge on such a platform?

NAThompson:
@evanmiller : It won't fail to converge, the filter coefficients will just not be as accurate. For digital filters, we generally want half-ulp accuracy precomputed once, under the assumption that applying the filter to data will dominate the execution time.

evanmiller:
@NAThompson Is it reasonable to ask that this coefficient computation respect the `promote_double` policy? It is important for my application that the same result is returned on all platforms, even if that result is not as accurate as it could be.

NAThompson:
[...]

PR #402, opened 2020-07-15
Remove more long doubles

evanmiller:
Eliminate unwanted long doubles from `unchecked_factorial` as well as `make_big_value`. The behavior of both functions should be identical (since the values were demoted to back to a `double` anyway) but these changes will remove references to `long double` from compiled code that doesn't want it.

See #398.


include/boost/math/tools/big_constant.hpp
```diff
@@ -33,9 +33,12 @@ struct numeric_traits<__float128>
    static const int max_exponent = 16384;
    static const bool is_specialized = true;
 };
-#else
+#elif LDBL_DIG > DBL_DIG
 typedef long double largest_float;
 #define BOOST_MATH_LARGEST_FLOAT_C(x) x##L
+#else
+typedef double largest_float;
+#define BOOST_MATH_LARGEST_FLOAT_C(x) x
```
jzmaddock, line 41: Surely this breaks long double support for those that want it?

evanmiller: This `else` condition is met when double and long double are identical (`LDBL_DIG == DBL_DIG`) - so either would qualify as the largest float on those platforms. I do not believe this will have any run-time effect or break any compiles but I could be wrong (maybe if the result of `BOOST_MATH_LARGEST_FLOAT_C` is passed to a templated function?).

I'm basically indifferent to this particular change - but it does remove the last of the `long double`s from my debug symbols.

jzmaddock: Ah yes, my bad: I was viewing the diff and the PP logic got a bit confused at least in my head! ;)

PR #404, opened 2020-07-17
Engel expansions

NAThompson:


jzmaddock:
That looks like fun ;)  Just out of curiosity do you have a use case in mind?

It all looks fine BTW.

NAThompson:
The goal is to have a large number of ways to represent each number; so we now have simple continued fractions, centered continued fractions, Luroth expansions, and now Engel expansions. This allows us to decide is a number is "interesting" in some way; for example exp(1) has a very simple and unusual Engel expansion, whereas the Engel expansion of pi looks "random" (albeit with exponential growth of terms).

I'm using these utilities in the PSLQ PR; see [here](https://github.com/boostorg/multiprecision/blob/3711b54b62465c425489e68f18599bcacbfb3dfb/example/representations/dottie.cpp) for an example.

PR #407, opened 2020-07-24
Reciprocal Fibonacci constant and Dottie number.

NAThompson:


NAThompson:
@jzmaddock : Do you understand this performance behavior?

```
Dottie/512                       0.080 us        0.080 us     17392255
Dottie/1024                      0.109 us        0.109 us     12780486
Dottie/2048                      0.117 us        0.117 us     11917226
Dottie/4096                      0.088 us        0.088 us     16042701
Dottie/8192                      0.098 us        0.098 us     14220373
Dottie/16384                     0.130 us        0.130 us     10628203
Dottie/32768                     0.245 us        0.245 us      5869713
Dottie/65536                   1723078 us      1722178 us            1
Dottie/131072                  4771532 us      4769465 us            1
Dottie/262144                 12902627 us     12895118 us            1
```

The jump at 65000 bits is pretty jarring.

jzmaddock:
@NAThompson : Do I understand your benchmark correctly that you're starting at 500 decimal digit accuracy and going up from there?  If so it will be calculating the constant fresh each time so we should be comparing like with like.  65536 decimal places is a heck of a lot of precision though - and a lot of memory allocation per mpfr_float instance.  Has it run the machine out of actual memory and started swapping like crazy?

NAThompson:
> Do I understand your benchmark correctly that you're starting at 500 decimal digit accuracy and going up from there?

That's correct; I guess I could start at 128 and still trigger the calculation each call.

>  65536 decimal places is a heck of a lot of precision though - and a lot of memory allocation per mpfr_float instance. Has it run the machine out of actual memory and started swapping like crazy?

The swapping might be the correct explanation, because I tried this benchmark on a machine with a smaller cache and saw the same behavior, but at 16,000 bits of precision.

Issue #406, opened 2020-07-23
Crash / assertion failure calculating T-distribution quantile


evanmiller:
This is a customer-reported bug and so I don't have a full test case. However, I have the following backtrace:

```
Application Specific Information:
Assertion failed: (*p_derivative >= 0), function ibeta_a_step, file /usr/include/boost/math/special_functions/beta.hpp, line 739.
 

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff6b70d33a __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff6b7c9e60 pthread_kill + 430
2   libsystem_c.dylib             	0x00007fff6b694808 abort + 120
3   libsystem_c.dylib             	0x00007fff6b693ac6 __assert_rtn + 314
4   libcrit.dylib                 	0x000000010496bef4 double boost::math::detail::ibeta_a_step<double, boost::math::policies::policy<boost::math::policies::promote_double<false>, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy> >(double, double, double, double, int, boost::math::policies::policy<boost::math::policies::promote_double<false>, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy, boost::math::policies::default_policy> const&, bool, double*) + 196
[...]

PR #408, opened 2020-07-24
Fix bug in student's T quantile for denormalized degrees_of_freedom.

jzmaddock:
See https://github.com/boostorg/math/issues/406

### DIFF
--- a/include/boost/math/special_functions/detail/ibeta_inverse.hpp
+++ b/include/boost/math/special_functions/detail/ibeta_inverse.hpp
@@ -673,7 +673,28 @@ T ibeta_inv_imp(T a, T b, T p, T q, const Policy& pol, T* py)
          invert = !invert;
          xs = 1 - xs;
       }
-      T xg = pow(a * p * boost::math::beta(a, b, pol), 1/a);
+      if (a < tools::min_value<T>())
+         return invert ? 1 : 0; // nothing interesting going on here.
+      //
+      // The call to beta may overflow, plus the alternative using lgamma may do the same
+      // if T is a type where 1/T is infinite for small values (denorms for example).
+      //
+      T bet = 0;
+      T xg;
+      bool overflow = false;
+      try {
+         bet = boost::math::beta(a, b, pol);
+      }
+      catch (const std::runtime_error&)
+      {
+         overflow = true;
+      }
+      if (overflow || !(boost::math::isfinite)(bet))
+      {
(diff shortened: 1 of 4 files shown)


include/boost/math/special_functions/detail/ibeta_inverse.hpp
```diff
@@ -673,7 +673,28 @@ T ibeta_inv_imp(T a, T b, T p, T q, const Policy& pol, T* py)
...
+      try {
+         bet = boost::math::beta(a, b, pol);
+      }
+      catch (const std::runtime_error&)
+      {
+         overflow = true;
+      }
+      if (overflow || !(boost::math::isfinite)(bet))
+      {
+         xg = exp((boost::math::lgamma(a + 1, pol) + boost::math::lgamma(b + 1) - boost::math::lgamma(a + b + 1) + log(p) - log(b) + log(a + b)) / a);
```
evanmiller, line 694: Shouldn't the 2nd and 3rd calls to `lgamma` here include the policy as the 2nd argument?

jzmaddock:
>Shouldn't the 2nd and 3rd calls to lgamma here include the policy as the 2nd argument?

Doh!  Yes.

include/boost/math/special_functions/detail/ibeta_inverse.hpp
```diff
@@ -673,7 +673,34 @@ T ibeta_inv_imp(T a, T b, T p, T q, const Policy& pol, T* py)
...
+      try {
+         bet = boost::math::beta(a, b, pol);
+      }
+      catch (const std::runtime_error&)
+      {
+         overflow = true;
+      }
+      if (overflow || !(boost::math::isfinite)(bet))
+      {
+         xg = exp((boost::math::lgamma(a + 1, pol) + boost::math::lgamma(b + 1, pol) - boost::math::lgamma(a + b + 1, pol) + log(p) - log(b) + log(a + b)) / a);
```
evanmiller, line 700: Maybe I'm missing something, but why not reduce

`lgamma(b+1) - log(b) - lgamma(a+b+1) + log(a+b)`

To just `lgamma(b) - lgamma(a+b)`

?

jzmaddock: The branch deals with a,b < 1 and the problem occurs when one of a,b is so small that 1/a (for example) overflows, so I had issues with lgamma overflowing as well.... though in point of fact I guess that may be a bug as well :(

jzmaddock: It was a bug in lgamma: there was actually a fix for this already in place, it just didn't quite go far enough.

PR #415, opened 2020-08-07
Cohen acceleration

NAThompson:
Accelerates convergence of an alternating series by a method designed by Cohen, Villegas, and Zagier.


NAThompson:
@jzmaddock : You gotta give this one a try; it's amazing.

jzmaddock:
> You gotta give this one a try; it's amazing.

We used this for something - Owen's T I think - I'll look at your code later.  I do remember I contemplated boilerplating this, but had mixed results: it was brilliant when it worked, but in the difficult cases often failed (and it was the slow/non-converging cases I was interested in at the time).  Am I right in thinking this is method 1 from the paper?  I never quite got my head around the other methods :(

NAThompson:
> Am I right in thinking this is method 1 from the paper? I never quite got my head around the other methods :(

It is method 1. The other methods only are useful when calculation of the terms is extremely expensive. I thought about adding those, but my use case really didn't require that.

BTW, should this go in `series.hpp` or is it fine where it is?

doc/internals/cohen_acceleration.qbk
```diff
@@ -0,0 +1,105 @@
...
+
+    double computed = cohen_acceleration(gen, 5);
+    
+/Nota bene/: When experimenting with this option, we found that adding more terms was no guarantee of additional accuracy, and could not find an example where a user-provided number of terms outperformed the default.
+In addition, it is easy to generate intermediates which overflow if we let /n/ grow too large.
+Hence we recommend only playing with this parameter to /decrease/ the default number of terms to increase speed.
+
+[heading Performance]
+
+To see that Cohen acceleration is in fact faster than naive summation for the same level of relative accuracy, we can run the `reporting/performance/cohen_acceleration_performance.cpp` file:
```
jzmaddock, line 75: Can we mention what the test sum is in the docs here?  Thanks!

jzmaddock:
I think it's fine as it is IMO.  Other than the minor comment on the doc, looks all good to go?

PR #417, opened 2020-08-09
Improve ulps plot:

jzmaddock:
Use numeric_limits for precision comparison, plot out of bounds points at the clip boundary so we can still see that they're there.

### DIFF
--- a/include/boost/math/tools/ulps_plot.hpp
+++ b/include/boost/math/tools/ulps_plot.hpp
@@ -247,11 +247,16 @@ public:
                 }
                 if (plot.clip_ > 0 && static_cast<PreciseReal>(abs(ulp[j])) > plot.clip_)
                 {
-                    continue;
+                   CoarseReal x = x_scale(plot.coarse_abscissas_[j]);
+                   PreciseReal y = y_scale(static_cast<PreciseReal>(ulp[j] < 0 ? -plot.clip_ : plot.clip_));
+                   fs << "<circle cx='" << x << "' cy='" << y << "' r='1' fill='" << color << "'/>\n";
                 }
-                CoarseReal x = x_scale(plot.coarse_abscissas_[j]);
-                PreciseReal y = y_scale(static_cast<PreciseReal>(ulp[j]));
-                fs << "<circle cx='" << x << "' cy='" << y << "' r='1' fill='" << color << "'/>\n";
+                else
+                {
+                   CoarseReal x = x_scale(plot.coarse_abscissas_[j]);
+                   PreciseReal y = y_scale(static_cast<PreciseReal>(ulp[j]));
+                   fs << "<circle cx='" << x << "' cy='" << y << "' r='1' fill='" << color << "'/>\n";
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


jzmaddock:
@NAThompson : A quick question:

* What should we do with errors that are NaN's?  Don't these reflect a possible bug in the function?

Other than that this fixes the case where PreciseReal is a multiprecision type.

include/boost/math/tools/ulps_plot.hpp
```diff
@@ -433,9 +438,8 @@ template<class F, typename PreciseReal, typename CoarseReal>
 ulps_plot<F, PreciseReal, CoarseReal>::ulps_plot(F hi_acc_impl, CoarseReal a, CoarseReal b,
              size_t samples, bool perturb_abscissas, int random_seed)
 {
-    // This fails for heap allocated types like MPFR.
-    // Got to think on what to do about it.
-    static_assert(sizeof(PreciseReal) >= sizeof(CoarseReal), "PreciseReal must have larger size than CoarseReal");
+    // Use digits10 for this comparison in case the two types have differeing radixes:
+    static_assert(std::numeric_limits<PreciseReal>::digits10 >= std::numeric_limits<CoarseReal>::digits10, "PreciseReal must have higher precision that CoarseReal");
```
NAThompson, line 442: Yes, I've noticed this problem before for MPFR types; since the `sizeof` operator returns the size of a pointer. Thanks!

include/boost/math/tools/ulps_plot.hpp
```diff
@@ -247,11 +247,16 @@ class ulps_plot {
                 }
                 if (plot.clip_ > 0 && static_cast<PreciseReal>(abs(ulp[j])) > plot.clip_)
                 {
-                    continue;
+                   CoarseReal x = x_scale(plot.coarse_abscissas_[j]);
+                   PreciseReal y = y_scale(static_cast<PreciseReal>(ulp[j] < 0 ? -plot.clip_ : plot.clip_));
+                   fs << "<circle cx='" << x << "' cy='" << y << "' r='1' fill='" << color << "'/>\n";
```
NAThompson, line 252: Should we maybe color the clipped values differently? Suppose red for all clips?

NAThompson:
> What should we do with errors that are NaN's? Don't these reflect a possible bug in the function?

Yeah, but it could be a natural value to return at both precisions; e.g. sin(1/x) at x = 0. In this case there is no problem with the ulps plot, you just don't plot that value.

So the only problem is when only one of the two precisions is a nan. In this case, maybe put it at the clip? But if there is no clip specified, then I would probably do nothing; or a message to `std::cerr`. Graphical techniques can't be expected to ergonomically represent all classes of errors.

include/boost/math/tools/ulps_plot.hpp
NAThompson: One further point: If the condition number goes infinite somewhere, which is very common, the top of the plot is going to be a bloodbath, even if there's really nothing wrong with the implementation.

I'm not sure what to do about this, since it seems like something no one wants an option for!

jzmaddock: +1 on red for clipped values, and yes the top of the plot may be a bloodbath, but my feeling is it's better to know than not about this stuff.

jzmaddock:
OK yes, I see that we need to support NaN's when the function goes out of domain or whatever.  It's tough to do the right thing in all cases here!

cosurgi:
> function goes out of domain

actually a helper function which returns `true`, `false` about whether given `x` is in a function domain would be awesome.

NAThompson:
@jzmaddock : I like these diffs; should I squash merge them?

jzmaddock:
>I like these diffs; should I squash merge them?

Yes please do, thanks!

Issue #424, opened 2020-08-25
autodiff::fvar doesn't work with Eigen


wgledbetter:
Currently, the templated copy constructor for `autodiff::fvar` interferes with Eigen's lazy evaluation, and, in my experience, produces two types of errors:
1. Ambiguous operators
- Example error: `ambiguous overload for ‘operator*’ (operand types are ‘Eigen::Transpose<Eigen::Matrix<boost::math::differentiation::autodiff_v1::detail::fvar<double, 1>, 4, 1> >’ and ‘Eigen::Matrix<boost::math::differentiation::autodiff_v1::detail::fvar<double, 1>, 4, 1>’)`
2. Failed attempts to `static_cast` a matrix into an fvar
- Example error: `error: invalid ‘static_cast’ from type ‘const Eigen::CwiseUnaryOp<Eigen::internal::scalar_cast_op<double, boost::math::differentiation::autodiff_v1::detail::fvar<double, 1> >, const Eigen::Matrix<double, 4, 4> >’ to type ‘double’`

A solution would be to only define the constructor for types that are convertible to the fvar's RealType. A colleague and I worked out a potential fix using SFINAE that I'm currently testing against the existing `test_autodiff` files. Tests 1-3 have succeeded with the existing travis configuration, but I'm having issues in test_autodiff_4 with `boost::has_right_shift` as called from `boost::lexical_cast`.

Is there any interest in merging a solution to this problem? I can post the work I've done so far.

NAThompson:
@wgledbetter : Yes, we're interested.

@pulver : Can your review the PR when it shows up?

pulver:
The build errors @wgledbetter forewarned of are on https://travis-ci.org/github/boostorg/math/builds/721216811

```
../../../boost/static_assert.hpp:31:45: error: static assertion failed: Target type is neither std::istream`able nor std::wistream`able
 #     define BOOST_STATIC_ASSERT_MSG( ... ) static_assert(__VA_ARGS__)
                                             ^
../../../boost/lexical_cast/detail/converter_lexical.hpp:243:13: note: in expansion of macro ‘BOOST_STATIC_ASSERT_MSG’
             BOOST_STATIC_ASSERT_MSG((result_t::value || boost::has_right_shift<std::basic_istream<wchar_t>, T >::value),
             ^~~~~~~~~~~~~~~~~~~~~~~
```
Guessing: Does this require a fix to/generalization of `boost::has_right_shift`?

[...]

PR #425, opened 2020-08-26
Only allow fvar constructor for convertible types

wgledbetter:
#424 

Use SFINAE to enable copy construction and assignment only from valid types. Allows use in Eigen library.

`test_autodiff` 1, 2, and 3 succeed, but number 4 fails due to `boost::has_right_shift` required by `boost::lexical_cast`.


wgledbetter:
Added overload of `operator>>` that behaves like `operator=` to solve issue with `has_right_shift`.

NAThompson:
The build failures do look like a pre-existing problem, but unfortunately I don't have time to investigate; have they failed in any other builds?

I have read the diffs and think they look good.

I'm going to let @jzmaddock squash merge this since he understands `lexical_cast` much better than I do.

NAThompson:
@pulver : How do I get the cool "Approved" button you put on the PR?

pulver:
> have they failed in any other builds?

Yes: https://ci.appveyor.com/project/jzmaddock/math/builds/34818710
This is from a few days ago prior to any commits related to this PR.

> How do I get the cool "Approved" button you put on the PR?

`Files changed` tab > green `Review changes` button > `Approve` radio button.

jzmaddock:
Looks good to me too - do we have a test case for the original Eigen issue?  We do have some Eigen interoperability tests in Multiprecision which could be copied here (which is to say I'll sort out the Jamfile stuff if you don't want to go there).

I'll try to sort out the Appveyor issue too - we really need to get that fixed - not really sure how it crept in to be honest.

jzmaddock:
I'm hoping this will fix the outstanding develop test failure: https://github.com/boostorg/math/commit/526fc6eef47cbdedc09d0895c9ce97ed383dc45a

wgledbetter:
@NAThompson and @pulver Thanks for the quick review!

@jzmaddock I'll work on some Eigen tests over the weekend if you can help me put them into the existing test framework.

NAThompson:
So there is a broader question now: This now explicitly uses Eigen in the boost source directory, but traditionally we only test that our code is "generic enough" to interoperate with Eigen.

Do we have a compelling reason to use Eigen explicitly, rather than (say) writing generic code that will work with Eigen and Boost.ublas?

wgledbetter:
@NAThompson boost multiprecision has an eigen.hpp compatibility file that directly includes Eigen, and I was referencing that code for the recent updates. It provides an overload of Eigen::NumTraits that is necessary for full compatibility. The other overload it provides is Eigen::ScalarBinaryOpTraits. This one is proving troublesome because the libeigen3-dev package on travis:xenial is out of date. That reason alone is probably enough to remove the Eigen include, since end-users may still be using the older version.

The commit I'm about to push relocates the Eigen code from the source directory to the test directory.

NAThompson:
Merged in #1485 with @wgledbetter as coauthor.

PR #426, opened 2020-08-27
Add missing include

camolezi:
This patch makes the header able to be built standalone, making possible C++ clang modules builds.


pulver:
Hi @camolezi - thank you for the PR.

Is it possible you are/were using an older version of boost? This PR #313 from earlier this year replaced all instances of `decay_t` (with no namespace) with `std::decay`. Or are you still encountering the same issue with current `master`?

camolezi:
Sorry, we are indeed not using the current version of boost. Seems like on the current version we don't need this include anymore, thanks for review.

PR #427, opened 2020-08-27
Add missing include

camolezi:
This patch makes the header able to be built standalone, making possible C++ clang modules builds.@vgvassilev


NAThompson:
I admit to not understanding what is going on.

In `boost/math/tools/atomic.hpp`, we `#include <atomic>` if it exists.

Apparently this doesn't play nice with clang module builds?

jzmaddock:
Exactly, all the correct headers are included at the proper times: note that we support situations where there is no threading support and no <atomic> (even though we may be in C++11 or later).  I also wonder how stable modules are yet?

camolezi:
Hello, thanks for the reviews and insights.
Sorry, this was a mistake on my part, we are actually excluding this file on our module map. So I'm closing this PR.

PR #434, opened 2020-09-10
Implement Policies in Statistics

mborland:
This PR is working through implementing execution policies for uni-variate statistics. As with the STL there will be additional overloads for each function, and the default policy will be sequential.


NAThompson:
This is definitely useful; did you see #248? I never finished for reasons I can't recall.

Could you make sure that your parallel reductions agree with the algorithms presented here?

Bennett, Janine, et al. "Numerically stable, single-pass, parallel statistics algorithms." Cluster Computing and Workshops, 2009. CLUSTER'09. IEEE International Conference on. IEEE, 2009.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -12,75 +12,101 @@
...
+            double i = 1;
+            for(auto it = first; it != last; ++it) {
+                mu = mu + (*it - mu)/i;
+                i += 1;
+            }
+            return mu;
+        }
+        else
+        {
+            return std::reduce(exec, first, last, 0.0) / std::distance(first, last);
```
NAThompson, line 42: Take a read the Higham's "Accuracy and Stability of Numerical Algorithms", equation 1.6a,b to see why I did it this way and not with `std::reduce`.

Another good reference is:

Robert F Ling. Comparison of several algorithms for computing sample means and variances. Journal of the American Statistical Association, 69(348): 859–866, 1974

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -12,75 +12,101 @@
...
     else if constexpr (std::is_same_v<typename std::iterator_traits<ForwardIterator>::iterator_category, std::random_access_iterator_tag>)
     {
-        size_t elements = std::distance(first, last);
-        Real mu0 = 0;
-        Real mu1 = 0;
-        Real mu2 = 0;
-        Real mu3 = 0;
+        std::size_t elements {std::distance(first, last)};
+        std::valarray<Real> mu {0, 0, 0, 0};
+        std::valarray<Real> temp {0, 0, 0, 0};
```
NAThompson, line 49: Isn't `valarray` on the verge of deprecation?

mborland:
> This is definitely useful; did you see #248? I never finished for reasons I can't recall.
> 
> Could you make sure that your parallel reductions agree with the algorithms presented here?
> 
> Bennett, Janine, et al. "Numerically stable, single-pass, parallel statistics algorithms." Cluster Computing and Workshops, 2009. CLUSTER'09. IEEE International Conference on. IEEE, 2009.

I did; I was scrolling through stale PRs, and Issues to see what would be worthwhile to work next. I can give it a read this weekend.

include/boost/math/statistics/univariate_statistics.hpp
mborland: What about for integer arithmetic? From [26.7.1](http://www.open-std.org/Jtc1/sc22/wg21/docs/papers/2016/p0024r1) ```reduce applies binary_op in an unspecified order, which yields a non-deterministic result for non-associative or non-commutative binary_op such as floating-point addition.``` I will have to look at those references for the other cases.

include/boost/math/statistics/univariate_statistics.hpp
mborland: Not that I am aware of. The implementation in the Intel MKL has some pretty serious optimization. It certainly has its idiosyncrasies but it is good at what it does.

NAThompson:
@evanmiller : I believe I now recall what happened: I did some performance benchmarks and found it very hard to get the parallel performance to exceed the single-threaded-the arithmetic intensity is pretty low for many of these. Also, compiler support was tough on Ubuntu; had to install tbb via github!

Also, IIRC the projections had very poor compiler support.

But maybe things have stabilized a bit . . .

include/boost/math/statistics/univariate_statistics.hpp
NAThompson: Yeah, for integer arithmetic this won't work; then again the result needs to be cast to a floating point type anyway.

If indeed we wished to support integer arithmetic then we'd need to figure out whether we want to cast every element to double and use the floating point algorithm or use the naive algorithm and then cast the sum to double. Overflow is the danger of the latter.

include/boost/math/statistics/univariate_statistics.hpp
NAThompson: Cool; but let's check the generated asm on godbolt and make sure it looks good.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -109,25 +110,43 @@ inline auto mean(Container const & v)
...
+                M = M + tmp/k;
+                k += 1;
+            }
+            return Q/(k-1);
+        }
+        else
+        {
+            std::atomic<double> M {*first};
+            std::atomic<double> Q {0};
+            std::atomic<double> k {2};
```
NAThompson, line 139: I dunno man! Using atomics here is gonna be expensive!

I'd get some performance benchmarks written before you get too far into this.

Also, the Monte-Carlo integration class has an algorithm for pooling variance that you might enjoy looking at.

mborland: I just wanted to see how this would work with automatic partitioning from `std::for_each`. The answer is not super well.

NAThompson: Got it; probably should give you some time before sniping at everything you try!

In any case, looks like I never committed a google benchmark for these functions; do you mind doing that? Should be easy now that you've got the first one done.

mborland: Can do. I need to validate that the usefulness anyway.

mborland:
@NAThompson have you seen the [latest draft of the stats proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1708r2.pdf)l? I think the stats class would be a useful addition. I am not sure we could implement it as constexpr without pushing the standard forward to c++20 to have constexpr iterators, but it still seems like it would be nice to input a data set then have everything calculated and stored.

NAThompson:
@mborland : I like that they use the bitset to specify which statistics they want to use; it actually solves a design problem I had in this.

Once you get the execution policies working it might be good to write the authors of that proposal with the results.

mborland:
Some initial benchmarks:
```
-------------------------------------------------------------------------------------------------------
Benchmark                                                             Time             CPU   Iterations
-------------------------------------------------------------------------------------------------------
mean<int>/real_time_BigO                                           5.02 N          5.01 N    
mean<int>/real_time_RMS                                               0 %             0 % 
parallel_mean<int>/real_time_BigO                                  0.12 N          0.12 N    
parallel_mean<int>/real_time_RMS                                      1 %             1 %  
mean<double>/real_time_BigO                                        1.10 N          1.10 N    
mean<double>/real_time_RMS                                            0 %             0 %  
parallel_mean<double>/real_time_BigO                               1.04 N          1.04 N    
parallel_mean<double>/real_time_RMS                                   1 %             1 % 

variance<int>/real_time_BigO                                       5.16 N          5.15 N    
variance<int>/real_time_RMS                                           0 %             0 %  
parallel_variance<int>/real_time_BigO                              1.58 N          0.04 N    
parallel_variance<int>/real_time_RMS                                 26 %            86 %
variance<double>/real_time_BigO                                    5.16 N          5.16 N    
variance<double>/real_time_RMS                                        0 %             0 %  
parallel_variance<double>/real_time_BigO                           1.62 N          0.04 N    
parallel_variance<double>/real_time_RMS                              25 %            85 %

median<int>/real_time_BigO                                         1.45 N          1.45 N    
median<int>/real_time_RMS                                            14 %            15 % 
parallel_median<int>/real_time_BigO                                1.01 N          1.01 N    
parallel_median<int>/real_time_RMS                                    1 %             1 %
median<double>/real_time_BigO                                      1.48 N          1.48 N    
median<double>/real_time_RMS                                          7 %             7 %   
parallel_median<double>/real_time_BigO                             1.22 N          1.22 N    
parallel_median<double>/real_time_RMS                                 3 %             3 %

median_absolute_deviation<int>/real_time_BigO                      3.07 N          3.07 N    
median_absolute_deviation<int>/real_time_RMS                          9 %             9 % 
parallel_median_absolute_deviation<int>/real_time_BigO             2.32 N          2.32 N    
parallel_median_absolute_deviation<int>/real_time_RMS                 1 %             1 % 
median_absolute_deviation<double>/real_time_BigO                   2.61 N          2.60 N    
median_absolute_deviation<double>/real_time_RMS                       8 %             8 %  
parallel_median_absolute_deviation<double>/real_time_BigO          1.40 N          1.40 N    
parallel_median_absolute_deviation<double>/real_time_RMS              2 %             2 % 

gini_coefficient<int>/real_time_BigO                              10.12 N         10.11 N    
gini_coefficient<int>/real_time_RMS                                  14 %            14 % 
parallel_gini_coefficient<int>/real_time_BigO                     27.73 N         27.71 N    
parallel_gini_coefficient<int>/real_time_RMS                          7 %             7 % 
gini_coefficient<double>/real_time_BigO                           11.38 N         11.37 N    
gini_coefficient<double>/real_time_RMS                               18 %            18 %  
parallel_gini_coefficient<double>/real_time_BigO                  30.03 N         29.99 N    
parallel_gini_coefficient<double>/real_time_RMS                      12 %            12 % 

interquartile_range<double>/real_time_BigO                         2.41 N          2.41 N    
interquartile_range<double>/real_time_RMS                             8 %             8 % 
parallel_interquartile_range<double>/real_time_BigO                2.40 N          2.40 N    
parallel_interquartile_range<double>/real_time_RMS                    4 %             4 % 
```

mborland:
@NAThompson When you have time I believe this is in good shape to be reviewed. Almost all of the implementations use your original algorithms and recursively decompose the data set. The only error that I thought was unacceptably high was the calculation for M2 in `first_four_moments` for `long double` and `cpp_bin_float_50` which has some cascading effects. I am hoping a second set of eyes will catch a small mistake.

NAThompson:
Wow this looks really good!

I'm confused though; I feel like the CPU time should = threads*time, but it looks like it's the exact opposite?

jzmaddock:
Google bench is IMO confusing for multithreaded tests, I just quickly tried this:

```
void thread_proc(unsigned n)
{
   auto f = [](unsigned n)
   {
      std::chrono::duration<unsigned, std::ratio<1, 10> > d(n);
      std::this_thread::sleep_for(d);
   };
   std::thread t(f, n);
   t.join();
}

void thread_test(benchmark::State& state)
{
   unsigned n = state.range(0);
   for (auto _ : state)
   {
      thread_proc(n);
   }
   state.SetComplexityN(state.range(0));
}


BENCHMARK(thread_test)->RangeMultiplier(2)->Range(1 << 1, 1 << 20);
```

and got:

```
thread_test/2                                   202543985 ns        0.000 ns          100
thread_test/4                                   402490410 ns        0.000 ns           10
thread_test/8                                   802324240 ns        0.000 ns           10
thread_test/16                                 1602629310 ns        0.000 ns           10
thread_test/32                                 3201796500 ns        0.000 ns            1
```

So time looks to be reliable, but cpu not so much?

mborland:
I believe this is the issue:
`
By default, the CPU timer only measures the time spent by the main thread. If the benchmark itself uses threads internally, this measurement may not be what you are looking for. Instead, there is a way to measure the total CPU usage of the process, by all the threads.
`

If you run the thread_test with: 
`BENCHMARK(thread_test)->RangeMultiplier(2)->Range(1 << 1, 1 << 20)->MeasureProcessCPUTime()->UseRealTime();`

the results are a more reasonable:
```
2020-11-19T21:45:54+03:00
Running ./stats_benchmark
Run on (12 X 4500 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 1.22, 1.33, 0.78
-------------------------------------------------------------------------------------
Benchmark                                           Time             CPU   Iterations
-------------------------------------------------------------------------------------
thread_test/2/process_time/real_time        200176299 ns        81472 ns            3
thread_test/4/process_time/real_time        400235147 ns        83652 ns            2
thread_test/8/process_time/real_time        800256134 ns       103176 ns            1
thread_test/16/process_time/real_time      1600253804 ns        82747 ns            1
thread_test/32/process_time/real_time      3200307348 ns       152016 ns            1
thread_test/64/process_time/real_time      6400255888 ns       103569 ns            1
```

Here is the full snippet from the docs:
```
// Measure the time spent by the main thread, use it to decide for how long to
// run the benchmark loop. Depending on the internal implementation detail may
// measure to anywhere from near-zero (the overhead spent before/after work
// handoff to worker thread[s]) to the whole single-thread time.
BENCHMARK(BM_OpenMP)->Range(8, 8<<10);

// Measure the user-visible time, the wall clock (literally, the time that
// has passed on the clock on the wall), use it to decide for how long to
// run the benchmark loop. This will always be meaningful, an will match the
// time spent by the main thread in single-threaded case, in general decreasing
// with the number of internal threads doing the work.
BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->UseRealTime();

// Measure the total CPU consumption, use it to decide for how long to
// run the benchmark loop. This will always measure to no less than the
// time spent by the main thread in single-threaded case.
BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->MeasureProcessCPUTime();

// A mixture of the last two. Measure the total CPU consumption, but use the
// wall clock to decide for how long to run the benchmark loop.
BENCHMARK(BM_OpenMP)->Range(8, 8<<10)->MeasureProcessCPUTime()->UseRealTime();
```

mborland:
Here is the current performance of the last remaining function (mode):

```
----------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations
----------------------------------------------------------------------------------
mode<int>/real_time_BigO                      5.52 N          5.51 N
parallel_mode<int>/real_time_BigO           337.80 N        337.31 N 
mode<double>/real_time_BigO                   9.08 N          9.07 N
parallel_mode<double>/real_time_BigO        319.07 N        318.71 N
```
Obviously the current implementation of pre-processing all the data using an `unordered_map` did not work out like I had hoped. I am not sure how to decompose the problem like is done for most of the other functions. Any input/papers would be appreciated. I have not seen much on the topic. Maybe making this parallel is not particularly useful or in demand?

mborland:
AppVeyor results look like they should. Travis completed the most recent build 8 days ago, but has not started processing the next item in the queue. Is this a known issue? I think I saw something on the mailing list recently about Travis hanging in other libraries.

NAThompson:
> Is this a known issue?

Not a known issue, but maybe related to the AWS downtime?

mborland:
[Status shows they did not have AWS issues](https://www.traviscistatus.com/#day). Does boost have a paid organizational account? If not the queue for open source linux build time being 10x what they process at a time would explain the delay.

NAThompson:
I'm pretty sure it's a free tier account.

mborland:
Travis finally got around to doing its thing, and the results are clean. Pending any further review this is good to go from my end.

NAThompson:
@mborland : Gimme a few days to get around to this; please reping me if I don't have it done by Friday!

mborland:
@NAThompson I forgot that I left a TODO in this, so if you have started reviewing I made a trivial change to one of the assertions.

include/boost/math/statistics/univariate_statistics.hpp
```diff
@@ -231,10 +230,9 @@ inline auto first_four_moments(ExecutionPolicy&& exec, ForwardIterator first, Fo
         }
         else
         {
-            static_assert(!std::is_same_v<Real, long double>, "Error for parallel calculation using long double exceeds 100 Epsilon");
+            static_assert(!std::is_same_v<Real, long double> ^ !std::is_trivially_constructible_v<Real>, 
+                          "Error for parallel calculation using long double or aribitrary precision types is excessivly large (>100 epsilon).");
```
NAThompson, line 234: Do we have a good idea of why this is happening just for multiprecision types?

For instance, is it happening for mpfr types as well?

NAThompson:
There's a syntax error somewhere in the `statistics/univariate_statistics.qbk` file:

```
math/doc$ ../../../b2
statistics/univariate_statistics.qbk:397: error: Element not allowed in this context.
Error: Error count: 1.
```

I tried to chase it down, but failed. I don't believe it's in line 397, but rather a previous error . . .

NAThompson:
Do we understand why this is happening?

> The use of parallel execution policies (std::execution::par or par_unseq) generally doubles the error introduced in the computation, but is still minimal.

doc/statistics/univariate_statistics.qbk
```diff
@@ -121,12 +212,17 @@ In general, you can store your data in an Eigen array, and Armadillo vector, `st
...
 
     std::vector<double> v{1,2,3,4,5};
     double mu = boost::math::statistics::mean(v.cbegin(), v.cend());
     // Alternative syntax if you want to use entire container:
     mu = boost::math::statistics::mean(v);
+    // Alternative syntax if you want to use parallel execution:
+    mu = boost::math::statistics::mean(std::execution::par, v);
 
 The implementation follows [@https://doi.org/10.1137/1.9780898718027 Higham 1.6a].
 The data is not modified and must be forward iterable.
```
NAThompson, line 228: I think it now has to be random access, not just forward iterable.

doc/statistics/univariate_statistics.qbk
```diff
@@ -121,12 +212,17 @@ In general, you can store your data in an Eigen array, and Armadillo vector, `st
 These routines are usable in float, double, long double, and Boost.Multiprecision precision, as well as their complex extensions whenever the computation is well-defined.
 For certain operations (total variation, for example) integer inputs are supported.
 
+/Nota bene/: The default execution policy for every function is std::execution::seq.
+The use of parallel execution policies (std::execution::par or par_unseq) generally doubles the error introduced in the computation, but is still minimal.
```
NAThompson, line 216: Might be good to say "Parallel routines must use random access containers, not just forward iterable."

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -0,0 +1,384 @@
...
+    {
+        std::size_t elements {std::distance(first, last)};
+        std::valarray<ReturnType> mu {0, 0, 0, 0};
+        std::valarray<ReturnType> temp {0, 0, 0, 0};
+        ReturnType i = 1;
+        const auto end {last - (elements % 4)};
+
+        for(auto it {first}; it != end; it += 4)
+        {
+            const ReturnType inv {ReturnType(1) / i};
```
NAThompson, line 38: In the random access case, you make the reciprocal approximation, but not in the general case below. That could be one reason the error rates are higher.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -0,0 +1,384 @@
...
+        const ReturnType num1 = ReturnType(elements - (elements %4))/ReturnType(4);
+        const ReturnType num2 = num1 + ReturnType(elements % 4);
+
+        for (auto it = end; it != last; ++it)
+        {
+            mu[3] += (*it-mu[3])/i;
+            i += 1;
+        }
+
+        return (num1 * std::valarray<ReturnType>(mu[std::slice(0,3,1)]).sum() + num2 * mu[3]) / ReturnType(elements);
```
NAThompson, line 54: This should reference the appropriate equation in the reference for parallel reduction of means.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -0,0 +1,384 @@
...
+        k += 1;
+        Real delta_2 = *it - M;
+        M2 += delta_1 * delta_2;
+    }
+
+    return std::make_tuple(M, M2, Q/(k-1));
+}
+
+// Global thread counter required for recursive calls. Reset before each parallel method is called
+static std::atomic<unsigned> thread_counter {1};
```
NAThompson, line 98: I admit to not understanding why this needs a global static.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -0,0 +1,384 @@
...
+    {
+        if(thread_counter + 2 <= num_threads && range_a > 10)
+        {
+            return parallel_variance_impl<ReturnType>(std::next(first, range_a), last);
+        }
+        else
+        {
+            return variance_sequential_impl<ReturnType>(std::next(first, range_a), last);
+        }
+    })};
```
NAThompson, line 134: Are you sure this needs to be done via recursion? It looks like the only thing that buys you is the ability to use the formula to pool two variances to pool all the variances. But the Pebay paper shows how to pool n variances without issue.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -0,0 +1,384 @@
...
+ReturnType parallel_first_four_moments_impl(ForwardIterator first, ForwardIterator last)
+{
+    using Real = typename std::iterator_traits<ForwardIterator>::value_type;
+    static unsigned num_threads {std::thread::hardware_concurrency()};
+
+    const auto elements {std::distance(first, last)};
+    const auto range_a {std::floor(elements / 2)};
+    const auto range_b {elements - range_a};
+
+    thread_counter.fetch_add(2);
```
NAThompson, line 188: Looks like you're using the same global static thread counter for this as you do for the variance, which could lead to race conditions.

NAThompson:
I get unused typedef warning and a compile fail on the performance tests:

```
$ make reporting/performance/univariate_statistics_performance.x
g++-10 -g -Wall -Wextra --std=gnu++17 -fno-finite-math-only -march=native -Wfatal-errors -MMD -O2 -I../../ -I/usr/local/include -o reporting/performance/univariate_statistics_performance.x reporting/performance/univariate_statistics_performance.cpp -L/usr/local/lib -pthread -lbenchmark -lbenchmark_main -lmpfr -lquadmath
In file included from ../../boost/math/statistics/univariate_statistics.hpp:10,
                 from reporting/performance/univariate_statistics_performance.cpp:8:
../../boost/math/statistics/detail/single_pass.hpp: In function 'ReturnType boost::math::statistics::detail::parallel_variance_impl(ForwardIterator, ForwardIterator)':
../../boost/math/statistics/detail/single_pass.hpp:105:11: warning: typedef 'using Real = typename std::iterator_traits<_II>::value_type' locally defined but not used [-Wunused-local-typedefs]
  105 |     using Real = typename std::iterator_traits<ForwardIterator>::value_type;
      |           ^~~~
../../boost/math/statistics/detail/single_pass.hpp: In function 'ReturnType boost::math::statistics::detail::parallel_first_four_moments_impl(ForwardIterator, ForwardIterator)':
../../boost/math/statistics/detail/single_pass.hpp:181:11: warning: typedef 'using Real = typename std::iterator_traits<_II>::value_type' locally defined but not used [-Wunused-local-typedefs]
  181 |     using Real = typename std::iterator_traits<ForwardIterator>::value_type;
      |           ^~~~
reporting/performance/univariate_statistics_performance.cpp: In function 'std::vector<_RealType> generate_random_vector(std::size_t, std::size_t)':
reporting/performance/univariate_statistics_performance.cpp:49:31: error: 'is_complex' is not a member of 'boost'
   49 |     else if constexpr (boost::is_complex<T>::value)
      |                               ^~~~~~~~~~
compilation terminated due to -Wfatal-errors.
make: *** [reporting/performance/univariate_statistics_performance.x] Error 1
```

NAThompson:
Could you also compile both the unit tests and the performance tests with `-fsanitize=thread` and make sure they are clean?

mborland:
> Could you also compile both the unit tests and the performance tests with `-fsanitize=thread` and make sure they are clean?

Unit tests and performance tests are both clean with asan and tsan.

include/boost/math/statistics/detail/single_pass.hpp
mborland: The thread counter and `fetch_add` are both atomic.

doc/statistics/univariate_statistics.qbk
mborland: Forward iterators are all that is required; parallel tests with `std::forward_list` still pass. The only changes I made to your battery of unit tests was to adjust the tolerances.

include/boost/math/statistics/detail/single_pass.hpp
mborland: It guarantees thread-safe initialization vs using a local static variable. It is also atomic so that each thread can safely increment it in order to avoid over-subscription since I am using `std::launch::async` policy.

include/boost/math/statistics/detail/single_pass.hpp
mborland: It is done via recursion because I found success using Chan's pairwise algorithms. I did not try Pebay's method.

include/boost/math/statistics/univariate_statistics.hpp
mborland: I tried it with `cpp_bin_float_50` and `float128`, and for both types the residual error for M2 is 5.55112e-17. As for the why? I am not quite sure. That error cascaded to others like kurtosis that use `first_four_moments` where the error remained in the realm of e-17.

jzmaddock: When you see that, it's strongly suggestive of a stray double precision value in the calculation somewhere.  What's the precision for 80-bit long double?  One way to try and track this down is to set a breakpoint in cpp_bin_float's convert-from-double code (see assign_float in cpp_bin_float.hpp:295) and see if it get's triggered and with what values.

mborland: @jzmaddock You are correct. `auto` was deducing the wrong type in the calculation of M2. Everything looks fine with explicit types.

pabristow:
I think that this is in line 258  range [0,1] and  [0, 1 - 1/ /n/].

You need a \ before square brackets to get a real bracket

https://www.boost.org/doc/libs/master/doc/html/quickbook/syntax/phrase.html

Single char escape
The backslash may be used to escape a single punctuation character. The punctuation immediately after the backslash is passed without any processing. This is useful when we need to escape QuickBook punctuations such as [ and ]. For example, how do you escape the triple quote? Simple: \'\'\'

Been there, done that - often :-(

Aside  \math\doc\html4_symbols.qbk has quickbook templates for lots of non-ascii chars, and may aid readability.

[template eacute[]'''&#xE9;'''] [/ e acute ]

or by using

Unicode escape
You can enter any 16-bit unicode character by using \u followed by its 4 digit hexadecimal code, or a 32-bit character by using \U followed by an 8 digit hexadecimal code. eg.

\u03B1 + \u03B2

include/boost/math/statistics/detail/single_pass.hpp
NAThompson: Just because the counter is atomic doesn't mean race conditions are impossible. The counter can become incorrect. For example, try launching two threads running this algorithm and the variance at the same time. Each will increment the atomic counter, but then read a value that is incorrect.

doc/statistics/univariate_statistics.qbk
NAThompson: @mborland : Doesn't the complexity of the algorithm increase for `std::forward_list`?

mborland: Yes, it will.

include/boost/math/statistics/detail/single_pass.hpp
mborland: In the event that you run this and variance simultaneously the number of threads will be synced globally to avoid over-subscription. The latest commit offers better synchronization.

include/boost/math/statistics/detail/single_pass.hpp
mborland: Commit cf08af7 changes the implementation slightly so that the random access case now handles all cases. The general case has been removed.

mborland:
Once this is in I have another PR in a similar vein almost ready to go. I realized that since each of the sequential implementations now relies on a specified return type I could reduce the required language standard. With some ugly template syntax all of the sequential methods can be made available with C++11.

mborland:
@NAThompson If you have already taken off for the holidays I can merge my c++11 bindings into this PR. It is nothing ground breaking, but will increase the diff stat by ~1500 LOC. If nothing else it'll give Travis some time since it seems to be quite slow right now.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,255 @@
+# Copyright 2020 Evan Miller
+# Copyright 2020 Matt Borland
```
NAThompson, line 2: Shouldn't this file be part of a separate PR?

.gitignore
```diff
@@ -7,3 +7,6 @@ inspect.html
 test/cuda
```
NAThompson, line 7: I believe this diff is already merged. I would squash merge this PR onto the HEAD another so as to not generate conflicts.

doc/math.qbk
```diff
@@ -1,13 +1,13 @@
...
     [/purpose ISBN 0-9504833-2-X  978-0-9504833-2-0, Classification 519.2-dc22]
     [license
         Distributed under the Boost Software License, Version 1.0.
         (See accompanying file LICENSE_1_0.txt or copy at
         [@http://www.boost.org/LICENSE_1_0.txt])
     ]
-    [authors [Agrawal, Nikhar], [Bikineev, Anton], [Bristow, Paul A.], [Holin, Hubert], [Guazzone, Marco], [Kormanyos, Christopher], [Lalande, Bruno], [Maddock, John], [Miller, Evan], [Murphy, Jeremy W.], [Pulver, Matthew], [Råde, Johan], [Sobotta, Benjamin], [Sewani, Gautam], [Thompson, Nicholas], [van den Berg, Thijs], [Walker, Daryle], [Zhang, Xiaogang]]
+    [authors [Agrawal, Nikhar], [Bikineev, Anton], [Borland, Matthew], [Bristow, Paul A.], [Holin, Hubert], [Guazzone, Marco], [Kormanyos, Christopher], [Lalande, Bruno], [Maddock, John], [Miller, Evan], [Murphy, Jeremy W.], [Pulver, Matthew], [Råde, Johan], [Sobotta, Benjamin], [Sewani, Gautam], [Thompson, Nicholas], [van den Berg, Thijs], [Walker, Daryle], [Zhang, Xiaogang]]
```
NAThompson, line 10: I think you should add this in a separate PR; you've contributed enough even without this PR.

.github/workflows/ci.yml
mborland: Yes; this is identical to PR #476. I had it in here to speed up CI cycles but will pull it out.

.gitignore
mborland: It has been. Git shows no conflicts with base branch but I'll merge develop to be safe.

mborland:
@NAThompson Your requested clean-up is complete. The only remaining CI failure was `C1128: number of sections exceeded object file format limit: compile with /bigobj`. I commented out all of the `std::execution::par_unseq` tests to drop the size by 1/3. The only implementations that vary based on `par` vs `par_unseq` are STL algorithms.

doc/statistics/univariate_statistics.qbk
```diff
@@ -203,7 +203,7 @@ compatible with C++11. In order to pass an ExecutionPolicy to the methods and ac
 Many of these functionals have trivial naive implementations, but experienced programmers will recognize that even trivial algorithms are easy to screw up, and that numerical instabilities often lurk in corner cases.
 We have attempted to do our "due diligence" to root out these problems-scouring the literature for numerically stable algorithms for even the simplest of functionals.
 
-/Nota bene/: Some similar functionality is provided in [@https://www.boost.org/doc/libs/1_68_0/doc/html/accumulators/user_s_guide.html Boost Accumulators Framework].
+/Nota bene/: Some similar functionality is provided in [@https://www.boost.org/doc/libs/1_75_0/doc/html/accumulators/user_s_guide.html Boost Accumulators Framework].
```
NAThompson, line 206: @jzmaddock : Is there a way to always link to the latest boost version?

test/univariate_statistics_test.cpp
```diff
@@ -924,200 +924,202 @@ void test_mode(ExecutionPolicy&& exec)
...
+    //test_integer_mode<int64_t>(std::execution::par_unseq);
     test_integer_mode<uint32_t>(std::execution::seq);
     test_integer_mode<uint32_t>(std::execution::par);
-    test_integer_mode<uint32_t>(std::execution::par_unseq);
+    //test_integer_mode<uint32_t>(std::execution::par_unseq);
 
     test_mode<float>(std::execution::seq);
     test_mode<float>(std::execution::par);
-    test_mode<float>(std::execution::par_unseq);
+    //test_mode<float>(std::execution::par_unseq);
```
NAThompson, line 1116: Why are the `par_unseq` commented out?

mborland: The only remaining CI failure was `C1128: number of sections exceeded object file format limit: compile with /bigobj`. I commented out all of the std::execution::par_unseq tests to drop the size by 1/3. The only implementations that vary based on par vs par_unseq are STL algorithms.

NAThompson: Ok, I'm cool with that. Want to just delete this rather than comment it out?

mborland: I have deleted them.

doc/statistics/univariate_statistics.qbk
jzmaddock: @NAThompson : use https://www.boost.org/doc/libs/release/doc/html/accumulators.html

NAThompson: @mborland : Could you make that change?

mborland: @NAThompson Done.

mborland:
@NAThompson If you have no other changes this should be good to go; CI is finally clean.

mborland:
@NAThompson Ping. Once this is merged in I have the stats class outlined in the [WG21 Statistics Library Proposal ](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1708r2.pdf) ready for review.

include/boost/math/statistics/detail/single_pass.hpp
NAThompson: A global counter is going to cause problems. This counter can be shared by two entirely independent calculations running at the same time. Using Janine Bennet's algorithm will get you past this.

include/boost/math/statistics/detail/single_pass.hpp
NAThompson: I have no problem with Chan's pairwise algorithm, but if it requires a global it just isn't going to work.

include/boost/math/statistics/detail/single_pass.hpp
mborland: The intent here was that it is shared by two independent calculations to avoid over-subscription. Looking back the case where a user specifies a parallel execution policy from two different threads is unlikely. I can make it locally static without any real issue. The other option would be to delete the thread counter entirely, and change the async launch policy to let the runtime deal with thread synchronization. Do you have a particular opinion either way?

NAThompson: >  Looking back the case where a user specifies a parallel execution policy from two different threads is unlikely.

I don't understand why this is particularly unlikely; nor is probability a useful framing as predicting how things will be used is very difficult. But nonetheless, imagine a Rest API where the server is multithreaded and each thread uses this class to launch threads.

> The other option would be to delete the thread counter entirely, and change the async launch policy to let the runtime deal with thread synchronization. Do you have a particular opinion either way?

I admit I don't see how this is gonna work, though if you flesh it out in code I might be able to understand. Personally I think the pairwise recursive decomposition is going to cause problems-some of which you've already encountered-but I feel it's very fundamental to the nature of the algorithm, and [Bennet's](https://link.springer.com/article/10.1007/s00180-015-0637-z) algorithm will be simpler.

mborland: I should have something up in the next day. Chan only described the calculation of M1 and M2. The current implementation of `first_four_moments` follows EQN 3.1 out of the linked paper (I had previously seen it from [Terriberry](https://web.archive.org/web/20140423031833/http://people.xiph.org/~tterribe/notes/homs.html)). Converting from recursive decomposition to linear should not be overly strenuous.

mborland:
@NAThompson The requested `first_four_moments_parallel_impl` using linear decomposition instead of recursion is in.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -0,0 +1,370 @@
...
+*/
+
+template<typename ReturnType, typename ForwardIterator>
+ReturnType first_four_moments_parallel_impl(ForwardIterator first, ForwardIterator last)
+{
+    using Real = typename std::tuple_element<0, ReturnType>::type;
+
+    const auto elements = std::distance(first, last);
+    static unsigned num_threads;
+    if(elements/10 >= std::thread::hardware_concurrency())
```
NAThompson, line 182: I think 10 is probably too low. Launching a thread should take at least [10 microseconds](https://lemire.me/blog/2020/01/30/cost-of-a-thread-in-c-under-linux/).

Your benchmark shows that a sequential variance takes 5.16N nanoseconds, so just to ensure that you can't complete the computation before you launch the thread, you need N >= 10^4/5.16 ~ 2000.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -0,0 +1,370 @@
...
+        Real M1_b = std::get<0>(temp);
+        Real M2_b = std::get<1>(temp);
+        Real M3_b = std::get<2>(temp);
+        Real M4_b = std::get<3>(temp);
+        Real range_b = std::get<4>(temp);
+
+        const Real n_ab = range_a + range_b;
+        const Real delta = M1_b - M1_a;
+        
+        M1_a = (range_a * M1_a + range_b * M1_b) / n_ab;
```
NAThompson, line 232: Could you put a reference to these equations?

mborland: Added at the top of the function.

include/boost/math/statistics/detail/single_pass.hpp
mborland: Makes sense to me; I figured 10 was going to be too low. `std::async` should attempt to use a thread pool with the launch policy here, but ultimately I believe it's implementation defined.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -106,90 +106,28 @@ ReturnType first_four_moments_sequential_impl(ForwardIterator first, ForwardIter
...
 {
     using Real = typename std::tuple_element<0, ReturnType>::type;
 
     const auto elements = std::distance(first, last);
     static unsigned num_threads;
-    if(elements/10 >= std::thread::hardware_concurrency())
+    
+    // https://lemire.me/blog/2020/01/30/cost-of-a-thread-in-c-under-linux/
+    if(elements/2000 >= std::thread::hardware_concurrency())
     {
```
NAThompson, line 121: What happens if `std::thread::hardware_concurrency()` returns 0?

Also, might be better to do `elements >= 2000*std::thread::hardware_concurrency()` than do the division.

include/boost/math/statistics/detail/single_pass.hpp
```diff
@@ -106,90 +106,28 @@ ReturnType first_four_moments_sequential_impl(ForwardIterator first, ForwardIter
...
-
+// https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Higher-order_statistics
+// EQN 3.1: https://www.osti.gov/servlets/purl/1426900
 template<typename ReturnType, typename ForwardIterator>
 ReturnType first_four_moments_parallel_impl(ForwardIterator first, ForwardIterator last)
 {
     using Real = typename std::tuple_element<0, ReturnType>::type;
 
     const auto elements = std::distance(first, last);
     static unsigned num_threads;
```
NAThompson, line 117: I'd remove the `static` here unless you have a compelling reason to have it.

include/boost/math/statistics/detail/single_pass.hpp
NAThompson: Am I correct in thinking that this model assumes that each thread generation costs 10us, sequentially, for 10j us of cost? I believe it costs the same to launch 10 threads as it does to launch 1.

NAThompson: Ok, assuming it take 10µs to launch j threads, independent of j, the parallel calculation is faster whenever

10 + 5.13e-3 N/j <= 5.13e-3N => N >= 10^4j/5.13(j-1).

The j = 0 and j=1 case should shunt through the sequential case, obviously.

NAThompson:
Other than the two small issues identified above, I'm happy with this.

mborland:
@NAThompson I should have the fixes in soon; working through how to fix all these CI failures. By and large they seem to be related to [TBB](https://github.com/oneapi-src/oneTBB).

NAThompson:
@mborland : You're not the first one to struggle with TBB!

include/boost/math/statistics/detail/single_pass.hpp
mborland:
```
diff --git a/include/boost/math/statistics/detail/single_pass.hpp b/include/boost/math/statistics/detail/single_pass.hpp
index 00b2d9e67..8559d0b31 100644
--- a/include/boost/math/statistics/detail/single_pass.hpp
+++ b/include/boost/math/statistics/detail/single_pass.hpp
@@ -114,20 +114,32 @@ ReturnType first_four_moments_parallel_impl(ForwardIterator first, ForwardIterat
     using Real = typename std::tuple_element<0, ReturnType>::type;
 
     const auto elements = std::distance(first, last);
-    static unsigned num_threads;
+    const unsigned max_concurrency = std::thread::hardware_concurrency() == 0 ? 2u : std::thread::hardware_concurrency();
+    unsigned num_threads = 2u;
     
+    // Threading is faster for: 10 + 5.13e-3 N/j <= 5.13e-3N => N >= 10^4j/5.13(j-1).
+    const auto parallel_lower_bound = 10e4*max_concurrency/(5.13*(max_concurrency-1));
+    const auto parallel_upper_bound = 10e4*2/5.13; // j = 2
+
     // https://lemire.me/blog/2020/01/30/cost-of-a-thread-in-c-under-linux/
-    if(elements/2000 >= std::thread::hardware_concurrency())
+    if(elements < parallel_lower_bound)
     {
-        num_threads = std::thread::hardware_concurrency();
+        return detail::first_four_moments_sequential_impl<ReturnType>(first, last);
     }
-    else if(elements < 4000)
+    else if(elements >= parallel_upper_bound)
     {
-        return detail::first_four_moments_sequential_impl<ReturnType>(first, last);
+        num_threads = max_concurrency;
     }
     else
     {
-        num_threads = elements/2000;
+        for(unsigned i = 3; i < max_concurrency; ++i)
+        {
+            if(parallel_lower_bound < 10e4*i/(5.13*(i-1)))
+            {
+                num_threads = i;
+                break;
+            }
+        }
```
Here is a quick diff. It should phase in the number of threads up to the maximum.

mborland:
@NAThompson CI is clean. The only failure is: `multiprc_concept_check_1.cpp(42): fatal error C1060: compiler is out of heap space`. I haven't seen this before, and this PR does not affect the test set it came from. More recent CI cycles are showing green still so I do not think it is overly concerning.

NAThompson:
Looks good to me!

PR #436, opened 2020-09-17
fix typo in README.md

suyash-patil:
Here are the snapshots:

Before:
before

After:
after


NAThompson:
@suyash-patil : Yup, looks good, thanks!

PR #455, opened 2020-11-10
At least make error dimensionally correct.

### LINKED ISSUE #449: tanh_sinh integral is re-scaled internally but the tolerance isn't
The convergence criteria for the tanh_sinh quadratures looks like `error < tol * L1` (https://github.com/boostorg/math/blob/develop/include/boost/math/quadrature/detail/tanh_sinh_detail.hpp#L377).

The problem is that depending on the endpoints of the integral, `error` and `L1` are scaled differently than the actual result. For instance, here's a rescaling if both endpoints are [finite](https://github.com/boostorg/math/blob/develop/include/boost/math/quadrature/tanh_sinh.hpp#L204):

```
result_type Q = diff*m_imp->integrate(u, error, L1, function, left_min_complement, right_min_complement, tolerance, levels);

if (L1) {
  *L1 *= diff;
}
```

There's a different one [here](https://github.com/boostorg/math/blob/develop/include/boost/math/quadrature/tanh_sinh.hpp#L114).

So whenever I run a quadrature, if I want to check if it is converged, I cannot just double check `error < tol * L1` like the docs say (https://www.boost.org/doc/libs/1_74_0/libs/math/doc/html/math_toolkit/double_exponential/de_levels.html), I gotta check `error < tol * L1 * scaling`, where `scaling` is the internal rescaling of the integrals.

It would be nice if the convergence check was all in the units of the user-facing integral. I don't know what other quadratures this applies to, just hit it on `tanh_sinh`.

NAThompson:
@bbbales2: Could you make sure this fixes your issue?

I now wonder how many places I've made this same mistake . . . 

Fixes #449

### DIFF
--- a/include/boost/math/quadrature/tanh_sinh.hpp
+++ b/include/boost/math/quadrature/tanh_sinh.hpp
@@ -113,6 +113,10 @@ auto tanh_sinh<Real, Policy>::integrate(const F f, Real a, Real b, Real toleranc
           {
              *L1 *= 2;
           }
+          if (error)
+          {
+             *error *= 2;
+          }
 
           return Q;
        }
@@ -139,6 +143,10 @@ auto tanh_sinh<Real, Policy>::integrate(const F f, Real a, Real b, Real toleranc
           if (L1)
           {
              *L1 *= 2;
+          }
+          if (error)
+          {
+             *error *= 2;
           }
           return Q;
        }
@@ -203,6 +211,10 @@ auto tanh_sinh<Real, Policy>::integrate(const F f, Real a, Real b, Real toleranc
           {
              *L1 *= diff;
           }
+          if (error)
+          {
+             *error *= diff;
+          }
           return Q;
        }
     }
@@ -240,6 +252,10 @@ auto tanh_sinh<Real, Policy>::integrate(const F f, Real a, Real b, Real toleranc
       if (L1)
       {
          *L1 *= diff;
+      }
+      if (error)
+      {
+         *error *= diff;
       }
       return Q;
    }


bbbales2:
Yeah this fixes it and makes sense to me. Thanks!

So Boost is currently 1.74.0. If this pull request goes through, does that mean I should expect this change when 1.75.0 appears?

NAThompson:
@jzmaddock : Is 1.75 closed for merging?

NAThompson:
I now wonder if we should map the tolerance via tolerance -> tolerance/diff as well.

jzmaddock:
> Is 1.75 closed for merging?

Yes, the beta candidates are out, there may be an opportunity to merge bug fixes post-beta but the timing will be tight!

NAThompson:
It appears that we are now seriously constrained by CI resources. It looks like Travis is only doing one job at a time; moreover it doesn't look like it moves from one job to the next immediately.

bbbales2:
@NAThompson okay cool. I will just put a fix for this in our code for now. Lemme know when you know what version this might appear in, and I can roll back the fix in our code when we upgrade boost.

NAThompson:
@jzmaddock : Looking at the build log, it appears that only bug fixes have gone in since the merge to master:

```
f3952b8b41ddb737155104f1738b544da77e9c8b (HEAD -> develop, origin/develop, origin/HEAD) At least make error dimensionally correct. (#455)
6cbd2259946184e020a564fd57f2f1948f61ab23 Merge pull request #454 from boostorg/issue453
8e3c2bc77d042136e8c3a5f4982e6ff617d41787 Merge pull request #452 from boostorg/issue451
c25cae2067abc689997881781c985b7e2ddbf9e2 (origin/issue453) Correct testing mistake in test_legendre.hpp.
889620031aaad68bdc6eba46be80ee4517e563e5 Fix up legendre_p(n, n-1, x) as special case. Add tests. Fixes https://github.com/boostorg/math/issues/453.
56c4c890d107b23064fdaf384dc9e085aa62ef36 (origin/issue451) Fix spurious overflow in gamma_p/q calculation. Add tests. Fixes: https://github.com/boostorg/math/issues/451
6b8c2691d38b44ec61f2664a9635948058aaca13 Correct expected error rate - epsilon should be as a percentage for BOOST_CHECK_CLOSE.
83f1aa42e11577396b0f4f7ae49a0ac96e1c0ae6 Merge pull request #448 from suyash-patil/develop
01888dc92970db2117dad9b98795dcef340d4e79 fixed CSS property
a5ff4371daabf0dfcde6d89ad843196e4a811f46 Merge pull request #447 from suyash-patil/develop
eb532dc219fac300a10b84d49f9f34710dbb8cfe small typo fixed
992794d591dbb894ee31be90726adfb87c2820e3 (origin/master) Merge branch 'master' into develop
```

Does this qualify for a post-beta merge?

jzmaddock:
Indeed, I don't see why not - note however, that the beta isn't quite out yet so master is still closed.  Which is OK as it lets the CI tests cycle ;)

Issue #459, opened 2020-12-20
beta distribution encounters overflow error for beta<0 and x=1


mckib2:
The beta distribution PDF has interesting cases at x=0 and x=1.  x=0 seems to be handled correctly, but when beta<1, x=1 it throws an overflow error:
```cpp
#include <iostream>
#include <boost/math/distributions/beta.hpp>

int main() {
    auto d = boost::math::beta_distribution<double>(2.3098496451481823, 0.6268795430096368);
    std::cout << "pdf(1.0): " << boost::math::pdf(d, 1.0) << std::endl;
    return 0;
}
```

Error:
```
terminate called after throwing an instance of 'boost::wrapexcept<std::overflow_error>'
  what():  Error in function ibeta_derivative<double>(double,double,double): Overflow Error
```

jzmaddock:
Confirmed.

mborland:
@jzmaddock I have a PR about ready for this one.

jzmaddock:
@mborland : that was my first thought, but I don't think that's quite the right fix.  We need to address this upstream in ibeta_derivative here: https://github.com/boostorg/math/blob/db42e6bced785fa5d7fcb92ad7406f139a4ca10b/include/boost/math/special_functions/beta.hpp#L1414

But hold on, now that I look at this again, if the PDF is:

`x^(a-1) (x-1)^(b-1) / beta(a, b)`

Then surely that *is* infinite for x == 1, b < 1 ?

I'm sure I'm missing something obvious though....

jzmaddock:
Ah... my bad, the answer is here: https://www.wolframalpha.com/input/?i=PDF%5BBetaDistribution%5Ba%2C+b%5D%2C+x%5D, the PDF is defined as zero at x >= 1 and the usual formula no longer applies.  OK, your fix is correct then :)  Thanks Matt!

mckib2:
#460  appears to fix the issue, thanks for looking into this!

PR #460, opened 2020-12-20
Fix for issue #459

mborland:
Fixes issue #459: corner case where check_x is valid [0, 1], but PDF is 0 for x = 0 or x =1


mborland:
@jzmaddock This should be good to go. The only CI failure was an error on Travis' end.

jzmaddock:
Thanks Matt!

Issue #237, opened 2019-08-10
Deprecated header and underflow warnings


NAThompson:
```
In file included from test_1F0.hpp:13:0,

                 from test_1F0.cpp:6:

../../../boost/test/floating_point_comparison.hpp:14:152: note: #pragma message: This header is deprecated. Use This header is deprecated. Please use <boost/test/tools/floating_point_comparison.hpp> instead. instead.
```

I'll just fix this whenever I get the chance; low priority.

Also, thousands of lines of:

```
hypergeometric_1f1_large_regularized.ipp:1150:7: warning: floating constant truncated to zero [-Woverflow]

       { SC_(2.2103403323709479866940105602712347713881e-31), SC_(1.2220855000000000000000000000000000000000e+06), SC_(1.0184045000000000000000000000000000000000e+06), SC_(1.1011379385218373966166313545729000750778e-6908211) }, 
```

in [this](https://travis-ci.org/boostorg/math/jobs/569821331) log. Actually, I don't know what to do about that. It's clearly harmless, and it does clearly underflow, so in a sense the pain is only generated by the compiler  . . .

NAThompson:
I'm working on a fix for the header deprecations; should be straightforward.

NAThompson:
I think I see the problem for the literal range:

```
lang-darwin.compile.c++ ../../../bin.v2/libs/math/test/test_1F1_float.test/clang-darwin-10.0/debug/link-static/threading-multi/visibility-hidden/test_1F1.o
warning: unknown warning option '-Wno-literal_range'; did you mean '-Wno-literal-range'? [-Wunknown-warning-option]
````

NAThompson:
Fixed in [https://github.com/boostorg/math/commit/0445f5c2cc20f5be2e15fb8b22bc0d0e85844c53](0445f5c).

evanmiller:
The underflow warnings still appear with GCC, e.g. here:

https://travis-ci.org/github/boostorg/math/jobs/746061269

They are now tripped by the `-Woverflow` flag. I have a ham-fisted fix [here](https://github.com/evanmiller/math/commit/30fadae9781a442253c988a595ffab30a2538732) which simply replaces the offending literals with zeroes. Happy to open a pull request with it, but I didn't want to contribute needlessly to the Travis CI Holiday Traffic Jam.

jzmaddock:
[...]

PR #469, opened 2020-12-23
Replace underflowing literals with zeroes

evanmiller:
GCC emits a large number of warnings when `-Woverflow` is enabled. Replace floating point literals that are less than 1e-10000 (roughly the square of the smallest representable number in quad precision) with zero in the hypergeometric 1f1 regularized table.

Tests are passing locally. See #237 for more context.


evanmiller:
Gently bumping before this goes stale (@jzmaddock)

jzmaddock:
Forgot about that one - thanks for the reminder - merging now!

Issue #462, opened 2020-12-22
Error at linear_regression_test.cpp:test_scaling_relations with Clang++-10 and C++17


evanmiller:
I've been experimenting with building Boost math with new-ish compiler versions. I get the following error running when running the "misc" test suite with Clang 10.0.0 and __cplusplus = 201703L:

```
**passed** ../../../bin.v2/libs/math/test/test_t_test.test/clang-linux-10.0.0/debug/link-static/threading-multi/visibility-hidden/test_t_test.test
clang-linux.link ../../../bin.v2/libs/math/test/linear_regression_test.test/clang-linux-10.0.0/debug/link-static/threading-multi/visibility-hidden/linear_regression_test
testing.capture-output ../../../bin.v2/libs/math/test/linear_regression_test.test/clang-linux-10.0.0/debug/link-static/threading-multi/visibility-hidden/linear_regression_test.run
====== BEGIN OUTPUT ======
Error at linear_regression_test.cpp:test_scaling_relations:151:
  ULP distance in float precision is 56, which exceeds 50, error/ulps  = 1.12.
  Expected: +3.205449104 = +0x1.9a4c28p+1
  Computed: +3.205462456 = +0x1.9a4c98p+1
  Mollified relative error: +4.16523244e-06
Error at linear_regression_test.cpp:test_scaling_relations:151:
  ULP distance in long double precision is 60, which exceeds 50, error/ulps  = 1.2.
  Expected: +3.205454929075602930715 = +0xc.d262c6e4ba1342cp-2
  Computed: +3.205454929075602943725 = +0xc.d262c6e4ba13468p-2
  Mollified relative error: +4.0588391843581566924e-18
Error count: 2, total ulp distance = 116

EXIT STATUS: 1
====== END OUTPUT ======
```

You can see the full output here (error message starts around line 2190):

https://github.com/evanmiller/math/runs/1595605908

The host system is Ubuntu 18.04, and the code base is an up-to-date fork of `develop`.

evanmiller:
Running more tests, I see the same error with G++-10. It also appears with both compilers under C++20.

mborland:
Confirmed on Fedora 33 with clang and GCC. Replacing `auto` with explicit types seems to do the trick. PR to follow.

jzmaddock:
This is showing up in the newly merged github-actions CI: https://github.com/boostorg/math/runs/1681924587?check_suite_focus=true

NAThompson:
[...]

Issue #430, opened 2020-09-03
Clang issues many -Wimplicit-int-float-conversion warnings from round, trunc etc


pabristow:
Running the Boost.math tests using Clang10.0.0, I note many warnings from all uses of round, trunc (and others?).

`In file included from ..\..\..\boost/math/concepts/real_concept.hpp:28:
..\..\..\boost/math/special_functions/round.hpp:118:12: warning: implicit conversion from 'long long' to 'long double' changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-int-float-conversion]
   if((r > (std::numeric_limits<boost::long_long_type>::max)()) || (r < (std::numeric_limits<boost::long_long_type>::min)()))
         ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
`

Warnings only with Clang, not GCC nor MSVC, but the very many warning are cluttering the log files.

Should I deal with this by quieting the warning, perhaps by adding 

`
<toolset>clang:<cxxflags>-Wno-implicit-int-float-conversion
`
to the /test jamfile?

Or is there some more subtle change required?

evanmiller:
I can't speak for the authors of the code but here may be some subtle issues with numbers here at the outer reaches of the `long long` type. For example, looking at the cited `llround` code:

https://github.com/boostorg/math/blob/5a1a3027c8399705200988c3551cb5071eadfbb3/include/boost/math/special_functions/round.hpp#L113-L121

Let's follow the logic of the warning and say that `r` comes in as `9223372036854775808`. The number `max` should be `9223372036854775807`, but according to the warning it is converted to `9223372036854775808` before comparison. The numbers compare equal, and so instead of raising an error, a static cast to `long long` is performed on `9223372036854775808`, which is unrepresentable by `long long`. The function in this case will return an undefined result.

Based on the foregoing analysis, I'd say the warnings are legitimate and likely merit additional attention. The test code does appear to cover the long long max case – but it could be falling into the trap of comparing the undefined integer result of the function with the same undefined test value (`long long` max cast to a `long double`). I don't know if anyone has run an undefined behavior sanitizer on the test suite, but they can be very helpful for uncovering edge cases like these at run-time.

evanmiller:
See also #435

mborland:
This should be resolved by PR #474

mborland:
Found another instance stemming from beta:

```
In file included from test_t_test.cpp:13:
[...]

PR #471, opened 2020-12-26
Fix for issue #462

mborland:
Fix for issue #462. Replace instances of `auto` with `Real`. c1 is over-estimated for all types so compensate ε as constant error coefficient. Clang-11.0.0 passes tests, but issues `-Wimplicit-const-int-float-conversion` warning as noted in issue #430.

### DIFF
--- a/include/boost/math/statistics/linear_regression.hpp
+++ b/include/boost/math/statistics/linear_regression.hpp
@@ -14,6 +14,7 @@
 #include <tuple>
 #include <iterator>
 #include <limits>
+#include <stdexcept>
 #include <boost/math/statistics/univariate_statistics.hpp>
 #include <boost/math/statistics/bivariate_statistics.hpp>
 
@@ -54,8 +55,15 @@ template<typename RandomAccessContainer, typename Real = typename RandomAccessCo
 auto simple_ordinary_least_squares_with_R_squared(RandomAccessContainer const & x, RandomAccessContainer const & y)
     -> std::tuple<Real, Real, Real>
 {
-    BOOST_ASSERT_MSG(std::size(x) > 2, "At least 2 samples are required to perform a linear regression.");
-    BOOST_ASSERT_MSG(std::size(x) == std::size(y), "The same number of samples must be in the independent and dependent variable.");
+    if (std::size(x) <= 1)
+    {
+        throw std::domain_error("At least 2 samples are required to perform a linear regression.");
+    }
+
+    if (std::size(x) != std::size(y))
+    {
+        throw std::domain_error("The same number of samples must be in the independent and dependent variable.");
+    }
 
     const std::tuple<Real, Real, Real> temp = boost::math::statistics::means_and_covariance(x, y);
     const Real mu_x = std::get<0>(temp);
@@ -63,7 +71,10 @@ auto simple_ordinary_least_squares_with_R_squared(RandomAccessContainer const & 
     const Real cov_xy = std::get<2>(temp);
     const Real var_x = boost::math::statistics::variance(x);
 
-    BOOST_ASSERT_MSG(var_x > 0, "Independent variable has no variance; this breaks linear regression.");
+    if (var_x <= 0) 
+    {
+        throw std::domain_error("Independent variable has no variance; this breaks linear regression.");
+    }
 
     // c1 is over-estimated for all types without compensation
     const Real c1 = cov_xy/(var_x - std::numeric_limits<Real>::epsilon());
--- a/include/boost/math/statistics/linear_regression.hpp
+++ b/include/boost/math/statistics/linear_regression.hpp
@@ -11,6 +11,9 @@
 #include <cmath>
 #include <algorithm>
 #include <utility>
+#include <tuple>
+#include <iterator>
+#include <limits>
 #include <boost/math/statistics/univariate_statistics.hpp>
 #include <boost/math/statistics/bivariate_statistics.hpp>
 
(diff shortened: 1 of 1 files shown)


include/boost/math/statistics/linear_regression.hpp
```diff
@@ -46,35 +49,29 @@ auto simple_ordinary_least_squares(RandomAccessContainer const & x,
...
-    }
-
-    if (x.size() != y.size())
-    {
-        throw std::domain_error("The same number of samples must be in the independent and dependent variable.");
-    }
-    auto [mu_x, mu_y, cov_xy] = boost::math::statistics::means_and_covariance(x, y);
-
-    auto var_x = boost::math::statistics::variance(x);
+    BOOST_ASSERT_MSG(std::size(x) > 2, "At least 2 samples are required to perform a linear regression.");
```
jzmaddock, line 57: I'm not sure that asserts are correct here: asserts normally verify our own logic, where these are checking input values.  In particular these checks disappear in release mode which rather worries me here.

mborland: Makes sense; I'll revert the asserts.

As an aside has anything gone in recently that would have touched [tr1](https://ci.appveyor.com/project/jzmaddock/math/builds/37002310/job/lhb7gba4fatmsyen?fullLog=true#L1302) or [nonfinite_num_facets](https://ci.appveyor.com/project/jzmaddock/math/builds/37002310/job/5e2eaaq6o4gjb8gc?fullLog=true#L3885)? CI is showing a litany of failures with those two (see links). tr1 is ODR linker errors.

jzmaddock: Nothing has changed at our end - the serialization errors look like they're coming from that lib, the link errors are more worrying as they suggest that lib and exe were built with differing build options.

jzmaddock: I can reproduce after updating Boost.Build to latest develop, but have no idea what's going on yet :(

mborland: Right before this went to CI Boost.Build had [two](https://github.com/boostorg/build/commit/30a7e738ad2757b811720a56173dd3ba3453eaef) [commits](https://github.com/boostorg/build/commit/8d37b68e1db40ef05ee974e6f263071cd42e6d79) on develop that changed linking on MSVC.

mborland:
@jzmaddock Minus all of the strange linker errors from Appveyor the CI results for this looks good.

jzmaddock:
I'll let @NAThompson double check before merging this one.

include/boost/math/statistics/linear_regression.hpp
```diff
@@ -46,35 +50,39 @@ auto simple_ordinary_least_squares(RandomAccessContainer const & x,
...
+    if (var_x <= 0) 
+    {
         throw std::domain_error("Independent variable has no variance; this breaks linear regression.");
     }
 
-
-    Real c1 = cov_xy/var_x;
-    Real c0 = mu_y - c1*mu_x;
+    // c1 is over-estimated for all types without compensation
+    const Real c1 = cov_xy/(var_x - std::numeric_limits<Real>::epsilon());
```
NAThompson, line 80: I don't understand this change. Do you have a reference for this?

Could you test it with data that is all < 1e-16? (e.g., the LIGO data is all < 1e-16).

include/boost/math/statistics/linear_regression.hpp
```diff
@@ -46,35 +50,39 @@ auto simple_ordinary_least_squares(RandomAccessContainer const & x,
...
-auto simple_ordinary_least_squares_with_R_squared(RandomAccessContainer const & x,
-                                   RandomAccessContainer const & y)
+template<typename RandomAccessContainer, typename Real = typename RandomAccessContainer::value_type, 
+         typename Size = typename RandomAccessContainer::size_type>
+auto simple_ordinary_least_squares_with_R_squared(RandomAccessContainer const & x, RandomAccessContainer const & y)
+    -> std::tuple<Real, Real, Real>
 {
-    using Real = typename RandomAccessContainer::value_type;
-    if (x.size() <= 1)
+    if (std::size(x) <= 1)
```
NAThompson, line 58: I don't have a problem with `std::size`, and in fact it solves a real problem. But it also has historically caused problems which is why I avoid it.

Did you have a type that requires `std::size`?

include/boost/math/statistics/linear_regression.hpp
```diff
@@ -46,35 +50,39 @@ auto simple_ordinary_least_squares(RandomAccessContainer const & x,
     return std::make_pair(c0, c1);
 }
 
-template<class RandomAccessContainer>
-auto simple_ordinary_least_squares_with_R_squared(RandomAccessContainer const & x,
-                                   RandomAccessContainer const & y)
+template<typename RandomAccessContainer, typename Real = typename RandomAccessContainer::value_type, 
+         typename Size = typename RandomAccessContainer::size_type>
+auto simple_ordinary_least_squares_with_R_squared(RandomAccessContainer const & x, RandomAccessContainer const & y)
+    -> std::tuple<Real, Real, Real>
```
NAThompson, line 56: If `Real` is a set of integers, then doesn't this break?

include/boost/math/statistics/linear_regression.hpp
mborland: I have no compelling reason other than Scott Myers et. al. recommend it, and I was already changing things.

include/boost/math/statistics/linear_regression.hpp
mborland: I do not have a reference, but empirically it seems to be the case. If you multiply everything by 1e-22 (LIGO-Hanford Range) and represent with `float_quad`, `float_50`, and `float_100` the results are similarly improved.

include/boost/math/statistics/linear_regression.hpp
mborland: Like in the event the container passed in is a Matrix? As long is it supports `operator/` and subtraction by a constant than it should behave normally, and if not it would break anyway. `means_and_covariance` returns `std::tuple<Real, Real, Real>` so that should be changed as well if integral types need to return `double`. I know that is the case in almost all of univariate statistics and would be easy enough to change.

include/boost/math/statistics/linear_regression.hpp
NAThompson: @mborland : Representing with `float_quad` defeats the purpose. There is no reason to use `float_quad` on LIGO-Handford data; the precision isn't there; double is fine. In this case the denominator becomes negative and the result is wrong.

I still don't understand why this improves the result.

include/boost/math/statistics/linear_regression.hpp
NAThompson: Ok, maybe the ecosystem has improved a bit since I first tried it. I'm fine with this change.

include/boost/math/statistics/linear_regression.hpp
NAThompson: Ok, I do need to check this then. I had a goal of supporting integer data and if I haven't then that needs to be fixed.

include/boost/math/statistics/linear_regression.hpp
mborland: 50% detection probability is at [3.2e-20](https://image2.slideserve.com/4246672/detection-efficiency-for-narrow-band-bursts-l.jpg) which is why I tested with quadruple and greater precision. The denominator would only go negative if `var_x` is 0 since its one epsilon of Real. The tests for variance only specify <5 epsilon and the result typically has positive error.

include/boost/math/statistics/linear_regression.hpp
mborland: I can open another PR to add integer support; that would be easy at this point. Replacing the structured binding in `covariance` with an tuple would drop the required language standard from 17 to 11 as well.

include/boost/math/statistics/linear_regression.hpp
NAThompson: 3.2e-20 is fine for double precision. Double precision can represent numbers down to 1e-300.

NAThompson: > The denominator would only go negative if var_x is 0 since its one epsilon of Real.

This is incorrect. You should read up on the rounding model of floating point arithmetic; see Higham, Accuracy and Stability of Numerical Algorithms. (Your statement is correct for *fixed-point* arithmetic.)

mborland: I'll have to find a copy and give it a read. I can close this as well because it is clearly not correct.

PR #473, opened 2020-12-27
Update to gitignore

mborland:
Some small updates to the .gitignore. `**/.temps/*` is for b2 build artifacts and `build/*` is for all others (e.g. gcc with -MMD flag). `.vscode` for VSCode config files.


NAThompson:
I'm cool with all these changes.

@mborland : Will try to take a look at your other PRs after Christmas holiday is over.

PR #474, opened 2020-12-27
Fix for issue #430

mborland:
Fix for issue #430 . Added explicit casting to eliminate `-Wimplicit-int-float-conversion` warnings on clang. Added test case of `std::numeric_limits<long long>::max() + 1` which passes ubsan.


mborland:
@jzmaddock The CI on this one is clean and should be good to go.

PR #480, opened 2021-01-04
Add integer support to bivariate statistics

mborland:
Add support for integers to bivariate statistics. Add unit tests with same tolerances as Real types. Remove uses of structured bindings and `std::size()` to reduce the required language standard from 17 to 11.

### DIFF
--- a/include/boost/math/statistics/bivariate_statistics.hpp
+++ b/include/boost/math/statistics/bivariate_statistics.hpp
@@ -1,4 +1,5 @@
 //  (C) Copyright Nick Thompson 2018.
+//  (C) Copyright Matt Borland 2021.
 //  Use, modification and distribution are subject to the
 //  Boost Software License, Version 1.0. (See accompanying file
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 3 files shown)


mborland:
@NAThompson I realize the irony that I am turning around and taking the opposite stance on the `u.size()` vs `std::size(u)` argument. I forgot that it was a c++17 language feature.

include/boost/math/statistics/bivariate_statistics.hpp
```diff
@@ -8,25 +9,28 @@
...
+#include <cstddef>
 #include <boost/assert.hpp>
 
 
-namespace boost{ namespace math{ namespace statistics {
+namespace boost{ namespace math{ namespace statistics { namespace detail {
 
-template<class Container>
-auto means_and_covariance(Container const & u, Container const & v)
+template<class ReturnType, class Container>
```
NAThompson, line 20: This will be a breaking change.

mborland: It is a breaking change so it got renamed and moved into the detail namespace. The user facing functions remain the same and offer return type deduction via `std::enable_if` (lines 93-103). I can move implementation details into a separate file to make this more clear. I did not do that already because this file is still pretty small.

mborland:
CI is clean minus 1 test where it tried to compile in c++03 mode and failed with error `error C4519: default template arguments are only allowed on a class template`. I added requires `cxx11_sfinae_expr` to the Jamfile to resolve.

NAThompson:
Ok, the structured bindings still work on the *user* end right? You just removed them so the tests would pass in C++11 mode?

If so, I'm happy with these changes.

@jzmaddock : Anything I missed?

mborland:
> Ok, the structured bindings still work on the _user_ end right? You just removed them so the tests would pass in C++11 mode?

That is correct; structured bindings still work and the user is free to use them.

jzmaddock:
@NAThompson : The interface changes look good to me, I haven't looked in detail at the implementation details.

Issue #446, opened 2020-10-27
Initializing nested autodiff::fvar


wgledbetter:
This might just be me misunderstanding how the `fvar` type is meant to be used.

This code works:
```
fvar<double, 1> x1 = make_fvar<double, 1>(1.5);
auto y1 = myFunction(x1);
cout << "y1 deriv = " << y1.derivative(1) << endl;

// Output:
    y1 deriv = 12.345
```

I would like to be able to do:
```
fvar<fvar<double, 1>, 1> x2 = make_fvar<fvar<double, 1>, 1>(1.5);
auto y2 = myFunction(x2);
cout << "y2 deriv = " << y2.derivative(1) << endl;

// Output:
    y2 deriv = depth(1)(12.345, 98.765)
```
However, `make_fvar<fvar<double, 1>, 1>()` throws a compile-time error that there is no matching constructor. Reading further, it is trying to convert from `const fvar<double,1>` to `const double`. I understand that if you down-cast you would lose the derivative information, but would it make sense to have a workaround in this case?

Alternatively, am I missing something?

pulver:
What is it that you are trying to do, mathematically?

It looks like you are trying to declare `x2` to be an independent variable of differentiation from `x1`. If that is the case, then here are a few ways to do this (but only the last 2 ways are recommended.)

### Method 1 (Not recommended - don't call fvar directly)

```
#include <boost/math/differentiation/autodiff.hpp>
#include <iostream>

using namespace boost::math::differentiation;

template <typename RealType, size_t Order>
using fvar = detail::fvar<RealType, Order>;

auto myFunction = [](auto x1, auto x2) {
  return x1*x1 + x2*x2;
};

int main() {
  fvar<double, 1> x1 = fvar<double, 1>(2, true);
  fvar<fvar<double, 1>, 1> x2 = fvar<fvar<double, 1>, 1>(3, true);
  auto y2 = myFunction(x1, x2);

  std::cout << "y2.derivative(0,0) = " << y2.derivative(0,0) << std::endl; // 13
  std::cout << "y2.derivative(0,1) = " << y2.derivative(0,1) << std::endl; // 6
  std::cout << "y2.derivative(1,0) = " << y2.derivative(1,0) << std::endl; // 4
  std::cout << "y2.derivative(1,1) = " << y2.derivative(1,1) << std::endl; // 0
  return 0;
}
```
[...]

PR #483, opened 2021-01-06
Make "v" accessible to derived classes

wgledbetter:
I'd like to be able to modify the underlying "v" array inside fvar, partially in response to #446

### DIFF
--- a/include/boost/math/differentiation/autodiff.hpp
+++ b/include/boost/math/differentiation/autodiff.hpp
@@ -125,6 +125,7 @@ using get_type_at = typename type_at<RealType, Depth>::type;
 // https://www.boost.org/libs/math/doc/html/math_toolkit/real_concepts.html
 template <typename RealType, size_t Order>
 class fvar {
+ protected:
   std::array<RealType, Order + 1> v;
 
  public:


NAThompson:
@pulver

pulver (approved):
Looks good to me.

Would be interested to see how this is used in practice.

wgledbetter:
@pulver Here's how I'm using it.

I wrote a class `fvar_editor` so I can shift the elements of the 'v' array. I wanted to be able to create `yPrime` and `yPrimePrime` such that:
```
y.derivative(n+2) == yPrime.derivative(n+1) == yPrimePrime.derivative(n)
```

It's implemented to solve my templated derivative issue from #446 like this:

```
template <class F>
struct Derivative {
  F f;

  template <typename T>
  T calculate(T x) {
    if (is_fvar<T>::value) {
      const int Order = T::order_sum;
      using Type = typename T::root_type;
      auto xx = fvar<Type, Order+1>(x);
      auto yy = f.calculate(xx);
      T yPrime = fvar_editor<Type, Order>::shiftDeriv(yy, 1);
    } else {
      auto xx = make_fvar<T, 1>(x);
      auto yy = f.calculate(xx);
      T yPrime = yy.derivative(1);
      return yPrime;
    }
  }
};
```
for any class F that has a templated `calculate` method.

PR #484, opened 2021-01-07
Add integer support to linear regression

mborland:
Add support for integer types to linear regression. Will be compatible with C++11 once PR #434 is merged since it is dependent on univariate statistics becoming C++11 compatible. Does not fix issue #462.


mborland:
@NAThompson This should be good to go. The only CI failure was fixed by commit 605f563.

NAThompson:
Looks good to me!

Issue #489, opened 2021-01-13
CI Failure in makima_test


mborland:
Reported by @NAThompson but moved from the PR into issues:

```
makima_test.cpp:168:26:   required from here
../../../boost/math/interpolators/makima.hpp:26:80: internal compiler error: in dependent_type_p, at cp/pt.c:23654
            Real left_endpoint_derivative = std::numeric_limits<Real>::quiet_NaN(),
                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-7/README.Bugs> for instructions.
```

Looking at the travis test battery only g++-5, 6, and 8 were tested. To suppress this issue I can add the flag `-ffinite-math-only` but it would not solve the underlying issue. Thoughts?

NAThompson:
Is the compiler running out of memory on the CI system? This might be a random failure . . .

mborland:
@NAThompson I looked through the runs of the CI system on my fork, and this error pops up consistently. I would file a report on the GCC bugzilla but only 8, 9, and 10 are supported and there are no issues with those.

NAThompson:
@jzmaddock : I'd prefer to just support gcc's supported set on our own CI. What's your opinion?

jzmaddock:
Since it's just this test, I would prefer to continue testing with everything post gcc-6, maybe even gcc-5 as it has good C++11 support, and just mark this test as failing (add some pp-logic to the test to disable it for old gcc versions).

PR #490, opened 2021-01-14
Fix for issue 489

mborland:
Issue #489


NAThompson:
Just grepped through the logs; this does some the issue it claims it does, but there are a few more errors which I hopefully fixed on develop.

PR #491, opened 2021-01-14
Skip test_autodiff_8 with clang 5

mborland:
CI tests using clang 5 fail at test_autodiff_8 for all language standards. Not an issue with clang 6 and on.

### DIFF
--- a/test/test_autodiff_8.cpp
+++ b/test/test_autodiff_8.cpp
@@ -8,8 +8,6 @@
 
 BOOST_AUTO_TEST_SUITE(test_autodiff_8)
 
-#if __clang_major__ > 5 || __GNUC__ > 5 || defined(_MSC_VER)
-
 BOOST_AUTO_TEST_CASE_TEMPLATE(hermite_hpp, T, all_float_types) {
   using test_constants = test_constants_t<T>;
   static constexpr auto m = test_constants::order;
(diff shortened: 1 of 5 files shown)


pulver (approved):
Looks good, thank you!

test/test_autodiff_8.cpp
```diff
@@ -8,6 +8,8 @@
 
 BOOST_AUTO_TEST_SUITE(test_autodiff_8)
 
+#if __clang_major__ > 5 || __GNUC__ > 5 || defined(_MSC_VER)
```
pulver, line 11: Should the `#if` block surround the entire test suite? Otherwise as you may have seen:
```
testing.capture-output ../../../bin.v2/libs/math/test/test_autodiff_8.test/clang-linux-5.0.0/release/link-static/threading-multi/visibility-hidden/test_autodiff_8.run
====== BEGIN OUTPUT ======
Test setup error: test tree is empty

EXIT STATUS: 200
====== END OUTPUT ======
```

pulver (changes requested):

NAThompson:
@mborland : Do you mind rebasing on develop? I just fixed a pile of issues as well are merged your other fix, so there's a chance of this going green.

test/test_autodiff_8.cpp
mborland: Surrounding the entire test suite yields the same error. I moved the `#if` block to right above the test that fails.

mborland:
@NAThompson Rebased. Makes the commit history look strange but there are no merge conflicts.

mborland:
@NAThompson Not green, but down to single digit errors. Linux is clean.

Issue #465, opened 2020-12-23
Test failure in ooura_fourier_transform_test with Apple Clang 12.0 and C++17


evanmiller:
Full log here:

https://github.com/evanmiller/math/runs/1597837166

Relevant portion:

```
testing.capture-output ../../../bin.v2/libs/math/test/ooura_fourier_integral_test.test/clang-darwin-12.0/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden/ooura_fourier_integral_test.run
====== BEGIN OUTPUT ======
Running 1 test case...
Testing integral of cos(x)/(x*x+1) on type float
Testing integral of cos(x)/(x*x+1) on type double
Testing integral of cos(x)/(x*x+1) on type long double
Testing integral of exp(-a*x) on type float
Testing integral of exp(-a*x) on type double
Testing integral of exp(-a*x) on type long double
ooura_fourier_integral_test.cpp:298: error: in "ooura_fourier_transform_test": difference{2.96692e-16} between Is{0.124999999999999962914} and exact{0.125} exceeds 1.08420217248550443401e-17

...

*** 1 failure is detected in the test module "test_ooura_fourier_transform"

EXIT STATUS: 201
====== END OUTPUT ======
```

jzmaddock:
@NAThompson : Error rate is suspiciously close to double precision, so one would suspect a stray double constant or double-precision function call?

NAThompson:
@evanmiller : Is your `long double` 64 bits? That would explain the problem . . . 

My apologies to everyone; I can't look at any of these until January . . .

evanmiller:
long double is 80 bits

NAThompson:
Should be fixed in [this commit](https://github.com/boostorg/math/commit/76631a42b611a49d897ce6a5190243e9f87b7bd8).

evanmiller:
@NAThompson The reported difference is nearly 30X the accepted tolerance (3e-16 vs 1e-17). I'll wait to see what the test says but a 5X tolerance bump won't be enough, will it?

NAThompson:
@evanmiller : Bah; I should've taken the opportunity to convert the test to using ULPs. . . I'll fix again if needed.

evanmiller:
@NAThompson Looks like it's needed...

https://github.com/boostorg/math/runs/1704804139?check_suite_focus=true#step:12:603

evanmiller:
[...]

PR #492, opened 2021-01-14
Fix for MSVC CI Failures

mborland:
Should reduce noise from the CI system while waiting for https://github.com/boostorg/random/pull/76 to be merged


mborland:
@NAThompson This suppressed a good number of the Windows failures. Previously all MSVC versions and language standards were failing for the `../tools` tests

mborland:
@NAThompson Do you have a solution for issue #465? If it is just increasing the tolerance I can throw that in here. Once that is fixed CI should finally be green.

jzmaddock:
> Do you have a solution for issue #465? If it is just increasing the tolerance I can throw that in here. Once that is fixed CI should finally be green.

Has the code been fed through `std_real_concept` to check there are no mistakenly called double precision std library functions?

NAThompson:
Yeah I need to actually think about that problem; will try to get to it soon!

mborland:
I merged develop just to verify, but I believe this PR resolves all the outstanding CI failures except the two stemming from that issue.

evanmiller:
@mborland Looking at the logs, the two `ooura_fourier_integral_test` failures appear to be unrelated. The Windows failure looks like another `bigobj` problem, so you should be able to fix it the same way the Daubechies tests were fixed.

However, it looks like the big object problems can be fixed with `-Wa,-mbig-obj` instead of skipping the compilation via ifdef. This is already done for a few tests, see e.g.

https://github.com/boostorg/math/blob/develop/test/Jamfile.v2#L490

Might be worth investigating if you haven't done so already.

mborland:
@evanmiller Good call. The jamfile already had `/bigobj` for `daubechies_scaling_test` and `daubechies_wavelet_test` using MSVC; just needed to add it for mingw runs.

mborland:
This should be good to go. Only two remaining errors are issue #465.

PR #494, opened 2021-01-16
Add Z-test

mborland:
Adds one and two sample z-tests. The implementation is nearly identical to t-test (as it should be).


doc/statistics/z_test.qbk
```diff
@@ -0,0 +1,114 @@
...
+[section:z_test /z/-tests]
+
+[heading Synopsis]
+
+```
+#include <boost/math/statistics/z_test.hpp>
+
+namespace boost::math::statistics {
+
+template<typename Real, typename Size = Real>
```
NAThompson, line 17: Is `typename Size = Real` a good naming convention here? I feel like this default may be incorrect.

test/Jamfile.v2
```diff
@@ -979,6 +979,7 @@ test-suite misc :
    [ run anderson_darling_test.cpp : : : [ requires cxx17_if_constexpr cxx17_std_apply ] ]
    [ run ljung_box_test.cpp : : : [ requires cxx17_if_constexpr cxx17_std_apply ] ]
    [ run test_t_test.cpp : : : [ requires cxx17_if_constexpr cxx17_std_apply ] ]
+   [ run test_z_test.cpp : : : [ requires cxx17_if_constexpr cxx17_std_apply ] ]
```
NAThompson, line 982: Do you need the `requires cxx17_*` here? It looks like you've avoided type deduction using C++17 `if constexpr`.

test/test_z_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    auto [computed_statistic, computed_pvalue] = boost::math::statistics::one_sample_z_test(Real(10), Real(2), Real(100), Real(10));
+    CHECK_ULP_CLOSE(Real(0), computed_statistic, 5);
+    CHECK_MOLLIFIED_CLOSE(Real(0), computed_pvalue, 1e-22);
+
+    auto [computed_statistic_2, computed_pvalue_2] = boost::math::statistics::one_sample_z_test(Real(10), Real(2), Real(100), Real(5));
+    CHECK_ULP_CLOSE(Real(25), computed_statistic_2, 5);
+
+    auto [computed_statistic_3, computed_pvalue_3] = boost::math::statistics::one_sample_z_test(Real(1.0/2.0), Real(10), Real(100), Real(1.0/3.0));
+    CHECK_ULP_CLOSE(Real(1.0/6.0), computed_statistic_3, 5);
```
NAThompson, line 26: This will never be more accurate than double precision. Use `Real(1)/6`.

test/test_z_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+template<typename Real>
+void test_two_sample_z()
+{
+    std::vector<Real> set_1 {1,2,3,4,5};
+    std::vector<Real> set_2 {2,3,4,5,6};
+
+    auto [computed_statistic, computed_pvalue] = boost::math::statistics::two_sample_z_test(set_2, set_1);
+    CHECK_ULP_CLOSE(Real(1), computed_statistic, 5);
+    CHECK_MOLLIFIED_CLOSE(Real(0), computed_pvalue, 1e-22);
```
NAThompson, line 48: Where is the `1e-22` coming from here?

doc/statistics/z_test.qbk
mborland: It is incorrect. I was toying with not having to call `Real(sample_size)` but removed it.

test/Jamfile.v2
mborland: There is a dependency on univariate statistics which is currently C++17. Once PR #434 is merged everything I've recently edited in the stats library will drop to C++11.

test/test_z_test.cpp
mborland: Changed to `5*std::numeric_limits<Real>::epsilon()`.

mborland:
@NAThompson Once a message get posted on the ML that build is fixed I'll cycle this through CI.

test/test_z_test.cpp
```diff
@@ -0,0 +1,82 @@
...
+void test_one_sample_z()
+{
+    auto [computed_statistic, computed_pvalue] = boost::math::statistics::one_sample_z_test(Real(10), Real(2), Real(100), Real(10));
+    CHECK_ULP_CLOSE(Real(0), computed_statistic, 5);
+    CHECK_MOLLIFIED_CLOSE(Real(0), computed_pvalue, 5*std::numeric_limits<Real>::epsilon());
+
+    auto [computed_statistic_2, computed_pvalue_2] = boost::math::statistics::one_sample_z_test(Real(10), Real(2), Real(100), Real(5));
+    CHECK_ULP_CLOSE(Real(25), computed_statistic_2, 5);
+
+    auto [computed_statistic_3, computed_pvalue_3] = boost::math::statistics::one_sample_z_test(Real(1.0/2.0), Real(10), Real(100), Real(1.0/3.0));
```
NAThompson, line 26: Same problem here: 1.0/3.0 is only double precision accurate.

doc/statistics/z_test.qbk
```diff
@@ -0,0 +1,114 @@
...
+]
+
+[section:z_test /z/-tests]
+
+[heading Synopsis]
+
+```
+#include <boost/math/statistics/z_test.hpp>
+
+namespace boost::math::statistics {
```
NAThompson, line 15: Says below that these are C++11 compatible, so maybe remove the C++17 syntax.

NAThompson:
For some reason the LaTex came out looking horrible for me:

Screen Shot 2021-01-16 at 5 49 31 PM

Does it look like this for you?

mborland:
It has been replaced by a Latex generated SVG. That does not look great.

mborland:
@NAThompson The only CI failure has been fixed in commit 66f3a5f. Should be good to go.

NAThompson:
@mborland : Docs look good to me now. For reference, I use this tool to create my equations: https://viereck.ch/latex-to-svg/

Probably many similar tools are out there . . .

jzmaddock:
>Probably many similar tools are out there . . .

This site works well IMO: http://www.tlhiv.org/ltxpreview/

PR #497, opened 2021-01-22
static_assert when mingw is used in a way not supported.

jzmaddock:
IT has issue with thread_local and non-trivial types.

### DIFF
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -470,6 +470,14 @@ namespace boost{ namespace math{
 #else
 #  define BOOST_MATH_THREAD_LOCAL
 #endif
+//
+// Some mingw flavours have issues with thread_local and types with non-trivial destructors
+// See https://sourceforge.net/p/mingw-w64/bugs/527/
+//
+#if !defined(BOOST_NO_CXX11_THREAD_LOCAL) && (defined(__MINGW32__) || defined(__MINGW64__)) && !defined(_REENTRANT)
+#  define BOOST_MATH_NO_THREAD_LOCAL_WITH_NO_TRIVIAL_TYPES
+#endif
+
 
 //
 // Can we have constexpr tables?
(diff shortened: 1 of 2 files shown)


ckormanyos:
I wrote some thought-providing detection logic [here:](https://github.com/boostorg/multiprecision/pull/291#issuecomment-765362563)

Something like (maybe?) this:

```
#if defined(__GNUC__) && defined(__MINGW__)

#if (__GNUC__ < 9) && (__GNUC_MINOR__ < 2) && defined(_REENTRANT)
static_assert(false, "Error: bad behavior");
#endif

#endif
```

jzmaddock:
I don't think we can say that it's fixed post gcc-9 because there are no gcc-9 win32-thread releases yet.  Maybe there will be, maybe it's a dead project I don't know.  MSys builds all works fine, as do older releases built with the POSIX thread model.  BTW I think your _REENTRANT logic is backwards (it *is* defined for mingw builds that are OK, and not for builds with the issue).

jzmaddock:
Refs https://github.com/boostorg/multiprecision/pull/291

include/boost/math/tools/config.hpp
```diff
@@ -470,6 +470,14 @@ namespace boost{ namespace math{
 #else
 #  define BOOST_MATH_THREAD_LOCAL
 #endif
+//
+// Some mingw flavours have issues with thread_local and types with non-trivial destructors
+// See https://sourceforge.net/p/mingw-w64/bugs/527/
+//
```
ckormanyos, line 476: Oh... Now I see... You already have worked out the logic. I could use this in the multiprecision branch for CI...

jzmaddock: Maybe not: I notice the failing CI run *does* define _REENTRANT so there's still something different going on than on my local machine :(

ckormanyos: > ... there's still something different going on than on my local machine

An interesting point. The virtual machine on CI runs on `[windows-latest]` which is an alias for Windows Server 2019. Perhaps there is a different C-Library or threading model on the MinGW that somehow got put together on the machine.

It is not immediately clear to me how this might be, but the C runtime library is now shipped as an integral port of a Windows distro, so there might be slight variations in that part...?

PR #499, opened 2021-01-23
Stats Class

mborland:
My cut at implementing the statistics class outlined in [P1708](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1708r2.pdf). Until meta-classes are added there is not an obvious (to me) way to offer more interfaces under `stats` instead of resorting to using `integer_stats`

### DIFF
--- a/doc/math.qbk
+++ b/doc/math.qbk
@@ -606,6 +606,7 @@ and as a CD ISBN 0-9504833-2-X  978-0-9504833-2-0, Classification 519.2-dc22.
 
 [mathpart statistics Statistics ]
 [include statistics/univariate_statistics.qbk]
+[include statistics/stats_set.qbk]
 [include statistics/bivariate_statistics.qbk]
 [include statistics/signal_statistics.qbk]
 [include statistics/anderson_darling.qbk]
--- a/doc/statistics/stats_set.qbk
+++ b/doc/statistics/stats_set.qbk
@@ -1,0 +1,104 @@
+[/
+Copyright (c) 2021 Matt Borland
+Use, modification and distribution are subject to the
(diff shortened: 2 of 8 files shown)


NAThompson:
@mborland : Is the goal to just implement the spec listed above? I'm just not sure what the benefit of this is over the stats in boost/accumulators and therefore I feel like this PR should maybe go directly into libstdc++.

mborland:
It is derived from the spec above since `univariate_statistics` has more functionality than the proposal. The earliest something similar would go into the language standard is 2023, so in the meantime it seems like a pretty ergonomic way to using the math stats library.

NAThompson:
> The earliest something similar would go into the language standard is 2023

It could go in an `#include <experimental/stats.hpp>` no?

mborland:
> > The earliest something similar would go into the language standard is 2023
> 
> It could go in an `#include <experimental/stats.hpp>` no?

It could, but aren't the authors still at the mercy of the LEWG/WG21 to push the proposal far enough forward to get to that point? I would say this is more akin to filesystem, type_traits, etc. where a similar or expanded version exists in boost first that can be transplanted into the STL when the time comes.

NAThompson:
I do worry that this will bloat our own library. The single-pass, parallel, numerically stable stats routines are a compelling feature, and I don't see how this really improves on that. (The reason for the bitset to specify which stats are wanted was to allow for computation of all desired stats in one pass, but this doesn't exploit that.)

I'll let @jzmaddock make a decision on this one, there's a case for it I admit, but it's really not something I'm interested in since it appears to be syntactic sugar on top of what we already have.

PR #502, opened 2021-01-26
Fix MSVC MINMAX CI Failures

mborland:
Many of these errors continue to stem from https://github.com/boostorg/random/pull/76 not being resolved.

### DIFF
--- a/test/test_carlson.hpp
+++ b/test/test_carlson.hpp
@@ -4,6 +4,7 @@
 //  Boost Software License, Version 1.0. (See accompanying file
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+#define NOMINMAX
 #include <boost/math/concepts/real_concept.hpp>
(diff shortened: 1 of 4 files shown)


mborland:
@jzmaddock This gets us back to green.

test/test_carlson.hpp
```diff
@@ -4,6 +4,15 @@
...
 
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#ifdef min
+#undef min
+#endif
+#ifdef max
+#undef max
+#endif
```
jzmaddock, line 15: Point of pedantry: isn't this superfluous now?

mborland: You are correct.

mborland:
@jzmaddock This is good with the superfluous macros removed. I will open a PR with random to hit all the minmax violations.

jzmaddock:
Thanks Matt!

PR #503, opened 2021-01-28
Bivariate Stats Policies

mborland:
Adds execution policies to bivariate statistics. Benchmarks:

```
------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
seq_covariance<double>/real_time_BigO           5.22 N          5.22 N    
seq_covariance<double>/real_time_RMS               1 %             1 % 
par_covariance<double>/real_time_BigO           0.75 N          0.28 N    
par_covariance<double>/real_time_RMS              45 %            94 %
 
seq_correlation<double>/real_time_BigO          3.22 N          3.22 N    
seq_correlation<double>/real_time_RMS              3 %             3 %
par_correlation<double>/real_time_BigO          0.65 N          0.27 N    
par_correlation<double>/real_time_RMS             42 %            91 %
```

### DIFF
--- a/include/boost/math/statistics/bivariate_statistics.hpp
+++ b/include/boost/math/statistics/bivariate_statistics.hpp
@@ -77,10 +77,10 @@ ReturnType means_and_covariance_parallel_impl(ForwardIterator u_begin, ForwardIt
     const unsigned max_concurrency = std::thread::hardware_concurrency() == 0 ? 2u : std::thread::hardware_concurrency();
     unsigned num_threads = 2u;
     
-    // TODO(mborland): Benchmark and replace 5.13N with results
-    // Threading is faster for: 10 + 5.13e-3 N/j <= 5.13e-3N => N >= 10^4j/5.13(j-1).
-    const auto parallel_lower_bound = 10e4*max_concurrency/(5.13*(max_concurrency-1));
-    const auto parallel_upper_bound = 10e4*2/5.13; // j = 2
+    // 5.16 comes from benchmarking. See boost/math/reporting/performance/bivariate_statistics_performance.cpp
+    // Threading is faster for: 10 + 5.16e-3 N/j <= 5.16e-3N => N >= 10^4j/5.16(j-1).
+    const auto parallel_lower_bound = 10e4*max_concurrency/(5.16*(max_concurrency-1));
+    const auto parallel_upper_bound = 10e4*2/5.16; // j = 2
 
     // https://lemire.me/blog/2020/01/30/cost-of-a-thread-in-c-under-linux/
     if(u_elements < parallel_lower_bound)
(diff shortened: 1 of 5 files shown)


include/boost/math/statistics/bivariate_statistics.hpp
```diff
@@ -10,72 +10,183 @@
...
+    {
+        return means_and_covariance_seq_impl<ReturnType>(u_begin, u_end, v_begin, v_end);
+    }
+    else if(u_elements >= parallel_upper_bound)
+    {
+        num_threads = max_concurrency;
+    }
+    else
+    {
+        for(unsigned i = 3; i < max_concurrency; ++i)
```
NAThompson, line 96: Can't you just calculate the `num_threads` directly instead of looping?

NAThompson:
@mborland : Really like these diffs.

Quick question: Could you run the parallel benchmarks under `perf record`, and create a [flamegraph](https://github.com/brendangregg/FlameGraph) and a timechart using [hotspot](https://github.com/KDAB/hotspot)? I'm interested in how effectively the threads are being utilitized?

mborland:
@NAThompson I am not sure if this is totally what you are looking for. 
Flamegraph
Timegraph

include/boost/math/statistics/bivariate_statistics.hpp
mborland: I can't think of a way besides the substitution used here to calculate the minimum number of threads.

NAThompson:
That does tell me what I need to know, namely that there aren't any particularly weird threading things consuming a bunch of time.

Could you screenshot the timeline at the bottom?

mborland:
Here is the longest sample. There are over 150k processes in the list.
Timeline_2

NAThompson:
@mborland : Wut! Why?

mborland:
I tried again and ran this reduced benchmark only:
```
BENCHMARK_TEMPLATE(par_covariance, double)->RangeMultiplier(2)->Range(1 << 20, 1 << 20)->Complexity(benchmark::oN)->UseRealTime();
```
The timeline shows 6072 processes, and a counter shows the covariance was calculated 506 times. 6072/506 = 12 which is the number of threads my machine has. Only the main thread has a complete timeline.

Reduced_Timeline

NAThompson:
@mborland : That makes sense, and the flamegraph does look good.

Does the Schubert paper have a better error estimate than Bennet's paper?

mborland:
Schubert is much more pragmatic than Bennet. While he does not offer an equation for error bounds like she does he offers a lot more data comparing a large number of implementations.

test/bivariate_statistics_test.cpp
```diff
@@ -38,7 +38,314 @@ using boost::multiprecision::cpp_complex_50;
 using  boost::math::statistics::means_and_covariance;
 using  boost::math::statistics::covariance;
 
-template<class Real>
+#if (__cplusplus > 201700 || _MSVC_LANG > 201700) && (__GNUC__ > 9 || (__clang_major__ > 9 && defined __GLIBCXX__)  || _MSC_VER > 1927)
+#include <execution>
```
NAThompson, line 42: Shouldn't this only need to be in the library code, not the user code?

mborland: This is here so that I could keep all the tests in one file. It is one giant if-else macro with a main for C++11&14 and a main for C++17&20. Normal users would not need this.

PR #504, opened 2021-01-29
Remove Appveyor

mborland:
Tests increasingly fail (C++03) leading to a red X for every CI run even if actions passes. All tests covered by Actions.


jzmaddock:
Yep, been meaning to do this...

PR #516, opened 2021-02-03
Fixes for test_std_lib CI failures in float128

mborland:
Fix for large number of CI failures in `test_std_lib` in float128 tests. [Example here](https://travis-ci.org/github/mborland/math/jobs/757277541#L873).


jzmaddock:
Good catch!

The using declarations are superfluous but harmless: as the non-member functions will be found via ADL.

mborland:
@jzmaddock This is clean. It's nice to see the green check mark for CI showing up again

PR #517, opened 2021-02-03
Fix for unable to find numeric literal operator Q

mborland:
Disable several tests in `test_cstdfloat` if GNU extensions are not enabled on GCC. Several issues across other repos in github that point back to this. Example in CI [here](https://travis-ci.org/github/mborland/math/jobs/757277544#L1991).


test/test_cstdfloat.cpp
```diff
@@ -304,6 +304,8 @@ namespace test_cstdfloat
       }
       s = 8 * std::numeric_limits<T>::epsilon();
       val = 2.5;
+
+      #ifndef __STRICT_ANSI__
```
jzmaddock, line 308: We have a macro `BOOST_MATH_LARGEST_FLOAT_C(val)` declared in boost/math/tools/big_constant.hpp which declares `val` as the largest supported literal.

But there's another issue here, the function test() appears to be never called?  @ckormanyos ?

ckormanyos:
> the function test() appears to be never called? 

It appears as though when `BOOST_FLOAT128_C` is defined, then `test_cstdfloat::extend_check_128_func()` near the bottom of the main test routine should subsequently call `test<boost::float128_t>()`.

If memory serves, then `BOOST_FLOAT128_C` is automatically set when the compiler can detect the presence of an existing 128-bit buit in floating-point type such as `__float128`.

We wrote this code, however, quite a long time ago, so it would not really surprise me if the logic on some of those compiler switches needs to be adusted or re-adjusted. I can try a few dedicated compilation runs here at my workstation in a standalone project and report findings.

ckormanyos:
> It appears as though when `BOOST_FLOAT128_C` is defined, ...

Or is the use of `BOOST_FLOAT128_C` not optimal for reaching Boolean decisions on preprocessor level? Years ago we had some dialog regarding which preprocessor definition to use for activating/deactivating the quad-math stuff and i think we decided for that one.

mborland:
This does resolve the errors. New run with this commit [here](https://travis-ci.org/github/mborland/math/jobs/757568362).

PR #521, opened 2021-02-06
Remove use of MPL in fraction

mborland:
Now that C++11 will be the minimum requirement the remaining uses of boost::mpl can be removed. Additional minor changes to use the C++11 standards and style.


jzmaddock:
Thanks Matt, looks fine.

mborland:
@jzmaddock CI is clean. Should be good to go.

PR #523, opened 2021-02-07
Remove MPL from constants

mborland:
Remove MPL, and other C++11 changes.


jzmaddock:
We probably should remove use of boost::integral_constant, but it will need a coordinated change inside Multiprecision.

mborland:
I will open a PR in multiprecision to make the change, and put `std::integral_constant` back in here.

mborland:
Both this and https://github.com/boostorg/multiprecision/pull/294 replace `boost::integral_constant` with `std::integral_constant` now.

jzmaddock:
Merged, both this and Multiprecision will be shown as failing until the boost super-project updates.

PR #526, opened 2021-02-09
Remove MPL from lambert_w

mborland:
Remove MPL, and make other C++11 changes (e.g. boost::integral_constant to std::integral_constant).

### DIFF
--- a/include/boost/math/special_functions/lambert_w.hpp
+++ b/include/boost/math/special_functions/lambert_w.hpp
@@ -68,7 +68,6 @@ BOOST_MATH_INSTRUMENT_LAMBERT_W_SMALL_Z_SERIES_ITERATIONS  // Show evaluation of
 #include <limits>
 #include <exception>
 #include <type_traits>
-#include <string>
 #include <cstdint>
 
 // Needed for testing and diagnostics only.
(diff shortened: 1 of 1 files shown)


include/boost/math/special_functions/lambert_w.hpp
```diff
@@ -1223,19 +1215,19 @@ T lambert_w_negative_rational_float(T z, const Policy& pol)
 }
 
 //! Lambert_w0 @b 'float' implementation, selected when T is 32-bit precision.
-template <class T, class Policy>
-inline T lambert_w0_imp(T z, const Policy& pol, const boost::integral_constant<int, 1>&)
+template <typename T, typename Policy>
+inline T lambert_w0_imp(T z, const Policy& pol, const std::integral_constant<int, 1>&)
 {
-  static const char* function = "boost::math::lambert_w0<%1%>"; // For error messages.
+  static std::string function = "boost::math::lambert_w0<%1%>"; // For error messages.
```
jzmaddock, line 1221: What's the motivation here?  A string literal avoids the (tiny) runtime overhead of initializing a std::string which we don't use (as that type) anyway?

mborland: Mostly to remove non-smart pointers that are floating around as part of the modernization campaign. If there is a `BOOST_CXX20_CONSTEXPR` macro it could become `constexpr`.

NAThompson:
So will this put us on a path so that boost.math can be consumed in isolation from the rest of boost?

jzmaddock:
>So will this put us on a path so that boost.math can be consumed in isolation from the rest of boost?

Yes.  We would probably need a separate "standalone" mode that disables some features, but yes :)

include/boost/math/special_functions/lambert_w.hpp
mborland: I reverted it to `const char*`; I would say that is more idiomatic than a C style array.

PR #527, opened 2021-02-09
Replace integral_constant in calculate_constants

mborland:


mborland:
@jzmaddock This is clean if it looks good to you.

PR #534, opened 2021-02-13
Remove MPL and type_traits from promotion

mborland:


mborland:
@jzmaddock All of these MPL based PRs should be good to go. Once they are all in I can make the final push to remove MPL use from the policies.

include/boost/math/tools/promotion.hpp
```diff
@@ -101,51 +91,51 @@ namespace boost
...
             >::type
-         >::type type;
+         >::type;
 
 #ifdef BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
          //
          // Guard against use of long double if it's not supported:
          //
-         BOOST_STATIC_ASSERT_MSG((0 == ::boost::is_same<type, long double>::value), "Sorry, but this platform does not have sufficient long double support for the special functions to be reliably implemented.");
+         BOOST_STATIC_ASSERT_MSG((0 == std::is_same<type, long double>::value), "Sorry, but this platform does not have sufficient long double support for the special functions to be reliably implemented.");
```
jzmaddock, line 138: These BOOST_STATIC_ASSERT's could be real `static_assert`'s now, but we can do that later ;)

PR #537, opened 2021-02-14
Remove MPL from traits

mborland:


mborland:
@jzmaddock This one should be good to go.

PR #540, opened 2021-02-15
Remove MPL from policies

mborland:
Also removes dependencies from boost types (e.g. boost/assert and boost/type_traits) that were implemented in C++11.


include/boost/math/policies/policy.hpp
```diff
@@ -837,54 +834,54 @@ struct precision
...
 
 #endif
 
 namespace detail{
 
 template <class T, class Policy>
-inline BOOST_MATH_CONSTEXPR int digits_imp(boost::true_type const&) BOOST_NOEXCEPT
+inline constexpr int digits_imp(std::true_type const&) noexcept
 {
 #ifndef BOOST_NO_LIMITS_COMPILE_TIME_CONSTANTS
```
jzmaddock, line 847: Just a heads up that BOOST_NO_LIMITS_COMPILE_TIME_CONSTANTS can be assumed to NOT be set post C++11, so these pp-branches can be chopped down.

mborland:
@jzmaddock The current failures are all related to multiprecision gmp and mpfr related using older versions of clang (5-8) on Ubuntu. They are all similar to:

```
In file included from ../tools/bessel_derivative_data.cpp:14:
In file included from ../../../boost/multiprecision/mpfr.hpp:11:
../../../boost/multiprecision/gmp.hpp:291:11: error: no template named '_mp_size'; did you mean 'mp11::mp_size'?
      if (mpf_sgn(m_data) == 0)
          ^
/usr/include/x86_64-linux-gnu/gmp.h:2239:26: note: expanded from macro 'mpf_sgn'
#define mpf_sgn(F) ((F)->_mp_size < 0 ? -1 : (F)->_mp_size > 0)
                         ^
../../../boost/mp11/list.hpp:45:19: note: 'mp11::mp_size' declared here
67
template<class L> using mp_size = typename detail::mp_size_impl<L>::type;
```

Is this a known issue? I don't expect naming collision because the majority of tests pass.

jzmaddock:
Someone else has experienced this: https://stackoverflow.com/questions/57857572/why-does-boost-log-break-boost-gmp-multiprecision, which leads on to this clang bug: https://bugs.llvm.org/show_bug.cgi?id=43266

Not quite sure what we do about though :(

mborland:
I guess I should have looked at SO before asking you. Good news is now that I replaced MP11 with internal tools the error seems to have gone away.

jzmaddock:
I should have said this before, but the purpose of the main meta-program in policies.hpp is to take a random policy and normalise it in some way so that if a function is instantiated with 2 different policy types, that actually specify the same thing, then internally we call the implementation level functions with the same policy type (ie code bloat suppression).  

The point is, it doesn't really matter how that is achieved, or what the normalised form is, as long as there is one.

Currently we:

* Remove any policy that is the same as the default, and then
* Sort whatever is left.

So there may well be a simpler/neater way to do this without re-implementing the MPL-like machinery.

mborland:
The structure of the program made sense, and has been around for quite some time so I did not change a whole lot in that respect. Not all of the stuff in the new meta-programming tools header is needed so I can pare it back to the minimum if you want. I thought it would be good to have to avoid re-introducing MPL or MP11 down the road. I believe this is very close to going green. These last few commits have been odd failures on old compilers.

mborland:
This is finally green. Pending any further review it should be good to go.

mborland:
@jzmaddock Any changes required on this one?

jzmaddock:
There's a couple of sanity checks I want to do on the new policy code, but I'll merge, and file a bug report if I find anything.

Once again, many thanks for this!

PR #541, opened 2021-02-15
Re-vitalize e_float (2021) bindings for c++11

ckormanyos:
This draft PR exposes the re-vitalization of math bindings for `e_float` (2021) upgraded for modern Boost multiprecision and C++11.


ckormanyos:
There is a C++11 draft of the [new `e_float` bindings](https://github.com/boostorg/math/blob/e_float_2021/include/boost/math/bindings/e_float.hpp) available.

Some preliminary tests running now on GCC and VC are [here](https://github.com/ckormanyos/e_float-2021/blob/main/libs/e_float/test_boost/test_boost.cpp). These tests exercise non-trivial calculations using `tgamma()` from math and a generalized Legendre function on the real axis in the unit disk. Tests to be extended include real concept (TBD).

A plan for testing in CI still needs to be worked out. The `e_float` library requirec either pre-compilation of a lib or compilation on the spot of about 5 cpp files.

ckormanyos:
> Tests to be extended include real concept (TBD).

So I've refined the `e_float` bindings to the extent where it's time to get something running on CI. There are several math tests and real-concept tests that I have run successfully on a few compilers locally.

One of the archtectural questions is... How should I integrate the tests of the bindings in math whrer these tests need a few cpp files to either be built on the spot or pulled from a pre-built LIB?.

Some ideas include:
- Checkout the e_float-2021 repo, build the selected files together with math test code and run. I would not at the moment know exactly how to do each step of this in bjam right now, but that might be feasible.
- Roll some pre-built libs and publish tem in the e_float-2021 repo to be gotten for the math tests.
- Do no testing of the e_float bindings whatsoever and leave testing up to me in my own repo.

Any thoughts on how to get this stuff tested?

ckormanyos:
> In the CI of math, checkout the e_float-2021 repo, build the selected files together with math test code and run.

A prototype of this added CI for testing `e_float-2021` bindings is now running in CI in this draft PR.

At the moment, the jobs in the matrix run the _efx_ (native decimal base-10) and _gmp_ floating point backends on C++11,14,17,2a on g++ and clang++ on Ubuntu latest. The build commands are expressed in bash commands on the command line without the use of shell scripts and not explicitly using b2.

This prototypes this general kind of CI, whereby more consistent use of b2 or shell scripts might be desired.

mborland:
[This is the current list](https://github.com/mborland/math/issues/2) of dependencies in Boost.Math once PR #540 gets merged. I am actively stripping out dependencies so that Math can come closer to being consumed in isolation like Boost.JSON. The only current dependency on Boost.Multiprecision is a tool I added for benchmarking that can be removed since I am the only user. I think adding bindings for Multiprecision would be useful if they exist behind a guard that can disable them and the associated dependency if the end user doesn't need or want them. My failed attempt at implementing a prime sieve would have offered explicit bindings for multiprecision that would not have been hard to hide.

ckormanyos:
> I am actively stripping out dependencies so that Math can come closer to being consumed in isolation like Boost.JSON. The only current dependency on Boost.Multiprecision is a tool I added for benchmarking that can be removed since I am the only user.

That is great. Nonetheless, this is not exactly what I mean. This draft PR actually needs to build parts of the `e_float-2021` repo and link them with boost test code, as this legacy project is not header-only and not part of Boost. Thus it is so unusual to actually clone a third party repo (even if it is my own) for some Boost testing.

My questions mean:

Should we either:
- Support `e_float`bindings.
- Get something sensible running on CI.
- Update any docs needed like [here](https://www.boost.org/doc/libs/1_75_0/libs/math/doc/html/math_toolkit/high_precision/e_float.html).

Or..
- Eliminate `e_float` bindings from code headers, docs and CI

This PR goes for the former. I could live with either way, but having broken `e_float`bindings pointing to a non-supported project is a state I*d recommend resolving.

As to your dependency list, it's great! In fact, I could imagine making a little GHA that clones non-recursively, updates only those submodules needed, bootstraps, runs `b2 headers` and gets you ready to run CI using Boost.Math and Boost.Multiprecision. I just did these steps manually (probably including a few too many submodules, bot OK for now).

mborland:
That is my mistake. I agree that having out of tree dependencies is far from ideal especially since it is not something I would be able to get through my system package manager like FFTW3. I would argue for removing the bindings here and calling it a day. If the repo the current bindings point to is unsupported I would assume there are not many users unless you are seeing issues and PRs against that repo.

ckormanyos:
> having out of tree dependencies is far from ideal

I suggest the following compromise.

- The new and supported [`e_float-2021` repo](https://github.com/ckormanyos/e_float-2021) supports Boost.Multiprecision bindings for `e_float`.
- This newly supported Repo also performc CI steps with Boost.Math and Boost.Multiprecision develop branches [here](https://github.com/ckormanyos/e_float-2021/blob/main/.github/workflows/e_float_boost_test.yml)
- Manage `e_float`bindings in my own repo, do CI in the same.
- Eliminate `e_float` bindings from Boost.Math and modify the Boost.Math docs accordingly to mention them in the `e_float`repo.

Agreed?

mborland:
Agreed; that plan makes sense to me.

ckormanyos:
> Agreed; that plan makes sense to me.

OK. Good. Then that's what we will do. I can handle this in this PR.

ckormanyos:
This recent push is intended to eliminate `e_float` explicit bindings and tests from Boost.Math.
These will be handled in the [`e_float-2021` repository](https://github.com/ckormanyos/e_float-2021).

ckormanyos:
This is green now.

In this PR, the `e_float` bindings (which are also now a full-fledged Multiprecision backend) are supported solely in the [`e_float-2021` repo](https://github.com/ckormanyos/e_float-2021).

The docs in this checkin attempt to clarify the `e_float` bindings. These docs could be re-clafified or the small section on `e_float` could harmlessly be excluded altogether, depending on style or choice.

Since the `e_float` bindings are actually now a Multiprecision backend, a note of them might be helpful in the multiprecision docs to assist others when trying to fulfill the Multiprecision backend requirements?

Don't know if @jzmaddock has followed this discussion, but do you have any comments?

ckormanyos:
The docs regarding `e_float` bindings have been further reduced in scope and depth.

I would like to finish this issue and this PR and would like to merge after drone cycles.

Are there any rejections or futher comments or suggestions for improving technical content or clarity?

ckormanyos:
I've re-opened this one, as I forgot to merge develop prior to merging the PR.
Corrections are cycling in CI.

Issue #535, opened 2021-02-13
linear_regression_test fails at 128-bit real precisiokn


jzmaddock:
@NAThompson : I'm seeing:

```
Error at libs/math/test/linear_regression_test.cpp:test_permutation_invariance:167:
  Mollified relative error in __float128 precision is 0.00238, which exceeds 0.002, error/tol  = 1.19.
  Expected: -7.200000000000000177635683940025046468 = -0x1.ccccccccccccdp+2
  Computed: -7.217158929845008151987144669539711967 = -0x1.cde5ee916dfea99579fa451efbfep+2
  ULP distance: +22273564382513600491790226357246
Error at libs/math/test/linear_regression_test.cpp:test_scaling_relations:211:
  Mollified relative error in __float128 precision is 0.00536, which exceeds 0.005, error/tol  = 1.07.
  Expected: +3.200000000000000177635683940025046468 = +0x1.999999999999ap+1
  Computed: +3.182841070154992203284223210510388672 = +0x1.976756105735ecd50c0b75c20818p+1
  ULP distance: +44547128765027200983580452714472
Error count: 2
```

from linear_regression_test.cpp on Linux/s390x.

Are the existing error bounds arbitrary?  They only just fail, so it's temping to just them a bit.

jzmaddock:
A related issue for the whittaker shannon test:

```
Error at libs/math/test/whittaker_shannon_test.cpp:test_bump:117:
  Mollified relative error in __float128 precision is 2.05e-29, which exceeds 1.93e-33, error/tol  = 1.07e+04.
  Expected: +0.360803088275448657029385009164666528 = +0x1.71765d61c162afbf43d7d05ea90ap-2
  Computed: +0.360803088275448657029385009185199204 = +0x1.71765d61c162afbf43d7d0652ad9p-2
  ULP distance: +426447
  Problem occurred at abscissa 0.138032129403380054031217804789231206
```

and lots more the same.

NAThompson:
@jzmaddock : Wowsers the linear regression test looks bad. The Whittaker-Shannon also looks like it's beyond reasonable "fudge the tolerance bounds." Will try to look into this soon.

mborland:
Both of those tests call `std::normal_distribution` which is supposed to only have defined behavior for float, double and long double. On architectures like s390x that substitute __float128 for long double does this become a mixed precision/UB error? If you try to add a __float128 test on x86_64 you get the following error:

```
[...]

PR #543, opened 2021-02-20
Enable c++11,14 use on compilers with <execution>

mborland:
BOOST_NO_CXX17_HDR_EXECUTION fails as a guard if using std=c++11 or c++14 on a compiler that has `<execution>` e.g. GCC10 (default is c++14). Found while poking around on issue #535. @jzmaddock I know we have been going back and forth on this logic, but I think this addition gets it right.

### DIFF
--- a/include/boost/math/statistics/bivariate_statistics.hpp
+++ b/include/boost/math/statistics/bivariate_statistics.hpp
@@ -17,11 +17,10 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
-#include <boost/assert.hpp>
 
 // Support compilers with P0024R2 implemented without linking TBB
 // https://en.cppreference.com/w/cpp/compiler_support
-#ifndef BOOST_NO_CXX17_HDR_EXECUTION
+#if !defined(BOOST_NO_CXX17_HDR_EXECUTION) && (__cplusplus > 201700L || _MSVC_LANG > 201700L)
 #include <execution>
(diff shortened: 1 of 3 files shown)


include/boost/math/statistics/bivariate_statistics.hpp
```diff
@@ -17,11 +17,10 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
-#include <boost/assert.hpp>
 
 // Support compilers with P0024R2 implemented without linking TBB
 // https://en.cppreference.com/w/cpp/compiler_support
-#ifndef BOOST_NO_CXX17_HDR_EXECUTION
+#if !defined(BOOST_NO_CXX17_HDR_EXECUTION) && (__cplusplus > 201700L || _MSVC_LANG > 201700L)
```
jzmaddock, line 23: This is Boost.Config's responsibility to get right, plus this should already be handled here: https://github.com/boostorg/config/blob/1c45d765f107e0fcfc45a075c55bd66780e98188/include/boost/config/detail/suffix.hpp#L1096 

What am I missing?

mborland: That looks correct. Is that new to the develop branch? I link against release 1.75 and it only has optional, string_view, and variant.

jzmaddock: Yes develop.

PR #545, opened 2021-02-21
Change used exponent/scaling type from int to long long in the hyperg…

jzmaddock:
…eometric functions.

Allows use with type which have a greatly extended exponent range.
Also changes all itrunc's to lltrunc.


jzmaddock:
Failed checks were connection timeouts, so merging...

PR #547, opened 2021-02-21
2D Akima Interpolation

mborland:
As requested by @NAThompson in the wishlist. Currently just the interface shell so it will be a draft for a bit. It looked to me like this could get computationally expensive, but would be relatively easy to thread. Being able to design for threading up front is significantly easier than going back and adding it in later.


NAThompson:
@mborland : Highly recommend *against* threading; these interpolators are often called from a multithreaded context; e.g., try to render a surface z = f(x,y) where f(x,y) is this interpolator.

And thanks! I'm super excited for this.

NAThompson:
Here's the interface I had imagined for the (uniform grid) 2D Akima interpolation:

```cpp
#ifndef BOOST_MATH_INTERPOLATORS_AKIMA_2D_UNIFORM_HPP
#define BOOST_MATH_INTERPOLATORS_AKIMA_2D_UNIFORM_HPP

namespace boost::math::interpolators {

template<typename Matrix>
class akima_2d_uniform {
public:
    using Real = Matrix::value_type;
    // The data is expected to be listed in row-major order, so that data(0,0) corresponds to (some decision here)
    akima_2d_uniform(Real origin_x, Real origin_y, Real dx, Real dy, Matrix&& data);

    Real operator()(Real x, Real y) const;

    Real gradient(Real x, Real y) const;

    Real laplacian(Real x, Real y) const;
};

}
#endif
```

Ostensibly the `Matrix` type should be compatible with both Eigen and Boost.ublas.

@jzmaddock: Is that how you would do it?

mborland:
Not adding threading support in the first place is easy if it is unnecessary. [If you scroll down to appendix B](https://www.its.bldrdoc.gov/DownloadPublications.ashx?id=OT-75-70.pdf) you can see the authors original interface in FORTRAN. It is not too far off of what you are suggesting.

NAThompson:
> Not adding threading support in the first place is easy if it is unnecessary.

I think threading the constructor would be a good idea if you had to assemble a KD-tree. But if you take on the uniform grid case first, there should be no issue.

include/boost/math/interpolators/detail/2d_akima_detail.hpp
```diff
@@ -0,0 +1,159 @@
...
+#include <cmath>
+
+namespace boost { namespace math { namespace interpolators { namespace detail {
+
+template <typename Matrix, typename VectorType = typename Matrix::value_type, typename Real = typename VectorType::value_type>
+class akima_2d_uniform_impl
+{
+private:    
+    const Matrix data_;
+    const std::size_t n_interpolation_points_;
```
NAThompson, line 27: Wait, do I have to decide beforehand how many points I want to interpolate at?

include/boost/math/interpolators/detail/2d_akima_detail.hpp
```diff
@@ -0,0 +1,159 @@
...
+#include <stdexcept>
+#include <utility>
+#include <algorithm>
+#include <iostream>
+#include <cstddef>
+#include <cmath>
+
+namespace boost { namespace math { namespace interpolators { namespace detail {
+
+template <typename Matrix, typename VectorType = typename Matrix::value_type, typename Real = typename VectorType::value_type>
```
NAThompson, line 22: The `Matrix::value_type` for Eigen matrices will be (say) `float` or `double`, not a vector.

include/boost/math/interpolators/detail/2d_akima_detail.hpp
```diff
@@ -0,0 +1,159 @@
...
+    const std::size_t n_interpolation_points_;
+    VectorType x_interpolation_points_;
+    VectorType y_interpolation_points_;
+    VectorType z_interpolation_points_;
+
+    void validate_inputs() const;
+    VectorType apply_uniform_grid(const VectorType& vals);
+
+public:
+    akima_2d_uniform_impl(Matrix&& data, std::size_t n) : data_{std::move(data)}, n_interpolation_points_{n}
```
NAThompson, line 36: So you can only do nxn data? I expect the most common case is in fact an nxm.

test/test_2d_akima.cpp
```diff
@@ -0,0 +1,25 @@
...
+    std::vector<double> x {11.16, 24.20, 19.85, 10.35, 19.72,  0.00, 20.87, 19.99, 10.28,  4.51,  0.00, 16.70,  6.08, 25.00, 14.90,  
+                            0.00,  9.66,  5.22, 11.77, 15.10, 25.00, 25.00, 14.59, 15.20,  5.23,  2.14,  0.51, 25.00, 21.67,  3.31};
+    
+    std::vector<double> y { 1.24, 16.23, 10.72,  4.11,  1.39, 20.00, 20.00,  4.62, 15.16, 20.00,  4.48, 19.65,  4.58, 11.87,  3.12,
+                            0.00, 20.00, 14.66, 10.47, 17.19,  3.87,  0.00,  8.71,  0.00, 10.72, 15.03,  8.37, 20.00, 14.36,  0.13};
+    
+    std::vector<double> z {22.15,  2.83,  7.97, 22.33, 16.83, 34.60,  5.74, 14.72, 21.59, 15.61, 61.77,  6.31, 35.74,  4.40, 21.70,
+                           58.20,  4.73, 40.36, 13.62, 12.57,  8.74, 12.00, 14.81, 21.60, 26.50, 53.10, 49.43,  0.60,  5.52, 44.08};
+
+    std::vector<std::vector<double>> xyz {x, y, z};
```
NAThompson, line 22: A vector of vectors is a pretty bad data structure in general; why not an `Eigen` matrix?

include/boost/math/interpolators/detail/2d_akima_detail.hpp
mborland: That was how the author's implementation works, but honestly I have not made it far enough to know if it's a hard requirement or not. I am implementing the steps of the procedure as I go down the paper as non-member functions so there is flexibility in interface. Same generally goes for your other comments.

NAThompson: Yeah, that's a restriction of Fortran, not of this particular method. So your goal is to first do a faithful tranlation of the paper and the C++ it? (If so I'll shut up for a while.)

mborland: He details a step by step procedure using equations that I am using to implement this.  There is a not so great scan copy of his code that is written in FORTRAN 66 which I would prefer not to translate. I am sure there are efficiencies that can be gained from the last 50 years.

test/test_2d_akima.cpp
mborland: Is it worth supporting a vector of vectors or array of arrays? With some abuse of `std::conditional` it can happen but I would need to build an adapter that would transverse an iterator across the vectors to mimic the behavior of say ublas iterators.

NAThompson: I would say it is not worth supporting vector of vector. Array of arrays is more debatable because the storage is contiguous, but the syntax will be incompatible with all matrix classes, so I'd say don't try supporting array of arrays.

jzmaddock: Just lurking here.... but the thing to do is decide on a matrix *concept*, and then use traits classes etc to map disparate concrete types (uBlas, Eigen, contain of container etc) to something that your code can understand.

Back to lurking...

NAThompson: @jzmaddock : Please lurk more!

But are you saying both a vector of vectors and an Eigen matrix can work at the same time?

jzmaddock: In the sense that one should be capable of being mapped to the other via an abstract interface/traits class.  I wouldn't see it as necessary to support *everything* from the get go though.  We should also try to have one eye on whatever might come along in the future, I'm somewhat surprised that there's still no std lib matrix/array support.

NAThompson: The matrix has been a sore point since you can't do:

```
operator[](int64_t i, int64_t j) const;
```

and write `M[i,j]`. People I guess want this before the matrix class can become a real boy.

PR #552, opened 2021-02-23
Newton-Raphson instrumented message improvements

NAThompson:
The Newton-Raphson method has an option to log convergence via `BOOST_MATH_INSTRUMENT`. However, too much information was give about some things which were not super helpful, and too little about residuals.

In addition, remove buffer flushes as they slow down the process dramatically.

### DIFF
--- a/include/boost/math/tools/roots.hpp
+++ b/include/boost/math/tools/roots.hpp
@@ -10,29 +10,16 @@
 #pragma once
 #endif
 
-#include <iostream>
 #include <utility>
 #include <cmath>
-#include <stdexcept>
+#include <tuple>
 
 #include <boost/math/tools/config.hpp>
 #include <boost/cstdint.hpp>
-#include <boost/assert.hpp>
-#include <boost/throw_exception.hpp>
 #include <boost/math/tools/cxx03_warn.hpp>
-
-#ifdef BOOST_MSVC
-#pragma warning(push)
-#pragma warning(disable: 4512)
-#endif
-#include <boost/math/tools/tuple.hpp>
-#ifdef BOOST_MSVC
-#pragma warning(pop)
-#endif
 
 #include <boost/math/special_functions/sign.hpp>
 #include <boost/math/special_functions/next.hpp>
-#include <boost/math/tools/toms748_solve.hpp>
 #include <boost/math/policies/error_handling.hpp>
 
 namespace boost {
--- a/include/boost/math/tools/roots.hpp
+++ b/include/boost/math/tools/roots.hpp
@@ -9,6 +9,7 @@
 #ifdef _MSC_VER
 #pragma once
 #endif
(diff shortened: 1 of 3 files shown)


NAThompson:
@jzmaddock : I believe that all the boost files not in boost.math in roots.hpp are now unnecessary. Is this correct?

include/boost/math/tools/roots.hpp
```diff
@@ -9,11 +9,10 @@
 #ifdef _MSC_VER
 #pragma once
 #endif
-#include <boost/math/tools/complex.hpp> // test for multiprecision types.
 
 #include <iostream>
 #include <utility>
-#include <boost/config/no_tr1/cmath.hpp>
+#include <cmath.hpp>
```
jzmaddock, line 15: Should be <cmath> ?

NAThompson: omg. . .

jzmaddock:
>  I believe that all the boost files not in boost.math in roots.hpp are now unnecessary. Is this correct?

Looks like it, yes.

iotream can go (unless BOOST_MATH_INSTRUMENT is set, but really this should be handled by boost/math/tools/config.hpp already.
stdexcept can go (handled by policies/error_handling.hpp)

I'm also not sure what #include <boost/math/tools/toms748_solve.hpp> is doing in there!

NAThompson:
What about the tuples? They are part of C++11 now . ..

jzmaddock:
Yes, replacing `boost/math/tools/tuple.hpp` with `<tuple>` is on the list to do.  Note that the former is header just includes `<tuple>` anyway in C++11 and later.

jzmaddock:
Actually... we probably don't need to include `<tuple>` at all, all we use is the concept of a tuple, not the actual type?

NAThompson:
The boost::uint32_t might also be able to go; is there a way to remove it in a backwards compatible way?

jzmaddock:
I only see uintmax_t - that should be a straight replacement boost::uintmax_t => std::uintmax_t.   Hmmm, but those might be different (but same width) types?  And we should make that replacement library wide in one go probably.

NAThompson:
> I'm also not sure what #include <boost/math/tools/toms748_solve.hpp> is doing in there!

It defines a class `eps_tolerance`:

```
$ grep -nr 'eps_tolerance' ../include/boost/math/tools
../include/boost/math/tools/toms748_solve.hpp:31:class eps_tolerance
../include/boost/math/tools/toms748_solve.hpp:34:   eps_tolerance()
../include/boost/math/tools/toms748_solve.hpp:38:   eps_tolerance(unsigned bits)
```

NAThompson:
It looks like the [documentation](https://github.com/boostorg/math/blob/develop/doc/roots/roots_without_derivatives.qbk) lists it under the `root.hpp` header.

Issue #549, opened 2021-02-21
Can we make repository adding more robust?


jzmaddock:
The answer to this may well be "no", but I'm seeing lots of CI failures due to:

```
gpg: keyring `/tmp/tmphatqqn4x/secring.gpg' created
gpg: keyring `/tmp/tmphatqqn4x/pubring.gpg' created
gpg: requesting key BA9EF27F from hkp server keyserver.ubuntu.com
Error: retrieving gpg key timed out.
gpg: keyserver timed out
gpg: keyserver receive failed: keyserver error
Error: Process completed with exit code 1.
```

This PR has 15 jobs failed with this: https://github.com/boostorg/math/pull/545 and it's a pain to check through them all.  

Is there any way we can auto-retry on failure a few times?

mborland:
This looks to be working as intended now

![Retry Repo](https://user-images.githubusercontent.com/3745830/109188370-7e04c000-7758-11eb-8d70-856e5015faa3.png)

jzmaddock:
Good spot!  Closing down, thanks again for this.

PR #553, opened 2021-02-23
Retry adding repos

mborland:
From https://github.com/boostorg/math/issues/549. There is currently [no official way](https://github.community/t/retry-for-failed-steps/17136) to retry steps so this adds extra steps that only run on failure.


jzmaddock:
That's a neat solution - thanks for this Matt - hopefully we can cut down the churn a bit!

mborland:
@jzmaddock This is green, but it does not look like any of the retry steps ran. At least we know that skipping those steps on success works.

jzmaddock:
OK, lets merge and see!  Thanks for this.

ckormanyos:
This is cool, ... I was having difficulty moulding and contorting my mindset to _didn't run is the new green_.

PR #555, opened 2021-02-24
Deprecate boost::tuple and various tuple workarounds.

NAThompson:


jzmaddock:
I realise it's a bigger change, but shouldn't we remove tools/tuple.hpp altogether and use std::tuples directly?

NAThompson:
@jzmaddock : Yes, I was wanting to do that, however, I didn't want to take that one in one bite because I figured I'd get myself into trouble.

pabristow:
As Ryan Air CEO might say, isn't removing /tools/tuple.hpp also "unnecessarily pissing-off the customers?"  Won't it break existing customers working code and force them to update to use std::tuple?

(Nothing to stop using replace all Boost.Math's and Boost.Multiprecision's use of math's own tuple with std::tuple, of course).

PR #557, opened 2021-02-25
Remove multiprecision dependency

mborland:
Removes the only dependency on Boost.Multiprecision. This has not gone out in a release, and the only place it is called is a benchmark I wrote.

### DIFF
--- a/include/boost/math/tools/random_vector.hpp
+++ b/include/boost/math/tools/random_vector.hpp
@@ -8,11 +8,6 @@
 #include <random>
 #include <type_traits>
 #include <cstddef>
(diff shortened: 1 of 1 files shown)


NAThompson:
Looks a good diff to me.

Hopefully we can remove all dependencies on the rest of boost and then make a boost.math package which is consumable from (say) a github download page.

mborland:
That is the goal. [Here is the list](https://github.com/mborland/math/issues/2) of dependencies that I pulled a couple of days ago. There are a fair amount of easy fixes.

jzmaddock:
Fine by me.

mborland:
@jzmaddock This one is green.

PR #559, opened 2021-02-25
Remove MP11 dependency

mborland:
Removes the only MP11 dependency. Does not need to make it in for release.

### DIFF
--- a/include/boost/math/tools/mp.hpp
+++ b/include/boost/math/tools/mp.hpp
@@ -337,7 +337,7 @@ using mp_remove_if_q = mp_remove_if<L, Q::template fn>;
 
 // Index sequence
 // Use C++14 index sequence if available
-#ifndef BOOST_NO_CXX14_VARIABLE_TEMPLATES
+#if defined(__cpp_lib_integer_sequence) && (__cpp_lib_integer_sequence >= 201304)
 #include <utility>
 template<std::size_t... I>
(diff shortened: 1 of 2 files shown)


include/boost/math/tools/mp.hpp
```diff
@@ -335,6 +335,108 @@ using mp_remove_if = typename detail::mp_remove_if_impl<L, P>::type;
 template<typename L, typename Q> 
 using mp_remove_if_q = mp_remove_if<L, Q::template fn>;
 
+// Index sequence
+// Use C++14 index sequence if available
+#ifndef BOOST_NO_CXX14_VARIABLE_TEMPLATES
+#include <utility>
+template<std::size_t... I>
+using index_sequence = std::index_sequence<I...>;
```
mborland, line 343: @jzmaddock `std::index_sequence` is a C++14 feature, but I didn't see anything in config that specifically called it out. Is there something better to put here?

jzmaddock: Try `defined(__cpp_lib_integer_sequence) && (__cpp_lib_integer_sequence >= 201304)`

mborland: That does the trick locally, hopefully CI agrees. Thanks.

mborland:
@jzmaddock This should be good to go.

jwakely:
This is badly broken, see #670

PR #561, opened 2021-02-27
Remove boost.algorithm dependency

mborland:
@NAThompson

### DIFF
--- a/include/boost/math/tools/ulps_plot.hpp
+++ b/include/boost/math/tools/ulps_plot.hpp
@@ -15,9 +15,9 @@
 #include <list>
 #include <random>
 #include <stdexcept>
+#include <filesystem>
 #include <boost/math/tools/condition_numbers.hpp>
 #include <boost/random/uniform_real_distribution.hpp>
-#include <boost/algorithm/string/predicate.hpp>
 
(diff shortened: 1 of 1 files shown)


NAThompson:
@mborland : thanks looks good!

mborland:
@NAThompson I am not sure why this feature took till C++20 to make it in, but oh well. CI is clean.

NAThompson:
@mborland : Do you think this should go into an anonymous namespace so we don't break anyone else who has hand-coded and `endswith`?

jzmaddock:
@NAThompson A "detail" namespace would be better than an anonymous one: anonymous namespaces don't actually hide anything, just change the linkage.

mborland:
@jzmaddock Exactly my thought

mborland:
@NAThompson I added the detail namespace. Do you think it is worthwhile to bring this version of `ends_with` up into a higher level header in tools (something like adding a boost/math/tools/algorithms.hpp) or fine here since its the only use?

NAThompson:
@mborland : Not right now. I believe this is the only thing that could use it. @jzmaddock can overrule at his leisure.

PR #562, opened 2021-02-27
Remove boost fenv dependency

mborland:
I suspect a lot more of this can be pared back, but lack the knowledge and numerous platforms to test everything.

### DIFF
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -249,20 +249,9 @@
 //
 // noexcept support:
 //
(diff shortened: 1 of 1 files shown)


jzmaddock:
That all looks fine too.

mborland:
@jzmaddock this is good to go.

NAThompson:
It also appears that `#define BOOST_MATH_IS_FLOAT(T) (std::is_floating_point<T>::value)` is superfluous now. Doesn't need to be done in the PR of course.

Issue #544, opened 2021-02-21
Inaccurate Non-Central Student's T Test Data


jzmaddock:
This is mostly for future reference... with fixes coming soon....

While testing Math on platforms with 128-bit long doubles (s390x) all the non-central T tests fail with large errors.

Main test case is:

v = 45487064
l = 4.548706300511184963397681713104248e-09
t = -0.594260692596435546875

And when calculating the CDF we see:

```
Found:    0.27616888228432564767719590104835779
Expected: 0.27616888228432562302216979333024938
                            ^ Differ here
```

Increasing the precision of the calculation yields:

```
35 digit precision:  0.27616888228432564767719590104835779
100 digit precision: 0.27616888228432564767719590104835781
```

So nearly in agreement, but that doesn't prove the test data is in error, and Wolfram Alpha declined to give an answer :(

So.... as a first start I coded up a naive version of the CDF from wikipedia here: https://en.wikipedia.org/wiki/Noncentral_t-distribution.

![image](https://user-images.githubusercontent.com/5011768/108624564-94033f80-743d-11eb-89e9-50ebe0266b71.png)

Code:

```
big_t nc_t_F(big_t v, big_t nc, big_t t)
{
   unsigned j = 0;
   big_t sum = 0;
   big_t tol = std::numeric_limits<big_t>::epsilon();
   do
   {
      big_t x = t * t / (t * t + v);
      big_t p = exp(-nc * nc / 2) * pow(nc * nc / 2, j) / (2 * boost::math::factorial<big_t>(j));
      big_t q = (nc / 2) * exp(-nc * nc / 2) * pow(nc * nc / 2, j) / (boost::math::constants::root_two<big_t>() * boost::math::tgamma(j + big_t(3) / 2));
      big_t term = p * boost::math::ibeta(big_t(j) + 0.5, v / 2, x) + q * boost::math::ibeta(big_t(j + 1), v / 2, x);
      ++j;

      sum += term;

      if (fabs(sum * tol) > fabs(term))
         break;
   } while (true);

   return sum + boost::math::constants::half<big_t>() * (1 + boost::math::erf(-nc / boost::math::constants::root_two<big_t>()));
}
big_t nc_t(big_t v, big_t nc, big_t t)
{
   return t < 0 ? 1 - nc_t_F(v, -nc, t) : nc_t_F(v, nc, t);
}
```

And this gave a value in agreement with our current implementation.  However, they share the same incomplete beta code, so there could still be an issue lurking somewhere.

[...]

PR #564, opened 2021-03-01
Fix up non-central T test data for 128-bit floats.

jzmaddock:
Fixes: https://github.com/boostorg/math/issues/544.
Also fixes uncovered issue in tgamma_ratio where we were previously relying on sub-normals to get the correct result.


NAThompson:
I believe this is the first time I've ever heard someone say subnormals actually did something useful.

Wowsers!

jzmaddock:
Failure are GHA SNAFU's, merging.

PR #565, opened 2021-03-10
Remove boost.array dependency

mborland:


include/boost/math/tools/rational.hpp
```diff
@@ -187,7 +188,7 @@ inline V evaluate_polynomial_c_imp(const T* a, const V& val, const Tag*) BOOST_M
 template <class T, class U>
 inline U evaluate_polynomial(const T* poly, U const& z, std::size_t count) BOOST_MATH_NOEXCEPT(U)
 {
-   BOOST_ASSERT(count > 0);
```
jzmaddock, line 190: I'd prefer to keep BOOST_ASSERT's for now as it has some (small) advantages over assert.

The long term plan would be to search and replace BOOST_ASSERT->BOOST_MATH_ASSERT and then set BOOST_MATH_ASSERT to BOOST_ASSERT when Boost is present, and to assert when it's not (ie "standalone" mode).

mborland:
Resolved the merge conflicts; good to go now.

jzmaddock:
Thanks Matt!

PR #567, opened 2021-03-13
Remove boost.smart_ptr dependency

mborland:


mborland:
@jzmaddock This is green, and should be good to go.

PR #568, opened 2021-03-14
Remove boost.lambda dependency

mborland:


include/boost/math/tools/polynomial.hpp
```diff
@@ -338,13 +327,12 @@ class polynomial
       }
    }
 #ifdef BOOST_MATH_HAS_IS_CONST_ITERABLE
```
jzmaddock, line 329: I suspect BOOST_MATH_HAS_IS_CONST_ITERABLE is always set in C++11 land?

Other than that, this all looks good to me,, thanks!

mborland: BOOST_MATH_HAS_IS_CONST_ITERABLE requires cxx14_variable_templates so it has to stay put for now.

jzmaddock: Ah, shame.  No matter.

mborland:
This one should also be good to go. CI is clean.

PR #569, opened 2021-03-14
Remove boost.range dependency

mborland:


include/boost/math/distributions/hyperexponential.hpp
```diff
@@ -282,12 +277,13 @@ class hyperexponential_distribution
     // Two arg constructor from 2 ranges, we SFINAE this out of existence if
     // either argument type is incrementable as in that case the type is
     // probably an iterator:
-    public: template <typename ProbRangeT, typename RateRangeT>
+    public: template <typename ProbRangeT, typename RateRangeT,
```
jzmaddock, line 280: Sorry Matt, but this is going to be a lot more complex: we have two sets of 2-arg constructors, one accepts 2 ranges, conceptually that's anything that has begin() and end() member functions, but now that C++20 is here, we should probably take the opportunity to check that we're compatible with native range concepts.  The other constructor accepts 2 iterators.  So the use of has_pre_incremement is a simplistic way to check for an iterator concept, and I'm afraid is_pointer doesn't match this.

BTW the "modern" way to implement concept-like-traits is via is_detected: https://www.boost.org/doc/libs/1_75_0/libs/type_traits/doc/html/boost_typetraits/reference/is_detected.html  GCC has it's own version but it's not in the std sadly so other compilers do not :(

mborland: I saw the comment about using a pointer and was hopeful that would work. `std::iterator_traits` is only SFINAE friendly in C++14 and on or that would be the easy button. I'll have to take a look at your link and come back with a more robust solution.

mborland:
Using SFINAE on the minimum requirements for an iterator (pre-increment and derefernce) seems to have solved the issue from using `is_pointer`. I also added tests for using `std::ranges::begin/end` and a native range; both pass locally without needing any additional constructors.

jzmaddock:
> I also added tests for using std::ranges::begin/end and a native range; both pass locally without needing any additional constructors.

Cool!

mborland:
This is good to go. I added an implementation of `is_detected` like you recommended; it will get further use removing boost.type_traits.

jzmaddock:
That's weird, merging this has triggered a new failure in univariate_statistics_test: https://github.com/boostorg/math/runs/2162144720?check_suite_focus=true

mborland:
Maybe a spurious failure? Nothing here would have affected gini coefficient calculations.

jzmaddock:
>Maybe a spurious failure? Nothing here would have affected gini coefficient calculations.

No.  I worry more that we have a non-deterministic test which sometimes passes and sometimes fails?

PR #570, opened 2021-03-14
Remove boost.assert and boost.static_assert dependencies

mborland:
Convert all uses of `BOOST_ASSERT` and `BOOST_STATIC_ASSERT` to `BOOST_MATH_ASSERT` and `BOOST_MATH_STATIC_ASSERT` respectively. Macros expand depending on if `BOOST_MATH_STANDALONE` is defined.


config/has_128bit_floatmax_t.cpp
```diff
@@ -4,13 +4,13 @@
...
 #include <boost/cstdfloat.hpp>
-#include <boost/static_assert.hpp>
+#include <boost/math/tools/assert.hpp>
 
 #ifndef BOOST_FLOAT128_C
 #error "There is no 128 bit floating point type"
 #endif
 
-BOOST_STATIC_ASSERT(sizeof(boost::floatmax_t) * CHAR_BIT == 128);
+BOOST_MATH_STATIC_ASSERT(sizeof(boost::floatmax_t) * CHAR_BIT == 128);
```
jzmaddock, line 13: Is there any reason not to just use `static_assert` here?

mborland: The message in `static_assert` did not become optional until C++17.

jzmaddock: Yes I know, we would have to invent some descriptive error messages.

mborland: That's fair.

mborland:
This should be good to go now. There are fixes for the new failures in the misc test set; I am not sure why those started popping up.

jzmaddock:
I'm still seeing instances of BOOST_MATH_STATIC_ASSERT_MSG that were missed?

mborland:
> I'm still seeing instances of BOOST_MATH_STATIC_ASSERT_MSG that were missed?

The only header they are still in is `TR1.hpp` in the below as an example:

```
template<class T> int fpclassify NO_MACRO_EXPAND(T x)
{ BOOST_MATH_STATIC_ASSERT(sizeof(T) == 0); return false; } // must not be instantiated
```

Should the message be file a compiler bug? There should be nothing that yields `sizeof(T) == 0`

jzmaddock:
I'm seeing them in for example  example/root_elliptic_finding.cpp, plus include/boost/math/cstdfloat/cstdfloat_types.hpp and many others?

This code:

```
template<class T> int fpclassify NO_MACRO_EXPAND(T x)
{ BOOST_MATH_STATIC_ASSERT(sizeof(T) == 0); return false; } // must not be instantiated
```

Is used to force a hard error when a template which should never be instantiated is done so.  In this case, only the float/double/long double specializations are supported.  Looking at TR1, I *think* this probably comes under "undefined behaviour", so an error message along those lines should be fine.

mborland:
> I'm seeing them in for example example/root_elliptic_finding.cpp, plus include/boost/math/cstdfloat/cstdfloat_types.hpp and many others?

They are gone now. Easy enough because `BOOST_MATH_STATIC_ASSERT_MSG` expands to `static_assert` cleanly.

> This code:
> 
> ```
> template<class T> int fpclassify NO_MACRO_EXPAND(T x)
> { BOOST_MATH_STATIC_ASSERT(sizeof(T) == 0); return false; } // must not be instantiated
> ```
> 
> Is used to force a hard error when a template which should never be instantiated is done so. In this case, only the float/double/long double specializations are supported. Looking at TR1, I _think_ this probably comes under "undefined behaviour", so an error message along those lines should be fine.

I added a message about it being UB.

jzmaddock:
All green - good to go?

mborland:
> All green - good to go?

Yes; it is good to go.

PR #575, opened 2021-03-20
Remove boost.concept_check dependency

mborland:


test/compile_test/distribution_concept_check.cpp
```diff
@@ -5,52 +5,86 @@
...
-   function_requires<DistributionConcept<normal_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<pareto_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<poisson_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<rayleigh_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<skew_normal_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<students_t_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<triangular_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<uniform_distribution<RealType, custom_policy> > >();
-   function_requires<DistributionConcept<weibull_distribution<RealType, custom_policy> > >();
+   DistributionConcept<bernoulli_distribution<RealType, custom_policy>> bernoulli;
```
jzmaddock, line 22: I'm not sure, but I don't think this change actually instantiates the concept check at all?

mborland: It looks like I need to actually call `constraints` with this change.

mborland: The only files that include `boost/math/concepts/distributions.hpp` are the concept check tests. Rather than re-inventing the wheel on this one how do you feel about just wrapping this header to be disabled if `BOOST_MATH_STANDALONE` is defined?

jzmaddock: Personally I think it's fine to have dependencies to other boost libraries in our _tests_: it's not at all unreasonable to have to install other dependencies in order to run the tests IMO.

mborland:
This one should be good to go.

PR #576, opened 2021-03-21
Remove boost.type_traits dependency

mborland:


mborland:
@jzmaddock Since all of these [failures involve Multiprecision](https://github.com/boostorg/math/pull/576/checks?check_run_id=2163312687#step:18:341) have you seen anything like it before? The fact that they're only affecting G++5 and 6 makes me think that there is a compiler/language defect that was resolved.

jzmaddock:
>Since all of these failures involve Multiprecision have you seen anything like it before? The fact that they're only affecting G++5 and 6 makes me think that there is a compiler/language defect that was resolved.

No.  But I'm guessing that the implementation is a little too cute :)  There's another issue: class `number` *always* has `real()` and `imag()` member functions (no choice really on that) so the default versions of this trait will always be true... except that there is a partial specialization here: https://github.com/boostorg/multiprecision/blob/1c8fcaea68943f86e94070a86c003513794633a7/include/boost/multiprecision/detail/number_base.hpp#L1653 which should be taken in this case, but I'm guessing is the cause of the ambiguity?

The only fix I can see is to revert to using a partial specialization on std::complex for older gcc?

jzmaddock:
Ah, a better fix might be:

```
template <class T> is_complex_type : public is_complex_type_imp {};
```

With `is_complex_type_imp` the same as your current `is_complex_type`.  That should remove the ambiguity and retain the old interface, while keeping the new functionality... I hope ;)

mborland:
This is good to go.

PR #577, opened 2021-03-21
Remove boost.throw_exception dependency

mborland:
Contains `#define BOOST_MATH_STANDALONE` so that the CI system will test standalone mode. Will remove before merge.

### DIFF
--- a/doc/overview/error_handling.qbk
+++ b/doc/overview/error_handling.qbk
@@ -12,6 +12,10 @@ Handling of errors by this library is split into two orthogonal parts:
 [warning The default error actions are to throw an exception with an informative error message.
 [role red If you do not try to catch the exception, you will not see the message!]]
 
+The default behavior is for exceptions to be thrown using boost::throw_exception.
+This facility has special handling for exception free environments.
+If BOOST_MATH_STANDALONE is defined standard library facilities (e.g. throw) are used for exception handling.
+
 The kinds of errors that can be raised are:
 
 [variablelist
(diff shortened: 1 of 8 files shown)


jzmaddock:
Looks good, thanks Matt.

mborland:
This is clean, but I have one question before it goes in. For metal environments like @ckormanyos is using is it worth adding another option that is something to the affect of the following:

```
#if defined(BOOST_MATH_STANDALONE) && defined(BOOST_MATH_NO_THROW)
#define BOOST_MATH_THROW_EXCEPTION(expr) std::abort() or std::exit(EXIT_FAILURE);
#endif
```

NAThompson:
@mborland : Not sure, but certainly a separate pull request. . .

jzmaddock:
>Not sure, but certainly a separate pull request. . .

+1.  Lets get some sort of standalone mode working and then I'm sure Chris will chime in with whatever else is required, everything is macro-ised so it's an easy change down the road.  Probably we should let the user define BOOST_MATH_THROW_EXCEPTION on the command line as one option?

ckormanyos:
> For metal environments like @ckormanyos is using is it worth adding another option that is something to the affect of the following

Thank you for thinking of this. I think that no special provisions should be made for metal. Bare metal is a new concept that we will work on from time to time as we progress. The main flow of Boost.Math should, in my opinion, in no way be hindered or modified by this.

> +1. Lets get some sort of standalone mode working and then I'm sure Chris will chime in with whatever else is required,

Yes. I absolutely agree. Please proceed without any concern whatsoever about such matters. proceed please full speed ahead with what we would consider the normal and usual PC approach without any considerations for metal adaptions, which we can handle _after_ reaching stability.

Kind regards, Chris

PR #579, opened 2021-03-22
Widen gini coefficient test tolerance

mborland:
Reported in https://github.com/boostorg/math/pull/569 and failure [seen here](https://github.com/boostorg/math/runs/2162144720?check_suite_focus=true#step:12:1493). Increase test tolerance to be match rest of tests with 10*epsilon as acceptable tolerance. Should head off random failures in the future.


mborland:
@jzmaddock This should be good to go. The 2x internal failures are both GHA internal errors. The 6x queued tests are all duplicates? Not sure what is going on with the system, but on the actual GHA page it shows the run as complete.

PR #580, opened 2021-03-24
Fix for beta link error

mborland:
Fixes the compile error noted in the ML, and address a couple more instances of missing fpclassify.


jzmaddock:
One failure is a spurious timeout.

PR #582, opened 2021-03-25
Remove boost.predef dependency

mborland:


mborland:
@jzmaddock This should be good to go. The next PR will be config which will include testing for the standalone functionality.

include/boost/math/special_functions/detail/fp_traits.hpp
```diff
@@ -25,8 +25,36 @@ With these techniques, the code could be simplified.
...
+#else // Does not have compliant C++20
+
+#ifdef _WIN32
+
+#define BOOST_MATH_ENDIAN_BIG_BYTE 1
+#define BOOST_MATH_ENDIAN_LITTLE_BYTE 0
+
+#else
+
+#define BOOST_MATH_ENDIAN_BIG_BYTE (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
```
jzmaddock, line 52: I think this is nearly but not quite correct: these macros are GCC specific I assume?  We should probably have another #elif branch here for the case where `__BYTE_ORDER__` is not defined and a reasonably helpful #error message explaining that either the header needs to be fixed, or they should not use standalone mode.

mborland: Those macros work with Clang as well. Adding an error message if they are not defined makes sense.

Issue #585, opened 2021-03-28
Lot's of non-deterministic failures in univariate_statistics_test.cpp


jzmaddock:
These seem to be related to gini calculation, I changed all the BOOST_TEST's to BOOST_TEST_LT so we could see how close the failures were, here's a typical run with msvc-14.2:

```
D:\data\boost\boost\libs\math\test\univariate_statistics_test.cpp(694): test 'abs(gini - expected) < Real(0.03)' ('0.0487317' < '0.03') failed in function 'void __cdecl test_gini_coefficient<float,const class std::execution::parallel_policy&>(const class std::execution::parallel_policy &)'
D:\data\boost\boost\libs\math\test\univariate_statistics_test.cpp(694): test 'abs(gini - expected) < Real(0.03)' ('0.0593992' < '0.03') failed in function 'void __cdecl test_gini_coefficient<long double,const class std::execution::parallel_policy&>(const class std::execution::parallel_policy &)'
D:\data\boost\boost\libs\math\test\univariate_statistics_test.cpp(659): test 'abs(gini - expected) < tol' ('0.666667' < '5.34553e-50') failed in function 'void __cdecl test_gini_coefficient<class boost::multiprecision::number<class boost::multiprecision::backends::cpp_bin_float<50,10,void,int,0,0>,0>,const class std::execution::parallel_policy&>(const class std::execution::parallel_policy &)'
D:\data\boost\boost\libs\math\test\univariate_statistics_test.cpp(682): test 'abs(gini) < tol' ('0.222222' < '5.34553e-50') failed in function 'void __cdecl test_gini_coefficient<class boost::multiprecision::number<class boost::multiprecision::backends::cpp_bin_float<50,10,void,int,0,0>,0>,const class std::execution::parallel_policy&>(const class std::execution::parallel_policy &)'
D:\data\boost\boost\libs\math\test\univariate_statistics_test.cpp(694): test 'abs(gini - expected) < Real(0.03)' ('0.144181' < '0.03') failed in function 'void __cdecl test_gini_coefficient<class boost::multiprecision::number<class boost::multiprecision::backends::cpp_bin_float<50,10,void,int,0,0>,0>,const class std::execution::parallel_policy&>(const class std::execution::parallel_policy &)'
[...]

PR #586, opened 2021-03-28
Fix for issue #585

mborland:
Provides thread synchronization if using the MSVC implementation of the Parallel TS.

### DIFF
--- a/include/boost/math/statistics/detail/single_pass.hpp
+++ b/include/boost/math/statistics/detail/single_pass.hpp
@@ -21,6 +21,11 @@
 #include <functional>
 #include <vector>
 
+#if !defined(BOOST_NO_CXX17_HDR_EXECUTION) && !defined(_MSC_VER)
+#include <mutex>
+#include <execution>
+#endif
+
 namespace boost { namespace math { namespace statistics { namespace detail {
 
 template<typename ReturnType, typename ForwardIterator>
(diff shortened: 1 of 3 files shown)


NAThompson:
What's the performance implications of this diff? Could you run the benchmarks?

mborland:
I expect that the performance penalty will be high maybe to the point where it is better off just running the sequential method. I have been meaning to create a MSVC environment; should have benchmarks up in the next day or two.

jzmaddock:
After a minor correction to the lambda, this still fails I'm afraid :(

mborland:
> After a minor correction to the lambda, this still fails I'm afraid :(

OK. I will get MSVC up and running on my end to troubleshoot and re-benchmark.

mborland:
The new implementation takes the decomposition method from the first four moments calculation. This maintains the required sorting. Clean on tsan with both Clang 12 and GCC 11. New benchmarks:

```
gini_coefficient<double>/real_time_BigO                   1.09 N          1.08 N    
gini_coefficient<double>/real_time_RMS                      13 %            13 %
parallel_gini_coefficient<double>/real_time_BigO          0.88 N          0.82 N    
parallel_gini_coefficient<double>/real_time_RMS             36 %            39 %
```

mborland:
Any changes for this one? The only CI failure is from CircleCI from before the config was committed.

jzmaddock:
Looks good to me, @NAThompson ?

NAThompson:
I think this is now *correct*, but I'm not sure the parallelism is worth it; it seems the speedup  is not so impressive.

What about using the parallel sort from the standard library? If you use that first, then you can reason much more clearly about a parallel computation of the Gini coefficient.

(Note: This is just a suggestion. I believe this PR can be merged if @mborland desires; we don't need to make every PR a research program.)

mborland:
> I think this is now _correct_, but I'm not sure the parallelism is worth it; it seems the speedup is not so impressive.
> 
> What about using the parallel sort from the standard library? If you use that first, then you can reason much more clearly about a parallel computation of the Gini coefficient.

Passing an execution policy to `std::is_sorted` and `std::sort` is [used here.](https://github.com/boostorg/math/blob/1924e3179c24ac4d73181587d752774ad72c0ccb/include/boost/math/statistics/univariate_statistics.hpp#L416-L424) I agree the speedup is not impressive. Another interface could be added where a distribution function is taken as input to directly calculate the gini coefficient (much like is done to compare against a uniform distribution in the test cases). I do not know if implementing all of those would be value added.

PR #587, opened 2021-03-29
Make gini calculation serial only for now.

jzmaddock:
See https://github.com/boostorg/math/issues/585


mborland:
I was going to suggest this so bad code didn't go out in 1.76. Thanks.

PR #588, opened 2021-03-29
Remove boost.integer dependency

mborland:


mborland:
@jzmaddock This is clean on CI pending review.

include/boost/math/common_factor.hpp
```diff
@@ -10,10 +10,14 @@
...
+#ifndef BOOST_MATH_STANDALONE
 #include <boost/math/common_factor_ct.hpp>
 #include <boost/math/common_factor_rt.hpp>
+#include <boost/math/tools/header_deprecated.hpp>
 
-BOOST_HEADER_DEPRECATED("<boost/integer/common_factor.hpp>");
-
+BOOST_MATH_HEADER_DEPRECATED("<boost/integer/common_factor.hpp>");
+#else
+#error Common factor is not available in standalone mode because it requires boost.integer.
```
NAThompson, line 20: Isn't gcd part of the standard library now?

include/boost/math/tools/polynomial_gcd.hpp
```diff
@@ -14,8 +14,57 @@
...
+    {
+        a %= b;
+        swap(a, b);
+    }
+    return a;
+}
+
+} // namespace gcd_detail
+} // namespace boost::integer
+#error polynomial gcd can only be used in standalone mode with C++17 or higher
```
jzmaddock, line 65: This #error appears to be misplaced - missing a #else branch?

More pertinently, this header is included by polynomial.hpp, so we don't want hard #error's here, probably better to static_assert in it's interface somewhere if the gcd code is actually instantiated.  BTW I'm *reasonably* sure that static_assert's need to depend on the template parameter, if you do a `static_assert(false, "message")` then the compiler get get ultra-cute and realises that the assert will always trigger and issue an error even if the template is never instantiated.  That's why I've used `static_assert(sizeof(T) == 0, "message")` in the past, even though it's much less readable.

include/boost/math/common_factor.hpp
mborland: Yes; `std::gcd` was added in C++17.

mborland:
The CI system seems to have timed out on this one. Still shows it as in the queue, but all of the failures are: "This request was automatically failed because there were no enabled runners online to process the request for more than 1 days."

mborland:
Consolidated into #597

PR #589, opened 2021-03-29
Remove boost.lexical cast and boost.core dependencies

mborland:


mborland:
All the CI failures on this one are: "GitHub Actions has encountered an internal error when running your job." Should be good to go.

jzmaddock:
@mborland : try compiling :

```
g++ -I. -c -DBOOST_MATH_STANDALONE libs/math/test/compile_test/*.cpp
```

and there are gazillions of errors.  The first is caused by a `static_assert(false` which is *always* triggered by gcc, not just when the containing function is instantiated, the condition has to depend on a dependent template type parameter, the next by use of BOOST_STRINGIZE.

mborland:
> @mborland : try compiling :
> 
> ```
> g++ -I. -c -DBOOST_MATH_STANDALONE libs/math/test/compile_test/*.cpp
> ```
> 
> and there are gazillions of errors. The first is caused by a `static_assert(false` which is _always_ triggered by gcc, not just when the containing function is instantiated, the condition has to depend on a dependent template type parameter, the next by use of BOOST_STRINGIZE.

You are correct. I have a branch that contains this PR and #588 so that I could begin fully testing Standalone. It has several little fixes like the ones you described. Are you good with me closing both of those PRs and rolling them into one?

mborland:
Consolidated into #597

jzmaddock:
> Are you good with me closing both of those PRs and rolling them into one?

Nod.  Sounds good to me!

PR #593, opened 2021-03-30
Fix gini coefficient calculation.

jzmaddock:
Temporary fix to avoid using broken parallel implementation.


glenfe (approved):
Ok to merge

jzmaddock:
Thanks Glen!

PR #596, opened 2021-03-31
Add Sanitizers to CI Tests

jzmaddock:


jzmaddock:
@mborland , @NAThompson , Just a heads up that I'm merging this PR which adds, `-fsanitize=address, undefined, thread, integer` tests into the mix (on drone).  There is one sporadic failure from naive_monte_carlo_test_4 under the thread sanitizer (https://github.com/boostorg/math/issues/598), but other than that all is green.  

There are also quite a number of tests which are disabled for sanitizer runs: some just make no sense (compile tests, concept checks), but some just gobble up so much memory that they kill the system.  Which is to say, you can't completely rely on these to catch everything, but they should catch the most grievous errors in most things.

Finally, while there are no false positives so far, I wouldn't rule these out down the road...

NAThompson:
@jzmaddock : I've seen a few false positives with TSAN before, but the signal to noise is very high. The most common problem is not false positives, but true positives that you really have no control over, say, in MPI libraries.

PR #600, opened 2021-04-04
Fix TSAN failures in naive monte carlo

mborland:
Fix for issue #598.


jzmaddock:
Thanks Matt, it's a shame we can have atomic shared_ptr's, but not exception_ptr's

mborland:
This should be good to go; I re-arranged the test slightly to fix the one CI failure.

Issue #571, opened 2021-03-15
GCC 4.8/4.9 real_concept errors


Kojoley:
https://www.boost.org/development/tests/develop/developer/output/teeks99-dkr-dg4-8-boost-bin-v2-libs-spirit-test-karma-karma_real1-test-gcc-4-8~c++11-debug-threading-multi-visibility-hidden.html
https://www.boost.org/development/tests/develop/developer/output/teeks99-dkr-dg4-9-boost-bin-v2-libs-spirit-test-karma-karma_real1-test-gcc-4-9~c++11-debug-threading-multi-visibility-hidden.html
```
    "g++-4.9"  -ftemplate-depth-512 -fvisibility-inlines-hidden -std=c++11 -fPIC -m64 -pthread -O0 -fno-inline -Wall -g -fvisibility=hidden  -DBOOST_ALL_NO_LIB=1 -DBOOST_BEAST_ALLOW_DEPRECATED -DBOOST_BEAST_SEPARATE_COMPILATION  -I".." -I"../libs/spirit/test/karma"  -c -o "/var/boost/run/results/boost/bin.v2/libs/spirit/test/karma/karma_real1.test/gcc-4.9~c++11/debug/threading-multi/visibility-hidden/real1.o" "../libs/spirit/test/karma/real1.cpp"

In file included from ../boost/config.hpp:61:0,
                 from ../boost/core/lightweight_test.hpp:26,
                 from ../boost/detail/lightweight_test.hpp:15,
                 from ../libs/spirit/test/karma/real.hpp:14,
                 from ../libs/spirit/test/karma/real1.cpp:7:
../boost/math/tools/precision.hpp:41:83: error: redeclaration 'int boost::math::tools::digits()' differs in 'constexpr'
 inline BOOST_MATH_CONSTEXPR int digits(BOOST_MATH_EXPLICIT_TEMPLATE_TYPE_SPEC(T)) BOOST_NOEXCEPT
                                                                                   ^
In file included from ../boost/math/policies/error_handling.hpp:21:0,
                 from ../boost/math/special_functions/round.hpp:14,
                 from ../boost/math/concepts/real_concept.hpp:28,
                 from ../libs/spirit/test/karma/real.hpp:25,
                 from ../libs/spirit/test/karma/real1.cpp:7:
../boost/math/policies/policy.hpp:25:15: error: from previous declaration 'constexpr int boost::math::tools::digits()'
[...]

PR #601, opened 2021-04-05
Make real_concept constructor constexpr

mborland:
Make real_concept constructors and tools `constexpr` to avoid collision failures from issue #571.

### DIFF
--- a/include/boost/math/concepts/real_concept.hpp
+++ b/include/boost/math/concepts/real_concept.hpp
@@ -60,24 +60,24 @@ class real_concept
 {
 public:
    // Constructors:
-   real_concept() : m_value(0){}
-   real_concept(char c) : m_value(c){}
-   real_concept(wchar_t c) : m_value(c){}
-   real_concept(unsigned char c) : m_value(c){}
-   real_concept(signed char c) : m_value(c){}
-   real_concept(unsigned short c) : m_value(c){}
-   real_concept(short c) : m_value(c){}
-   real_concept(unsigned int c) : m_value(c){}
-   real_concept(int c) : m_value(c){}
-   real_concept(unsigned long c) : m_value(c){}
-   real_concept(long c) : m_value(c){}
-   real_concept(unsigned long long c) : m_value(static_cast<real_concept_base_type>(c)){}
-   real_concept(long long c) : m_value(static_cast<real_concept_base_type>(c)){}
-   real_concept(float c) : m_value(c){}
-   real_concept(double c) : m_value(c){}
-   real_concept(long double c) : m_value(c){}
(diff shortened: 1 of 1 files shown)


jzmaddock:
Morning Matt: I'm not sure this is the correct fix - it changes the interface to the concept, and in general, we _do_ want to allow non-constexpr constructors etc, as in the Multiprecision types for example.

mborland:
> we _do_ want to allow non-constexpr constructors etc, as in the Multiprecision types for example.

Wouldn't that still be permissible with this change as long as they are used outside of a `constexpr` context? A broader question is how much effort should we put into fixes for ancient compilers like this issue? Is it worth putting in the notes that the oldest supported versions of GCC/Clang/MSVC are X.Y.Z? I feel like the current testing of GCC 5-10 (and I test on GCC11 locally) is pretty thorough.

jzmaddock:
It's certainly pointless unless we're testing those compilers, otherwise we'll just see more regressions down the road.  I've added tentative testing to drone on https://github.com/boostorg/math/pull/602

jzmaddock:
> A broader question is how much effort should we put into fixes for ancient compilers like this issue? Is it worth putting in the notes that the oldest supported versions of GCC/Clang/MSVC are X.Y.Z? I feel like the current testing of GCC 5-10 (and I test on GCC11 locally) is pretty thorough.

I'm beginning to think that gcc-4.8 and 4.9 are too broken to support:  I've had a good try at getting things working in https://github.com/boostorg/math/pull/602, but every time you fix one issue something else pops up.  Currently stalled because `is_complex_type<>` fails completely with these compilers.  That's fixable, but I suspect other issues will follow.

ckormanyos:
> I'm beginning to think that gcc-4.8 and 4.9...

Personally, I draw the line at 5, preferably 5.2, which features decent C++11 support. As it turns out, I use quite a few embedded compilers around that 5 zone, and even one single exotic architecture with a 4.9.

But I have found it very difficult to roll today's code, the way we write it in modern C++11 prior to the GCC5 barrier.

mborland:
> > I'm beginning to think that gcc-4.8 and 4.9...
> 
> Personally, I draw the line at 5, preferably 5.2, which features decent C++11 support. As it turns out, I use quite a few embedded compilers around that 5 zone, and even one single exotic architecture with a 4.9.
> 
> But I have found it very difficult to roll today's code, the way we write it in modern C++11 prior to the GCC5 barrier.


I agree with @ckormanyos. I think the only real workaround for GCC 5/6 was in `is_complex_type`.

jzmaddock:
Closing, gcc < 5 is a lost cause for us now.

PR #602, opened 2021-04-06
Attempt to test older gcc versions on drone.

jzmaddock:


jzmaddock:
Closing, GCC < 5 is going to be a lost cause.

Issue #594, opened 2021-03-30
Polution of global namespace with "mp"


jzmaddock:
@mborland : Apologies I should have spotted this before, but there are several places where `namespace mp = boost::math::tools::meta_programming;` occurs at global scope: this is a big no-no, everything should be contained with boost::math, just moving the namespace alias inside our namespaces should do the trick I hope!

I just hit this while testing a Multiprecision bug report which used `mp` as an alias for `boost::multiprecision` and clashed.

mborland:
Locally moving it into the boost::math namespace worked fine. The were both included in the giant policy PR.

jzmaddock:
@mborland can this be closed now, I think it's fixed?

mborland:
This can be closed

PR #604, opened 2021-04-08
Fix for issue #594

mborland:
@jzmaddock I realize it may be too late for this to make release, but if possible it should. I cherry picked the commit off develop to add to master.


jzmaddock:
Ah... according to the calender the last day for post-beta changes was yesterday :(

mborland:
This also unexpectedly made the quadrature tests unhappy? Oh well; there is always next release.

PR #606, opened 2021-04-09
Standalone tests

mborland:
Adds missing include tests and fixes for failures of the former. Tests run by regular CI using Jamfile and standalone CI using CMake.


include/boost/math/distributions/arcsine.hpp
```diff
@@ -85,7 +86,11 @@ namespace boost
       { // Check x_min < x_max
         if (x_min >= x_max)
         {
+          #ifndef BOOST_MATH_STANDALONE
           std::string msg = "x_max argument is %1%, but must be > x_min = " + lexical_cast<std::string>(x_min) + "!";
```
NAThompson, line 90: @jzmaddock : Does this still need lexical_cast? It looks like we could just get away with `std::to_string`. . .

mborland:
@NAThompson Do you know why [this error ](https://github.com/boostorg/math/runs/2309694662?check_suite_focus=true#step:18:217)has started popping up? I first saw it on [this PR](https://github.com/boostorg/math/pull/604) the other day.

NAThompson:
@mborland : That's generally due to some minor variations in rounding modes, or some multiprecision diffs of no real consequence

Hopefully fixed in 2553ecf00ee234d7d5.

include/boost/math/distributions/arcsine.hpp
jzmaddock: +1, I think that's correct, yes.  And it allows us to remove the #if logic entirely.

jzmaddock:
Really big thanks for this Matt: there's some good stuff here, and lot's of errors caught and fixed which is great!

One busy-body comment: the new files really should have your name on the copyright and the current date on them ;)

One more substantive comment: some of the include tests don't do anything except #include the header: unfortunately I've found from bitter experience that that is not enough to catch all errors, to do that, you actually have to instantiate the templates as well.  This one has come up on the mailing list several times BTW, where someone suggests automating producing #include tests, but sadly this only catches errors in the interface, not for the most part, in the templates themselves.

We could push that to the long grass for a moment and merge this as is for now, but it might be easier to spot which tests need to be updated while the PR is around?

include/boost/math/distributions/arcsine.hpp
mborland: `std::to_string` will only work with the builtin arithmetic types. I can keep the else block which will pass tests using `real_concept` but offer a less helpful error message.

jzmaddock: Good point, glad one of us is paying attention! :)

mborland:
> We could push that to the long grass for a moment and merge this as is for now, but it might be easier to spot which tests need to be updated while the PR is around?

I think it would be easier to just keep going against this PR. It makes sense you need to instantiate the templates to get better results.

mborland:
Pending further review I believe this is good to go now.

include/boost/math/quadrature/gauss.hpp
```diff
@@ -80,7 +81,7 @@ class gauss_detail<T, 7, 0>
 public:
    static std::array<T, 4> const & abscissa()
    {
-      static const std::array<T, 4> data = {
+      static constexpr std::array<T, 4> data = {
```
NAThompson, line 84: Yup that's a good diff to make.

include/boost/math/tools/engel_expansion.hpp
```diff
@@ -56,15 +57,15 @@ class engel_expansion {
             computed += term;
         }
 
-        for (size_t i = 1; i < a_.size(); ++i)
+        for (size_t j = 1; j < a_.size(); ++j)
         {
             // Sanity check: This should only happen when wraparound occurs:
-            if (a_[i] < a_[i-1])
+            if (a_[j] < a_[j-1])
```
NAThompson, line 63: Did I accidently alias `i` in a nested loop?

include/boost/math/tools/header_deprecated.hpp
```diff
@@ -8,10 +8,10 @@
 
 #ifdef _MSC_VER
 // Expands to "This header is deprecated; use expr instead."
-#define BOOST_MATH_HEADER_DEPRECATED(expr) #prama message("This header is deprecated; use " expr " instead.")
```
NAThompson, line 11: Shouldn't the fix have been `#pragma` to remove the `#prama` typo?

include/boost/math/tools/luroth_expansion.hpp
```diff
@@ -29,17 +29,17 @@ class luroth_expansion {
...
         {
            d_.shrink_to_fit();
            return;
         }
         // This attempts to follow the notation of:
         // "Khinchine's constant for Luroth Representation", by Sophia Kalpazidou.
-        x = x - dn;
-        Real computed = dn;
+        x = x - dn1;
+        Real computed = dn1;
```
NAThompson, line 42: I'm confused about what I did on this one that needed the variable name change.

NAThompson:
I added some trivial comments; I think it's looking pretty good. @jzmaddock will have an orthogonal set of suggestions but I'm happy.

include/boost/math/quadrature/gauss.hpp
mborland: In this situation no; you only need `constexpr const` for reference variables.

include/boost/math/tools/engel_expansion.hpp
mborland: The change suppressed shadowing warnings because `i` is [defined here](https://github.com/boostorg/math/blob/63a16971afc470e1772ed99da2ed2a2835a8954c/include/boost/math/tools/engel_expansion.hpp#L41).

include/boost/math/tools/header_deprecated.hpp
mborland: That may have worked. In the GCC case the spelling was right but the `#warning` did not expand correctly so I used `_Pragma` instead

include/boost/math/tools/luroth_expansion.hpp
mborland: `dn` is defined [again here](https://github.com/boostorg/math/blob/63a16971afc470e1772ed99da2ed2a2835a8954c/include/boost/math/tools/luroth_expansion.hpp#L52). Changed the name to suppress shadowing warnings.

test/compile_test/cstdfloat_cmath_incl_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <boost/math/cstdfloat/cstdfloat_cmath.hpp>
+//
+// Note this header includes no other headers, this is
+// important if this test is to be meaningful:
+//
+#include "test_compile_result.hpp"
+
+void compile_and_link_test()
+{
+    #ifdef __float128
```
jzmaddock, line 18: __float128 is a type not a macro?

Looking back at our docs on this, they look to be a bit confused, or indeed flat out wrong in places, but I *think* the thing to do is if BOOST_FLOAT128_C is defined, then `boost::float128` is available and can be tested as you have here.

test/compile_test/cstdfloat_limits_incl_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <boost/math/cstdfloat/cstdfloat_limits.hpp>
+//
+// Note this header includes no other headers, this is
+// important if this test is to be meaningful:
+//
+#include "test_compile_result.hpp"
+
+void compile_and_link_test()
+{
+    #ifdef __float128
```
jzmaddock, line 18: as above __float128 is not a macro.

test/compile_test/interpolators_cardinal_trig_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/cardinal_trigonometric.hpp>
```
jzmaddock, line 9: Do we still need to instantiate something here?

test/compile_test/interpolators_cubic_hermite_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/cubic_hermite.hpp>
```
jzmaddock, line 9: Also need to instantiate something here?

test/compile_test/interpolators_makima_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/makima.hpp>
```
jzmaddock, line 9: Instantiation?

jzmaddock: Something like:

```
// we can't help but include a std lib header here:
#include <vector>

int main()
{
  std::vector<double> x{1, 5, 9 , 12};
  std::vector<double> y{8,17, 4, -3};
  using boost::math::interpolators::makima;
  makima<std::vector<double>> spline(std::move(x), std::move(y));
  // evaluate at a point:
  double z = spline(3.4);
  // evaluate derivative at a point:
  double zprime = spline.prime(3.4);
}
```

Which is straight out of the docs, might do it - I hope!!

The other interpolators all have the same interface... so hopefully just change the name ;)

test/compile_test/interpolators_pchip_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/pchip.hpp>
```
jzmaddock, line 9: Instantiation?

test/compile_test/interpolators_quintic_hermite_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/quintic_hermite.hpp>
```
jzmaddock, line 9: Instantiation?

test/compile_test/interpolators_septic_hermite_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/septic_hermite.hpp>
```
jzmaddock, line 9: Instantiation?

test/compile_test/interpolators_vector_barycentric_rational_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/vector_barycentric_rational.hpp>
```
jzmaddock, line 9: Instantiation?

test/compile_test/interpolators_whittaker_shannon_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// A sanity check that this file
+// #includes all the files that it needs to.
+//
+#include <boost/math/interpolators/whittaker_shannon.hpp>
```
jzmaddock, line 9: Instantiation?

test/compile_test/quad_exp_sinh_incl_test.cpp
```diff
@@ -3,7 +3,7 @@
 //  Boost Software License, Version 1.0. (See accompanying file
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 //
-// Basic sanity check that header <boost/math/special_functions/gamma.hpp>
+// Basic sanity check that header
 // #includes all the files that it needs to.
 //
 #include <boost/math/quadrature/exp_sinh.hpp>
```
jzmaddock, line 9: Instantiation?

Probably just copy from gauss test?

jzmaddock: Ooops ignore that - my bad!!

test/compile_test/tools_ulps_plot_incl_test.cpp
```diff
@@ -0,0 +1,9 @@
+//  Copyright Matt Borland 2021.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Basic sanity check that header
+// #includes all the files that it needs to.
+//
+#include <boost/math/tools/ulps_plot.hpp>
```
jzmaddock, line 9: Probably need to instantiate something on this one too?

jzmaddock:
@mborland : many thanks for all the hard work that's obviously gone into this - I've made some nit-picky comments above for a few of the new tests that don't instantiate anything at present.  Other than that, this is all really good stuff!

test/compile_test/interpolators_makima_incl_test.cpp
mborland: I should have mentioned I skipped the interpolators. I am sure they could be made to work with a c-array or pointers, but not sure it was worth the effort. Adding `std::vector` should work just fine.

test/compile_test/interpolators_vector_barycentric_rational_incl_test.cpp
mborland: @NAThompson `vector_barycentric_rational` is the only one of the interpolators not in the `interpolators` namespace. Do you want me to move it or leave it? Currently it is in just `boost::math`.

include/boost/math/cstdfloat/cstdfloat_cmath.hpp
```diff
@@ -879,7 +879,9 @@ namespace std
    using boost::math::cstdfloat::detail::fabs;
 
 #if !(defined(_GLIBCXX_USE_FLOAT128) && defined(__GNUC__) && (__GNUC__ >= 7))
+#if defined(__clang__) && !(!defined(__STRICT_ANSI__) && defined(_GLIBCXX_USE_FLOAT128)) // workaround for clang using libstdc++
```
mborland, line 882: @jzmaddock I am not sure if this is correct. `std_abs.h` defines `std::abs(__float128_)` and clang throws an error for collision that GCC does not. `!(!defined(__STRICT_ANSI__) && defined(_GLIBCXX_USE_FLOAT128))` is just negated of what allows the float128 specialty in libstdc++. Thoughts?

test/compile_test/interpolators_vector_barycentric_rational_incl_test.cpp
NAThompson: Yes, that was a screw up on my part.

NAThompson:
All my concerns are resolved.

include/boost/math/cstdfloat/cstdfloat_cmath.hpp
jzmaddock: It looks reasonably correct - there is some inevitable impiracle guess work here that I guess we just can't avoid.  The tests are all passing though, so yes, looks good to me.

jzmaddock:
All looks good to me too - @mborland any reason not to merge?

mborland:
@jzmaddock It have not run a full CI cycle since changing `cstdfloat_cmath` to fix standalone. Made a minor change to cycle to make sure it does not affect older versions of Clang/GCC. Should be good to if this run is clean.

mborland:
CI on this is clean again

PR #607, opened 2021-04-11
Allow Bernoulli numbers to support different precisions in different threads

jzmaddock:
Fixes issue found while working on thread safe variable precision multiprecision types.

Without this we have a race condition where a thread may spot that the cache is of the wrong precision and invalidate it while another thread is still using it.

### DIFF
--- a/include/boost/math/special_functions/detail/bernoulli_details.hpp
+++ b/include/boost/math/special_functions/detail/bernoulli_details.hpp
@@ -144,6 +144,40 @@ inline T tangent_scale_factor()
    return tools::min_value<T>() * 16;
 }
 //
+// Initializer: ensure all our constants are initialized prior to the first call of main:
+//
+template <class T, class Policy>
+struct bernoulli_initializer
+{
+   struct init
+   {
+      init()
+      {
+         //
+         // We call twice, once to initialize our static table, and once to
+         // initialize our dymanic table:
+         //
+         boost::math::bernoulli_b2n<T>(2, Policy());
+
+         try{
+            boost::math::bernoulli_b2n<T>(max_bernoulli_b2n<T>::value + 1, Policy());
+         } catch(const std::overflow_error&){}
+
+         boost::math::tangent_t2n<T>(2, Policy());
+      }
+      void force_instantiate()const{}
+   };
+   static const init initializer;
+   static void force_instantiate()
+   {
+      initializer.force_instantiate();
+   }
+};
+
+template <class T, class Policy>
+const typename bernoulli_initializer<T, Policy>::init bernoulli_initializer<T, Policy>::initializer;
+
+//
 // We need something to act as a cache for our calculated Bernoulli numbers.  In order to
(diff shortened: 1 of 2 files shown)


include/boost/math/special_functions/detail/bernoulli_details.hpp
```diff
@@ -530,18 +496,35 @@ class bernoulli_numbers_cache
...
+template <class T, class Policy>
+inline bernoulli_numbers_cache<T, Policy>& get_bernoulli_numbers_cache_imp(const std::integral_constant<int, 0>&)
+{
+   //
+   // If we get here, then the precision of T is unknown at compile time, which means
+   // there's a fair chance it's a variable precision type.  In that case we *must*
+   // make our cache thread local, that way we can have different threads operating
+   // at differing precisions:
+   //
+   static BOOST_MATH_THREAD_LOCAL bernoulli_numbers_cache<T, Policy> data;
```
mborland, line 518: It makes no difference, but `BOOST_MATH_THREAD_LOCAL` always expands to `thread_local`. I should have axed that macro like the other pre-C++11 workarounds.

jzmaddock: Nod.  I did wonder about that, I guess I was worrying about platforms which might not have any thread support at all, but I guess those can just `#define thread_local` and be done with it.

jzmaddock:
OK, we have a C++11 issue to think about here: it turns out that `thread_local` with gcc on windows is quite badly broken: in particular when dealing with non-trivial types.  It seems to be deleting the variable unconditionally and not checking to see if it was actually constructed :(    See for example https://github.com/boostorg/math/pull/607/checks?check_run_id=2318588556.

I also worry we may see other platforms with similar issues down the road.

I'll see if I can find a workaround, but anyone have any thoughts?

mborland:
`boost/math/tools/config.hpp` has a macro defined with the same linked issue:

```
//
// Some mingw flavours have issues with thread_local and types with non-trivial destructors
// See https://sourceforge.net/p/mingw-w64/bugs/527/
//
#if (defined(__MINGW32__) || defined(__MINGW64__)) && !defined(_REENTRANT) && !defined(__clang__)
#  define BOOST_MATH_NO_THREAD_LOCAL_WITH_NON_TRIVIAL_TYPES
#endif
```

Maybe just disable non-trival type usage on mingw? A similar issue was only [recently solved](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83562).

jzmaddock:
> Maybe just disable non-trival type usage on mingw (like in zeta)? 

Ha!  I'd forgotten all about that: thanks Matt!

> A similar issue was only recently solved.

Nod.  Strangely I'm unable to reproduce locally, so at least some mingw versions are OK, they all report as Mingw-8.1.0 though :(

jzmaddock:
Closed in favour of https://github.com/boostorg/math/pull/609.

PR #610, opened 2021-04-21
Fixes for special_fun suite warnings

mborland:
Several instances of `-Wdeprecated-copy` and `-Wunused-variable` plus one instance of `-Wimplicit-const-int-float-conversion`. I am sure there will be other PRs here and in Boost.Random to address instances of deprecated copy.

### DIFF
--- a/include/boost/math/special_functions/detail/hypergeometric_asym.hpp
+++ b/include/boost/math/special_functions/detail/hypergeometric_asym.hpp
@@ -44,7 +44,7 @@
         }
         else
         {
-           e = z > (std::numeric_limits<long long>::max)() ? (std::numeric_limits<long long>::max)() : lltrunc(z, pol);
+           e = z > static_cast<T>((std::numeric_limits<long long>::max)()) ? (std::numeric_limits<long long>::max)() : lltrunc(z, pol);
            log_scaling += e;
            prefix = exp(z - e);
         }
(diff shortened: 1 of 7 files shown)


mborland:
@jzmaddock Pending further review this should be good to go.

include/boost/math/special_functions/detail/hypergeometric_asym.hpp
```diff
@@ -44,7 +44,7 @@
         }
         else
         {
-           e = z > (std::numeric_limits<long long>::max)() ? (std::numeric_limits<long long>::max)() : lltrunc(z, pol);
+           e = z > static_cast<T>((std::numeric_limits<long long>::max)()) ? (std::numeric_limits<long long>::max)() : lltrunc(z, pol);
```
jzmaddock, line 47: What's the warning on this one?

My only (very slight) concern, is that for multiprecision types it is in general cheaper to compare to an integer, than to create a temporary T and compare.  In this particular case I very much doubt that anyone will notice the difference in a hypergeometric calculation, but in general we need to be a bit careful about casting integers to T's just out of habit.

mborland:
```
../../../boost/math/special_functions/detail/hypergeometric_asym.hpp:47:21: warning: implicit conversion from 'long long' to 'double' changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-const-int-float-conversion]
           e = z > ((std::numeric_limits<long long>::max)()) ? (std::numeric_limits<long long>::max)() : lltrunc(z, pol);
```

jzmaddock: I bet there are a few of those, lets ignore it for now and merge - thanks for this!

PR #615, opened 2021-04-26
Fix for ambiguous abs(__float128) with old gcc

mborland:
An example is [here](https://drone.cpp.al/boostorg/math/180/86/2).


jzmaddock:
Thanks Matt, weird this didn't show up before?

mborland:
The failure only shows up on Drone which does not run on PRs from forks. Is it possible to enable those runs?

jzmaddock:
>The failure only shows up on Drone which does not run on PRs from forks. Is it possible to enable those runs?

That's a mistake, hopefully fixed now.

PR #616, opened 2021-04-26
Remove vc71 fix.

NAThompson:
This is a "discussion commit", because I don't know what this was for. It might be useful, or was useful at one time.

But now that boost.math will be distributed alone, I want the directory structure a little cleaner so new users don't get distracted.


mborland:
There is a [note about VC7.1](https://beta.boost.org/doc/libs/1_33_1/tools/build/v1/vc-7_1-tools.html) on an ancient version of the boost build docs. A compiler that old of a version is in no danger of compiling the current library.

jzmaddock:
Nod.  This can all go now since vc7.1 is no longer supported.

Issue #614, opened 2021-04-25
CMake superproject tests are failing because of Math


pdimov:
```
Scanning dependencies of target compile_tests
[ 17%] Building CXX object libs/math/test/compile_test/CMakeFiles/compile_tests.dir/compl_abs_incl_test.cpp.o
/home/travis/build/boostorg/boost/libs/math/test/compile_test/compl_abs_incl_test.cpp:9:39: fatal error: boost/math/complex/fabs.hpp: No such file or directory
compilation terminated.
libs/math/test/compile_test/CMakeFiles/compile_tests.dir/build.make:62: recipe for target 'libs/math/test/compile_test/CMakeFiles/compile_tests.dir/compl_abs_incl_test.cpp.o' failed
make[2]: *** [libs/math/test/compile_test/CMakeFiles/compile_tests.dir/compl_abs_incl_test.cpp.o] Error 1
CMakeFiles/Makefile2:14516: recipe for target 'libs/math/test/compile_test/CMakeFiles/compile_tests.dir/all' failed
make[1]: *** [libs/math/test/compile_test/CMakeFiles/compile_tests.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
The command "cmake --build ." exited with 2.
```
https://travis-ci.org/github/boostorg/boost/jobs/768322753

In addition, the `compile_tests` target needs to be prefixed with `boost_math-` by convention; CMake targets are global.

jzmaddock:
@mborland , @pdimov is this now fixed?

mborland:
The CMake build does not make it to math anymore so I am not sure. [Currently fails at Boost.JSON.](https://travis-ci.org/github/boostorg/boost/jobs/769225032#L1226)

pdimov:
I reported this failure to Boost.JSON. In the meantime you can check that Math works by
```
mkdir __build; cd __build
cmake -DBOOST_ENABLE_CMAKE=ON -DBUILD_TESTING=ON -DBOOST_INCLUDE_LIBRARIES=math ..
cmake --build .
```
This gives me a bunch of warnings
```
C:\boost-git\develop\libs\math\include\boost/math/tools/is_standalone.hpp(12,1): warning C4005: 'BOOST_MATH_STANDALONE'
: macro redefinition [C:\boost-git\develop\__build\libs\math\test\compile_test\boost_math-compile_tests.vcxproj]
[...]

PR #617, opened 2021-04-26
Fix for CMake superproject failure

mborland:
See issue #614.


mborland:
@jzmaddock Pending review I think this is good. I see that drone is running now on this PR so thank you for that.

jzmaddock:
My only question is have you run the super-project CMake project to check there are no longer conflicts?

mborland:
I did; locally the super-project build succeeds now.

jzmaddock:
Thanks Matt!

PR #618, opened 2021-04-27
Disable test_gamma_mp on sanitizer runs

mborland:
Sporadically the Drone [USAN test fails](https://drone.cpp.al/boostorg/math/183/2/2) at `test_gamma_mp` for out of memory error. This disables `test_gamma_mp` on sanitizer runs.


ckormanyos:
>  out of memory error

It appears as though the test allocates 5GB memory? I could try to take a look if I ever get a free _gamma moment_.

mborland:
@jzmaddock This should be good pending review. The CI failures are on Github's end.

PR #620, opened 2021-04-30
Minimax and distribution_explorer to examples.

NAThompson:
And while we're at it: What do these do? I've never seen these run or tested . . .


jzmaddock:
The Remez code is documented here: https://www.boost.org/doc/libs/1_42_0/libs/math/doc/sf_and_dist/html/math_toolkit/toolkit/internals2/minimax.html

I'd rather not loose it, since it contains the record of how the minimax approximations were derived (and is the starting point for new ones).  It could be moved under tools/minimax if we want to tidy up the root directory.

The distribution-explorer code is a .NET application that @pabristow wrote to illustrate the behaviour of our distribution classes.  I confess I'm unable to find any docs on it - Paul?

pabristow:
@jzmaddock - I deny writing it! (My eyes are dim, I cannot see ... ... ) But I may have influenced its production.  I have no reason to believe that it no longer works and it may still be a useful guide to anyone producing C# / .NET code that uses Boost.Math. Was one original purpose to show that this was possible?

Docs are at distexplorer.qbk and don't look too out of date. 

It's not hurting anyone?

NAThompson:
>  It could be moved under tools/minimax if we want to tidy up the root directory.

Done.

> It's not hurting anyone?

Just moved to the `examples` directory-not deleted.

Issue #621, opened 2021-05-02
We may need to be able to disable threading support after all.


jzmaddock:
Leaving aside @ckormanyos building on the metal, msvc with the /clr options does not support any of the std threading headers: `<mutex>`, `<thread>`, `<atomic>` etc.

Currently building in this mode triggers a #error in cxx03_warn.hpp which prevents **any** part of the library being used.

ckormanyos:
> on the metal, msvc with the /clr options does not support any of the std threading headers

Oh yes, I had completely forgotten that. It was really, really a long time ago. At that time, I wanted to export some functions to Microsoft's so-called _managed C++_, using `cli_ext`headers. This is indeed a limitation.

On my huge list of things to do was to make a cursory glance at what actually needs threads, muteses and atomics. For the special functions, it seems rather limited to `tgamma`and friends. I have not seriously looked into other parts like quadrature, etc.

I would be interested if we could pursue a way to handily and reliably disable threading, for metal, python exports, CLR/.NET exports and the lot.

mborland:
This seems to be pretty well supported at this time. Closing to cleanup our open issues.

PR #623, opened 2021-05-04
Add standalone BOOST_HAS_THREADS

mborland:
As mentioned in issue #621. Removes hard errors for not having thread support headers. Default assumption is that a particular platform has threads available.


ckormanyos:
This seems cool. What is planned for code actually using something like `std::mutex`? Will it simply be disabled with compiler switch or will the function be stubed/mocked with a dummy?

I have been reticent and slow to do my part on bare metal. But we're probably looking at something similar for I/O streaming and exceptions.

mborland:
> This seems cool. What is planned for code actually using something like `std::mutex`? Will it simply be disabled with compiler switch or will the function be stubed/mocked with a dummy?
> 

I think that disabling via compiler switching is the best route. There were previously branches in some functions (e.g. bernoulli numbers) that I can just put back with this change.

@jzmaddock Pending further review the CI is clean.

include/boost/math/tools/config.hpp
```diff
@@ -49,6 +49,17 @@
 #define BOOST_STRINGIZE(X) BOOST_DO_STRINGIZE(X)
 #define BOOST_DO_STRINGIZE(X) #X
 
+// Detect thread support via STL implementation
```
jzmaddock, line 52: Thanks Matt, however I suspect this is going to cause macro-redefinition conflicts unless guarded by a check that we're in standalone mode.

BTW BOOST_DISABLE_THREADS, while set inside Boost.Config is intended as a macro the user can set on the command line to force thread support off (for example if they just don't need it, even though it's supported, see https://www.boost.org/doc/libs/1_76_0/libs/config/doc/html/index.html#boost_config.configuring_boost_for_your_platform.user_settable_options), so there should be a check to see it's set already and configure appropriately.

A small point of pedantry: most std libraries will supply `<mutex>` etc even though they may be empty or non-functional in the current compilation mode.  Which is to say further down the road we may need to pull in some of the logic from Boost.Config here for popular platforms at least.

I might also be inclined to set BOOST_NO_CXX11_HDR_ATOMIC, BOOST_NO_CXX11_HDR_FUTURE, BOOST_NO_CXX11_HDR_MUTEX, BOOST_NO_CXX11_HDR_THREAD and BOOST_NO_CXX11_THREAD_LOCAL when BOOST_DISABLE_THREADS is set so that in code we can support partial conformance if needs be.  So I guess it would look something like:

```
#ifdef BOOST_MATH_STANDALONE
#ifdef BOOST_DISABLE_THREADS
// no threads!
#elif defined(__has_include)
#  if !__has_include(<thread>) || !__has_include(<mutex>) || !__has_include(<future>) || !__has_include(<atomic>)
#     define BOOST_DISABLE_THREADS
#  else
#     define BOOST_HAS_THREADS
#  endif 
#else
#  define BOOST_HAS_THREADS // The default assumption is that the machine has threads
#endif // Thread Support
//
// Compiler specific config may go here later....
//
// When threads are disabled, set the individual Boost.Config thread related macros:
//
#ifdef BOOST_DISABLE_THREADS
#  define BOOST_NO_CXX11_HDR_ATOMIC
#  define BOOST_NO_CXX11_HDR_FUTURE
#  define BOOST_NO_CXX11_HDR_MUTEX
#  define BOOST_NO_CXX11_HDR_THREAD
#  define BOOST_NO_CXX11_THREAD_LOCAL
#endif // BOOST_DISABLE_THREADS
#endif // BOOST_MATH_STANDALONE
```

ckormanyos:
The following suggestion might seem ridiculous, it might be reasonable. I'm not really experienced enough to judge...

I was thinking we could mentally or via code search identify all uses of `<atomic>`, `<mutex>` and `<thread>` and see if we could write dummy placeholders for them and configure to use real function/dummy function at one place?

mborland:
@jzmaddock This now is clean with your requested changes.

jzmaddock:
Thanks Matt!

PR #624, opened 2021-05-06
Fixes for compile tests

mborland:
Fixes for failures from -Wzero-length-array (see #614 discussion) and other warnings.


NAThompson:
All the diffs I understand I approve of!

mborland:
This should be good to go. CI is clean

include/boost/math/special_functions/rsqrt.hpp
```diff
@@ -15,7 +15,7 @@ template<typename Real>
 inline Real rsqrt(Real const & x)
 {
     using std::sqrt;
-    if constexpr (std::is_same_v<Real, float> || std::is_same_v<Real, double> || std::is_same_v<Real, long double>)
+    if constexpr (std::is_floating_point_v<Real>)
```
jzmaddock, line 18: @NAThompson : was the original intention here to exclude `__float128`?  If so replacing with is_floating_point *might* fail in -std=gnu++ mode.   Just double checking...

include/boost/math/special_functions/nonfinite_num_facets.hpp
```diff
@@ -414,7 +414,7 @@ namespace boost {
         switch(peek_char(it, end, ct)) {
         case 'q':
         case 's':
-          if(flags_ && legacy)
+          if(flags_)
```
jzmaddock, line 417: Looking at the rest of the code, I assume this should be `flags_&legacy` ?  @pabristow ?

pabristow: I deny writing it! Looks reasonable, though the comments seem not very specific to me.  I can't see any reason to restrict to only legacy here. And the tests are thin.  And testing is difficult on account of the too many possible floating-point types and platforms.

Lesson to Standards Writers - make up your mind and don't duck out by making stuff implementation-dependent!

include/boost/math/special_functions/rsqrt.hpp
NAThompson: Amazingly enough I remember this! Yes, that was the goal.

As you can see in the `else` branch, the `rsqrt` iteration starts with the `long double` approximation and then refines it with the Newton method for `1/sqrt(x)`. I verified that it is slower to call `1/sqrt(x)` for `float128` than to use the `long double` approximation and Newton iterate.

mborland:
The one erroneous diff has been reverted so should be good to go pending further review.

PR #625, opened 2021-05-08
Put back single threaded bernoulli code path

mborland:


mborland:
@jzmaddock Pending review this should be good to go. It is not much more than copying and pasting the code that used to be in here.

jzmaddock:
Thanks again Matt!

PR #626, opened 2021-05-09
Add thread guard to polygamma

mborland:
Continued work on issue #621. The call to `std::mutex` and `std::lock_guard` were already wrapped so only the header inclusion needed changed. Also make the arrays `constexpr`


mborland:
@jzmaddock Pending review this should be good to go.

jzmaddock:
Cheers Matt!

PR #627, opened 2021-05-10
Update acosh.hpp

Taqaddusshafi:
Removed spaces


jzmaddock:
Why?  Formatting looks fine to me - remember those includes are inside a #if block.

PR #629, opened 2021-05-12
Thread guards in stats

mborland:
Allows for use of stats functions in single threaded environments. See issue #621

### DIFF
--- a/include/boost/math/statistics/bivariate_statistics.hpp
+++ b/include/boost/math/statistics/bivariate_statistics.hpp
@@ -11,8 +11,6 @@
 #include <tuple>
 #include <type_traits>
 #include <stdexcept>
-#include <future>
-#include <thread>
 #include <vector>
 #include <algorithm>
 #include <cmath>
@@ -22,8 +20,10 @@
 
 // Support compilers with P0024R2 implemented without linking TBB
 // https://en.cppreference.com/w/cpp/compiler_support
-#ifndef BOOST_NO_CXX17_HDR_EXECUTION
+#if !defined(BOOST_NO_CXX17_HDR_EXECUTION) && defined(BOOST_HAS_THREADS)
 #include <execution>
+#include <future>
+#include <thread>
 #define EXEC_COMPATIBLE
 #endif
 
@@ -59,6 +59,8 @@ ReturnType means_and_covariance_seq_impl(ForwardIterator u_begin, ForwardIterato
 
     return std::make_tuple(mu_u, mu_v, cov/i, i);
 }
+
+#ifdef EXEC_COMPATIBLE
 
 // Numerically stable parallel computation of (co-)variance
 // https://dl.acm.org/doi/10.1145/3221269.3223036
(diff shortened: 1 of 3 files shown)


mborland:
@jzmaddock I am not quite sure I understand the rash of clang 9 failures on this one. All of the C++20 failures look to be stemming from the standard library, and the quadrature failures are all memory access violations. All the drone tests pass (including ASAN) [except one](https://drone.cpp.al/boostorg/math/209/132/1) which is a failure on Drone's end.

jzmaddock:
It seems to be *all* the Monte Carlo tests that are failing, @NAThompson any ideas?

NAThompson:
Wow! I have *no* clue. It looks like this change doesn't have anything to do with the Monte-Carlo integration. . .

We did have an ASAN build that should've caught this before if it was in user-space, right?

Question: Is there utility in having a single-threaded Monte-Carlo integrator? I personally feel the design of that class is *fundamentally* multithreaded.

jzmaddock:
I have a dim memory that I disabled sanitizers for the Monte Carlo tests because they take way too long to run (and/or run the machine out of memory).  Might be worth running them locally to see if that gives any clue....  currently struggling with my own (unrelated) thread safety bug yours...

mborland:
> I have a dim memory that I disabled sanitizers for the Monte Carlo tests because they take way too long to run (and/or run the machine out of memory). Might be worth running them locally to see if that gives any clue.... currently struggling with my own (unrelated) thread safety bug yours...

Locally using Clang-12's ASAN yields clean results on the Monte Carlo tests.

NAThompson:
Ok, my vote is to *not* disable threads for Monte-Carlo integration. There's little sense in it-the design is fundamentally multithreaded.

mborland:
> Ok, my vote is to *not* disable threads for Monte-Carlo integration. There's little sense in it-the design is fundamentally multithreaded.

I can revert the change and have it throw an error in single threaded environments.

mborland:
> > Ok, my vote is to _not_ disable threads for Monte-Carlo integration. There's little sense in it-the design is fundamentally multithreaded.
> 
> I can revert the change and have it throw an error in single threaded environments.

So after review the thread guards were not added to Monte-Carlo integration in #628. Last time it was touched was over 6 weeks ago by #600. I am still stumped as to why clang-9 is so unhappy.

NAThompson:
I repeat my unpopular opinion: CI is here to serve us, we don't live to serve CI. If we've run it locally under thread sanitizer, and it's clean, it seems like this is unlikely any users will be bitten by it.

jzmaddock:
I have some inscrutable thread-based failures with clang-9 in Multiprecision also, so I've pushed this: https://github.com/boostorg/math/pull/631 to sanity check the std lib thread support.

jzmaddock:
CI has now been adjusted on develop, @mborland is this now good to go?

jzmaddock:
Nod.  If you don't mind merging develop into this just be on the safe side, that would great!

PR #631, opened 2021-05-17
Add threading sanity check to probe clang-9 failures.

jzmaddock:
See: https://github.com/boostorg/math/pull/629.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        compiler: [ g++-9, g++-10, clang++-9, clang++-10 ]
+        compiler: [ g++-9, g++-10, clang++-10 ]
         standard: [ c++11, c++14, c++17, c++2a ]
         suite: [ float128_tests, special_fun, distribution_tests, misc, quadrature, mp, interpolators, autodiff, ../example//examples, ../tools ]
     steps:
--- a/reporting/accuracy/Jamfile.v2
+++ b/reporting/accuracy/Jamfile.v2
@@ -80,42 +80,42 @@ rule all-tests {
      local result ;
      for local source in [ glob test*.cpp ]
      {
-         result += [ run $(source) /boost/regex//boost_regex /boost/system /boost/filesystem /boost/test//boost_unit_test_framework 
+         result += [ run $(source) /boost/system /boost/filesystem /boost/test//boost_unit_test_framework 
          : : : 
           [ check-target-builds has_gsl : <define>ALWAYS_TEST_DOUBLE : ] 
           <target-os>linux:<linkflags>-lpthread
           <target-os>linux:<linkflags>-lrt
           <toolset>gcc:<linkflags>$(OTHERFLAGS) ]
          ;
-         result += [ run $(source) /boost/regex//boost_regex /boost/system /boost/filesystem /boost/test//boost_unit_test_framework 
+         result += [ run $(source) /boost/system /boost/filesystem /boost/test//boost_unit_test_framework 
          : : : [ check-target-builds has_cxx17_cmath : <define>TEST_CXX17_CMATH : <build>no ] 
           <target-os>linux:<linkflags>-lpthread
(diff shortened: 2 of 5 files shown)


mborland:
So clang 9 looks broken. Same errors with the spaceship operator in the C++20 runs, and your sanity check segfaults.

ckormanyos:
> Same errors with the spaceship operator in the C++20 runs

If this helps, I also found this phenomenon in another completely unrelated project and simply removed clang 9 runs on C++20.

jzmaddock:
This is a bit of a bloodbath - @ckormanyos it's not just C++20, my simple Boost-free multi threading test segfaults in C++17 too.

Since this all used to work, I assume CI has updated something on the ubuntu image, we have it fixed to ubuntu-20.04 though, so I would expect this to be OK?  Plus, I've just checked my Ubunutu VirtualBox and that's 20.04 and up to date and seems to work OK.

ckormanyos:
> assume CI has updated something on the ubuntu image

That rings a bell @jzmaddock. To be honest, all of a sudden, my nightly build of real-time-c++ companion code stopped working. This is shown in the pic below. Instead of investigating the problem, I eliminated the offending compiler runs.

I also got the impression that something in the CI machine changed. But i side stepped the problem before even really looking into it.

I can dive a bit deeper into the nature of what's really going on.

grafik

ckormanyos:
Here is an [example of errors](https://github.com/ckormanyos/real-time-cpp/runs/2581454568?check_suite_focus=true#step:5:48) that i was getting on that day. I can see what's similar with the failures here.

ckormanyos:
The errors are the same. The ones I got and the ones in this project.

It starts off with

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/chrono:1345:24: error: unknown type name 'strong_ordering'
```

Then moves on to comparison failuse of time epochs with the spaceship operator.

I can also try and take a closer look

ckormanyos:
Sorry for going on, but I need to provide what I do know. I need to be more specific, the failures that seem to be in the latest run here and those that I encountered elsewhere were clang-9 C++2a, not 20.

jzmaddock:
I think I can tell you what the difference is, but not *why*.

clang-9 is passing on drone: https://drone.cpp.al/boostorg/math/212/33/2
Same configuration and Ubuntu version is failing on GHA: https://github.com/boostorg/math/pull/631/checks?check_run_id=2601142494

The latter has GNU libstdc++ version 20210427 the former GNU libstdc++ version 20191114 and not surprisingly, clang-9 can't cope with a libstdc++ released well after it was.  I'm not sure which gcc version 20210427 corresponds to, but it doesn't appear to be a gcc-9 flavour which is what I would have expected.  Even the gcc-10 builds have GNU libstdc++ version 20200808 on GHA.

I think since we have clang-9 working on drone, we'll just remove it from GHA?

ckormanyos:
> The latter has GNU libstdc++ version 20210427 the former GNU libstdc++ version 20191114 and not surprisingly, clang-9 can't cope with a libstdc++ released well after it was. I'm not sure which gcc version 20210427 corresponds to, but it doesn't appear to be a gcc-9 flavour which is what I would have expected. Even the gcc-10 builds have GNU libstdc++ version 20200808 on GHA.

Great description @jzmaddock that fits perfectly with my empirical observations as well.

One strange point is that this change appears to have happened on GHA literally _overnight_ a little less than a week ago.

> I think since we have clang-9 working on drone, we'll just remove it from GHA?

Yes. I would recommend to simply remove the offending compiler configurations, otherwise you're just testing the setup of the compiler(s) instead of the actual code.

Some folks actually make docker images of known-to-work compiler topologies for this exact reason, but i really think this is beyond the scope of work that I could invest in this CI.

jzmaddock:
>One strange point is that this change appears to have happened on GHA literally overnight a little less than a week ago.

The images are updated weekly with the latest package releases.  It's just weird I can't reproduce locally.

jzmaddock:
I forgot to say, drone images are updated less often, but they do get updated, we may yet see the same issue appear there, which would be super annoying :(

PR #632, opened 2021-05-20
Remove lanczos22.

jzmaddock:
It was only used for 34 and 35 digit counts and had terrible accuracy at fixed precision.
See https://github.com/boostorg/multiprecision/pull/326#issuecomment-844997628.
Fortunately the "MP" suffixed traits appear fine, though further testing inside Multiprecision will follow.


ckormanyos:
This is cool.

I'm not sure, but I wonder @jzmaddock if we remove `lanczos22`, does that influence multiprecision`s specialization, which references `lanczos22` [by name here](https://github.com/boostorg/multiprecision/blob/bb261007705b509db6b81e54127383e96abd1878/include/boost/multiprecision/detail/big_lanczos.hpp#L29)?

Do you think it might ultimately be possible to reduce that specialization in multiprecision altogether?

jzmaddock:
One failure is a network issue, so merging.

ckormanyos:
> I think we should be able to remove altogether yes, but lets get the change in Math develop and bash the tests into better shape first.

Understood. This is moving along very well and quickly. Thanks John.

PR #640, opened 2021-06-05
nonfinite_num_get facet fix: honor "legacy" flag to enable "nanq"+"nans"

jblumsch:
A hopefully obvious one-line fix to "Facets for Floating-Point Infinities and NaNs".

The documentation states that the "legacy" flag enables support for "nanq" and "nans", but any other flag is not supposed to do so.

The "legacy" flag got lost in 16c57be3 "Fixes for compile tests", but the affected code was broken from the very beginning.


NAThompson:
Did this slip through the cracks? I have really no understanding of what this code does.

@jzmaddock : Could you take a look as well?

jzmaddock:
Looks right to me, we should merge once CI completes.

Thanks for the heads up!

PR #641, opened 2021-06-15
Bilinear uniform interpolation

NAThompson:
Still an OOB access somewhere, but a good start on this one.


NAThompson:
Ok, don't know why I did to get those 5 other commits in there; will figure it out shortly . . .

Issue #639, opened 2021-06-03
hypergeometric kurtosis differs from Wolfram and scipy and Wikipedia


pabristow:
I have reported in response to a complaint 

https://github.com/scipy/scipy/pull/14153#issuecomment-853942708

@jzmaddock Should I prepare a PR to correct this? and provide a new test?

It could just be the patch in the link above, or some update/enhancement to the docs might be better?

(I won't be able to work on this until 15 Jun).

mckib2:
Appears to be a duplicate of #637, linking to this one

PR #642, opened 2021-06-16
Hypergeomkurtosis

pabristow:
In response to https://github.com/boostorg/math/issues/639 I've changed the expression to calculate the hypergeometric distribution kurtosis excess to use Wolfram Alpha version and confirmed that it agrees with a separate implementation of Wikipedia formula and agrees with Python scipy as well.  Used boost::multiprecision::cpp_bin_float_quad in a separate test.

Updated tests for moments using Wolfram values. Passes local tests msvc, gcc and clang. 

Matching changes to documentation, but unable to update equations as don't have MathML software (and adding the formula for kurtosis is tedious and error-prone so replace by a reference to Wikipedia and Wolfram).

Docs rebuilt and running boost.Inspect reveals no broken links.


jzmaddock:
That all looks fine, merging...

jzmaddock (approved):

jzmaddock:
Ah... not merging after all, it's marked as draft, so you will have to do it.

pabristow:
Sorry - I didn't realise the implications of being marking as draft.

Merged OK.

PR #644, opened 2021-06-19
solve.hpp: fabs -> std::abs in templated function

jwuttke:
correction needed to make boost::math::tools::solve<T> work with T=long double


NAThompson:
Maybe should be:

```cpp
using std::abs;
T err = abs(delta[i] / b[i]);
```

so it works with quad/multiprecision as well?

NAThompson:
Looks good to me. @jzmaddock ?

jzmaddock:
That's fine, thought probably just needed a `using std::fabs` ;)

NAThompson:
Build failure was a timeout.

PR #645, opened 2021-06-24
improve mode of the non central chi squared algorithm

vindex10:
Hello! I made a little research on the mode of the non-central chi2, and it turned out that the boost library could benefit from it. The main change is aimed at tuning the starting point of the iterative scheme that searches for the mode. Please have a look at the theoretical bases of the change and also at some benchmarking plots in the preprint: https://arxiv.org/abs/2106.12276

The corresponding code that I used for benchmarking you can find on the github repo here:
https://github.com/vindex10/ncxmax/tree/master/boost-test

### DIFF
--- a/include/boost/math/distributions/non_central_chi_squared.hpp
+++ b/include/boost/math/distributions/non_central_chi_squared.hpp
@@ -839,7 +839,7 @@ namespace boost
             Policy()))
                return (RealType)r;
          bool asymptotic_mode = k < l/4;
-         RealType starting_point = asymptotic_mode ? k + l - 3 : 1 + k;
+         RealType starting_point = asymptotic_mode ? k + l - RealType(3) : RealType(1) + k;
          return detail::generic_find_mode(dist, starting_point, function);
       }
 
--- a/include/boost/math/distributions/non_central_chi_squared.hpp
+++ b/include/boost/math/distributions/non_central_chi_squared.hpp
@@ -838,7 +838,9 @@ namespace boost
             &r,
             Policy()))
                return (RealType)r;
-         return detail::generic_find_mode(dist, 1 + k, function);
+         bool asymptotic_mode = k < l/4;
+         RealType starting_point = asymptotic_mode ? k + l - 3 : 1 + k;
+         return detail::generic_find_mode(dist, starting_point, function);
       }
 
       template <class RealType, class Policy>
@@ -994,6 +996,3 @@ namespace boost
 #include <boost/math/distributions/detail/derived_accessors.hpp>
 
 #endif // BOOST_MATH_SPECIAL_NON_CENTRAL_CHI_SQUARE_HPP
-
-
-
--- a/reporting/performance/test_distributions_mode.cpp
+++ b/reporting/performance/test_distributions_mode.cpp
@@ -53,7 +53,7 @@ BENCHMARK_TEMPLATE(test_mode_2param, long double, non_central_chi_squared_distri
 
 BENCHMARK_TEMPLATE(test_mode_2param, long double, non_central_chi_squared_distribution)->ArgsProduct({
     benchmark::CreateRange(4, 4096, /*multi=*/2),
-    {0, 30, 100, 500}
+    {1, 30, 100, 500}
 })->Name("fixed_nc");
 
 BENCHMARK_TEMPLATE(test_mode_2param, long double, non_central_chi_squared_distribution)
--- a/reporting/performance/test_distributions_mode.cpp
+++ b/reporting/performance/test_distributions_mode.cpp
@@ -36,9 +36,8 @@ void test_mode_2param(benchmark::State& state)
 }
 
 
-template<class T>
-static void fixed_ratio_2args(benchmark::internal::Benchmark* b, T left_div_right, std::vector<int64_t> lefts) {
-    for (const T &left: lefts) {
+static void fixed_ratio_2args(benchmark::internal::Benchmark* b, long double left_div_right, std::vector<int64_t> lefts) {
+    for (const long double &left: lefts) {
         b->Args({(int64_t)left, (int64_t)(left/left_div_right)});
     }
 }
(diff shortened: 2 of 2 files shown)


include/boost/math/distributions/non_central_chi_squared.hpp
```diff
@@ -838,7 +838,9 @@ namespace boost
             &r,
             Policy()))
                return (RealType)r;
-         return detail::generic_find_mode(dist, 1 + k, function);
+         bool asymptotic_mode = k < l/4;
```
mborland, line 841: Did you mean `(k/l) < RealType(1)/4`? Your paper mentions using (k/l) < 1/4 as the conservative threshold. As written `asymptotic_mode` will always evaluate to false, and the ternary operator will always evaluate `starting_point = 1 + k` which is what the original code had.

mborland:
For your bench-marking I highly recommend looking at [Google Benchmark](https://github.com/google/benchmark). There are a ton of examples in the reporting/performance folder.

include/boost/math/distributions/non_central_chi_squared.hpp
vindex10: Hi @mborland, thank you for your attention here!

Well, mathematically `k < l/4` and `k/l < 1/4` is the same :) Mainly, I tried to avoid dividing by the parameter, because it might turn out to be zero.

Could you please elaborate a bit more on why would the asymptotic mode always evaluate to `false`? Is it something to do with `RealType` divided by `int`?

mborland: You are correct, and can ignore my comment. I read as `k < 1/4` not `k < l/4`.

NAThompson:
Looks like we let this kinda stagnate for a while; just got the ball rolling again by approving the CI.

vindex10:
Hi all! Thank you for giving attention to the PR. I'm coming back with the plots!

## Few comments:

* Below `baseline` is: 67d25f4a40c56dbf3c49438c3e62f813e09d2e77 and `new` is: 78df4a61a820153ecb0cd91ef9f6ea235c7c014b
* Vertical line denotes the location where switch between approx and default mode is happening

## Plots

**Fixed number of d.o.f and we increase noncentrality λ:**

fixed_k_2
fixed_k_15
fixed_k_50

**Fixed noncentrality and we increase number of d.o.f k:**

fixed_nc_1
fixed_nc_30
fixed_nc_100
fixed_nc_500

**We fix scale k/λ and plot the performance along k:**

fixed_scale_0_05
fixed_scale_0_15
fixed_scale_0_25


## Technical info:

* Raw json files of the benchmarks:
[report.baseline.json.gz](https://github.com/boostorg/math/files/6908789/report.baseline.json.gz)
[report.new.json.gz](https://github.com/boostorg/math/files/6908864/report.new.json.gz)

* compilation command for the `test_distributions_mode.cpp`:

```
g++ -std=c++17 -I$HOME/opt/benchmark/include -L$HOME/opt/benchmark/lib64 -I../../include/ test_distributions_mode.cpp -pthread -lbenchmark
```
custom include paths for benchmark were used because I compiled the recent benchmark library. it had few arguments that I needed that were not present in the released version (though they claim their master is stable, so I think it is fine :)) Specifically:

`benchmark::CreateRange` and CLI argument: `--benchmark_context`

* The command I used to run the compiled executable:

```
./a.out --benchmark_enable_random_interleaving=true --benchmark_context=boost_math_revision=`git rev-parse HEAD` --benchmark_repetitions=200 --benchmark_report_aggregates_only=true --benchmark_format=json > report.json
```

Therefore, you can find the exact revision that was used for benchmarking in the file itself

vindex10:
Just in case someone would like to play with plots
[visualize_benchmarks.ipynb.gz](https://github.com/boostorg/math/files/6908906/visualize_benchmarks.ipynb.gz)

vindex10:
error bars look suspicious. I recompiled executables and reran on one particular test, they appeared to be fixed. so I reran full benchmarks, it takes several hours on 200 runs.. but I don't think error bars here play any role :) they are 100 times smaller even in the worst case.

jzmaddock:
I'll example this in more detail later, but it looks good to me.

NAThompson:
@jzmaddock: Should we ask for an ULPs plot as well? If not @vindex10 might well enjoy learning that concept anyway.

(And agreed, this does look good.)

vindex10:
interesting read about ULPs, thanks for the reference: https://www.boost.org/doc/libs/1_77_0_beta1/libs/math/doc/html/math_toolkit/ulps_plots.html

In the meanwhile I have more consistent plots.

fixed_k_2
fixed_k_15
fixed_k_50

fixed_nc_1
fixed_nc_30
fixed_nc_100
fixed_nc_500

fixed_scale_0_05
fixed_scale_0_15
fixed_scale_0_25


and sources:
[report.baseline.json.gz](https://github.com/boostorg/math/files/6910152/report.baseline.json.gz)
[report.new.json.gz](https://github.com/boostorg/math/files/6910153/report.new.json.gz)

NAThompson:
Very cool. The build failure is a network error so you might want to make a meaningless diff to get it clean.

jzmaddock:
> Should we ask for an ULPs plot as well?

No real need, all this does is improve the initial "guess" for the mode, before it gets refined, so the result will be unchanged whatever.

vindex10:
if somebody killed the previous build on purpose, please let me know ) otherwise I keep pushing empty commits before the green tick appears :)

jzmaddock:
I've approved the run: not sure why we need to keep approving stuff... but it is what it is I guess!

mborland:
> I've approved the run: not sure why we need to keep approving stuff... but it is what it is I guess!

For first time contributors it is the [default setting for GHA.](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks) You can [change it](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#configuring-required-approval-for-workflows-from-public-forks) if you so choose.

jzmaddock:
> For first time contributors it is the default setting for GHA.

Nod.  I thought perhaps once we'd approved the first run we wouldn't have to do so for later runs, but apparently not.

NAThompson:
>  I thought perhaps once we'd approved the first run we wouldn't have to do so for later runs, but apparently not.

That would be a better default.

vindex10:
@NAThompson now it is clean! :)

NAThompson:
Looks good to me. @jzmaddock?

vindex10:
Hi! Is there anything we can do to move forward?

jzmaddock:
Yes.... nag me ;)

PR #646, opened 2021-06-25
Add CMake Testing for Bilinear Uniform Interpolator

mborland:
@NAThompson Some benign changes based off standalone testing.


include/boost/math/interpolators/detail/bilinear_uniform_detail.hpp
```diff
@@ -42,7 +45,7 @@ class bilinear_uniform_imp
         y0_ = y0;
         dx_ = dx;
         dy_ = dy;
-        using std::to_string;
+
```
NAThompson, line 48: Wait, if we don't have the `using std::to_string`, then we won't get the ADL for multiprecision types on the error message on line 50 correct?

See https://github.com/boostorg/multiprecision/pull/339

NAThompson:
@mborland : Thanks for this PR!

NAThompson:
@mborland: Looks like the build failed due to a network problem; got a meaningless diff to kick off another build?

PR #650, opened 2021-06-26
Bezier polynomials.

NAThompson:


NAThompson:
@mborland , @jzmaddock : I need a real programmer to look at this. The evaluation uses [de Casteljau's algorithm](https://en.wikipedia.org/wiki/De_Casteljau%27s_algorithm) but the particular issue is that I have to allocate an entire container on every evaluation of the interpolant.

Is there a slick way around this? Should we ask the user to provide scratch space?

jzmaddock:
The only suggestions I have are:

```
static Container& get_storage()
{
   static threadlocal Container the_storage;
   return the_storage;
}
```

Then each thread has it's own storage, and if there are 50000 instances of the class, they all share the same storage.

The other possibility, if it's a common use case for the number of points to be small, is to use a local array as storage in that case.  That way there's no allocation at all in the common case.... if it is indeed common ;)

include/boost/math/interpolators/detail/bezier_polynomial_detail.hpp
```diff
@@ -12,6 +12,15 @@
 
 namespace boost::math::interpolators::detail {
 
+namespace {
+    template <class RandomAccessContainer>
+    static RandomAccessContainer& get_storage()
+    {
+        static thread_local RandomAccessContainer the_storage;
```
mborland, line 19: This may run into issues with Mingw since it only supports using `thread_local` with trivial types. I think the following should work:

```
static
#ifndef BOOST_MATH_NO_THREAD_LOCAL_WITH_NON_TRIVIAL_TYPES
    BOOST_MATH_THREAD_LOCAL
#endif
RandomAccessContainer the_storage;
```

@jzmaddock feel free to correct me if I am wrong.

NAThompson: @mborland : I've restricted to C++-17; will that not help?

mborland: It probably will; I forget which versions of GCC it impacted but they are older. If the CI is clean I wouldn't worry about it.

jzmaddock: I'm afraid I've had issues up to and including gcc-10 with this.  Sometimes you hit it and sometimes you don't :(

There are also those platforms which have zero thread support (no thread_local, <mutex> etc).

jzmaddock: For example: https://github.com/boostorg/math/issues/621

NAThompson:
@pabristow : Good idea; I did kinda decide it was a paranoid assert and removed it . . .

NAThompson:
Ok, now it's a time for compromises, because as predicted, the thread_local storage doesn't work on mingw.

A couple options:

- Just disable it on MingW; it does give a compiler warning so maybe that's ok?
- Use the Horner scheme discussed in "Bezier and B-Spline Techniques" rather than the de Casteljau recurrence. Unfortunately the polynomial coefficients then are products of geometric coordinates with binomial coefficients, so both instability and overflow are easy and possible.

Any ideas?

mborland:
> Ok, now it's a time for compromises, because as predicted, the thread_local storage doesn't work on mingw.
> 
> A couple options:
> 
>     * Just disable it on MingW; it does give a compiler warning so maybe that's ok?
> 
>     * Use the Horner scheme discussed in "Bezier and B-Spline Techniques" rather than the de Casteljau recurrence. Unfortunately the polynomial coefficients then are products of geometric coordinates with binomial coefficients, so both instability and overflow are easy and possible.
> 
> 
> Any ideas?

My recommendation would be to disable the use of `thread_local` storage for both MingW and environments without threading (see #621). `<boost/math/tools/config.hpp>` provides a definition of `BOOST_DISABLE_THREADS` so this would continue to work in standalone mode.

PR #651, opened 2021-06-27
Preperation for 1.77:

jzmaddock:
Rework hypergeometric distro equations.
Fix up Fibonacci docs.
Fix Fibonacci constexpr and noexcept usage and add to testing.
Regenerate docs.


jzmaddock:
@pabristow : this includes reworked hypergeometric distribution equations, can I get you to double check that I haven't fat-fingered them again?  See https://github.com/boostorg/math/blob/d4de174bbbd53f6e13e6d078312f5adfd0683310/doc/equations/hypergeometric6.svg

pabristow:
@jzmaddock Sadly, your Greek is rusty - kurtosis has a 'u' not a 'e' ?  https://en.wikipedia.org/wiki/Kurtosis 

Why they couldn't Pearson have called it tailedness? Obfuscation.

More seriously, I can't reconcile this equation with Wolfram 

https://www.wolframalpha.com/input/?i=kurtosis+hypergeometric+distribution

used in my PR  (I'd trust this one as it is computer generated.

or Wikipedia https://en.wikipedia.org/wiki/Hypergeometric_distribution

even allowing for the confusion caused by using different letter symbols for the variables :-(

Should I also implement this formula in C++ to confirm that it is also correct?

pabristow:
Also can we be explicit that this is the kurtosis and not the excess kurtosis?

jzmaddock:
>@jzmaddock Sadly, your Greek is rusty - kurtosis has a 'u' not a 'e' ? https://en.wikipedia.org/wiki/Kurtosis

:(

Well at least that one is an easy fix!!

>More seriously, I can't reconcile this equation with Wolfram
>
>https://www.wolframalpha.com/input/?i=kurtosis+hypergeometric+distribution
>
>used in my PR (I'd trust this one as it is computer generated.

I had trouble getting my head round that one so I used Wikipedia...

Maybe I'll look at wolfram again.

>or Wikipedia https://en.wikipedia.org/wiki/Hypergeometric_distribution

It should be identical to the Wikipedia definition - that's where I copied it from, it is also marked as Kurtosis Excess on the wikipedia page

>even allowing for the confusion caused by using different letter symbols for the variables :-(

Yeah, that sucks, it does help though that the formula are all symmetrical in n and r (or whatever you call them).

jzmaddock:
Bollocks, the wolfram version isn't symmetrical in n and r :(

jzmaddock:
Ah, kurtosis excess is: https://www.wolframalpha.com/input/?i=kurtosis+excess+hypergeometric+distribution

pabristow:
I did check that the Wolfram and wikipedia equations when translated into C++ give the same numerical result.

It feels odd to me that the number of selections should be 'allowed' to be more than the number of trials, as implied by symmetry in n and r? 

I started to add a Boost check that failures/successes/events <= trials, but then I discovered the symmetry. 

Or am I terminally confused as usual?

Since we (I) have used Wolfram, can't we just refer to the Wolfram alpha page, and duck the issue?

jzmaddock:
I think it's quite nice to have the equation inline in that it makes the text easier to read.

How about this, basically cut and pasted from the wolfram page: 
preview

jzmaddock:
>It feels odd to me that the number of selections should be 'allowed' to be more than the number of trials, as implied by symmetry in n and r?

If there are r defectives in a population of size N and we sample n of them, then necessarily we find some number k defectives with:

k <= min(r, n)
r <= N
n <= N

Notice the symmetry again, and no relation between n and r.

In fact the distribution is *necessarily* symmetrical in n and r if you look at it this way:  we have a population of size N, and two independent events happening to it: the number of failures within it, and the number of times it gets sampled.  View the distribution as the union between these two events, and remember that the names we've given them "failures" and "samples" are quite arbitrary, and the symmetry becomes obvious.

But I confess not to have thought about it before! ;)

pabristow:
This at last makes sense to me.  
" View the distribution as the union between these two events,"
is the key (except that using the word 'events' might confuse some readers of definitions using 'events' only for the failures/success).  

Thanks.

PS Another possibly interesting question is how can the kurtosis be useful?  But that's probably Off-Topic :-)

(Only very few events from a large population shifts the kurtosis from above about 3 - but that's obviously raising the outlier risk.)

PR #654, opened 2021-07-02
Add GCC11 to CI

mborland:


NAThompson:
While we're adding compilers to test, are there any which can be sensibly removed? We've got a really computationally expensive build  . . .

mborland:
> While we're adding compilers to test, are there any which can be sensibly removed? We've got a really computationally expensive build . . .

It looks like drone tests g++-6,8,10 and clang-10. We could run g++-5,7,9,11 and clang-5,6,7,8,9,11 using GHA. That would cut down a good bit. We could make further cuts and go odd versions numbers on GHA, and even numbers on drone. I do not know what the limitations for run time etc. are imposed by drone (if any).

Thoughts @NAThompson and @jzmaddock?

NAThompson:
I'd probably get rid of gcc-5 and clang-5.

NAThompson:
Build failure was a timeout.

PR #655, opened 2021-07-02
 github appears to have lost this commit.

NAThompson:
Patching it up.

PR #656, opened 2021-07-04
Add standalone testing to bezier polynomials

mborland:
@NAThompson This PR is against the branch in #655. Should solve the one outstanding CI failure.


NAThompson:
Ah, that was it! Thanks Matt!

PR #659, opened 2021-07-07
Fix build on platforms with "double-double" long double type

mscastanho:
Since commit 01f3b7d7721fd most tests started to fail on platforms that use a
"double-double" for long double, like powerpc64le:

```
../../../boost/math/tools/precision.hpp:148:137: error: expected primary-expression before ‘long’
  148 | inline constexpr long double epsilon<long double>(const std::true_type&
  BOOST_MATH_APPEND_EXPLICIT_TEMPLATE_TYPE(long double)) noexcept(long double)

../../../boost/math/tools/precision.hpp:148:137: error: expected ‘)’ before ‘long’
  148 | inline constexpr long double epsilon<long double>(const std::true_type&
  BOOST_MATH_APPEND_EXPLICIT_TEMPLATE_TYPE(long double)) noexcept(long double)

../../../boost/math/tools/precision.hpp:148:137: error: expected initializer before ‘long’
  148 | inline constexpr long double epsilon<long double>(const std::true_type&
  BOOST_MATH_APPEND_EXPLICIT_TEMPLATE_TYPE(long double)) noexcept(long double)
```

Fix this by replacing the noexcept expression with the contents of
`BOOST_MATH_NOEXCEPT`, which was the intention of commit 01f3b7d7721fd.

### DIFF
--- a/include/boost/math/tools/precision.hpp
+++ b/include/boost/math/tools/precision.hpp
@@ -145,7 +145,7 @@ inline constexpr T epsilon(const std::true_type& BOOST_MATH_APPEND_EXPLICIT_TEMP
 
 #if defined(__GNUC__) && ((LDBL_MANT_DIG == 106) || (__LDBL_MANT_DIG__ == 106))
 template <>
-inline constexpr long double epsilon<long double>(const std::true_type& BOOST_MATH_APPEND_EXPLICIT_TEMPLATE_TYPE(long double)) noexcept(long double)
+inline constexpr long double epsilon<long double>(const std::true_type& BOOST_MATH_APPEND_EXPLICIT_TEMPLATE_TYPE(long double)) noexcept(std::is_floating_point<long double>::value)
 {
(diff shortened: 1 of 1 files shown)


mborland (approved):
@jzmaddock This change looks good to me; fixes an error that I introduced.

ckormanyos:
Hi @mscastanho this is nice. Thanks for investigating this phenomenon.

@mborland and @jzmaddock I took the liberty of approving the CI run.

ckormanyos:
Added note. We are still a long ways away, but work to implement double-double BSL backend is underway in the 2021 GSoC [here](https://github.com/BoostGSoC21/multiprecision)

mborland:
This looks good to go. Should probably make it into #661.

jzmaddock:
Being somewhat pedantic: `std::is_floating_point<long double>::value` is necessarily always true.  Merging anyhow.

PR #660, opened 2021-07-10
FT interface with custom allocators

Lagrang3:


ckormanyos:
Hi @Lagrang3 looks good, but I think we need to re-target this PR to the Fork. I think you saw that as well. Should I do anything?

Lagrang3:
Hi @ckormanyos. I push it here by mistake. I'm already working on a PR to the Gsoc fork. I don't know if you can do something about this. Delete? I already closed it.

ckormanyos:
> already closed it

Hi @Lagrang3 I have done that I think two times already as well. Just like you, I also closed the PR.

Issue #664, opened 2021-07-19
Wrong definition of Pearson correlation for constant vectors


zbjornson:
The `correlation_coefficient` function has two special cases:

https://github.com/boostorg/math/blob/4cdc6450d5844b6dcd787f70ca4122fd66d4b855/include/boost/math/tools/bivariate_statistics.hpp#L75-L84

The correlation is undefined if either stddev is 0 however, so I think that should just be `if (Qu == 0 || Qv == 0) return 0./0.` (or similar). The current behavior is documented, but returning 0 and 1 make these states impossible to distinguish from true 0 or 1 correlation coefficients without an additional loop in the consumer code.

References:
* https://stats.stackexchange.com/questions/23676/normalized-correlation-with-a-constant-vector
* Mathematica and R both print a warning and don't evaluate the input.

NAThompson:
Think `return std::numeric_limits<Real>::quiet_NaN()` would be preferable?

zbjornson:
Yeah, I think that's preferable.

jeremy-murphy:
I agree that it's _preferable_, but have you weighed up the pros and cons of other options? I mean, such as throwing an exception or returning an `optional`? I ask this because ordinary floating point arithmetic might cause a qNaN due to over/under flow right? And now this special case is being lumped in with them, even though the algorithm knows the difference between them. Maybe it's fine, but, maybe not.

zbjornson:
TBH I don't know what's allowed or idiomatic in Boost. `optional` or throwing an exception seem reasonable as well, and are indeed clearer.

> ordinary floating point arithmetic might cause a qNaN due to over/under flow right?

I think this algorithm is very difficult to get to under/overflow, fwiw. (The code attributes Sandia Labs, but as far as I can tell both the mean and variance are Welford's 1962 methods, which have been extensively analyzed. Although not consistently found to be "good" in terms of precision, I've yet to see under/overflow come up as a criticism.)

But over/underflow will never generate a NaN as far as I know. If FTZ/DAZ are set, underflow would be 0; otherwise +/-2<sup>e<sub>min</sub></sup> or a denormal. Overflow goes to infinity or the largest finite number depending on rounding mode.

jzmaddock:
[...]

PR #665, opened 2021-07-20
Fix for #664

NAThompson:
Returns nan if either correlation coefficient is zero.

### DIFF
--- a/include/boost/math/tools/bivariate_statistics.hpp
+++ b/include/boost/math/tools/bivariate_statistics.hpp
@@ -73,12 +73,9 @@ auto correlation_coefficient(Container const & u, Container const & v)
         mu_v = mu_v + v_tmp/(i+1);
     }
 
-    // If both datasets are constant, then they are perfectly correlated.
-    if (Qu == 0 && Qv == 0)
-    {
-        return Real(1);
-    }
-    // If one dataset is constant and the other isn't, then they have no correlation:
+    // If one dataset is constant, then they have no correlation:
+    // See https://stats.stackexchange.com/questions/23676/normalized-correlation-with-a-constant-vector
+    // Thanks to zbjornson for pointing this out.
     if (Qu == 0 || Qv == 0)
     {
         return std::numeric_limits<Real>::quiet_NaN();
--- a/include/boost/math/tools/bivariate_statistics.hpp
+++ b/include/boost/math/tools/bivariate_statistics.hpp
@@ -8,6 +8,7 @@
 
 #include <iterator>
 #include <tuple>
+#include <limits>
 #include <boost/math/tools/assert.hpp>
 #include <boost/math/tools/header_deprecated.hpp>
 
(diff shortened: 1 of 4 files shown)


include/boost/math/tools/bivariate_statistics.hpp
```diff
@@ -80,7 +81,7 @@ auto correlation_coefficient(Container const & u, Container const & v)
     // If one dataset is constant and the other isn't, then they have no correlation:
     if (Qu == 0 || Qv == 0)
     {
-        return Real(0);
+        return std::numeric_limits<Real>::quiet_NaN();
```
zbjornson, line 84: I think lines 76-79L should be deleted also. If both datasets are constant, then the correlation is still undefined.

mborland:
@NAThompson I think a similar fix should be applied [here](https://github.com/boostorg/math/blob/develop/include/boost/math/statistics/bivariate_statistics.hpp#L186-L195) since `<boost/math/tools/bivariate_statistics.hpp>` is deprecated.

include/boost/math/tools/bivariate_statistics.hpp
NAThompson: Have you worked out the correlation_coefficient(x,y) when x and y are constant vectors contaminated by additive Gaussian white noise in the limit of the noise variance decaying to zero? I think this is a relevant calculation . . .

NAThompson:
@mborland : Good catch, will fix.

include/boost/math/tools/bivariate_statistics.hpp
zbjornson: I think that's 0, both empirically and intuitively, right?

```Mathematica
x = Table[
   {
    N@var,
    Correlation[
     RandomVariate[NormalDistribution[0, var], 100000],
     RandomVariate[NormalDistribution[0, var], 100000]
     ]
    },
   {var, 10^Range[0, -10, -1]}
   ];
x // TableForm
```
```
1.           0.000770873
0.1          0.00157435
0.01        -0.00303611
0.001        0.00186714
0.0001      -0.0000911521
0.00001      0.000377327
1.*10^-6    -0.0000343936
1.*10^-7    -0.00773857
1.*10^-8     0.000511923
1.*10^-9     0.000770162
1.*10^-10   -0.00422592
```
```
Mean@x[[All,2]]
-0.000841306
```

Are you advocating for returning 0 on this basis though?

NAThompson: > Are you advocating for returning 0 on this basis though?

Yes.

NAThompson: Actually I've come around to your point of view on this. Looking at the [wiki](https://en.wikipedia.org/wiki/Pearson_correlation_coefficient) there are tons of different ways to approach that limit, and I really don't see why I thought that additive white Gaussian noise was the "canonical" limit of interest.

zbjornson:
@NAThompson thanks, this looks great. I think there's one more place to change: include/boost/math/statistics/bivariate_statistics.hpp line 316

NAThompson:
@zbjornson : Nice catch; thanks!

NAThompson:
Thank you both for putting eyes on this; really big help.

PR #666, opened 2021-07-20
Type cast to suppress warnings: in bivariate_statistics.hpp

kojiynet:
There "i"s are size_t but "Real" will be usually floating point type. We may want to explicitly cast the type.
I did not see any error or bug on these, but we might prefer these changes to suppress warnings a compiler will issues; at least MSVC will issue warnings on these.


NAThompson:
By incredibly bad luck, there is a merge conflict. Could you close and apply these changes on top of the develop? The merge is surprisingly ugly.

kojiynet:
OK, I will try!!

PR #667, opened 2021-07-20
Constexpr sqrt

mborland:
A constexpr (C++11 definition) implementation of sqrt with associated helper functions. Will continue to work through the rest of cmath.


mborland:
@NAThompson I think this is ready for review. Needed to fix some small CI errors.

doc/sf/ccmath.qbk
```diff
@@ -0,0 +1,50 @@
...
+        inline constexpr bool isinf(T x);
+
+        template <typename T>
+        inline constexpr bool isnan(T x);
+
+        template <typename Real>
+        inline constexpr Real sqrt(Real x);
+
+        template <typename Z>
+        inline constexpr double sqrt(Real x);
```
NAThompson, line 31: Typo here: Should be `sqrt(Z x);`

doc/sf/ccmath.qbk
```diff
@@ -0,0 +1,50 @@
...
+[/
+[heading Description]
+
+The following functions require C++14 or greater. Placeholder.
+
+[heading Synopsis]
+
+[heading Description]
+
+The following functions require C++17 or greater. Placeholder.
```
NAThompson, line 44: Let's just do everything in C++17. This will keep us from being constrained if we need better functionality. Surveys indicate that basically nobody is on C++14.

include/boost/math/ccmath/isnan.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#ifndef BOOST_MATH_CCMATH_ISNAN
+#define BOOST_MATH_CCMATH_ISNAN
+
+namespace boost { namespace math { namespace ccmath {
+
+template <typename T>
+inline constexpr bool isnan(T x)
+{
+    return x != x;
```
NAThompson, line 14: Have you tested with a bunch of compiler flags? e.g., `-ffast-math -O3 -march=native -fno-finite-math-only`?

include/boost/math/ccmath/sqrt.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+#ifndef BOOST_MATH_CCMATH_SQRT
+#define BOOST_MATH_CCMATH_SQRT
+
+#include <limits>
+#include <type_traits>
+#include <boost/math/ccmath/isnan.hpp>
+#include <boost/math/ccmath/isinf.hpp>
+
+namespace boost { namespace math { namespace ccmath {
```
NAThompson, line 16: I'd do `namespace boost::math::ccmath` just so if you need to make backwards incompatible changes in C++14 mode you don't surprise anyone.

test/isnan_test.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+int main()
+{
+    test<float>();
+    test<double>();
+
+    #ifndef BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
+    test<long double>();
+    #endif
```
NAThompson, line 36: Can you get this to work with libquadmath/float128 as well?

test/sqrt_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+    using std::abs;
+    
+    constexpr Real tol = 2*std::numeric_limits<Real>::epsilon();
+    
+    constexpr Real test_val = boost::math::ccmath::sqrt(Real(2));
+    constexpr Real dummy = 1;
+    static_assert(test_val > dummy, "Not constexpr");
+
+    Real known_val = std::sqrt(Real(2));
+    BOOST_TEST(abs(test_val - known_val) < tol);
```
NAThompson, line 24: I would highly recommend using my `math_unit_test.hpp` for this one. The expectation will be that this is half-ulp accurace. Use:

`CHECK_ULP_CLOSE(test_val, known_val, 0)`

NAThompson:
How are you verifying that these are actually being computed at compile time? Could you provide some godbolt links?

I do wonder how this is achieved without using compiler facilities. . .

jzmaddock:
Thanks @NAThompson for giving this the once over - some good comments there.

>Finally, isn't there already an isnan, sqrt in boost.math? Couldn't we constexpr those?

Sadly not.  They examine the bit-pattern of the float, which makes them immune to -ffast-math like optimisations, but the type punning is illegal in a constexpr context.

> How are you verifying that these are actually being computed at compile time? Could you provide some godbolt links?

I notice the tests have a mix of runtime and compile tests, only those that explicitly create a constexpr value will be evaluated at compile time.

In fact testing constexpr functions is actually really hard, arguably impossibly so, we need to:

* Call the function in a constexpr context, otherwise we don't actually test that the function is constexpr safe, and....
* If the function has multiple branches, we should ideally have one constexpr test for each branch... otherwise unevaluated branches are simply discarded (when in a constexpr context), and could contain all manner of strange illegal code and we would never know.  Note that compiling and testing all the branches at runtime is NOT equivalent, since some legal runtime constructs are not legal when consstexpr :(

>Also; how should this interact with compiler flags? https://stackoverflow.com/questions/57894032/floating-point-comparison-with-quiet-nan-on-gcc-with-ffast-math-always-yields
>
>It appears there will be different behavior at runtime vs compile time which could be very surprising.

Ugh, don't get me started!  In theory, when in a constexpr context we should get more consistent behaviour, which is immune to the vagaries of strange compiler flag or architecture behaviour.  But without a *very* careful reading of the standard, I wouldn't want to be definitive about that, it wouldn't surprise me if there aren't still dark underspecified corners lurking somewhere...

This also brings up a slightly more philosophical design question:  How should these functions behave when called NOT in a constexpr context?

I see three options:

1) Call the std lib function, as in:

```
template <class F>
inline constexpr F foo(F f)
{
   if(!std::is_constant_evaluated()) return std::foo(f);
   else return constexpr_foo(f);
}
```

2) Call the constexpr implementation - this is what happens currently.

3) Make it an error, either by marking the function as `consteval`, or with a `if(!std::is_constant_evaluated()) static_assert(0, "Function foo called in non-constexpr context")` in the function body.

BTW, while `consteval` requires building in C++20 mode, `std::is_constant_evaluated` can be simulated in C++11 mode and later on all recent compilers, see: https://github.com/boostorg/multiprecision/blob/d3c9d4a479a86b87b2fa56ddf494887e11018f31/include/boost/multiprecision/detail/number_base.hpp#L96-L132

Personally, I don't much like (2), as we potentially end up calling an inferior runtime implementation by mistake (sqrt is a good example of this - it's a single instruction on most architectures, but not if we call an all C++ constexpr version).

And finally... I note that implementing the whole std lib in this way is a very significant piece of work, that doesn't mean we can't concentrate on a few simple and frequently used functions in the first place.

mborland:
As always I appreciate the comments @jzmaddock and @NAThompson. This latest couple commits should get after most of the comments.

1) Replaced a lot of the instance of `BOOST_TEST` with `static_assert`. This should force a `constexpr` evaluation. 
2) Lifted the definition of `IS_CONSTANT_EVALUATED` from multiprecision. Now the functions will use the std lib for non-constexpr branches
3) Everything has been converted to C++17. This should help in the future because the definition of `constexpr` is a lot more liberal than it was in C++11.

include/boost/math/ccmath/isinf.hpp
```diff
@@ -6,16 +6,26 @@
...
+#include <boost/math/tools/is_constant_evaluated.hpp>
 
-namespace boost { namespace math { namespace ccmath {
+namespace boost::math::ccmath {
 
 template <typename T>
 inline constexpr bool isinf(T x)
 {
-    return x == std::numeric_limits<T>::infinity() || -x == std::numeric_limits<T>::infinity();
+    if(BOOST_MATH_IS_CONSTANT_EVALUATED(x))
```
NAThompson, line 18: Shouldn't this be an `if constexpr`?

mborland: I thought the same thing. Here is the [SO answer.](https://stackoverflow.com/questions/54251530/stdis-constant-evaluated-behavior)

jzmaddock: Yeah, it's a weird corner of the language for sure, but it must absolutely not be constexpr!

NAThompson: Wowsers, well good thing we have some real programmers eyes on this one! I would've Pied Piper'd us on this one. . .

mborland:
Opened https://github.com/boostorg/multiprecision/pull/348 in an effort to get this working with `float128` as requested by @NAThompson. CI is clean with the builtin floating point types.

mborland:
@jzmaddock Can you take a look at this [one outstanding failure](https://drone.cpp.al/boostorg/math/331/37/2)? It looks like g++8 is attempting to compile the branch that is not consteval. It looks like GCC8 has [ __builtin_constant_p](https://gcc.gnu.org/onlinedocs/gcc-8.4.0/gcc/Other-Builtins.html) but not `__builtin_constant_evaluated`. Is this worth restricting to g++9 and greater, or do you know of another workaround?

jzmaddock:
OK, the issue is that `BOOST_MATH_IS_CONSTANT_EVALUATED(x)` is called with x not a builtin type and ` __builtin_constant_p` only works for builtin types.  In other words it's a poor-man's substitute for `__builtin_constant_evaluated` which mostly works, but not in this case :(

I think I'd just disable testing float128 when `__builtin_constant_evaluated` is not available.

mborland:
Pending further review this is now clean on CI.

mborland:
Ping. Any other changes required for this one?

jzmaddock:
Looks good to me, merging...

jzmaddock:
fabs/abs, plus isfinite/isnormal would be trivial to add next?

mborland:
> fabs/abs, plus isfinite/isnormal would be trivial to add next?

Just opened a PR for fabs/abs. isfinite/isnormal will be next.

PR #669, opened 2021-07-25
Type cast (to suppress warnings)

kojiynet:
When means_and_covariance_seq_impl() and  correlation_coefficient_seq_impl() try to return a tuple, the type of "i" is assume to be Real, while actually its type is size_t. This makes VC issue warnings. To suppress this we might want to write "Real(i)" instead of "i".
This is also consistent with "Real(0)" and "Real(1)" in some of these return statements.


NAThompson:
Thanks!

kojiynet:
Great, thanks!

PR #671, opened 2021-08-02
Fix duplicate typedef in mp.hpp

jzmaddock:


mborland:
Thanks for the quick fix. My apologies for the low quality code.

PR #672, opened 2021-08-11
Constexpr abs/fabs

mborland:


mborland:
@jzmaddock and @NAThompson I think this is ready for review. The drone build failures are all network timeout issues.

doc/sf/ccmath.qbk
```diff
@@ -9,7 +9,10 @@ LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 [heading Description]
 
-Constexpr implementations of the functionality found in `<cmath>`. The following functions require C++17 or greater.
+Constexpr implementations of the functionality found in `<cmath>`.
+In a constexpr context the functions will use an implementation defined in boost.
```
NAThompson, line 13: Could you make sure `constexpr` is formatted as code?

test/ccmath_abs_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <boost/multiprecision/float128.hpp>
+#endif
+
+template <typename T>
+void test()
+{
+    static_assert(boost::math::ccmath::abs(T(3)) == 3);
+    static_assert(boost::math::ccmath::abs(T(-3)) == 3);
+    static_assert(boost::math::ccmath::abs(T(-0)) == 0);
+    static_assert(boost::math::ccmath::abs(-std::numeric_limits<T>::infinity()) == std::numeric_limits<T>::infinity());
```
NAThompson, line 23: Could some very simple expressions be joined here? eg

```
static_assert(abs(sin(1)) == ...)
```

mborland: The difficulty is the only functions guaranteed to be `constexpr` are `pow` and `sqrt`. I could do something heavily restricted like only testing your suggestion on GCC10+ where some of `cmath` is `constexpr`. Thoughts?

NAThompson: Yeah, I think if you have a guarantee like that, then you should *try* to test it. If it dies in CI, well, not a big deal.

NAThompson:
Unrelated point: I feel like after this we should consider doing another standalone release.

test/ccmath_abs_test.cpp
```diff
@@ -28,6 +29,18 @@ void test()
...
+template <typename T>
+void gpp_test()
+{
+    constexpr T tol = std::numeric_limits<T>::epsilon();
+
+    static_assert(std::sin(T(0)) == 0);
+    
+    constexpr T sin_1 = boost::math::ccmath::abs(std::sin(T(-1)));
+    static_assert(sin_1 > 0);
+    static_assert(boost::math::ccmath::abs(sin_1 - 0.8414709848078965066525l) < tol);
```
NAThompson, line 41: I believe this should be bitwise equal to the correctly rounded result, so `abs(sin_1 - expected) < tol` is too loose. Maybe use a hexadecimal floating point literal and check for bitwise equality?

NAThompson: Nice test, btw.

mborland:
@NAThompson All of the drone build failures are timeouts

NAThompson:
I'm happy with these changes, but this is something @jzmaddock understands better than me.

Also make sure to squash merge since there were a bunch of "fight the CI commits" which will make `git bisect` painful.

PR #674, opened 2021-08-17
Constexpr isfinite and isnormal

mborland:


mborland:
@NAThompson and @jzmaddock This should be good for review. The only CI failure is from `naive_monte_carlo_test`:

```
====== BEGIN OUTPUT ======
Running 1 test case...
Default hardware concurrency = 2
terminate called after throwing an instance of 'std::domain_error'
  what():  You have done something wrong.

Testing that a reasonable action is performed by the Monte-Carlo integrator when the integrand throws an exception on type float
unknown location(0): fatal error: in "naive_monte_carlo_test": signal: SIGABRT (application abort requested)
naive_monte_carlo_test.cpp(451): last checkpoint: "naive_monte_carlo_test" test entry

*** 1 failure is detected in the test module "naive_monte_carlo_test"

EXIT STATUS: 201
====== END OUTPUT ======
```
I thought that error had been fixed since I haven't seen it in quite a while now.

NAThompson:
I just read that error message, think "wow, that is the least helpful error message in the universe".

And I realized only I would right something like that.

NAThompson:
I'm happy with this code, BTW.

Issue #675, opened 2021-08-22
Swtich to BOOST_MATH_HEADER_DEPRECATED breaks Gentoo build


mborland:
As reported by @SoapGentoo [here](https://github.com/boostorg/math/commit/e8c7086afca6252979b38e8dedf1425e95400d39).

SoapGentoo:
On boost 1.77
```
#include <boost/math/common_factor_ct.hpp>

int main() {
	return 0;
}
```
leads to
```
/usr/include/boost/math/common_factor_ct.hpp:17:68: error: _Pragma takes a parenthesized string literal
   17 | BOOST_MATH_HEADER_DEPRECATED("<boost/integer/common_factor_ct.hpp>");
      |                                                                    ^
In file included from /usr/include/boost/math/common_factor_ct.hpp:15,
                 from test.cpp:1:
/usr/include/boost/math/common_factor_ct.hpp:17:1: error: ‘_Pragma’ does not name a type
   17 | BOOST_MATH_HEADER_DEPRECATED("<boost/integer/common_factor_ct.hpp>");
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
the synthesis of the warning pragma is wrong in `header_deprecated.hpp`. Restoring this to `#include <boost/config/header_deprecated.hpp>` and `BOOST_HEADER_DEPRECATED` fixes this.

mborland:
I will have a fix up shortly along those lines. The issue with just reverting is that it introduces a dependency that then breaks standalone mode.

PR #676, opened 2021-08-22
Fix for issue 675

mborland:
Fixes for #675. Allows for switching of `BOOST_MATH_HEADER_DEPRECATED` based on standalone context. Fixes behavior of standalone implementation.

### DIFF
--- a/include/boost/math/tools/header_deprecated.hpp
+++ b/include/boost/math/tools/header_deprecated.hpp
@@ -6,12 +6,22 @@
 #ifndef BOOST_MATH_TOOLS_HEADER_DEPRECATED
 #define BOOST_MATH_TOOLS_HEADER_DEPRECATED
 
-#ifdef _MSC_VER
+#ifndef BOOST_MATH_STANDALONE
+
+#   include <boost/config/header_deprecated.hpp>
+#   define BOOST_MATH_HEADER_DEPRECATED(expr) BOOST_HEADER_DEPRECATED(expr)
+
+#else
+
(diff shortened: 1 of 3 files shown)


NAThompson:
Ready to go?

mborland:
> Ready to go?

Yes; this is good.

jzmaddock:
Merging.  Do we need to run the new test in standalone mode as well?

thesamesam:
Thanks both. This indeed resolves the issue downstream for us in Gentoo. Cheers for the quick fix!

NAThompson:
@jzmaddock : I want to do a new standalone release upon resolution of #677 ; do we have a good CI solution for testing our standalone releases?

jzmaddock:
We would need to merge to master, and then just check that the CI run on master is clean.

mborland:
> @jzmaddock : I want to do a new standalone release upon resolution of #677 ; do we have a good CI solution for testing our standalone releases?

None of the current tests in the CI battery define `BOOST_MATH_STANDALONE`. I can open a PR for a new YAML file that runs against a different branch say `standalone_master`. That way you could just run it periodically instead of every time something is committed to develop. Currently the only thing that is tested in standalone mode is the compile tests using CMake. Thoughts?

NAThompson:
Our CI suite is already pretty expensive. Maybe one configuration on each merge request, and then a big run on `master` or `standalone_master`. 

We should probably see how @jzmaddock feels about having the standalone a bit out of sync with the Boost release cadence.

Issue #677, opened 2021-08-23
standalone: c++ usage in preprocessor evaluation


michael-herwig:
Compiling for windows on arm64 in standalone mode, I get an error `fatal error C1012: unmatched parenthesis: missing ')'` at https://github.com/boostorg/math/blob/564967785f2d01f755b1a38bb1f71c3484992d87/include/boost/math/special_functions/detail/fp_traits.hpp#L247-L253, causing the preprocessor to evaluate a c++ expression from https://github.com/boostorg/math/blob/564967785f2d01f755b1a38bb1f71c3484992d87/include/boost/math/special_functions/detail/fp_traits.hpp#L37-L43

I use the arm64 compiler shipped with Visual Studi 2022 Preview 3: `Microsoft (R) C/C++ Optimizing Compiler Version 19.30.30423 for ARM64`. To reproduce the error one may compile using the STL version including the `<bit>` header, [see github](https://github.com/microsoft/STL). To reproduce the error it is sufficient to include the corresponding header
```c++
#include "boost/math/special_functions/detail/fp_traits.hpp"
int main() { return 0; }
```
and compile it from the `arm64` development environment (ie. `cl /nologo -DBOOST_MATH_STANDALONE=1 -Istl\inc -Iboost-math\include /std:c++latest boost-math.cpp`).

NAThompson:
@mborland : Do you have a Windows machine that you can validate a fix on?

mborland:
> @mborland : Do you have a Windows machine that you can validate a fix on?

I have a windows machine, but I don't have any kind of development environment setup. I'll grab a copy of the referenced visual studio to try and get it to work.

NAThompson:
@michael-herwig : My apologies for the rarely-acceptable "submit a PR" trope, but if it's just closing a parenthesis it might get you unblocked faster . . .

michael-herwig:
No worries, I don't think it's a closing parenthesis. As far as I can see, it's the endianness detection stated as a macro used in a preprocessor `#if` directive. Actually, the offset(s) configured later are the only occurrences I found using standalone endianness detection. Thus one may replace that with a constexpr function or variable that depends on macros or the `::std::endian` value based on the environment. I can set up a pull request later if you like or help set up the environment; let me know

StephanTLavavej:
I'm working on a fix.

NAThompson:
Release cut with the fix from @StephanTLavavej .

Thanks for your help!

PR #678, opened 2021-08-24
Fix standalone endianness macros

StephanTLavavej:
* Fix comment typos: "endinaness" to "endianness"
* Change the C++20 checks (for both `__cplusplus` and `_MSVC_LANG`) to be strict: `> 202000L` to `>= 202002L`
* Add an `#error` when C++20 `<bit>` is missing, just in case.
* Fix the `_WIN32` definitions. All Windows platforms are little-endian (regardless of x86/x64/ARM/ARM64), `<boost/predef/other/endian.h>` correctly reports little-endian, and MSVC's STL simply hardcodes `enum class endian { little = 0, big = 1, native = little };`, see https://github.com/microsoft/STL/blob/f75c7f596c7b491168fefb5eff5164ab7ae36867/stl/inc/bit#L271 .
* Add a `static_assert` to verify that exactly one of `BOOST_MATH_ENDIAN_BIG_BYTE` or `BOOST_MATH_ENDIAN_LITTLE_BYTE` is true. This avoids the need for "Endian type could not be identified" consistency checks below.
* Fix boostorg/math#677 by not testing `BOOST_MATH_ENDIAN_BIG_BYTE` and `BOOST_MATH_ENDIAN_LITTLE_BYTE` with the preprocessor, as `std::endian::native` isn't a preprocessor constant. Now, this simply expects `BOOST_MATH_ENDIAN_BIG_BYTE` to be usable in a constant expression, and assumes (due to the consistency check above) that if we aren't big-endian, we must be little-endian.

### DIFF
--- a/include/boost/math/special_functions/detail/fp_traits.hpp
+++ b/include/boost/math/special_functions/detail/fp_traits.hpp
@@ -27,34 +27,40 @@ With these techniques, the code could be simplified.
 #include <boost/math/tools/is_standalone.hpp>
 #include <boost/math/tools/assert.hpp>
 
-// Determine endinaness
+// Determine endianness
 #ifndef BOOST_MATH_STANDALONE
 
 #include <boost/predef/other/endian.h>
 #define BOOST_MATH_ENDIAN_BIG_BYTE BOOST_ENDIAN_BIG_BYTE
 #define BOOST_MATH_ENDIAN_LITTLE_BYTE BOOST_ENDIAN_LITTLE_BYTE
 
-#elif (__cplusplus > 202000L || _MSVC_LANG > 202000L)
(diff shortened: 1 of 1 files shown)


NAThompson:
Respect for your PR description game. Will merge once the CI cycles and deploy another standalone.

AlexGuteniev:
I'm wondering why did it fail on specifically ARM64?

The check is here:
https://github.com/boostorg/math/blob/564967785f2d01f755b1a38bb1f71c3484992d87/include/boost/math/special_functions/detail/fp_traits.hpp#L221-L222

MSVC is not C++Builder, it supports in-class member initialization, and there should be `int_64t` on every platform MSVC supports. One of the constants is a lie. The endianness-aware code shouldn't have been ever used at all.

PR #680, opened 2021-08-24
Complete CI test with standalone mode

mborland:
Re-allocates the G++10 CI run to test the complete CI battery using standalone mode.


mborland:
Now that Boost.Multiprecision has a beta of it's own standalone mode I am having more success in getting the tests to pass locally without butchering everything.

NAThompson:
wowsers what a diff.

Everything I managed to read looks good.

ckormanyos:
> wowsers what a diff.

Amazing (ly cool) indeed.

That's also kind of what I found. it was not possible to embedd larger parts of math on the metal in a standalone fashion until multiprecision got close to standalone. This is a true unison endeavor i think.

test/centered_continued_fraction_test.cpp
```diff
@@ -8,6 +8,7 @@
 #include "math_unit_test.hpp"
 #include <boost/math/tools/centered_continued_fraction.hpp>
 #include <boost/math/constants/constants.hpp>
+#include <boost/core/demangle.hpp>
```
NAThompson, line 11: I have added a demangler to `math_unit_test.hpp` to facilitate standalone; it's available via `detail::demangle`.

include/boost/math/tools/config.hpp
```diff
@@ -33,7 +33,9 @@
 
 #if (__cplusplus > 201700L || _MSVC_LANG > 201700L)
 #define BOOST_IF_CONSTEXPR if constexpr
-#if !__has_include(<execution>)
+
+// Clang 13 on mac provides the execution header with none of the functionality. TODO: Check back on this
+#if !__has_include(<execution>) || (defined(__APPLE__) && defined(__clang__))
```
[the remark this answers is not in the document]
mborland: I edited the comment to make it more clear since it looks like no version of apple clang supports it ([see Standardization of Parallelism TS](https://en.cppreference.com/w/cpp/compiler_support)).

include/boost/math/tools/polynomial.hpp
```diff
@@ -301,7 +301,7 @@ class polynomial
 
    template <class I>
    polynomial(I first, unsigned length)
-      : m_data(first, std::next(first, length))
+      : m_data(first, std::next(first, length + 1))
```
NAThompson, line 304: Wait . . what happened here?

test/math_unit_test.hpp
```diff
@@ -29,12 +30,12 @@ namespace detail {
     {
         int status = 0;
         std::size_t size = 0;
-        std::string s = name;
 #if BOOST_MATH_HAS_CXX_ABI
-        return abi::__cxa_demangle( name, NULL, &size, &status );
+        std::string s {abi::__cxa_demangle( name, NULL, &size, &status )};
 #else
-        return name;
+        std::string s {name};
```
NAThompson, line 36: Apparently there is a demangler on Windows that would make this a bit cleaner. Do you have a Windows box?

include/boost/math/tools/polynomial.hpp
mborland: I sized this constructor wrong and it broke the tests; the pointer constructor uses `length + 1`. MSVC needed a constructor for iterators.

test/math_unit_test.hpp
mborland: I don't right now because I am about to move back across the pond. I am going to revert back to `boost::core::demangle` because I am still getting a memory leak on the GCC CI test that I can't replicate with clang.

NAThompson: Oh yeah . . . I thought std::string taking ownership of malloc'd char* was standardized now, but I guess not . . .

mborland:
@jzmaddock Does [this linker error](https://github.com/boostorg/math/runs/4740775982?check_suite_focus=true#step:18:271) from multi-precision make sense to you?

jzmaddock:
> Does this linker error from multi-precision make sense to you?

I think it's a missing #include of `<boost/math/special_functions/factorial.hpp>`, was probably getting implicitly included before?

NAThompson:
This looks good to me. Since it's a huge diff I think @jzmaddock should also sign off.

include/boost/math/special_functions/detail/unchecked_factorial.hpp
```diff
@@ -979,6 +979,8 @@ inline T unchecked_factorial(unsigned i)
    return unchecked_factorial_imp<T>(i, tag_type());
 }
 
+#endif // BOOST_NO_LEXICAL_CAST
+
```
jzmaddock, line 983: I don't think this is quite right yet, but we'll look at this again after merging as I think the multiprecision case could in general be improved here.

include/boost/math/tools/lexical_cast.hpp
```diff
@@ -17,11 +20,54 @@
 #endif
 
 namespace boost {
-    template <typename T1, typename T2>
-    inline T1 lexical_cast(const T2)
+
+    template <typename T, typename std::enable_if<std::is_same<T, float>::value, bool>::type = true>
+    inline T lexical_cast(const char* s)
```
jzmaddock, line 25: Are there ever any situations where we are lexical-casting to a float?  That would seem to be a mistake on our part if so?

include/boost/math/tools/polynomial_gcd.hpp
```diff
@@ -19,38 +19,19 @@
...
-    }
-    return std::make_pair(d, first);
-}
+namespace boost { namespace integer {
 
 namespace gcd_detail {
 
 template <typename EuclideanDomain>
-inline EuclideanDomain Euclid_gcd(EuclideanDomain a, EuclideanDomain b) noexcept(std::is_arithmetic_v<EuclideanDomain>)
+inline EuclideanDomain Euclid_gcd(EuclideanDomain a, EuclideanDomain b) noexcept(std::is_arithmetic<EuclideanDomain>::value)
```
jzmaddock, line 34: Can this not defer to std::gcd when available?  Euclid's method is very straightforward, but other methods (such as binary gcd) are usually faster.

include/boost/math/tr1.hpp
```diff
@@ -106,8 +106,12 @@ namespace boost{ namespace math{ namespace tr1{ extern "C"{
 #  endif
 #  include <boost/config/auto_link.hpp>
 #endif
-#else
-#error Auto linking not supported in standalone mode.
+#else // Standalone mode
+#  ifdef _MSC_VER
```
jzmaddock, line 110: I don't think we need this - auto_linking is msvc only, and the user can still use the code without it, they just need to manually link.  At most a warning for msvc would be enough I think?

jzmaddock:
Serious Kudos for taking this on Matt!  There are a few minor issues which could would be good to address prior to merging, and unchecked_factorial I'll look at post merge.  Other than my comments above this all looks good though!

include/boost/math/tools/big_constant.hpp
```diff
@@ -55,6 +56,19 @@ inline T make_big_value(largest_float, const char* s, std::false_type const&, st
 {
    return boost::lexical_cast<T>(s);
 }
+#else
+template <typename T>
```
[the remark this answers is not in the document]
mborland: If I replace this with a static assert it hits when `largest_float` = double which is easy to convert to. I added in an assertion if `std::numeric_limits<largest_float>::digits > LDBL_DIG` that standalone will need to be disabled.

include/boost/math/tools/lexical_cast.hpp
mborland: There were a couple of these that I remember hitting in the CI so I added all of the trivial facilities in case there was a hidden case or a need down the line for some reason.

include/boost/math/tools/big_constant.hpp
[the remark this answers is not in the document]
jzmaddock: Not so fast: this should *never* be called with type double, as it is only called when std::numeric_Limits<T>::is_specialized is false (that's the first std::false_type argument amongst other conditions), *and* when T is *NOT* constructible from a string (the second false_type argument).

So the intention of that overload is to handle the case where we have a multiprecision type with either a) more precision that the largest float or b) unknown (runtime) precision.  AND not constructible from a string (so not anything in Boost.Multiprecision).

So mpfr_class would be an example, as would our real_concept type as that's a trivial model for the "lowest common denominator that we support.

If this is really being called for type double, then something has gone badly wrong somewhere else.  

Do you have a test case?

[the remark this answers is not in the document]
mborland: Here is an example if I have only the `static_assert`:

```
...failed clang-darwin.compile.c++ ../../../bin.v2/libs/math/test/test_0F1_1.test/clang-darwin-13/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden/test_0F1.o...
...skipped <p../../../bin.v2/libs/math/test/test_0F1_1.test/clang-darwin-13/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden>test_0F1_1 for lack of <p../../../bin.v2/libs/math/test/test_0F1_1.test/clang-darwin-13/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden>test_0F1.o...
...skipped <p../../../bin.v2/libs/math/test/test_0F1_1.test/clang-darwin-13/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden>test_0F1_1.run for lack of <p../../../bin.v2/libs/math/test/test_0F1_1.test/clang-darwin-13/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden>test_0F1_1...
...removing outdated ../../../bin.v2/libs/math/test/test_0F1_1.test/clang-darwin-13/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden/test_0F1_1.test
clang-darwin.compile.c++ ../../../bin.v2/libs/math/test/test_1F1_regularized_float.test/clang-darwin-13/debug/cxxstd-17-iso/link-static/threading-multi/visibility-hidden/test_1F1_regularized.o
In file included from test_1F1_regularized.cpp:8:
In file included from ./test_1F1_regularized.hpp:9:
In file included from ../../../boost/math/concepts/real_concept.hpp:29:
../../../boost/math/tools/big_constant.hpp:63:5: error: static_assert failed due to requirement 'sizeof(double) == 1' "Type is unsupported in standalone mode. Please disable and try again."
    static_assert(sizeof(largest_float) == 1, "Type is unsupported in standalone mode. Please disable and try again.");
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Looks like all the errors come propagate from testing `real_concept` where we can process `largest_float` as a double trivially unless I am seriously missing something here.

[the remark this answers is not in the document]
jzmaddock: Yes, your static_assert doesn't depend on a template parameter so the condition `sizeof(largest_float) == 1` is evaluated as soon as the function is parsed (and fails), not when the template is instantiated.  That's why there's no instantiation backtrace in your error messages.

Try `static_assert(sizeof(T) == 0, "This should fail");`

mborland:
@jzmaddock This should be good now; the last test is hung.

Issue #682, opened 2021-08-30
ppc64le failing tests with IEEE128


mscastanho:
I've been testing boost Math compiled with GCC 12 (dev) using IEEE long double on ppc64le (instead of IBM long double format, the default). There are quite a few tests failing because the maximum error is higher than the tolerance used, most cases are just slightly off, others have errors a few orders of magnitude larger than the tolerance. I'd be happy to provide more info if needed.

Besides these, there is one test that is failing at the compilation step: `cstdfloat_cmath_incl_test` 
```
compile_test/cstdfloat_cmath_incl_test.cpp: In function ‘void compile_and_link_test()’:
compile_test/cstdfloat_cmath_incl_test.cpp:56:42: error: ‘nanq’ is not a member of ‘std’; did you mean ‘nanl’?
   56 |     check_result<boost::float128_t>(std::nanq(0));
      |                                          ^~~~
      |                                          nanl
...failed updating 1 target...
```

On ppc64le, when GCC is compiled with `--with-long-double-128 --with-long-double-format=ieee`, `__float128` becomes an alias to `long double`. As a consequence, libstdc++ does not define `_GLIBCXX_USE_FLOAT128`  anymore (see gcc-mirror/gcc@7c1e7eed891a7) , and so neither  `BOOST_HAS_FLOAT128` nor `BOOST_MATH_USE_FLOAT128` are defined. For this reason, `nanq` is not being defined in `cstdfloat_cmath.hpp`.

Do you have any pointers on how I could solve these issues? I'm still trying to find my way around the codebase.

ckormanyos:
> have any pointers on how I could solve these issues?

At the moment no. But...

> I'm still trying to find my way around the codebase.

This is a good time to lookinto (or try to solve) this since we are actually working on the general area of `<cstdfloat>` at the moment in #506.

I do not know how to get that compiler. Any suggestions?

mscastanho:
@ckormanyos The issue can be reproduced with a GCC built from latest source with the proper flags to make ieee128 the default long double format. But you will likely need a ppc64le machine to run the tests.
```
$ git clone git://gcc.gnu.org/git/gcc.git ~/src/gcc

$ mkdir -p ~/build/gcc
$ cd ~/build/gcc

[...]

PR #685, opened 2021-09-01
Fix cstdfloat_math_incl_test on ppc64le with ieee128

mscastanho:
When switching the default long double format on ppc64le to ieee128, this test
does not compile as nanq is not available. In such cases nanl should be used
instead.

Also, calling nanl with argument 0 raises warnings as it expects a char*. Change
the argument to an empty string, which keeps the purpose of the test and avoids
new warnings.

Related to #682


mscastanho:
Ping?

PR #686, opened 2021-09-03
constexpr frexp

mborland:
@ckormanyos This was borrowed heavily from your wide integer implementation with some added handling described in the standard (e.g. NaN). Do you have a sample `constexpr` test you use?  I am not sure how to get around `exp` having external visibility in a `constexpr` context. I could change to something like `frexp(T arg, void)`, but then the behavior would be different from the standard.

### DIFF
--- a/doc/sf/ccmath.qbk
+++ b/doc/sf/ccmath.qbk
@@ -12,6 +12,7 @@ LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 `Constexpr` implementations of the functionality found in `<cmath>`.
 In a `constexpr` context the functions will use an implementation defined in boost.
 If the context is not `constexpr` the functionality will be directly from the STL implementation of `<cmath>` used by the compiler.
+All functions that take an `Integer` type and return a `double` simply cast the `Integer` argument to a `double`.
 All of the following functions require C++17 or greater.
 
 [heading Synopsis]
@@ -30,8 +31,8 @@ All of the following functions require C++17 or greater.
         template <typename Real>
         inline constexpr Real sqrt(Real x);
 
-        template <typename Z>
-        inline constexpr double sqrt(Z x);
+        template <typename Integer>
+        inline constexpr double sqrt(Integer x);
 
         template <typename T>
         inline constexpr T abs(T x);
@@ -54,8 +55,8 @@ All of the following functions require C++17 or greater.
         template <typename Real>
         inline constexpr Real frexp(Real arg, int* exp);
 
-        template <typename Z>
-        inline constexpr double frexp(Z arg, int* exp);
+        template <typename Integer>
+        inline constexpr double frexp(Integer arg, int* exp);
 
     } // Namespaces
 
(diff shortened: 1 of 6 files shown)


ckormanyos:
> Do you have a sample `constexpr` test you use?

Hi Matt, thanks for pushing ahead on this function and the `constexpr` drive as an entire body of work.

My tests only used my implementation within the context of larger `constexpr` algorithms. So I do not, per se, have any explicit tests of `constexpr` `frexp`.

I am, however, convinced that the addess operator of the signed integer exponent argument (i.e., as in `double frexp(double, int*)`) rules out C++11 `constexpr`-ness entirely and requires `constexpr` from C++14 and higher standards. So this means that the word `constexpr` might not be allowed on that function when compiling with C++11. I was only able to achieve `constexpr`-ness on `frexp`-like and `ldexp`-like functions for C++14 or higher.

I will need to look at your implementation more closely. Once a differentiation is made between C++11/14-and-beyond `constexpr`-ness, my first thought is to start with some _really_ simple examples such as the one below.

```
  // N[125/32, 30]
  // 3.90625000000000000000000000000

  int n;

  // 0.976562500000000000000000000000 * 2^2
  const double x = std::frexp(3.90625, &n);
```

Unlike the traditional code above, however, you could use the `constexpr` version and as a first step check if `x` can be created and queried within a `constexpr` context.

mborland:
@ckormanyos This code is all written using C++17 so the definition of `constexpr` should not be an issue. The following trivial test yields `error: modification of ‘n’ is not a constant expression`

```
template <typename T>
void test()
{
    int n;
    constexpr T x = boost::math::ccmath::frexp(T(3.90625), &n);
}
```

I am not entirely sure this error can be avoided.

jzmaddock:
> I am not entirely sure this error can be avoided.

You can't use the function in that manner: it would have to be within the body of a larger constexpr function.

mborland:
Pending review this is clean on CI. Next in the queue is `ldexp` by request. Adding @NAThompson.

doc/sf/ccmath.qbk
```diff
@@ -51,6 +51,12 @@ All of the following functions require C++17 or greater.
         template <typename T>
         inline constexpr int fpclassify(T x);
 
+        template <typename Real>
+        inline constexpr Real frexp(Real arg, int* exp);
+
+        template <typename Z>
```
jzmaddock, line 57: Trivial quibble, but I when I first read this I didn't grok that this was limited to integer arguments, maybe use `<typename Integer>` and perhaps a comment that this is for integer arguments?

Other than that, I'm happy for this to be merged, and you'll be pleased to know that ldexp should be noticeably easier! :)

NAThompson: @jzmaddock : That was my (perhaps poor) convention; mimicking ℤ.

Is it possible to do a concept without breaking C++17 compatibility?

jzmaddock: Not really, although the enable_if does much the same thing.  It's debatable whether the enable_if condition should be part of the declared interface - it clutters up the docs, but does at least make it clear what the requirements on Z are.

mborland: Doc suggestions have been incorporated in the latest commit.

As for concepts, you can avoid breaking C++17 compatibility using this macro (taken from [good concepts](https://www.stroustrup.com/good_concepts.pdf)): 

```
#ifdef GOOD_COMPILER 
#define REQUIRES requires 
#elseif 
#define REQUIRES // 
#endif 
```

PR #688, opened 2021-09-06
constexpr ldexp

mborland:
As requested by @ckormanyos

### DIFF
--- a/include/boost/math/ccmath/ldexp.hpp
+++ b/include/boost/math/ccmath/ldexp.hpp
@@ -11,11 +11,9 @@
 #include <type_traits>
 #include <stdexcept>
 #include <boost/math/tools/is_constant_evaluated.hpp>
-#include <boost/math/special_functions/pow.hpp>
 #include <boost/math/ccmath/abs.hpp>
 #include <boost/math/ccmath/isinf.hpp>
 #include <boost/math/ccmath/isnan.hpp>
-#include <boost/math/ccmath/isfinite.hpp>
 
 namespace boost::math::ccmath {
 
@@ -37,7 +35,7 @@ inline constexpr T negative_exp(int exp)
         {
             throw std::underflow_error("Exponent leads to a value that is smaller than minimum positive subnormal value");
         }
-        
+
         result /= 2;
         --abs_exp;
     }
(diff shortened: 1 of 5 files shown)


ckormanyos:
> As requested by @ckormanyos

Thank you @mborland. This is very useful. It is amazing, how quickly and reliably you're implementing, testing and documenting these functions

mborland:
@NAThompson and @jzmaddock Pending review this should be good. The two CI failures are network errors.

include/boost/math/ccmath/ldexp.hpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if(result < denorm_min_limit)
+        {
+            throw std::underflow_error("Exponent leads to a value that is smaller than minimum positive subnormal value");
+        }
+
+        result /= 2;
+        --abs_exp;
+    }
+
```
NAThompson, line 42: This whole loop is `constexpr`!?

include/boost/math/ccmath/ldexp.hpp
```diff
@@ -0,0 +1,120 @@
...
+template <typename Real, std::enable_if_t<!std::is_integral_v<Real>, bool> = true>
+inline constexpr Real ldexp(Real arg, int exp)
+{
+    if(BOOST_MATH_IS_CONSTANT_EVALUATED(arg))
+    {
+        return boost::math::ccmath::abs(arg) == Real(0) ? arg :
+               boost::math::ccmath::isinf(arg) ? arg :
+               boost::math::ccmath::isnan(arg) ? arg :
+               exp == 0 ? arg :
+               arg * boost::math::ccmath::detail::intexp2<Real>(exp);
```
jzmaddock, line 91: I think there's an issue with this approach: lets say arg is very large and exp is large and negative, in that situation the result of intexp2 can go to zero even though the result is expected to be finite.

I would simplify this and do something more like:

```
while(exp > 0)
{
   arg *=2;
   --exp;
}
while(exp < 0)
{
   arg /= 2;
   ++exp;
}
```

Which should return 0 or INF in the edge cases anyway?

mborland:
This is clean on CI with the updated implementation.

PR #691, opened 2021-09-12
constexpr div

mborland:
Includes: `div`, `ldiv`, `lldiv`, and `imaxdiv`


mborland:
@jzmaddock and @NAThompson pending review this should be good to go.

NAThompson:
@mborland : I'm squash merging, but a few of these commit messages could've been done via amend commits to make the git log more readable.

PR #694, opened 2021-09-18
constexpr floor and ceil

mborland:
Implements: `floor`, `floorf`, `floorl`, `ceil`, `ceilf`, and `ceill`


mborland:
@jzmaddock and @NAThompson This is clean pending review. Next up will be `trunc` and `modf` with their overloads.

PR #695, opened 2021-09-20
constexpr trunc

mborland:
Implements: `trunc`, `truncf`, and `truncl`


mborland:
@jzmaddock and @NAThompson pending review this is good to go.

PR #696, opened 2021-09-23
constexpr modf

mborland:
Implements: `modf`, `modff`, and `modfl`


mborland:
@jzmaddock and @NAThompson This is clean pending review.

Issue #211, opened 2019-05-23
In the catmull_rom method, how is the max_parameter() defined?


xarthurx:
I'm testing the newly added interpolation function `catmull_rom` where I'm a little confused by the `max_parameter()` of a spline.

For a simple case that to interpolate 11 points `Vector2d(x, x)` where x = 0 - 10.
Then I tested the intepolated catmull_rom crv:
```
    for (size_t i = 0; i <= 11; i++) {
      cout << "input: " << i << " output: " << cr(i).transpose() << endl;
    }
    cout << "max pram: " << cr.max_parameter() << endl;
```
the output is:
```
input: 0 output: 0 0
input: 1 output: 0.81961 0.81961
input: 2 output: 1.68179 1.68179
input: 3 output: 2.52269 2.52269
input: 4 output: 3.36359 3.36359
input: 5 output: 4.20448 4.20448
input: 6 output: 5.04538 5.04538
input: 7 output: 5.88627 5.88627
input: 8 output: 6.72717 6.72717
input: 9 output: 7.56807 7.56807
input: 10 output: 8.40896 8.40896
input: 11 output: 9.29669 9.29669
max pram: 11.8921
```

The result is quite strange to me...

Is the parameter `param` of `cr(param)` supposed to be the length on the spline?
If will it be possible to add a normalised function so that the parameter range is `[0, 1]`?

Currently, I don't understand how to get the value of an interpolated point on the spline.

NAThompson:
> Is the parameter param of cr(param) supposed to be the length on the spline?

No. Let's start off with the chordal parametrization: Then parameter at p0 is zero, and the parameter at p1 is |p1 - p0|. But of course this isn't the length of the curve, as |p1-p0| is the straight line distance between p0 and p1, and we know that Catmull-Rom splines are cubic curves within segments. The interpretation gets even more difficult when we use the centripedal parametrization, because then s1 = |p1-p0|^{1/2}. 

This is why the method `parameter_at_point` is provided. Say you want to interpolate between points three and four on your curve.

You would write something like:

```cpp
double s3 = cr.parameter_at_point(3);
double s4 = cr.parameter_at_point(4);
for (double s = s3; s < s3; s += 0.01) {
    auto P = cr(s);
}
```

> If will it be possible to add a normalised function so that the parameter range is [0, 1]?

Sadly, no. If this was done the length of the tangent vectors would be totally wrong for the choral parametrization.

xarthurx:
@NAThompson 
[...]

PR #698, opened 2021-09-29
Expand the catmull-rom spline doc

oleg-alexandrov:
It is not clear from the current documentation how the curve parameter is relates to how to interpolate between any two of the input points, which is one reason for using spline interpolation. Here a blurb is added to the doc, following https://github.com/boostorg/math/issues/211.


NAThompson:
Thanks! This is a definite improvement.

PR #714, opened 2021-10-25
Improve tanh_sinh.

jzmaddock:
tanh_sinh: improve thrash detection.

Significantly reduces the number of iterations in many cases.


NAThompson:
This is a major improvement to tanh_sinh. Is it realistic to get it into 1.78?

jzmaddock:
> This is a major improvement to tanh_sinh. Is it realistic to get it into 1.78?

That's a good question.  I still have further improvements lurking on my HD (pruning the tails when they are trivially small), but I need to test them at various different precisions before signing off on those.

I'm also not completely convinced we don't have a bug somewhere, something doesn't seem quite right, but I don't see what.

That said, maybe we should merge what we have and continue this at a later date.

PR #718, opened 2021-11-04
Quartic roots.

NAThompson:


NAThompson:
@jzmaddock , @mborland : Could we disable the non-ASCII character lint in the comments? It seems to me to be a net negative as the legibility of the comments is dramatically reduced.

PR #720, opened 2021-11-21
Add MSVC-2022 to CI Testing

mborland:


NAThompson:
Does the one support modules?

jzmaddock:
> Does the one support modules?

Yes, as does the existing VS2019.

mborland:
@jzmaddock As the MSVC guru does the failure in `test_next` make sense to you? Some kind of change to default rounding mode in the new release?

jzmaddock:
No, but I see it's officially released now so I'm downloading as I speak...

BTW, C++11 testing is redundant for msvc - it only supports C++14/17/20.

jzmaddock:
I'm pretty much stumped on this, I'm unable to reproduce locally at all (which is to say all tests pass with msvc-14.3, compiler version matches that in the test suite).  Can anyone else have a try?

mborland:
I installed a copy of Visual Studio 2022, and am also unable to reproduce this error locally.

jzmaddock:
Now that VS2022 is released, lets kick this off again....

jzmaddock:
Matt the failures here I've just been hit with in multiprecision as well: the meaning of "windows-latest" has just changed, we need to change that to "windows-2019", otherwise everything breaks! :(

jzmaddock:
Never mind, apparently I can commit direct to this branch via the web-interface, CI updated and we'll see what happens.

NAThompson:
> Never mind, apparently I can commit direct to this branch via the web-interface, CI updated and we'll see what happens.

@jzmaddock : Try `gh pr checkout 720` using [this](https://cli.github.com/manual/gh); I believe that Matt always clicks the "allow edits from maintainers" button so it's easy to collaborate.

mbs-c:
I investigated a bit further and the root cause seems to be that `ldexp(long double, int)`, which is called from `float_prior_imp`, returns different values. When called from `test_next`, `std::ldexp(1.0L, -1075)` returns `4.94066e-324` inside the Windows container and `0` outside.

In both cases, the `real_concept` test case prints "Denormals are flushed to zero." in the beginning.

PR #721, opened 2021-11-25
Constexpr hypot

mborland:


mborland:
@NAThompson This should be good to go.

include/boost/math/ccmath/detail/swap.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace boost::math::ccmath::detail {
+
+template <typename T>
+inline constexpr void swap(T& x, T& y) noexcept
+{
+    T temp = x;
+    x = y;
+    y = temp;
+}
```
NAThompson, line 17: @mborland : Is there no constexpr `std::swap`?

NAThompson:
@mborland : If you do find a `constexpr std::swap` just open up another PR. I figure you probably checked this . . .

include/boost/math/ccmath/detail/swap.hpp
mborland: `std::swap` only became `constexpr` in [C++20.](https://en.cppreference.com/w/cpp/algorithm/swap)

PR #723, opened 2021-11-28
CCMATH Module

mborland:
@NAThompson This is a minimal working example of turning ccmath into a module. On GCC you can compile the test with:
```
gcc ccmath_module_test.cpp ../include/boost/math/ccmath/ccmath.cpp -std=c++20 -fmodules-ts
```
I would effectively copy all existing code into one giant file that would serve as the entire module. Let me know what you think.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
           fetch-depth: '0'
       - uses: mstachniuk/ci-skip@v1
         with:
-          commit-filter: '[skip ci];[ci skip];[CI SKIP];[SKIP CI];***CI SKIP***;***SKIP CI***;[windows];[Windows];[WINDOWS];[apple];[Apple];[APPLE];[standalone];[STANDALONE]'
+          commit-filter: '[skip ci];[ci skip];[CI SKIP];[SKIP CI];***CI SKIP***;***SKIP CI***;[windows];[Windows];[WINDOWS];[apple];[Apple];[APPLE];[standalone];[STANDALONE];[module];[MODULE]'
           commit-filter-separator: ';'
           fail-fast: true
       - name: Set TOOLSET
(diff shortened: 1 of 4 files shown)


NAThompson:
This is cool. Looks like most of the work will be to get the config into CI!

NAThompson:
@mborland : Have you hit this one?

```
~/m/test ❯❯❯ g++ ccmath_module_test.cpp -I../include  ../include/boost/math/ccmath/ccmath.cpp -std=c++20 -fmodules-ts                      ✘ 1 ccmath_module ◼

ccmath_module_test.cpp:6:8: fatal error: module 'boost_math_ccmath' not found
import boost_math_ccmath;
~~~~~~~^~~~~~~~~~~~~~~~~
1 error generated.
../include/boost/math/ccmath/ccmath.cpp:102:72: error: template parameter redefines default argument
template <typename T, std::enable_if_t<!std::is_unsigned_v<T>, bool> = true>
                                                                       ^
../include/boost/math/ccmath/ccmath.cpp:32:72: note: previous default template argument defined here
template <typename T, std::enable_if_t<!std::is_unsigned_v<T>, bool> = true>
                                                                       ^
../include/boost/math/ccmath/ccmath.cpp:118:71: error: template parameter redefines default argument
template <typename T, std::enable_if_t<std::is_unsigned_v<T>, bool> = true>
                                                                      ^
../include/boost/math/ccmath/ccmath.cpp:35:71: note: previous default template argument defined here
template <typename T, std::enable_if_t<std::is_unsigned_v<T>, bool> = true>
                                                                      ^
2 errors generated.
```

mborland:
> @mborland : Have you hit this one?

I get that error if I compile with Clang 13.0.0, but not with GCC 11.2.1.

NAThompson:
@mborland : Ok, looks like modules are a bit of a way off; I suggest we just keep the PR open and periodically try it out with new compilers . . .

mborland:
Superseded by #783

Issue #724, opened 2021-12-12
Apple M1 Arch Failures


mborland:
`test_bessel_j`, `test_root_iterations`, and `test_bessel_i_prime` all fail on Apple M1 architecture. Below is a dump of the log:

```
testing.capture-output ../../../bin.v2/libs/math/test/test_bessel_j.test/clang-darwin-13/debug/cxxstd-11-iso/link-static/threading-multi/visibility-hidden/test_bessel_j.run
====== BEGIN OUTPUT ======
Running 1 test case...
Tests run with Clang version 13.0.0 (clang-1300.0.29.3), libc++ version 12000, Mac OS
Testing Bessel J0: Mathworld Data with type float
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cyl_bessel_j<float> Max = 0 RMS Mean=0


Testing Bessel J0: Mathworld Data (Tricky cases) with type float
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cyl_bessel_j<float> Max = 0 RMS Mean=0


Testing Bessel J1: Mathworld Data with type float
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cyl_bessel_j<float> Max = 0 RMS Mean=0


Testing Bessel J1: Mathworld Data (tricky cases) with type float
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cyl_bessel_j<float> Max = 0 RMS Mean=0


Testing Bessel JN: Mathworld Data with type float
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cyl_bessel_j<float> Max = 0.6104 RMS Mean=0.148
    worst case at row: 3
    { 2, 0.0099999998, 1.2499896e-05 }


Testing Bessel J0: Mathworld Data (Integer Version) with type float
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #725, opened 2021-12-13
Fix tests for Apple M1

mborland:
From #724

@jzmaddock M1 does not define `__arm__`, but does define `__aarch64__`. The tests now pass locally.


mborland:
@jzmaddock This is good to go

jzmaddock:
Cool, merging....

PR #727, opened 2021-12-28
.github/workflows/ci.yml: Add cygwin

mkoeppe:
Add a CI for testing on Cygwin.

It can be seen in action at https://github.com/mkoeppe/math/actions/runs/1629431113, revealing errors from `BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS`.

We encountered this in SageMath when building SciPy 1.7.x on Cygwin. https://trac.sagemath.org/ticket/33080


mborland:
@NAThompson and @jzmaddock I think this is a reasonable add as well. I don't see anything we can obviously de-scope to add this in, but this is only 1 compiler and 1 language standard.

mborland:
Merged into #729 to test with your other change.

PR #728, opened 2021-12-28
Cygwin, FreeBSD: Do not define BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS

mkoeppe:
Conditionals for these platforms in `include/boost/math/tools/config.hpp` appear to be outdated.

FreeBSD patches it out in its port of SciPy (which vendors boost headers) - https://cgit.freebsd.org/ports/tree/science/py-scipy/files/patch-scipy___lib_boost_boost_math_tools_config.hpp

In SageMath, we have adopted this patch and made the same update to Cygwin - https://trac.sagemath.org/ticket/33080

The Cygwin platform can be tested via #727.

### DIFF
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -81,7 +81,7 @@
 
 #include <boost/math/tools/user.hpp>
 
-#if (defined(__CYGWIN__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__EMSCRIPTEN__)\
+#if (defined(__NetBSD__) || defined(__EMSCRIPTEN__)\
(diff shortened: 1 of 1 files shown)


mborland:
@NAThompson and @jzmaddock I think this is a reasonable change if one of you could approve the CI run. Some of the assumptions made in config are very old and worth relooking.

NAThompson:
CI run approved. Change looks reasonable.

jzmaddock:
What I suggest we do is merge this and the Cygwin CI PR into a new branch/PR and then fix up any issues so we have clean CI again before merging the whole lot to develop.

mborland:
Merged into #729 for testing with Cygwin in the CI system.

PR #729, opened 2022-01-03
Cygwin, FreeBSD and Long Doubles

mborland:
Consolidates #727 and #728 to allow for other CI fixes before merge.


NAThompson:
@mborland , @jzmaddock : I noticed that we haven't done a standalone for 1.78 yet; would it make sense to use this fix to tag a standalone?

jzmaddock:
> I noticed that we haven't done a standalone for 1.78 yet; would it make sense to use this fix to tag a standalone?

We could do, there are quite a few cygwin-related CI failures showing though.  Some are trivial limits-set-to-low, others look a bit more worry-some at first glance.

mborland:
@mkoeppe Does SageMath use the standalone version of Boost.Math? If not I think we are safe to cut the standalone release without this.

mkoeppe:
> @mkoeppe Does SageMath use the standalone version of Boost.Math?
No

ckormanyos:
Hi Matt (@mborland) this is not directly related to the exact build issue...

Nonetheless, I've been wondering if the explicit use of the commit filter in the YAML script is still helpful or up-to-date or needed. My understanding is that GHA now comprehends the `[ci skip]` expression in commits. If we agree to establish use of `[ci skip]`, could we remove these commit filter artifacts? Or do you think we need them for another process that I'm not thinking of.

mborland:
@jzmaddock Can you take a look at this? The error rates that Cygwin has for the special functions are in line with what MinGW produces so I don't see any serious problems there. I think for our purposes we can consider Cygwin as having appropriate long double support.

mborland:
Ping @jzmaddock

jzmaddock:
Looks good to me, just out of interest, what's the issue with the tests that are disabled for cygwin?

mborland:
> Looks good to me, just out of interest, what's the issue with the tests that are disabled for cygwin?

They all require big object support which still failed when the flag was passed due to a lack of support.

mkoeppe:
Thanks a lot for working on this!

PR #730, opened 2022-01-03
Constexpr fdim

mborland:
Implements: `fdim`, `fdimf`, and `fdiml`


mborland:
@NAThompson This should be good to go. All of the failures are on the CI side.

PR #734, opened 2022-01-06
Constexpr fma

mborland:
Implements: `fma`, `fmaf`, and `fmal`

### DIFF
--- a/include/boost/math/ccmath/fma.hpp
+++ b/include/boost/math/ccmath/fma.hpp
@@ -1,0 +1,115 @@
+//  (C) Copyright Matt Borland 2022.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_MATH_CCMATH_FMA_HPP
+#define BOOST_MATH_CCMATH_FMA_HPP
+
+#include <cmath>
+#include <limits>
+#include <type_traits>
+#include <boost/math/tools/is_constant_evaluated.hpp>
+#include <boost/math/ccmath/isinf.hpp>
+#include <boost/math/ccmath/isnan.hpp>
+
+namespace boost::math::ccmath {
+
+namespace detail {
+
+template <typename T>
+inline constexpr T fma_imp(const T x, const T y, const T z) noexcept
+{
+    #if __GNUC__ < 10
+        return (x * y) + z;
+    #else
+    if constexpr (std::is_same_v<T, float>)
+    {
+        return __builtin_fmaf(x, y, z);
+    }
+    else if constexpr (std::is_same_v<T, double>)
+    {
+        return __builtin_fma(x, y, z);
+    }
+    else if constexpr (std::is_same_v<T, long double>)
+    {
+        return __builtin_fmal(x, y, z);
+    }
+    else // e.g. Boost.Multiprecision types where no built-in exists
+    {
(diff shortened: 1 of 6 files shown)


jzmaddock:
On the one hand this looks trivial.... on the other hand do we not run the risk of compile-time and runtime fma's returning subtly differing values?

mborland:
If the compiler doesn't optimize this to an `FMA` asm call and the builtin is not available then you are correct that it could be off by one rounding. C++20 allows for constexpr asm if you would prefer that I hard code it for major architectures.

NAThompson:
@jzmaddock : Is an fma for multiprecision infeasible or just a PITA?

mborland:
@NAThompson This is clean and good for review. The implementation is trivial, but with the documented compiler flag the results are exact. [Here](https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/detail/default_ops.hpp#L3107) is the prototype for ima support in Boost.Multiprecision.

mborland:
@NAThompson This is clean and good for review. The implementation is trivial, but with the documented compiler flag the results are exact. [Here](https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/detail/default_ops.hpp#L3107) is the prototype for ima support in Boost.Multiprecision.

include/boost/math/ccmath/fma.hpp
```diff
@@ -0,0 +1,128 @@
...
+        return __builtin_fma(x, y, z);
+    }
+    else if constexpr (std::is_same_v<T, long double>)
+    {
+        return __builtin_fmal(x, y, z);
+    }
+    #endif
+    
+    // If we can't use compiler intrinsics hope that -fma flag optimizes this call to fma instruction
+    return (x * y) + z;
```
NAThompson, line 39: So std::fma won't work here?

Also, why not just place the asm directly in the code if we detect x86 >= ivy bridge?

include/boost/math/ccmath/fma.hpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+} // Namespace detail
+
+template <typename Real, std::enable_if_t<!std::is_integral_v<Real>, bool> = true>
+constexpr Real fma(Real x, Real y, Real z) noexcept
+{
+    if (BOOST_MATH_IS_CONSTANT_EVALUATED(x))
+    {
+        if (x == 0 && boost::math::ccmath::isinf(y))
```
NAThompson, line 49: Is this not a bit too branchy here? I worry about the case where x is constant evaluated but y and z are runtime and this becomes slow as a result.

include/boost/math/ccmath/fma.hpp
mborland: `std::fma` is likely not constexpr until C++23. [According to this accepted C++20 proposal](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1668r1.html) inlining asm is not needed if `std::is_constant_evaluated` is true

include/boost/math/ccmath/fma.hpp
mborland: x is only a place holder for old versions (GCC 6-8) of gcc's builtin constant evaluated. On GCC 9+ and MSVC this compiles down to `std::is_constant_evaluated()`. Either the whole thing can be constant evaluated or none of it.

NAThompson:
Cool; this looks good. Is the failed job a known issue?

mborland:
> Cool; this looks good. Is the failed job a known issue?

It's a sanitizer issue in autodiff that seems to be popping up consistently.

Issue #733, opened 2022-01-06
`boost::multiprecision` linker errors


mborland:
@jzmaddock I am not quite sure if this is from the recent `__float128` merge into multi-precision, but the drone runs are now showing linker errors like the below.

```
gcc.compile.c++ ../../../bin.v2/libs/math/test/test_tgamma_for_issue396_part1.test/gcc-10/debug/debug-symbols-off/link-static/threading-multi/visibility-hidden/test_tgamma_for_issue396_part1.o
1723s
2028	gcc.link ../../../bin.v2/libs/math/test/test_tgamma_for_issue396_part1.test/gcc-10/debug/debug-symbols-off/link-static/threading-multi/visibility-hidden/test_tgamma_for_issue396_part1
1724s
2029	/usr/bin/ld: ../../../bin.v2/libs/math/test/test_tgamma_for_issue396_part1.test/gcc-10/debug/debug-symbols-off/link-static/threading-multi/visibility-hidden/test_tgamma_for_issue396_part1.o: in function `boost::multiprecision::float128_procs::ldexp(__float128, int)':
1724s
2030	test_tgamma_for_issue396_part1.cpp:(.text._ZN5boost14multiprecision14float128_procs5ldexpEgi[_ZN5boost14multiprecision14float128_procs5ldexpEgi]+0x1e): undefined reference to `ldexpq'
1724s
2031	/usr/bin/ld: ../../../bin.v2/libs/math/test/test_tgamma_for_issue396_part1.test/gcc-10/debug/debug-symbols-off/link-static/threading-multi/visibility-hidden/test_tgamma_for_issue396_part1.o: in function `boost::multiprecision::float128_procs::frexp(__float128, int*)':
1724s
2032	test_tgamma_for_issue396_part1.cpp:(.text._ZN5boost14multiprecision14float128_procs5frexpEgPi[_ZN5boost14multiprecision14float128_procs5frexpEgPi]+0x21): undefined reference to `frexpq'
1724s
2033	/usr/bin/ld: ../../../bin.v2/libs/math/test/test_tgamma_for_issue396_part1.test/gcc-10/debug/debug-symbols-off/link-static/threading-multi/visibility-hidden/test_tgamma_for_issue396_part1.o: in function `boost::multiprecision::float128_procs::floor(__float128)':
1724s
[...]

PR #735, opened 2022-01-07
Fix for linker errors from multiprecision changes.

jzmaddock:
Fixes https://github.com/boostorg/math/issues/733.


ckormanyos:
Nice!

Just to be sure... So MSVC build(s) simply remain unchanged, right? And do not erroneously try to use 128-bit artifacts either?

jzmaddock:
>Just to be sure... So MSVC build(s) simply remain unchanged, right? And do not erroneously try to use 128-bit artifacts either?

Correct.  You only hit this if __float128 is an available type, and your precision is < 113 bits: in which case you should probably be using __float128 in the first place.

jzmaddock:
Failures are CI SNAFU's, merging...

PR #736, opened 2022-01-08
Constexpr comparison functions

mborland:
Implements: `isgreater`, `isgreaterequal`, `isless`, and `islessequal`


mborland:
@NAThompson This should be good to go. The CI failure was fixed by #733 and the force-push was a rebase onto that commit.

PR #745, opened 2022-01-20
constexpr isunordered

mborland:


mborland:
@NAThompson This is pretty trivial if you'd like to review before merge.

Issue #748, opened 2022-01-26
Building on MinGW fails due to missing "std::mutex"


yaronct:
MinGW-w64 (GCC targetting Windows) that uses the "win32" threading model, doesn't provide some threading stuff like "std::mutex" (I'm not sue about the "posix" threading model), therefore building fails with:

./boost/math/special_functions/detail/bernoulli_details.hpp:563:9: error: ‘mutex’ in namespace ‘std’ does not name a type
  563 |    std::mutex m_mutex;
      |         ^~~~~

This can probably be solved by using threading stuff from the boost thread library (in the case of MinGW), as I think boost thread gives an API compatible with that of the standard C++ library ("std"). So I think, e.g. using "boost::mutex" should be compatible with "std::mutex". I haven't tried it though. Alternately, building boost math can be disabled on MinGW (possibly only for the "win32" threading model).

Tested against boost 1.78.0.

mborland:
Can you try the linked PR, and let me know if it solves your issue?

yaronct:
Thanks, @jzmaddock, #753 solves the problem. Or at least eliminates it :smile:

PR #749, opened 2022-01-27
Fix for #748

mborland:


jzmaddock:
I suspect this may be a bit of a blunt instrument, I'm hoping this will get it: https://github.com/boostorg/math/pull/753

Issue #750, opened 2022-01-28
CI Hangs


mborland:
MSVC-14.2 running the `misc` test suite under C++17 fails for timeout every run now. Investigating as there have been enough runs to imply something has gone wrong.

PR #751, opened 2022-01-28
Fix for #750

mborland:


mborland:
Once https://github.com/boostorg/multiprecision/pull/422 is merged I will merge this and we should be back to green.

ckormanyos:
> Once (boostorg/multiprecision#422)[https://github.com/boostorg/multiprecision/pull/422] is merged ...

Done. Thanks everyone for driving forward so dedicatedly on this.

NAThompson:
@mborland : Sorry, do I need to do anything? I'm confused about what in the quartic roots code caused the issue. . . .

mborland:
@NAThompson I don't think anything needs to happen. I couldn't find any compelling reason that quartic roots would cause the issue either. I would wait and see if anybody reports and issue from the field.

jzmaddock:
> Sorry, do I need to do anything? I'm confused about what in the quartic roots code caused the issue

Unfortunately there is an issue (though I admit the msvc std lib is being a bit anal about this, if technically correct).

The problem is that from latest msvc-14.2 onwards, the std lib will assert if you attempt to sort anything that contains NaN's.  This is technically correct in that the behaviour is undefined in this case, we just haven't hit a std library which checked that before.  The problem here is that the quartic roots code, will attempt to sort a list containing Nan's in almost every case.... so yes, the code is now completely broken with latest msvc.  I would expect -fsanitize=undefined to trap this as well, but I haven't checked that.  IMO we need to be just sorting the N roots found, and not the NaN's at the end of the list.

Hope that all makes sense.

NAThompson:
@jzmaddock : Welp . . . yup this is a bug users will hit then. Will attempt a fix.

PR #752, opened 2022-01-29
Color Maps

mborland:
@NAThompson Here's my cut at implementing color tables without as much code duplication. I also think this is a good opportunity to start using C++20 concepts. [Compiler support](https://en.cppreference.com/w/cpp/compiler_support/20) seems pretty good.


NAThompson:
@mborland : I like the `concepts`. 

Two comments:

What about:

```
boost::math::tools::kindlemann
```

rather than

```
boost::math::tools::kindlemann_color_map
```

?

I wonder if the latter is too verbose.

Next, will we get any performance wins if we use:

```
std::array<std::array<Real, 3>, N>
```

rather than:

```
std::array<Real, N> r;
std::array<Real, N> g;
std::array<Real, N> b;
```

?

mborland:
1) I think using the full name describes the intent better, but agree it is verbose. Since it's an object type the user will only have to type that full thing once.
2) I don't think in this case we can gain anything because we can't convert this into a compressed matrix (that I know of at least)
3) Agreed this should be constexpr so now the entire class is. The test file validates use in constexpr contexts.
4) Not out of the box. You can write your own concepts which as we move forward will need to be done (e.g. to support boost.multiprecision types).

NAThompson:
>  I think using the full name describes the intent better, but agree it is verbose. Since it's an object type the user will only have to type that full thing once.

So you're thinking the users will do:

```
auto kindlemann = boost::math::tools::kindlemann_color_map()
```

and the less verbose syntax will occur in user code, rather than in ours?

> I don't think in this case we can gain anything because we can't convert this into a compressed matrix (that I know of at least)

Well, you might get gains from having the referenced elements being contiguous in memory, but of course it won't change any fundamental complexity . . .



... If these are all `constexpr` do they need to be classes? I feel like we may be able to get them to be straight function calls.

mborland:
> 
> So you're thinking the users will do:
> 
> ```
> auto kindlemann = boost::math::tools::kindlemann_color_map()
> ```
> 
> and the less verbose syntax will occur in user code, rather than in ours?
> 

Yes, then all the user has to do is call `kindlemann(scalar)` 

> > I don't think in this case we can gain anything because we can't convert this into a compressed matrix (that I know of at least)
> 
> Well, you might get gains from having the referenced elements being contiguous in memory, but of course it won't change any fundamental complexity . . .
> 
> ... If these are all `constexpr` do they need to be classes? I feel like we may be able to get them to be straight function calls.

To me derived classes was the easiest way to eliminate code repetition. It also allowed for an easy implementation of `custom_color_map` where the user can define whatever values they want. 

You mentioned using boost.GIL for an example. Do you have one written already I could edit to make this work? I have never used GIL before.

NAThompson:
> You mentioned using boost.GIL for an example. Do you have one written already I could edit to make this work? I have never used GIL before.

Actually I've never used it either. We can even write a .pnm to a textfile and then screenshot it for the docs.

Let me think of a 2D function which would nicely get plotted with these color maps. [This one](https://github.com/NAThompson/DifferentiableRaytracing/blob/master/figures/newton_fractal_z8_15z4_m16_viridis.png) might be good to show off viridis. . .

mborland:
@NAThompson I ended up modifying your example from DRT, and get the following output:

newton_fractal

NAThompson:
@mborland : That looks awesome! Looks like it wasn't too much work to get there either . . .

I agree that the calls aren't too verbose . . .

NAThompson:
@mborland : It occurs to me that the user might want to try multiple color tables without recompilation, say, via a command line flag. Giving that a try . . .

include/boost/math/tools/color_maps.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+namespace boost::math::tools {
+
+namespace detail {
+
+template <BOOST_MATH_FLOATING_POINT_TYPE Real>
+class color_map
+{
+private:
+    std::array<Real, 256> r;
```
NAThompson: generally the convention is the member variables have the underscore, e.g. `std::array<Real, 256> r_;`

include/boost/math/tools/color_maps.hpp
```diff
@@ -0,0 +1,156 @@
...
+    std::array<Real, 256> g;
```
NAThompson: One thing I worry about with this design is that there is no fundamental reason that users should be required to choose an array of length 256. You can see I simply chose one length from the [original reference](https://www.kennethmoreland.com/color-advice/), but the reference itself specifies that there is an analytic form to many of these which indicates that it might be very reasonable for a user to specify basically any continuous function from [0,1]->[0,1]^3.. .

include/boost/math/tools/color_maps.hpp
```diff
@@ -0,0 +1,156 @@
...
...9509923357,0.9509013387619204,0.9521193680507284,0.9532833326102016,0.9543957893212712,0.9554530885899511,0.9564564274893136,0.9574039803720095,0.9582930778369493,0.9591252451658657,0.9598979045875052,0.9606124721276129,0.9612663555645452,0.9618587615061226,0.9623887555724971,0.9628559105126825,0.9632594414845331,0.9635970174364671,0.9638686962955997,0.964073330664618,0.9642088982457305,0.9642762050566095,0.9642722292530789,0.9641978734579372,0.9640500241140576,0.9638285003224122,0.9635333683894992,0.9631614522931872,0.962712538770477,0.9621866932888067,0.9615816256657105,0.960894345402275,0.9601234461397462,0.959270237581007,0.958333448735586,0.9573126236007495,0.9562062303425974,0.9550135580216242,0.9537297610630108,0.9523531724953679,0.9508874276947369,0.9493337129963176,0.9476874510448859,0.9459406787196255,0.944103300221897,0.9421780061800054,0.9401426587007371,0.9380209885848168,0.9358000817994506,0.9334816723713976,0.9310720730839354,0.9285606186650813,0.9259594465149981,0.92327053647064,0.9204816890380406,0.9176163146663037,0.9146765909646813,0.9116700174610778,0.9086068264551928,0.9055122790291467,0.9024220298268155,0.8994080090565665,0.8965597700566809,0.894058310302958},
```
NAThompson: Now that I think of it, I doubt we're even coming close to achieving double precision resolution. Would it make sense to replace these character by a 32 bit hexfloat representation?

include/boost/math/tools/color_maps.hpp
mborland: I changed the base class to take a `std::size_t` instead of constraining ourselves to 256. The `custom_color_map` can now be specified by for example `custom_color_map<double, 512>(r, g, b)`, but our provided ones are not affected.

include/boost/math/tools/color_maps.hpp
mborland: We could. I don't think their are many displays in circulation that have a 32 bit color depth.

include/boost/math/tools/color_maps.hpp
```diff
@@ -0,0 +1,157 @@
...
+    constexpr std::array<Real, 3> operator()(Real scalar) const noexcept
+    {
+        scalar = std::clamp(scalar, static_cast<Real>(0), static_cast<Real>(1));
+
+        if (scalar == static_cast<Real>(1))
+        {
+            return {r_.back(), g_.back(), b_.back()};
+        }
+
+        Real s = (N-1) * scalar;
```
NAThompson: Isn't this incorrect? 255 is there as it's the max of the uint8_t right?

include/boost/math/tools/color_maps.hpp
```diff
@@ -14,6 +14,32 @@
 #include <utility>
 #include <boost/math/tools/concepts.hpp>
 #include <boost/math/ccmath/floor.hpp>
+#include <boost/math/tools/assert.hpp>
+
+#if __has_include("lodepng.h")
```
NAThompson, line 19: Actually this `__has_include` is a really nice trick . . .

NAThompson:
@mborland : I personally like huge images but the maintainers of the global boost package will complain if we add a 1.3MB file to the distribution. Could we make the image smaller?

ckormanyos:
Can the color maps be used on histogram data from fractal iterations such as Mandelbrot set?

ckormanyos:
>> You mentioned using boost.GIL for an example. Do you have one written already I could edit to make this work? I have never used GIL before.

> Actually I've never used it either. 

I have a non-published project that uses Boost.Gil in combination with a jpeg library that I made mostly to be portable as well. I can't find the origin of the original source code of my ported jpeg library any more. It seems to come from last century. I don't know if it can be redistributed.

Aside from the jpeg LIB potential availability/portability problem, I ended up going from `boost::gil::jpeg_write_view` directly to jpeg via `boost::gil::jpeg_write_view`. But this does mandate a jpeg eingine (LIB) in the project.

NAThompson:
> Can the color maps be used on histogram data from fractal iterations such as Mandelbrot set?

Of course!

mborland:
@NAThompson Here is the performance data

```
Unable to determine clock rate from sysctl: hw.cpufrequency: No such file or directory
2022-02-01T22:19:08+02:00
Running ./color_tables_performance
Run on (10 X 24.0833 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x10)
  L1 Instruction 128 KiB (x10)
  L2 Unified 4096 KiB (x5)
Load Average: 1.51, 1.51, 1.58
-----------------------------------------------------------------------------------------------
Benchmark                                                     Time             CPU   Iterations
-----------------------------------------------------------------------------------------------
color_table_benchmark<float>/64/real_time                  4144 ns         4144 ns       125506
color_table_benchmark<float>/128/real_time                 8230 ns         8230 ns        84417
color_table_benchmark<float>/256/real_time                16417 ns        16417 ns        42445
color_table_benchmark<float>/512/real_time                32776 ns        32776 ns        21361
color_table_benchmark<float>/1024/real_time               65499 ns        65499 ns        10596
color_table_benchmark<float>/2048/real_time              130932 ns       130932 ns         5319
color_table_benchmark<float>/4096/real_time              261722 ns       261716 ns         2671
color_table_benchmark<float>/8192/real_time              523603 ns       523604 ns         1322
color_table_benchmark<float>/16384/real_time            1046917 ns      1046917 ns          663
color_table_benchmark<float>/32768/real_time            2093535 ns      2093536 ns          334
color_table_benchmark<float>/65536/real_time            4185543 ns      4185545 ns          167
color_table_benchmark<float>/131072/real_time           8373275 ns      8373277 ns           83
color_table_benchmark<float>/262144/real_time          16747412 ns     16747405 ns           42
color_table_benchmark<float>/524288/real_time          33508554 ns     33508524 ns           21
color_table_benchmark<float>/1048576/real_time         66997783 ns     66997800 ns           10
color_table_benchmark<float>/real_time_BigO               63.90 N         63.90 N
color_table_benchmark<float>/real_time_RMS                    0 %             0 %
color_table_benchmark<double>/64/real_time                 4112 ns         4112 ns       170016
color_table_benchmark<double>/128/real_time                8185 ns         8185 ns        84702
color_table_benchmark<double>/256/real_time               16317 ns        16317 ns        42690
color_table_benchmark<double>/512/real_time               32595 ns        32595 ns        21467
color_table_benchmark<double>/1024/real_time              65187 ns        65187 ns        10642
color_table_benchmark<double>/2048/real_time             131284 ns       131265 ns         5228
color_table_benchmark<double>/4096/real_time             267619 ns       267287 ns         2637
color_table_benchmark<double>/8192/real_time             536032 ns       535327 ns         1286
color_table_benchmark<double>/16384/real_time           1059829 ns      1059212 ns          646
color_table_benchmark<double>/32768/real_time           2108699 ns      2107826 ns          327
color_table_benchmark<double>/65536/real_time           4198880 ns      4197747 ns          166
color_table_benchmark<double>/131072/real_time          8501609 ns      8492410 ns           83
color_table_benchmark<double>/262144/real_time         16936271 ns     16925381 ns           42
color_table_benchmark<double>/524288/real_time         33957917 ns     33924333 ns           21
color_table_benchmark<double>/1048576/real_time        67226867 ns     67208000 ns           10
color_table_benchmark<double>/real_time_BigO              64.27 N         64.24 N
color_table_benchmark<double>/real_time_RMS                   1 %             1 %
color_table_benchmark<long double>/64/real_time            4113 ns         4113 ns       170037
color_table_benchmark<long double>/128/real_time           8183 ns         8183 ns        84687
color_table_benchmark<long double>/256/real_time          16403 ns        16402 ns        42749
color_table_benchmark<long double>/512/real_time          32592 ns        32592 ns        21458
color_table_benchmark<long double>/1024/real_time         65143 ns        65143 ns        10656
color_table_benchmark<long double>/2048/real_time        130240 ns       130240 ns         5349
color_table_benchmark<long double>/4096/real_time        263375 ns       263214 ns         2681
color_table_benchmark<long double>/8192/real_time        522375 ns       522349 ns         1329
color_table_benchmark<long double>/16384/real_time      1041733 ns      1041734 ns          669
color_table_benchmark<long double>/32768/real_time      2082051 ns      2082048 ns          335
color_table_benchmark<long double>/65536/real_time      4164367 ns      4164363 ns          168
color_table_benchmark<long double>/131072/real_time     8332090 ns      8332096 ns           83
color_table_benchmark<long double>/262144/real_time    16670387 ns     16670262 ns           42
color_table_benchmark<long double>/524288/real_time    33341579 ns     33341571 ns           21
color_table_benchmark<long double>/1048576/real_time   66667208 ns     66667200 ns           10
color_table_benchmark<long double>/real_time_BigO         63.58 N         63.58 N
color_table_benchmark<long double>/real_time_RMS              0 %             0 %
```

NAThompson:
@mborland : Shouldn't the call be constant time?

mborland:
@NAThompson This is now clean with CI. Any other changes, or good to merge?

NAThompson:
I still want to check if we can make the color data an array of structs, rather than three arrays, and see how it affects the performance. Gimme a few days . . .

NAThompson:
Ok I managed to hack up a script to get the color tables into an array of structs:

```julia
using CSV
using DataFrames
csv_reader = CSV.File("viridis-table-float-0256.csv")
print("std::array<std::array<double, 3>, 256>> = {{ )
for row in csv_reader
    println("{$(row.RGB_r), $(row.RGB_g), $(row.RGB_b)}, ")
end
print("}};")
```

@mborland : Accidently amend committed on a shared branch; my bad. Could you rerun the performance data? I'm getting:


```
./a.out                                                                                           color_tables ✱ ◼
2022-02-06T10:52:23-08:00
Running ./a.out
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 256 KiB (x2)
  L3 Unified 3072 KiB (x1)
Load Average: 2.25, 1.68, 1.69
------------------------------------------------------------------------
Benchmark                              Time             CPU   Iterations
------------------------------------------------------------------------
color_table_benchmark<float>        2.30 ns         2.23 ns    303114574
color_table_benchmark<double>       6.56 ns         6.39 ns    108312187
```

NAThompson:
@mborland : Nice diffs. I think I'll put swatches for all the color maps in the docs and then we'll be done.

NAThompson:
@mborland : Could you put eyes on the docs for me the squash merge once you're happy?

mborland:
Docs look good to me now so merging.

PR #753, opened 2022-01-30
Rework thread-safety logic in bernoulli_details.hpp.

jzmaddock:
Fixes https://github.com/boostorg/math/issues/748.


mborland:
Thanks for taking this over. I expect it will also help #729 because I am seeing a lot of failures from `Bernoulli_details` probably from me taking a ham fisted approach.

PR #756, opened 2022-02-05
Add class template argument deduction guides for distributions

### LINKED ISSUE #754: Unexpected instantiation of `chi_squared_distribution` in C++17
The following is valid in C++17, AFAIK.
```c++
#include <iostream>
#include <boost/math/distributions/chi_squared.hpp>

int main()
{
    int dof=2;
    boost::math::chi_squared_distribution chiSqi(dof);
    boost::math::chi_squared_distribution<double> chiSq(dof);

    std::cout << "chiSqi cdf: " << boost::math::cdf(chiSqi,3.1415) << std::endl;
    std::cout << "chiSq cdf: " << boost::math::cdf(chiSq,3.1415) << std::endl;

    return 0;
}
```

The above code produces the following *unexpected* output under GCC and clang with `-std=c++17`.
```
chiSqi cdf: 0
chiSq cdf: 0.792111
```

In C++17, we may be able to omit the class template arguments ([CTAD](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction)) in favor of allowing the compiler to deduce them.
`chi_squared_distribution` is parameterized by a `RealType` template argument, defaulting to a `double`.
In the first case, where I don't explicitly provide the template argument, the compiler deduces the template parameter `RealType` to be an `int` based on the argument passed to the ctor.

I think it is unexpected/bad to allow instantiating a `chi_squared_distribution` with an *integral* `RealType`.
The ctor should happily accept an integral degrees of freedom, but the type used for the calculations shouldn't be integral.
Perhaps `chi_squared_distribution` could use `enable_if` to check that `RealType` `is_floating_point`?
This would force the user to explicitly provide the class template argument in cases where CTAD chooses the "wrong" type.

I'd be happy to submit a PR. Is the proposed solution okay? Is there a better solution?

Additionally, there may be distributions other than `chi_squared_distribution` that are affected by the same pattern.

os: Arch Linux, Linux 5.16.3
boost version: 1.78.0-1 (from Arch repo)
compiler: GCC 11.1.0, clang 13.0.0-4

jamesfolberth:
Fixes #754 by adding deduction guides for the case where the class template arguments are not given explicitly.
Follows the demonstration by @jzmaddock.

I didn't put in `static_assert` for silly cases, e.g., where the user writes `chi_squared_distribution<int>`.
If that's what they write, so be it.


jamesfolberth:
Putzing around in Godbolt, it looks like GCC-8 supports pre-final C++17 deduction guides (`__cpp_deduction_guides==201611`) and variadic templates, but they're not playing together well in `test_dist_deduction_guides.cpp`. Manually providing an overload of `test_deduction_guide` for the no arg case should hopefully do the trick.

jamesfolberth:
I'm not sure what's going on with `naive_monte_carlo_test` (case 5?) failing during the CI run ...

jamesfolberth:
Well, clang 6-8 are now unhappy with the "no arg" case. These use the default template arg, not the deduction guide, right? So I think we can just get rid of those tests.

I believe I've got some of the CI running locally, and removing the "no arg" tests seems to be working.

mborland:
@jamesfolberth I approved your PR to run against our CI so you should start seeing results.

mborland (approved):

mborland:
This PR looks good to me. Do you have anything else to add, and does it solve your problem?

NAThompson:
Looks good to me as well. Might want to let @jzmaddock weigh in though; I've pwned myself with argument deduction subtleties before . . .

jamesfolberth:
This was all I had planned for this PR. It does indeed solve the problem I
originally encountered.

Thanks for taking a look. Getting a few of the older compilers to be happy
with the test was a bit delicate.

include/boost/math/distributions/fisher_f.hpp
```diff
@@ -54,6 +54,13 @@ class fisher_f_distribution
 
 typedef fisher_f_distribution<double> fisher_f;
 
+#ifdef __cpp_deduction_guides
+template <class RealType>
+fisher_f_distribution(RealType)->fisher_f_distribution<typename boost::math::tools::promote_args<RealType>::type>;
```
jzmaddock, line 59: fisher_f doesn't have a single-arg constructor?

jzmaddock:
Other than the single comment on the fisher_f this looks good to go to me - many thanks to @jamesfolberth for seeing this through!

jamesfolberth:
Oh dang nice catch! Thanks all for the help. Happy to contribute!

jamesfolberth:
FWIW looks like CI is failing waiting on a runner.

jzmaddock:
Ugh, that's just silly.  Anyone any objections to merging anyway?  All looks good to me.

ckormanyos:
I've had the runner farm _go_ _on_ _strike_ a few times. I'm not exactly sure what causes that. Short of stopping and restarting the _entire_ run on GHA that run is toasted. So unless we restart entirely, which seems uwaranted, content gets the nod.

jamesfolberth:
I think we're safe, so I've no problem with merging anyway.

After my 3rd commit (6bd06cf77277f97ba53ca063a87e18d33aa77a7a) I ran locally `test_dist_deduction_guides` with all the compilers and language versions for `ubuntu-focal` and `ubuntu-bionic`. Given that that passed and that CI ran on my 2nd commit, I think we're in the clear.

PR #758, opened 2022-02-09
Add GCC-5 to drone config

mborland:
Based off conversation in https://github.com/boostorg/boost/issues/610 add GCC-5 back to our testing.


mborland:
@jzmaddock The [drone run is clean](https://drone.cpp.al/boostorg/math/591) with GCC-5. Is this a change you want to pursue?

CC: @ckormanyos

ckormanyos:
> with GCC-5

We never really formalized plans here. But for clarity's sake,...

- I would like to try for GCC-5 on both Math and Multiprecision for 1.79.
- In the summer, we can see how things look in CI for runners on 4.8 or 4.9.

If it 4.x goes well, we can include these as well. If not, then simply bail on that legacy support. We might have to change some of the PP logic to go back down to 4. or we might decide to let it go unsuccessfully if the work is not warranted by the gains.

PR #762, opened 2022-02-15
Logpdf support

mborland:


include/boost/math/distributions/arcsine.hpp
```diff
@@ -366,6 +366,34 @@ namespace boost
...
+        function,
+        lo, hi, x,
+        &result, Policy()))
+      {
+        return result;
+      }
+
+      using boost::math::constants::pi;
+      using boost::math::constants::half;
+      result = -half<RealType>() * log(-(x - 1) * x) - log(pi<RealType>());
```
jzmaddock, line 393: That assumes the support is `[0,1]` but is actually `[this->lo, this->hi]`

mborland: I realized I messed that up when I added the tests. Fixed in latest commit.

jzmaddock:
Thanks Matt, that's more or less along the lines I was thinking too.

I think we can safely add a default implementation:

```
template <class Distribution>
inline typename Distribution::value_type logpdf(const Distribution& dist, const typename Distribution::value_type& x)
{
  using std::log;
   return log(pdf(dist, x));
}
```

For those distributions that do not have exponential-like pdf's.  The Arcsine is actually a good case in point, do we need an actual implementation here, or is the default good enough?  I can see some arguments either way: if `this->hi` is super large then (x-lo)(hi-x) could overflow (but no one has ever complained), on the other hand, evaluating by logs as you do, may lead to cancellation error when the pdf is close to 1.  And I doubt there is much performance difference between them.

With regard to testing... we already have decent tests for the pdf's, so I suspect that a basic sanity check against the pdf may be good enough in thee cases?

test/test_normal.cpp
```diff
@@ -219,6 +221,13 @@ void test_spots(RealType)
...
+   RealType temp_tol = tolerance;
+
+   BOOST_IF_CONSTEXPR (std::is_same<long double, RealType>::value)
+   {
+      tolerance *= 100;
+   }
+
    BOOST_CHECK_CLOSE(
       logpdf(normal_distribution<RealType>(), static_cast<RealType>(0)),
       log(static_cast<RealType>(0.3989422804014326779399460599343818684759L)), // 1/sqrt(2*pi)
```
NAThompson, line 233: Isn't 1/sqrt(2pi) in boost::math::constants?

mborland:
@jzmaddock This is clean and good for review. Logpdf has been specialized when there are exponentials in the pdf and there rest will default to the naive `log(pdf)`. The only place where I had to adjust tolerances was in the test of long doubles on the normal distribution. The naive, specialized, nor hard coding the result from WolframAlpha got me into the existing tolerance band.

include/boost/math/distributions/extreme_value.hpp
```diff
@@ -122,6 +122,31 @@ inline RealType pdf(const extreme_value_distribution<RealType, Policy>& dist, co
...
+   if(0 == detail::check_finite(function, a, &result, Policy()))
+      return result;
+   if((boost::math::isinf)(x))
+      return 0.0f;
+   if(0 == detail::check_x(function, x, &result, Policy()))
+      return result;
+   RealType e = (a - x) / b;
+   if(e < tools::log_max_value<RealType>())
+      result = log(1/b) + e - exp(e);
+   // else.... result *must* be zero since exp(e) is infinite...
```
jzmaddock, line 146: Shouldn't the result be -INF if `e < tools::log_max_value<RealType>()` doesn't hold?

include/boost/math/distributions/chi_squared.hpp
```diff
@@ -131,6 +131,47 @@ RealType pdf(const chi_squared_distribution<RealType, Policy>& dist, const RealT
...
+      {
+         return -boost::math::constants::ln_two<RealType>();
+      }
+      else
+      {
+         return -std::numeric_limits<RealType>::infinity();
+      }
+   }
+
+   return log(pdf(dist, chi_square));
```
jzmaddock, line 172: Couldn't this rely on the default implementation since we do nothing special here?

include/boost/math/distributions/exponential.hpp
```diff
@@ -127,6 +127,24 @@ inline RealType pdf(const exponential_distribution<RealType, Policy>& dist, cons
...
 
+template <class RealType, class Policy>
+inline RealType logpdf(const exponential_distribution<RealType, Policy>& dist, const RealType& x)
+{
+   BOOST_MATH_STD_USING // for ADL of std functions
+
+   static const char* function = "boost::math::logpdf(const exponential_distribution<%1%>&, %1%)";
+
+   RealType lambda = dist.lambda();
+   RealType result = 0;
```
jzmaddock, line 138: We're going to have issues with copying over the special case handling in all of these I think, we may need to think about this, but to be consistent with taking the log of the pdf, we should start with result = -INF for these special cases.

include/boost/math/distributions/extreme_value.hpp
```diff
@@ -122,6 +122,31 @@ inline RealType pdf(const extreme_value_distribution<RealType, Policy>& dist, co
...
+template <class RealType, class Policy>
+inline RealType logpdf(const extreme_value_distribution<RealType, Policy>& dist, const RealType& x)
+{
+   BOOST_MATH_STD_USING // for ADL of std functions
+
+   static const char* function = "boost::math::logpdf(const extreme_value_distribution<%1%>&, %1%)";
+
+   RealType a = dist.location();
+   RealType b = dist.scale();
+   RealType result = 0;
```
jzmaddock, line 134: Special cases should probably be -INF again.

include/boost/math/distributions/gamma.hpp
```diff
@@ -147,6 +148,41 @@ inline RealType pdf(const gamma_distribution<RealType, Policy>& dist, const Real
...
+{
+   BOOST_MATH_STD_USING  // for ADL of std functions
+   using boost::math::tgamma;
+
+   static const char* function = "boost::math::logpdf(const gamma_distribution<%1%>&, %1%)";
+
+   RealType k = dist.shape();
+   RealType theta = dist.scale();
+
+   RealType result = 0;
```
jzmaddock, line 162: -INF

include/boost/math/distributions/inverse_gamma.hpp
```diff
@@ -195,6 +196,43 @@ inline RealType pdf(const inverse_gamma_distribution<RealType, Policy>& dist, co
...
+{
+   BOOST_MATH_STD_USING  // for ADL of std functions
+   using boost::math::tgamma;
+
+   static const char* function = "boost::math::logpdf(const inverse_gamma_distribution<%1%>&, %1%)";
+
+   RealType shape = dist.shape();
+   RealType scale = dist.scale();
+
+   RealType result = 0;
```
jzmaddock, line 210: -INF for all the special cases I *think*.  Need to change the `return 0;` to `return result;` as well.

include/boost/math/distributions/inverse_gamma.hpp
```diff
@@ -195,6 +196,43 @@ inline RealType pdf(const inverse_gamma_distribution<RealType, Policy>& dist, co
...
+   if(result < tools::min_value<RealType>())
+      return 0;  // random variable is infinite or so close as to make no difference.
+      
+   // x * x may under or overflow, likewise our result
+   if (!(boost::math::isfinite)(x*x))
+   {
+      return policies::raise_overflow_error<RealType, Policy>(function, "PDF is infinite.", Policy());
+   }
+
+   return shape * log(scale) + (-shape-1)*log(x) - log(tgamma(shape)) - (scale/x);
```
jzmaddock, line 233: lgamma

include/boost/math/distributions/inverse_gaussian.hpp
```diff
@@ -174,6 +174,43 @@ inline RealType pdf(const inverse_gaussian_distribution<RealType, Policy>& dist,
...
 } // pdf
 
+template <class RealType, class Policy>
+inline RealType logpdf(const inverse_gaussian_distribution<RealType, Policy>& dist, const RealType& x)
+{ // Probability Density Function
+   BOOST_MATH_STD_USING  // for ADL of std functions
+
+   RealType scale = dist.scale();
+   RealType mean = dist.mean();
+   RealType result = 0;
```
jzmaddock, line 184: -INF.

include/boost/math/distributions/laplace.hpp
```diff
@@ -150,6 +150,44 @@ inline RealType pdf(const laplace_distribution<RealType, Policy>& dist, const Re
    return result;
 } // pdf
 
+template <class RealType, class Policy>
+inline RealType logpdf(const laplace_distribution<RealType, Policy>& dist, const RealType& x)
+{
+   BOOST_MATH_STD_USING // for ADL of std functions
+
+   // Checking function argument
+   RealType result = 0;
```
jzmaddock, line 159: -INF, also a literal zero below to replace.

include/boost/math/distributions/normal.hpp
```diff
@@ -165,6 +166,42 @@ inline RealType pdf(const normal_distribution<RealType, Policy>& dist, const Rea
...
+inline RealType logpdf(const normal_distribution<RealType, Policy>& dist, const RealType& x)
+{
+   BOOST_MATH_STD_USING  // for ADL of std functions
+
+   RealType sd = dist.standard_deviation();
+   RealType mean = dist.mean();
+
+   static const char* function = "boost::math::logpdf(const normal_distribution<%1%>&, %1%)";
+
+   RealType result = 0;
```
jzmaddock, line 179: -INF and another literal zero.

include/boost/math/distributions/poisson.hpp
```diff
@@ -281,6 +282,42 @@ namespace boost
...
+    RealType logpdf(const poisson_distribution<RealType, Policy>& dist, const RealType& k)
+    {
+      BOOST_FPU_EXCEPTION_GUARD
+
+      BOOST_MATH_STD_USING // for ADL of std functions.
+      using boost::math::tgamma;
+
+      RealType mean = dist.mean();
+      // Error check:
+      RealType result = 0;
```
jzmaddock, line 295: -INF

include/boost/math/distributions/poisson.hpp
```diff
@@ -281,6 +282,42 @@ namespace boost
...
+      // Special case of mean zero, regardless of the number of events k.
+      if (mean == 0)
+      { // Probability for any k is zero.
+        return std::numeric_limits<RealType>::quiet_NaN();
+      }
+      
+      // Special case where k and lambda are both positive
+      if(k > 0 && mean > 0)
+      {
+        return -log(tgamma(k+1)) + k*log(mean) - mean;
```
jzmaddock, line 314: lgamma

include/boost/math/distributions/rayleigh.hpp
```diff
@@ -122,6 +122,32 @@ inline RealType pdf(const rayleigh_distribution<RealType, Policy>& dist, const R
    return result;
 } // pdf
 
+template <class RealType, class Policy>
+inline RealType logpdf(const rayleigh_distribution<RealType, Policy>& dist, const RealType& x)
+{
+   BOOST_MATH_STD_USING // for ADL of std function exp.
+
+   const RealType sigma = dist.sigma();
+   RealType result = 0;
```
jzmaddock, line 131: -INF and another literal zero.

include/boost/math/distributions/weibull.hpp
```diff
@@ -157,6 +157,40 @@ inline RealType pdf(const weibull_distribution<RealType, Policy>& dist, const Re
...
+         return log(1 / scale);
+      }
+      if(shape > 1)
+      {
+         return 0;
+      }
+      return policies::raise_overflow_error<RealType>(function, 0, Policy());
+   }
+   
+   result = log(shape * pow(scale, -shape) * pow(x, shape - 1)) - pow(x / scale, shape);
```
jzmaddock, line 189: Is there any reason not to expand this right out to:

`log(shape) - shape * log(scale) + log(x) * (shape - 1) - pow(x / scale, shape);`

jzmaddock:
There's a few comments above, I haven't checked the details of the formulas, and am replying on your tests.

We should also add logpdf to the DistributionConcept checks in test/compile_test/test_compile_result.hpp

Otherwise, looks good, much thanks for taking this on!

test/test_arcsine.cpp
```diff
@@ -292,6 +301,16 @@ void test_spots(RealType)
     BOOST_CHECK_CLOSE_FRACTION(pdf(arcsine_01, static_cast<RealType>(1) - tolerance),
       1 /(sqrt(tolerance) * boost::math::constants::pi<RealType>()), 2 * tolerance); //
 
+    // Log PDF
+    BOOST_CHECK_CLOSE_FRACTION(logpdf(arcsine_01, 0.000001), static_cast<RealType>(5.7630258931329868780772138043668005779060097243996L), tolerance);
```
NAThompson, line 305: One thing I worry about these tests is that if they fail at some later date, it's not clear why (say) `5.7630258931329868780772138043668005779060097243996L` is the correct value. Is there a property that can be tested instead?

test/test_gamma_dist.cpp
```diff
@@ -88,6 +88,14 @@ void check_gamma(RealType shape, RealType scale, RealType x, RealType p, RealTyp
          x),                                            // random variable.
          NaivePDF(shape, scale, x),                     // PDF
          tol);                                          // %tolerance.
+   
+   // LOGPDF:
+   BOOST_CHECK_CLOSE(
+      boost::math::logpdf(
+         gamma_distribution<RealType>(shape, scale),                             // distribution.
+         x),                                                                     // random variable.
+         log(boost::math::pdf(gamma_distribution<RealType>(shape, scale), x)),   // PDF
```
NAThompson, line 97: These are the sort of properties I'm thinking of--namely using two different implementations to compute the same thing. It's immediately obvious from this one that if the test fails something is *definitely* wrong.

mborland:
@jzmaddock Your review comments have been addressed; I apologize for the delay. I also ran the sonar lint and clang-tidy combs through this as suggested by @ckormanyos.

include/boost/math/distributions/gamma.hpp
```diff
@@ -147,6 +148,41 @@ inline RealType pdf(const gamma_distribution<RealType, Policy>& dist, const Real
...
+   }
+   
+   // The following calculation does not always work with float so take the naive road out
+   BOOST_IF_CONSTEXPR(std::is_same<RealType, float>::value)
+   {
+      result = log(pdf(dist, x));
+   }
+   else
+   {
+      result = -k*log(theta) + (k-1)*log(x) - log(tgamma(k)) - (x/theta);
```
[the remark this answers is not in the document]
jzmaddock: You still have log(tgamma(k)) rather than lgamma(k), this is probably the reason this didn't work with type float (which is to say tgamma would overflow much too easily?).

include/boost/math/distributions/inverse_gamma.hpp
jzmaddock: I'm still seeing log(tgamma(shape)) rather than lgamma(shape) ?

jzmaddock:
Hi Matt, thanks again for this, I just spotted a couple of location where you're still using `log(tgamma(x))` rather than `lgamma(x)` but other than that it all looks good to go to me.

include/boost/math/distributions/gamma.hpp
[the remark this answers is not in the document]
mborland: I'll go through and replace them. I realize now that I was just replacing `tgamma` with `lgamma` and could not figure out why it was failing. Guess I should have re-read the docs sooner.

NAThompson:
@HDembinski: Looks like this is basically finished. Let us know if you hit any bugs in the usage.

mborland:
CI is clean so merging and closing linked issue.

HDembinski:
Thanks for this. It looked fine, but perhaps to can remove some of the now duplicated code between pdf and logpdf. For some distributions (poisson, normal, exponential), the pdf can be computed from the logpdf.

jzmaddock:
>Thanks for this. It looked fine, but perhaps you can remove some of the now duplicated code between pdf and logpdf. For some distributions (poisson, normal, exponential), the pdf can be computed from the logpdf.

Technically yes, but would tend to loose precision due to cancellation error in the logpdf calculation?

NAThompson:
> Technically yes, but would tend to loose precision due to cancellation error in the logpdf calculation?

Yeah I'd like to see an ulps plot before going for that . . . also `log` and `exp` are not particularly cheap.

HDembinski:
Fair enough, but you can do it at least for the exponential distribution, no? In case of the normal, you may be right. In case of Poisson, I don't know.

NAThompson:
> Fair enough, but you can do it at least for the exponential distribution, no? 

For the rate λ, we'd have log(pdf(x)) = log(λ) - λx, and then exp(log(pdf)) = exp(log(λ) - λx). That gives two transcendental function calls over 1 using the code-duplicated λexp(-λx). Losing (or gaining!) precision on the exp(log(λ)) also seems possible, although generally you have half-ulp guarantees with those functions. Maybe I'm overthinking it but I just feel like an ulps plot would help convince me.

Issue #764, opened 2022-02-22
Test TR1 CI Failure


mborland:
The `test_tr1` is now failing with the change to window latest in GHA using MSVC 2022 [(error C1189: #error:  STL1003: Unexpected compiler, expected C++ compiler.)](https://github.com/boostorg/math/runs/5289281025?check_suite_focus=true#step:12:481)

PR #765, opened 2022-02-22
Fix for issue #764

mborland:


mborland:
Only failure if CI package installation so merging. Gets us back to green.

jzmaddock:
Can we not fix the source?  It looks like something is including C++ headers in C mode?

jzmaddock:
Unable to reproduce with msvc-14.3.0, once the huge download I have in progress completes I'll update to 14.3.1 and try again.

jzmaddock:
Reproduced, dealt with upstream in Boost.Config.

Issue #767, opened 2022-02-27
CMake build fails in standalone mode:


NAThompson:
```
~/math/build ❯❯❯ cmake ../ -G Ninja                                                                                                           develop ✭ ◼
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Boost.Math: standalone mode ON
-- Configuring done
-- Generating done
-- Build files have been written to: ~/math/build
~/m/build ❯❯❯ ninja                                                                                                                        develop ✭ ◼
[145/246] Building CXX object test/compile_test/CMakeFiles/boost_math-compile_tests.dir/interpolators_cardinal_trig_incl_test.cpp.o
FAILED: test/compile_test/CMakeFiles/boost_math-compile_tests.dir/interpolators_cardinal_trig_incl_test.cpp.o
/Library/Developer/CommandLineTools/usr/bin/c++ -DBOOST_MATH_STANDALONE=1 -I~/math/include -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -std=gnu++17 -MD -MT test/compile_test/CMakeFiles/boost_math-compile_tests.dir/interpolators_cardinal_trig_incl_test.cpp.o -MF test/compile_test/CMakeFiles/boost_math-compile_tests.dir/interpolators_cardinal_trig_incl_test.cpp.o.d -o test/compile_test/CMakeFiles/boost_math-compile_tests.dir/interpolators_cardinal_trig_incl_test.cpp.o -c ~/math/test/compile_test/interpolators_cardinal_trig_incl_test.cpp
In file included from ~/math/test/compile_test/interpolators_cardinal_trig_incl_test.cpp:9:
In file included from ~/math/include/boost/math/interpolators/cardinal_trigonometric.hpp:9:
~/math/include/boost/math/interpolators/detail/cardinal_trigonometric_detail.hpp:11:10: fatal error: 'fftw3.h' file not found
#include <fftw3.h>
         ^~~~~~~~~
1 error generated.
[156/246] Building CXX object test/compile_test/CMakeFiles/boost_math-compile_tests.dir/interpolators_cubic_spline_incl_test.cpp.o
ninja: build stopped: subcommand failed.
[...]

PR #768, opened 2022-02-28
Fix for #767

mborland:
Adds include guards for the usage of fftw3 and gives a hard error if it is unavailable.

### DIFF
--- a/include/boost/math/interpolators/detail/cardinal_trigonometric_detail.hpp
+++ b/include/boost/math/interpolators/detail/cardinal_trigonometric_detail.hpp
@@ -8,11 +8,18 @@
 #include <cstddef>
 #include <cmath>
 #include <stdexcept>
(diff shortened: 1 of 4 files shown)


mborland:
@NAThompson If this solves your problem it's good to merge. All the failures are on the CIs side.

NAThompson:
@mborland : Thanks!

NAThompson:
@mborland : Forgive my CMake stupidity, but how do you run the tests?

I'm trying:

```
build$ cmake ../ -G Ninja
build$ ninja
build$ ctest -V
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
No tests were found!!!
```

mborland:
It may be improper, but it works:

```
cd .../boost/libs/math
cmake .
make -j11
```

PR #774, opened 2022-03-10
Add further color map example

jzmaddock:


jzmaddock:
@mborland @NAThompson FYI this is just me experimenting with the color map code, there's a new example which plots out some of the 2-arg special functions as color maps, really just to test the code out and see how useful it is.

I ended up with plots like this:

extended_kindlmann_1F1

Which has obviously had the scales etc added afterwards, but looks somewhat useful to add?

mborland:
> ... but looks somewhat useful to add?

I think this is a useful addition. For the next release I can look into converting the examples from using `lodepng` to Boost.GIL or some other solution we can actually bundle.

NAThompson:
@jzmaddock : I'm a bit worried there; is it correct? I thought 1F1 was differentiable in b, but that looks discontinuous . . .

jzmaddock:
>  I'm a bit worried there; is it correct? I thought 1F1 was differentiable in b, but that looks discontinuous . . .

It's singular when `b = -k` for integer k.

> Also, how did you add the scales? That might be a nice feature . . .

It would be nice to do automatically, but I just edited in the GIMP.

Issue #776, opened 2022-03-12
second_order_root_finder (near) infinite loop


TFiFiE:
`halley_iterate` and `schroder_iterate` can get stuck in a (virtually) infinite loop even when given a limited `max_iter`, because in the function:
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L483
the variable `count` might already be zero when either of the following functions is called:
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L612
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L638
where it will wrap around by any of:
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L383
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L402
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L439
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L458
becoming `std::numeric_limits<boost::uintmax_t>::max()` before reaching:
https://github.com/boostorg/math/blob/b22105f141b44016b8257288f55bb874ba636a7f/include/boost/math/tools/roots.hpp#L662

PR #777, opened 2022-03-12
Fix logic error in roots.hpp

jzmaddock:
Fixes https://github.com/boostorg/math/issues/776

### DIFF
--- a/include/boost/math/tools/roots.hpp
+++ b/include/boost/math/tools/roots.hpp
@@ -373,7 +373,7 @@ namespace detail {
    {
       using std::fabs;
       if(count < 2)
-         return guess0 - (max + min) / 2; // Not enough counts left to do anything!!
+         return guess - (max + min) / 2; // Not enough counts left to do anything!!
       //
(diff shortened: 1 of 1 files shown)


TFiFiE:
This fix is itself somewhat problematic insofar that one can no longer reliably test whether the algorithm hit the iteration limit just by checking if the function changed `max_iter`. It can now also have been decremented by one.

jzmaddock:
>This fix is itself somewhat problematic insofar that one can no longer reliably test whether the algorithm hit the iteration limit just by checking if the function changed max_iter. It can now also have been decremented by one.

I don't follow - the main loop will still decrement max_iter, the change is that now, if there are insufficient iterations left, then we just don't try the alternate bracketing algorithm.

TFiFiE:
Before, `max_iter` didn't get decremented if it ended up being responsible for stopping the algorithm, because `count` was guaranteed to be decremented at least once in `bracket_root_towards_min` and `bracket_root_towards_max`. Although this change might technically be correct, it is now no longer the case that `max_iter` being changed necessarily implies that it had no hand in stopping the algorithm.

PR #782, opened 2022-04-18
Fix for issue 773

mborland:


mborland:
@NAThompson do you concur with this handling of NaNs? I thought this would be much lighter weigh than throwing an exception.

NAThompson:
@mborland : In computer graphics, nans always render black. So this is the correct behavior.

Why doesn't the std::isnan from std work?

mborland:
> Why doesn't the std::isnan from std work?

Isn't the behavior or `boost::math::isnan` more stable across compilers/optimizations? Passing a NaN to `std::clamp` should be UB since `clamp` requires the less than comparable concept. I was trying to avoid this with configurations that replace `std::isnan` with `false`.

NAThompson:
>  I was trying to avoid this with configurations that replace std::isnan with false.

Oh yeah that's just awful. . .

PR #783, opened 2022-05-02
Experiment with C++20 modules.

jzmaddock:
OK folks, here's a bit of an experiment with C++ modules, it works with msvc but gcc leads to a fairly hopeless spiral of internal compiler errors :(

There are 4 modules:

core.
constants
special_functions

and then an overarching "everything" module which just import/export's the others.

I haven't tried with clang - mostly because it's command line use is problematic for modules and doesn't integrate with the build system well.

msvc/gcc needed no modification to the build system to "just work", except that, module dependencies need to be explicitly specified to ensure there is a "builds before" relationship between module definition and user.  There's some considerable hot air expended on this subject on the web as it breaks parallel builds, and so may actually make the build time longer if you're not careful.

Anyhow, even though this is a somewhat "failed experiment" at present, I'm submitting as a draft PR for discussion, and we'll see what happens with the next gcc release.


NAThompson:
@jzmaddock : How do you build it?

jzmaddock:
> How do you build it?

With gcc/msvc the supplied Jamfile will build things, but the special_functions module will fail to build with gcc (core and constants work OK though).

Or in msvc IDE you can create a static lib project for the module files and set "build as module interface" under advanced settings.

clang, I haven't figured out the command line yet.

jzmaddock:
Update: I have managed to build the "core" module with clang-14, but not use it, I get:

```
In module 'boost.math.core' imported from ../../module/constants.cxx:28:
D:/compilers/msys64/mingw64/include/c++/11.3.0/string_view:146:2: error: 'std::basic_string_view::basic_string_view<_CharT, _Traits>' from module 'boost.math.core.<global>' is not present in definition of 'basic_string_view<_CharT, _Traits>' provided earlier
        basic_string_view(_It __first, _End __last)
        ^
D:/compilers/msys64/mingw64/include/c++/11.3.0/string_view:124:7: note: declaration of 'basic_string_view<_CharT, _Traits>' does not match
      basic_string_view() noexcept
      ^
D:/compilers/msys64/mingw64/include/c++/11.3.0/string_view:128:17: note: declaration of 'basic_string_view<_CharT, _Traits>' does not match
      constexpr basic_string_view(const basic_string_view&) noexcept = default;
                ^
D:/compilers/msys64/mingw64/include/c++/11.3.0/string_view:131:7: note: declaration of 'basic_string_view<_CharT, _Traits>' does not match
      basic_string_view(const _CharT* __str) noexcept
      ^
D:/compilers/msys64/mingw64/include/c++/11.3.0/string_view:137:7: note: declaration of 'basic_string_view<_CharT, _Traits>' does not match
      basic_string_view(const _CharT* __str, size_type __len) noexcept
      ^
D:/compilers/msys64/mingw64/include/c++/11.3.0/string_view:146:2: note: declaration of 'basic_string_view<_CharT, _Traits>' does not match
        basic_string_view(_It __first, _End __last)
```

I see other folks around the web see similar errors, but with no solutions thus far :(

NAThompson:
@jzmaddock : Has the module situation improved over the past year? I'm wondering if this PR might work now . .

jzmaddock:
Just tried it with GCC-12 and still see a slew of internal compiler errors.

NAThompson:
We'll try again in 2024!

mborland:
Looks like we might finally be in business: https://www.kitware.com/import-cmake-the-experiment-is-over/

mborland:
I pushed a couple of small updates to get this compiling under GCC-13. With C++20 and 23 you can now import std; so that version switch is in as well. I need to try under Clang-18 which came out last week and I think GCC-14 will be out in the next few weeks.

jzmaddock:
This *almost* works with gcc-13 and clang-18.

I'm going to try merging develop to this next, as I know that touches some of the problem areas anyway...

jzmaddock:
This now works (such as it is) with MSVC-17.9.5 and clang-18.  GCC-13 still ICE's.

mborland:
> This now works (such as it is) with MSVC-17.9.5 and clang-18. GCC-13 still ICE's.

Should we change the file extension of the module to `.ixx` (MSVC) or `.cppm` (other tutorials)? It looks like the current `.cxx` collides with one of the compilers are used so the module is not committed. We could then add `.gcm` (GCC), `.ifc` (MSVC), and `.cxx` (Presumably clang) to the gitignore.

jzmaddock:
This sucks, as different compilers require different naming conventions :(

I'm easy either way though.

mborland:
It looks like Dani went with `.ixx` in her demo she mentioned on the ML: https://github.com/DanielaE/CppInAction. Since that's proven to work it seems like the safe choice.

mborland:
I'm working the full scale library addition of modules today, but going to put it in a different PR. Modules seem to be working in other much smaller boost libs so those serve as good reference for this work. I will be heavily leaning on Claude Fable Max since this will be mindless addition of macros to every public function in the library.

Issue #785, opened 2022-05-12
Boots math tools nothrow exception


ckormanyos:
When I am putting math on systems lacking exception support, I keep encountering this line [here](https://github.com/boostorg/math/blob/4d5cc6972ea5576d0c7fbb7caee5a70512dea1b5/include/boost/math/tools/throw_exception.hpp#L18).

The conditions I would like are:
- Have `BOOST_MATH_STANDALONE`
- And Have `BOOST_NO_EXCEPTIONS`

But for some reason, the line mentioned above keeps wanting to throw something.

Is there any way (or is this even the right place) to offer an option there for the presence (or lack thereof) of `BOOST_NO_EXCEPTIONS` such that nothing is ever actually `throw`n?

jzmaddock:
I think perhaps we've goofed there.

Is this something that can be addressed by the policy support - ie by specifying a policy that will never throw anyway?

ckormanyos:
> Is this something that can be addressed by the policy support

In my investigations with a lot of bare metal attempts, I have thought that you could either:
- Cut the preprocessor macro at the end of the chain in [`tools/throw_exception.hpp`](https://github.com/boostorg/math/blob/4d5cc6972ea5576d0c7fbb7caee5a70512dea1b5/include/boost/math/tools/throw_exception.hpp#L18)
- Specializations in [`policies/policy.hpp`](https://github.com/boostorg/math/blob/4d5cc6972ea5576d0c7fbb7caee5a70512dea1b5/include/boost/math/policies/policy.hpp)
- Refactor [`policies/error_handling.hpp`](https://github.com/boostorg/math/blob/4d5cc6972ea5576d0c7fbb7caee5a70512dea1b5/include/boost/math/policies/error_handling.hpp)

Somehow, I'd like to do a combination of the second and third, because there are two things i do not like about `policies/error_handling.hpp`. Firstly it include< `<sstream>` which is very heave and I would like to optionally have that gone on the metal. Secondly, I'm not sure if the support for _no_ _RTTI_ is completely consistent throughout that file.

I do not, however, have the expertize needed to decide on the right scope of changes, nor the necessary skills to handle the scope of these changes alone.

But these are, in fact, the two main areas (other than the plethora of warnings, which I will address in separate threads) that hamper lightweight, bare-metal progress of Math.

jzmaddock:
>
>
>     Is this something that can be addressed by the policy support
>
> In my investigations with a lot of bare metal attempts, I have thought that you could either:
[...]

PR #786, opened 2022-05-12
Set BOOST_NO_EXCEPTIONS and BOOST_NO_RTTI for GCC and math in standal…

jzmaddock:
…one mode.

Disable everything not needed in error_handling.hpp when no exceptions are available.


ckormanyos:
Hi John (@jzmaddock) this does not entirely eliminate the problem(s).

I posted some sample code in the original post #785. If I play around, I can probably get some other tiny errors. But at the moment, i think Bernoulli might be throwing and is not caught in the expected mechanism.

ckormanyos:
Ahhh... Last but not least, this guy throws [here](https://github.com/boostorg/math/blob/4d5cc6972ea5576d0c7fbb7caee5a70512dea1b5/include/boost/math/special_functions/detail/bernoulli_details.hpp#L191).

Your fix is _almost_ complete for the moment. i will check for some other throwing activities in the general code base.

ckormanyos:
It seems like we have a couple of rogue throw potentials --- also a handful in my own creation of cstdfloat.

How should we deal with these?

grafik

jzmaddock:
I think I can see how to fix up bernoulli...

ckormanyos:
> fix up bernoulli

Yes. Many thanks John. Commit a6db2d19b7e96f802af751b039090d184343b3b5 is confirmed to fix my test case.

include/boost/math/special_functions/detail/bernoulli_details.hpp
```diff
@@ -184,17 +184,22 @@ struct fixed_vector : private std::allocator<T>
    const T& operator[](unsigned n)const { BOOST_MATH_ASSERT(n < m_used); return m_data[n]; }
    unsigned size()const { return m_used; }
    unsigned size() { return m_used; }
-   void resize(unsigned n, const T& val)
+   bool resize(unsigned n, const T& val)
    {
       if(n > m_capacity)
       {
+#ifndef BOOST_NO_EXCEPTIONS
```
ckormanyos, line 191: I'm not an expert, but just curious here... Why is this unexpected event handled differently than the ones below that go through _policies_? Is there no such fitting policy for this action?

ckormanyos:
For the next trimesters or so, this is more than I had hoped for, upon running green...

Thank you  John!

ckormanyos:
Hi John (@jzmaddock), based on [this minute compiler phenomenon](https://github.com/ckormanyos/wide-decimal/runs/6433159816?check_suite_focus=true#step:5:26) in my CI on an unrelated project on this branch, I took the liberty to add a tiny change.

Please give me your feedback on this at your convenience.

ckormanyos:
On another thought, and I'm not sure if I'd like to push for this right at this moment, but _yet_ _another_ CI job _could_ ensure that spec-fun runs with `-fno-exceptions` and `-fno-rtti` on at least one of each modern GCC/clang.

I think spec-fun is where some of the exception-free clients will see some benefit, as these folks might (unknowingly) be on the metal or could eventually use such a feature to gain savings.

jzmaddock:
>On another thought, and I'm not sure if I'd like to push for this right at this moment, but yet another CI job could ensure that spec-fun runs with -fno-exceptions and -fno-rtti on at least one of each modern GCC/clang.

That was my thought too, unfortunately, running the tests locally shows that Boost.Test doesn't support -fno-exceptions so *everything* fails to build :(

ckormanyos:
> Boost.Test doesn't support `-fno-exceptions` so everything fails to build

Thanks for looking into this John!

Ah yes, now that you mention it, that makes perfect sense.

I'm totally happy with the level reached so far and think if it's green, then it's good to go.

I consider us to be on a pathway toward reduced dependencies and flexibility. We might find a few things to improve (such in this PR here). Then we improve them. Maybe later, we find a few more. Maybe further down the road, we spend some time to build up a light-weight test environment, maybe not.

To re-iterate, I am really happy with thisw PR in its current state. I already ran my CI (doing a bunch of `tgamma`-like stuff) in another challenging, unrelated project on this branch. I ran with `-fno-exceptions` and `-fno-rtti` on GCC 4 through 11 and a bunch of clang versions and everything worked perfectly!

ckormanyos:
Wow! Even more awesomeness. Thanks John!

ckormanyos:
Hi John. I think one try-catch clause was missed in spec-fun no_eh in places like [here](https://github.com/boostorg/math/runs/6442199194?check_suite_focus=true#step:18:335).

The failures on Jammy look suspiciously like timeouts to me, although I'm not exactly sure what error code 143 means.

ckormanyos:
The problem in spec fun on no-eh is [here](https://github.com/boostorg/math/blob/4d5cc6972ea5576d0c7fbb7caee5a70512dea1b5/include/boost/math/special_functions/chebyshev_transform.hpp#L130), I believe.

I don't know what you'll find out on the failures on GCC. I have some very preliminary reports (from other projects) on what I might consider to be spurious behaviors in GCC12 on built-in memory operations, but that's super preliminary and not verified.

jzmaddock:
Github is timing out whenever I try to access the logs at present, I'll take a look later...

ckormanyos:
> Github is timing out whenever I try to access the logs at present, I'll take a look later

Yes. The pink unicorn is bucking and rearing its head

jzmaddock:
Still can't access the logs, so pushed a couple of speculative commits based on what I did manage to load!

ckormanyos:
Well, also this morning I'm still getting the pink Unicorn on log viewing.

ckormanyos:
Hi John, wonderful progress on this.

I reeally think this one is getting close. On this particular run, however, we get a hitch on no-eh [here](https://github.com/boostorg/math/runs/6572342065?check_suite_focus=true#step:18:839).

jzmaddock:
> I reeally think this one is getting close. On this particular run, however, we get a hitch on no-eh [here](https://github.com/boostorg/math/runs/6572342065?check_suite_focus=true#step:18:839).

Grrr, forgot to check those locally... at least this time we can actually see the logs!  Old run cancelled and fix pushed.

ckormanyos:
This is cool.

It seems _almost_ green. There does, however, seem to be an actual failure in a test case on Chatterjee in Drone on Ubuntu clang++-10 with language standard `-std=gnu++17`. It might be a tolerance, or the problem might be not yet known.

jzmaddock:
> It seems almost green. There does, however, seem to be an actual failure in a test case on Chatterjee in Drone on Ubuntu clang++-10 with language standard -std=gnu++17. It might be a tolerance, or the problem might be not yet known.

It's a tolerance thing: tolerance was 0.01 and the error found was 0.0103.  I suspect @NAThompson and @mborland missed that one because we have one consistently failing test (an autodiff sanitizer test runs the machine out of memory I suspect), I've disabled that one, and also removed a few redundant Github tests - it doesn't cut down the testing load by much to be honest - I can't decide whether we should cull a few more or not - it feels like we should, but I worry about missing something...!

ckormanyos:
> It's a tolerance thing: tolerance was 0.01 and the error found was 0.0103. I suspect @NAThompson and @mborland missed that one because we have one consistently failing test (an autodiff sanitizer test runs the machine out of memory I suspect), I've disabled that one, and also removed a few redundant Github tests - it doesn't cut down the testing load by much to be honest 

Nice!

> - I can't decide whether we should cull a few more or not - it feels like we should, but I worry about missing something...!

Matt (@mborland) recently expressed interest in this topic.

One thing on GHA is that you can now selectively re-run either failed tests only or all tests entirely. I've used this selective re-run feature on Multiprecision successfully to go green. It does not cut down any testing time. Also the fail we mention here wasn't on GHA but Drone. Anyway, I thought I'd mention that if it was not known.

Thank you very much John (@jzmaddock) for driving onward in this PR!

mborland:
> It's a tolerance thing: tolerance was 0.01 and the error found was 0.0103. I suspect @NAThompson and @mborland missed that one 

It is safe to bump up the tolerance. It's shooting for correlation on random numbers compared to an approximate published value. 

> I can't decide whether we should cull a few more or not - it feels like we should, but I worry about missing something...!

Can we offload more testing to drone? It seems like those runs generally finish in an hour right now as opposed to the 6+ it takes GHA.

jzmaddock:
@mborland I'll look into shifting some load onto drone later.

ckormanyos:
Thank you John (@jzmaddock) for sticking with this mammoth PR. It entailed many remarkably detailed and difficult nuances.

I've already checked this branch this all out on the metal and in a bunch of unrelated CI tasks on other projects that dis-allow exception handling.

If anything else arises, we can handle it in the next rimester. But this looks really great, light and clean for no-exception usage!

PR #788, opened 2022-05-23
Bump minimum language standard to C++14

mborland:


mborland:
@NAThompson and @ckormanyos I think this now matches your feedback. I will make a matching PR in multiprecision.

PR #789, opened 2022-05-27
Constexpr next

mborland:
Implements constexpr: `nextafter`, `nextafterf`, `nextafterl`, `nexttoward`, `nexttowardf`, and `nexttowardl`. Once this and #734 are complete [all functions pending standardization ](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0533r9.pdf)will be constexpr.


mborland:
@NAThompson Can you review this? It is almost a 1:1 copy of the existing nextafter implementation with standard library cmath functions being replaced by our ccmath functions. The results match exactly which makes me feel good about the other functions.

test/ccmath_next_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    // test_1 has already passed through static_asserts so we know it was calculated at compile time
+    // rather than farming out to std at run time.
+    const T existing_test_1 = boost::math::nextafter(T(1), T(1.5));
+    CHECK_EQUAL(test_1, existing_test_1);
+}
+
+int main(void)
+{
+    test_next<float>();
+    test_next<double>();
```
NAThompson, line 49: @mborland : Is there a way to test for float16?

NAThompson:
@mborland : Looks good; make sure to squash merge with a nice commit message.

test/ccmath_next_test.cpp
mborland: I don't see any macros in cstdfloat that allow a portable way of checking that `float16_t` is available. I believe that fixed width floating points are being added in C++23 with the reference implementation coming from Boost.Math.

include/boost/math/ccmath/next.hpp
```diff
@@ -0,0 +1,456 @@
...
+        diff = detail::get_smallest_value<T>();
+    }
+    return val + diff;
+}
+
+//
+// Special version for some base other than 2:
+//
+template <typename T>
+constexpr T float_next_imp(const T& val, const std::false_type&)
```
jzmaddock, line 199: Do we actually need to support bases other than 2 in the constexpr case?  I would be inclined to remove this and static_assert that the radix is 2.

PR #790, opened 2022-05-31
Move all GCC-5/6 testing to drone.

jzmaddock:
Remove C++11 testing for gcc where the compiler default is C++14.
Remove C++11 testing from msvc - it doesn't support it, only C++14.
All C++11 testing is now on gcc-5 and 6 and on drone.


jzmaddock:
One failure is unrelated to us (in Boost.Hash, fix in progress) so merging.

@mborland : this moves about one half of the testing load to drone, so we'll see if this speeds up the CI pipeline a bit.

PR #791, opened 2022-06-05
Try and consolidate more Github tests.

jzmaddock:
In the hopes of speeding up CI build times.


jzmaddock:
@mborland @NAThompson , this changes consolidates our CI - we still have the same breadth of testing but on many fewer jobs.  Should cycle much faster now I hope!

PR #792, opened 2022-06-14
Replace non-UTF-8 copyright character

jwakely:
Similar to https://github.com/boostorg/boost/pull/618 and https://github.com/boostorg/boost/pull/618 and https://github.com/boostorg/multi_index/pull/55 and https://github.com/boostorg/detail/pull/22


jzmaddock:
Thanks!

PR #793, opened 2022-07-01
constexpr signbit

mborland:


mborland:
@NAThompson This is good for review. I deviate slightly from the C++ standard (same as copysign) in that using UB to find the sign of a `NAN` is not allowable in constexpr contexts.

jzmaddock:
That all looks fine to me, I wonder how useful it will be if we can't detect signed zero though?

Hmm... just a thought, we can detect signed infinities right?  So how about checking the sign of `1/x` when x is zero?  Might need to verify that the type has infinities first, and I worry it might trigger compiler traps, but would make the facility more generally useful.

mborland:
> That all looks fine to me, I wonder how useful it will be if we can't detect signed zero though?
> 
> Hmm... just a thought, we can detect signed infinities right? So how about checking the sign of `1/x` when x is zero? Might need to verify that the type has infinities first, and I worry it might trigger compiler traps, but would make the facility more generally useful.

We can detect signed infinities, but division by 0 is not allowed in constexpr contexts because it is UB.

jzmaddock:
>We can detect signed infinities, but division by 0 is not allowed in constexpr contexts because it is UB.

Oh :(  I knew there had to be a catch somewhere.  Oh well, never mind, carry on as you were ;)

PR #794, opened 2022-07-02
Fix unsafe bool usage

mborland:
Replace unsafe bool usage in type promotion with `boost::math::tools::promote_args_t`. Additionally, refactor sonarlint warnings.

### DIFF
--- a/include/boost/math/ccmath/fmod.hpp
+++ b/include/boost/math/ccmath/fmod.hpp
@@ -1,4 +1,4 @@
-//  (C) Copyright Matt Borland 2021.
+//  (C) Copyright Matt Borland 2021 - 2022.
 //  Use, modification and distribution are subject to the
 //  Boost Software License, Version 1.0. (See accompanying file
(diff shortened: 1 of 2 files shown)


jzmaddock:
Question: do we want to remove the inline and noexcept qualifiers?  Is a likely use case that this may be used in code which is maybe or maybe not constexpr?  If not then maybe we should perhaps devise a static_assert that will fail when instantiated in a non-constexpr context, and/or make the qualifiers `consteval` in C++20 and later?

mborland:
The `inline` qualifier was removed because [it is redundant](https://rules.sonarsource.com/cpp/RSPEC-5408). The `noexcept` qualifier was removed because at compile time the function is `noexcept`, but at runtime the STL does not require it. This matches the now [accepted C++23 proposal](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0533r8.pdf) for making some of cmath constexpr.

PR #795, opened 2022-07-03
Cleanup Warnings

mborland:
Large scale cleanup of warnings by running `-Wall -Wextra` through the complete test set. Fixed sonarlint complaints while in the file cleaning up warnings.


mborland:
@jzmaddock This is good for review. The diff stat is misleading in that almost all of it is from whitespace changes in `octonion`

PR #796, opened 2022-07-04
More warning fixes

mborland:
Primarily unreachable code, and changing implicit conversions from floating point to int and vice versa to be explicit. As always addressed viable sonarlint comments while in the file.


ckormanyos:
Wow! Amazing clarity improvements, Matt (@mborland). You move so rapidly.

jzmaddock:
Looks good to me... merging.  Thanks!

PR #801, opened 2022-07-15
Corrected paths for cubic and quartic roots headers in doc

RL-S:
The current documentation lists `cubic_roots.hpp` and `quartic_roots.hpp` as located in `boost/math/`**`roots/`**, however, they are now both located in `boost/math/`**`tools`**.

It appears that these two doc files are the only ones affected. Checked with `grep -rn "roots/" . | grep -v "doc/roots"` in the `libs/math/doc/roots` directory.


jzmaddock:
Yup, that looks right to me, no need for a CI run on a small doc change so merging...

PR #807, opened 2022-07-26
Don't inline is_value_non_zero () to be evaluated at runtime.

dlaugt:
I'm compiling a program using boost math (1.79.0) with gcc 7.3.1. When I run the program with Valgrind, I have an overflow exception at startup. The following patch fixes the issue.


jzmaddock:
Ugh, that's the correct fix, but we need to tweak it to maintain compatibility with standalone mode.

PR #818, opened 2022-08-15
Von Mises Distribution

mborland:
#330 looked very close to completion so this is my attempt to foster it.

### DIFF
--- a/include/boost/math/distributions/detail/common_error_handling.hpp
+++ b/include/boost/math/distributions/detail/common_error_handling.hpp
@@ -11,6 +11,8 @@
 
 #include <boost/math/policies/error_handling.hpp>
 #include <boost/math/special_functions/fpclassify.hpp>
+#include <boost/math/constants/constants.hpp>
+
 // using boost::math::isfinite;
 // using boost::math::isnan;
 
@@ -91,6 +93,26 @@ inline bool check_location(
       *result = policies::raise_domain_error<RealType>(
          function,
          "Location parameter is %1%, but must be finite!", location, pol);
+      return false;
+   }
+   return true;
+}
+
+
+template <class RealType, class Policy>
+inline bool check_angle(
+      const char* function,
+      RealType angle,
+      RealType* result,
+      const Policy& pol)
+{
+   if(!(boost::math::isfinite)(angle)
+       || angle < -boost::math::constants::pi<RealType>() 
+       || angle > +boost::math::constants::pi<RealType>())
+   {
+      *result = policies::raise_domain_error<RealType>(
+         function,
+         "Angle parameter is %1%, but must be between -pi and +pi!", angle, pol);
       return false;
    }
    return true;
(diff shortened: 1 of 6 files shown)


mborland:
@NAThompson This should be good for review. The only major outstanding change from the last cut at this was removing all the double precision magic numbers which I have. I would have just made the edits on the linked PR, but since the guy deleted his account I can't access the branch.

doc/distributions/von_mises.qbk
```diff
@@ -0,0 +1,102 @@
...
+    }; // class von_mises_distribution
+   
+   }} // namespaces
+   
+The [@https://en.wikipedia.org/wiki/Von_Mises_distribution von Mises distribution], also known as a the circular normal distribution,
+is a continuous probability distribution on a circle.
+
+[h4 Member Functions]
+
+   von_mises_distribution(RealType mean = 0, RealType concentration = 1); ;
```
NAThompson, line 43: @jzmaddock : IIRC we now use `Real` instead of `RealType`? Obviously a nitpick.

doc/distributions/von_mises.qbk
```diff
@@ -0,0 +1,102 @@
...
+    RealType scale() const;
+
+Returns the /concentration/ parameter as the location is synonymous but used in some definitions.
+
+[h4 Non-member Accessors]
+
+All the [link math_toolkit.dist_ref.nmp usual non-member accessor functions]
+that are generic to all distributions are supported: __usual_accessors.
+
+The domain of the random variable is O <= x <= 2[pi].
```
NAThompson, line 72: Maybe use “≤” (U+2264) which I think is \u2264 in .qbk?

include/boost/math/distributions/detail/common_error_handling.hpp
```diff
@@ -96,6 +98,26 @@ inline bool check_location(
...
+      RealType* result,
+      const Policy& pol)
+{
+   if(!(boost::math::isfinite)(angle)
+       || angle < -boost::math::constants::pi<RealType>() 
+       || angle > +boost::math::constants::pi<RealType>())
+   {
+      *result = policies::raise_domain_error<RealType>(
+         function,
+         "Angle parameter is %1%, but must be between -pi and +pi!", angle, pol);
```
NAThompson, line 115: If the CDF is periodic, is it sensible to restrict the angle like this?

test/test_von_mises.cpp
```diff
@@ -0,0 +1,393 @@
...
+  // Basic sanity-check spot values.
+  // (Parameter value, arbitrarily zero, only communicates the floating point type).
+  test_spots(0.0F);   // Test float. OK at decdigits = 0 tolerance = 0.0001 %
+  test_spots(0.0);    // Test double. OK at decdigits 7, tolerance = 1e07 %
+  test_spots(0.0L); // Test long double.
+
+  // Check symmetry of PDF and CDF
+  test_symmetry(0.0F);
+  test_symmetry(0.0);
+  test_symmetry(0.0L);
```
NAThompson, line 383: Might we test the quadrature of the pdf = 1?

NAThompson:
@mborland : I've approved, but with the caveat that I don't know much about this. We'll probably want a second review from @jzmaddock .

AtariDreams:
Any update on this?

PR #819, opened 2022-08-19
Improve ccmath signbit and copysign

mborland:
Use `std::bit_cast` or `__builtin_bit_cast` when available to support signed zeros and NANs.

### DIFF
--- a/include/boost/math/ccmath/signbit.hpp
+++ b/include/boost/math/ccmath/signbit.hpp
@@ -7,6 +7,7 @@
 #define BOOST_MATH_CCMATH_SIGNBIT_HPP
 
 #include <cmath>
+#include <cstdint>
 #include <limits>
 #include <type_traits>
 #include <boost/math/tools/is_constant_evaluated.hpp>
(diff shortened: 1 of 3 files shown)


mborland:
@jzmaddock and @NAThompson this is good for review. It conditionally solves the issue of not supporting 0s and NANs brought up on the mailing list.

include/boost/math/ccmath/signbit.hpp
```diff
@@ -7,19 +7,173 @@
...
 #include <cmath>
+#include <cstdint>
 #include <limits>
 #include <type_traits>
 #include <boost/math/tools/is_constant_evaluated.hpp>
+#include <boost/math/special_functions/detail/fp_traits.hpp>
 #include <boost/math/ccmath/isnan.hpp>
+#include <boost/math/ccmath/abs.hpp>
+
+#if __cpp_lib_bit_cast >= 201806L
```
jzmaddock, line 18: I think we need to include <version> before checking for this macro (and use __has_include to check for <version>).

include/boost/math/ccmath/signbit.hpp
```diff
@@ -7,19 +7,173 @@
...
+                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+../../../boost/math/ccmath/signbit.hpp:62:24: note: constexpr bit_cast involving bit-field is not yet supported
+        const auto u = BOOST_MATH_BIT_CAST(float_bits, arg);
+                       ^
+../../../boost/math/ccmath/signbit.hpp:20:37: note: expanded from macro 'BOOST_MATH_BIT_CAST'
+#  define BOOST_MATH_BIT_CAST(T, x) __builtin_bit_cast(T, x)
+                                    ^
+*/
+
+#if defined(__clang__) && defined(BOOST_MATH_BIT_CAST)
```
jzmaddock, line 42: Is this all clang's or just Apples?

There is an alternative implementation which avoids bit fields: just store the left shift which puts the sign bit at index 0, then bit_cast to an appropriately sized integer, shift and bitwise & with 1u.

include/boost/math/ccmath/signbit.hpp
```diff
@@ -7,19 +7,173 @@
...
+        {
+            return false;
+        }
+        
+        return arg < static_cast<T>(0);
+    }
+}
+
+#else
+
```
jzmaddock, line 172: If bit_cast is unavailable, would it be better to static_assert if the argument is zero or NaN on the grounds that we just don't know what the result should be?

include/boost/math/ccmath/signbit.hpp
mborland: I just ran clang 12,13, and 14 on Ubuntu. Looks like strictly an Apple problem.

include/boost/math/ccmath/signbit.hpp
mborland: I think that is fair, and also matches what we discussed on the ML.

mborland:
@jzmaddock Would you like another look before I merge? I hit all of your comments in the latest commit.

PR #821, opened 2022-09-04
constexpr sin

mborland:


mborland:
@NAThompson Will you take a look at this? I think within an epsilon for floats and doubles at compile time is reasonable accuracy.

include/boost/math/ccmath/sin.hpp
```diff
@@ -0,0 +1,98 @@
...
+    constexpr auto a6 = static_cast<T>(+1.604729591825977403374012010065495e-10L);
+    constexpr auto a7 = static_cast<T>(-7.364589573262279913270651228486670e-13L);
+    
+    const T x2 = x * x;
+    const T x3 = x2 * x;
+    const T x4 = x2 * x2;
+    const T x8 = x4 * x4;
+    const T x9 = x8 * x;
+
+    const T A = x3 * (a1 + x2 * (a2 + x2 * a3));
```
NAThompson, line 42: Is the fma constexpr? If so, you might be able to do a `std::fma` here and improve the accuracy.

NAThompson:
Did you create an ULPs plot for this? Might be able to find some regions of lower accuracy. . . .

include/boost/math/ccmath/sin.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include <boost/math/ccmath/isnan.hpp>
+
+namespace boost::math::ccmath {
+
+namespace detail {
+
+template <typename T>
+[[nodiscard]] constexpr T sin_impl(T x) noexcept
+{
+    // Precalculated coeffs from the minmax polynomial for up to 128 bit long doubles
```
NAThompson, line 25: Are your long doubles 128 bits?

mborland:
> 

Is there a way to generate them with constexpr functions?

include/boost/math/ccmath/sin.hpp
mborland: No but they have sufficient precision. -mlong-double-128 is not supported on x86_64 because it implicitly calls `__float128`. In order to support that I would have to make the end user link `-lquadmath` anytime these functions are used.

NAThompson:
@mborland : You're worried that the constexpr branch will do something different than the runtime branch?

mborland:
> @mborland : You're worried that the constexpr branch will do something different than the runtime branch?

Yes; I default to the linked standard library for run time implementation.

NAThompson:
Well, you couldn't use our standard `ulps_plot` then, but you could generate a large `std::array` and then compute `sin(array)` at compile-time, and then compute the float distance to the value computed in higher precision.

jzmaddock:
A quick check with:

```
#include <boost/math/ccmath/sin.hpp>
#include <iostream>
#include <iomanip>

int main()
{
   constexpr double v = boost::math::ccmath::sin(3.1277929220348599);
   double x = std::sin(3.1277929220348599);
   std::cout << std::setprecision(18);
   std::cout << v << std::endl;
   std::cout << x << std::endl;
   return 0;
}
```

shows large errors as the input approaches +-PI.

Unfortunately, correct angle reduction is really hard, and requires an arbitrary number of digits for pi.  However, maybe we could at least do better in the range [-PI, +PI] by storing more digits for argument reduction as a pair of values?

mborland:
The accuracy has been improved now with argument reduction. Below is the plot for `boost::math::ccmath::sin - std::sin` with the x axis being fractions of pi and y axis being form -1 epsilon to 1 epsilon for double precision types. The average difference between the two is: -0.0574961 epsilon.

sin_plot

NAThompson:
LGTM. Might want to include that ULPS plot in the docs just so no one is surprised.

@jzmaddock : You agree?

mborland:
Better versions of this exist in the decimal repo. Need to port those over at some point.

PR #822, opened 2022-09-17
Modernize exception specification

apolukhin:
Done via `find include src/ example/ meta/ test/ tools/ include_private/ config/ -type f | xargs sed -i 's/throw()/BOOST_MATH_NOTHROW/g'` and by adding the include/boost/math/tools/nothrow.hpp


mborland:
Looks good to me. Sorry for the delay.

PR #831, opened 2022-10-04
Fix Clang-15 warnings

AtariDreams:
Redundant casts, noexcept for move constructors, and usage of default and delete make the code more performant and maintainable

### DIFF
--- a/include/boost/math/tools/polynomial_gcd.hpp
+++ b/include/boost/math/tools/polynomial_gcd.hpp
@@ -23,7 +23,6 @@
 #include <numeric>
 #include <utility>
 #include <iterator>
-#include <type_traits>
 #include <boost/math/tools/assert.hpp>
 #include <boost/math/tools/config.hpp>
 
--- a/include/boost/math/concepts/std_real_concept.hpp
+++ b/include/boost/math/concepts/std_real_concept.hpp
@@ -340,7 +340,6 @@ inline std::basic_istream<charT, traits>& operator>>(std::basic_istream<charT, t
 } // namespace concepts
 }}
 
-#include <boost/math/tools/precision.hpp>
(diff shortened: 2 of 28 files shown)


ckormanyos:
Hi @AtariDreams I took a quick glance at the drone run in CI (which runs even without admin approval).

To my rather quick glance it seems like a lot of tests fail with these changes. I'm not sure exactly what went wrong, but this should be investigated prior to running all of CI.

Cc: @jzmaddock and @NAThompson and @mborland

AtariDreams:
Fixed!

ckormanyos:
Hi @AtariDreams thanks for correcting these suggested corrections.

Could we wait a few hours or half a day or so before approving CI? John (@jzmaddock) has been working steadily and dilligently on some special functions correctins and these are cacling now.

I (or somebody) will approve this run after some of heavy action finishes up, which should be in a few hours.

ckormanyos:
To be more clear @AtariDreams to me it would seem nice to get #835 (cycling in CI now) merged (see work by @jzmaddock). Then merge that updated develop into your branch and then run all of CI.

Thoughts?

AtariDreams:
That makes sense. Perfect!

include/boost/math/tools/polynomial_gcd.hpp
```diff
@@ -23,7 +23,6 @@
 #include <numeric>
 #include <utility>
 #include <iterator>
-#include <type_traits>
```
jzmaddock, line 26: This header use std::is_arithmetic: surely we should keep this #include?

jzmaddock: Ah no I see, yes you're quite right!

PR #838, opened 2022-10-08
Fix quartic roots when depressed cubic only has single real root

NAThompson:


NAThompson:
Error is unrelated; merging.

PR #841, opened 2022-10-14
Remove warning disable warning 4800

AtariDreams:
We should do a comparison with 0 instead of forcing to bool


mborland:
What does this change if the warnings are already suppressed? How did you generate/find these?

AtariDreams:
I found these while browsing the code. It is best to get a bool directly via comparison than a conversion. It is considered to be better C++

mborland:
Without a specific bug/error/warning I am going to close this. Generally we are looking for compelling reasons to change code because of our large user base. If you are looking for ways to contribute there is a feature wishlist  https://github.com/boostorg/math/issues/303, or scan through our other open issues.

PR #842, opened 2022-10-14
Fix compiler format warnings

AtariDreams:


include/boost/math/special_functions/legendre.hpp
```diff
@@ -61,13 +61,10 @@ T legendre_imp(unsigned l, T x, const Policy& pol, bool second = false)
    if(l == 0)
       return p0;
 
-   unsigned n = 1;
-
-   while(n < l)
+   for (unsigned n = 1; n < l; ++n)
```
mborland, line 64: What warning is this going to fix?

AtariDreams: This looks better.

NAThompson: I confess to some discomfort with this reasoning. It seems like a very easy way to introduce regressions, moreover it could lead to a veritable infinitude of pull requests where we all make minor changes to each other's code based on aesthetic criteria which are not globally shared.

PR #844, opened 2022-10-15
Fix warnings in quadrature tests

mborland:
Removes all Boost.Math warnings from "-Wall -Wextra -Wpedantic" on Apple Clang 14.
- Conditionally unused variable
- Fix operator precedence with parentheses


NAThompson:
Yup, all looks good to me!

PR #846, opened 2022-10-16
Fix for issue 845

mborland:
Avoid division by zero in special cases

xref: https://github.com/scipy/scipy/issues/17101

### DIFF
--- a/include/boost/math/distributions/non_central_beta.hpp
+++ b/include/boost/math/distributions/non_central_beta.hpp
@@ -86,7 +86,12 @@ namespace boost
                }
                pois *= i / l2;
                beta += xterm;
-               xterm *= (a + i - 1) / (x * (a + b + i - 2));
+
+               if (a + b + i != 2)
+               {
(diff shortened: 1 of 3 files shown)


test/git_issue_845.cpp
```diff
@@ -0,0 +1,59 @@
+// Copyright Matt Boralnd, 2022
```
WarrenWeckesser, line 1: I see I'm not the only one who sometimes mistypes their own last name. 😄

include/boost/math/distributions/non_central_beta.hpp
```diff
@@ -86,7 +86,12 @@ namespace boost
                }
                pois *= i / l2;
                beta += xterm;
-               xterm *= (a + i - 1) / (x * (a + b + i - 2));
+
+               if (a + b + i != 2)
```
jzmaddock, line 90: Rather than putting an `if` inside the inner loop, it would be better to move calculation of xterm to the start of the loop: albeit it's inconvenient, and we have to be careful to get the algebra right ;)

Again this should wait till after the next release.

PR #847, opened 2022-10-16
Fix for scipy issue 17146

mborland:
xref: https://github.com/scipy/scipy/issues/17146

### DIFF
--- a/include/boost/math/distributions/binomial.hpp
+++ b/include/boost/math/distributions/binomial.hpp
@@ -483,11 +483,6 @@ namespace boost
         {
           return 1; // Probability = 1 = certainty.
         }
-        if (k == 0)
-        { // binomial coeffic (n 0) = 1,
-          // n ^ 0 = 1
-          return pow(1 - dist.success_fraction(), n);
-        }
         if (k == n)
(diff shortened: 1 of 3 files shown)


include/boost/math/distributions/binomial.hpp
```diff
@@ -483,11 +483,6 @@ namespace boost
         {
           return 1; // Probability = 1 = certainty.
         }
-        if (k == 0)
-        { // binomial coeffic (n 0) = 1,
-          // n ^ 0 = 1
-          return pow(1 - dist.success_fraction(), n);
```
jzmaddock, line 489: Thanks for taking this on Matt, and with apologies for being AWOL: I'm just catching up with everything while writing the release notes for 1.81.

Since ibeta_derivative is relatively heavyweight, and the alternative formula is so simple, I think it would be nice to have retained the alternative form here.  The trick would be to take logs and use `exp(n*log1p(-dist.success_fraction()))` in the case where `- dist.success_fraction()` is small.  Maybe try 0.125 as a cutoff?  Might need some experimenting to see where the errors start to rise.

This can certainly wait till after the next release though...

PR #850, opened 2022-10-19
Drone minor config update

sdarwin:
- A small configuration fix  
- and also testing drone builds


mborland:
Looks like everything is working again. Thanks for the help.

PR #854, opened 2022-10-25
Remove redundant casts to bool

AtariDreams:
They are booleans already. Why do we need to static cast or implicitly cast?


mborland:
This PR is identical to https://github.com/boostorg/math/pull/841 where I provided you feedback.

PR #863, opened 2022-10-30
Junk removal

AtariDreams:
Removal of junk headers, typos, or mistakenly duplicated keywords


example/color_maps_sf_example.cpp
```diff
@@ -580,7 +580,7 @@ int main(int argc, char** argv)
     if (debug)
     {
        for (int64_t i = image_width / 2; i < image_width; ++i)
-          points[image_width * (image_height - 1) + i] = i & 1 ? 1 : 0;
+          points[image_width * (image_height - 1) + i] = i & 1;
```
jzmaddock, line 583: This is a superfluous change IMO.

AtariDreams: The expression is superfluous to start with

jzmaddock: How so?

AtariDreams: i & 1 can only evaluate to 1 or 0, so the ternary operator is redundant

include/boost/math/distributions/poisson.hpp
```diff
@@ -554,7 +554,6 @@ namespace boost
 // for this distribution have been defined, in order to
 // keep compilers that support two-phase lookup happy.
 #include <boost/math/distributions/detail/derived_accessors.hpp>
-#include <boost/math/distributions/detail/inv_discrete_quantile.hpp>
```
jzmaddock, line 557: There is a genuine mistake here, based on the comment in the file, it looks like we should be removing the *first* rather than last occurrence of the #include.  My bad on this one!

include/boost/math/interpolators/detail/cardinal_cubic_b_spline_detail.hpp
```diff
@@ -170,7 +170,7 @@ cardinal_cubic_b_spline_imp<Real>::cardinal_cubic_b_spline_imp(BidiIterator f, B
     m_beta.resize(length + 2, std::numeric_limits<Real>::quiet_NaN());
 
     // Since the splines have compact support, they decay to zero very fast outside the endpoints.
-    // This is often very annoying; we'd like to evaluate the interpolant a little bit outside the
+    // This is often very annoying; we'd like to evaluate the interpolant a little outside the
```
jzmaddock, line 173: We could argue this one forever, but the original text makes perfect sense to me.

example/color_maps_sf_example.cpp
jzmaddock: Doh! My apologies I read the diff backwards and thought you were adding the ?: operator.

My bad again!

AtariDreams:
Done @jzmaddock

include/boost/math/distributions/poisson.hpp
AtariDreams: Ok so I removed that instead and now it doesn't build

include/boost/math/special_functions/nonfinite_num_facets.hpp
```diff
@@ -223,7 +223,6 @@ namespace boost {
           *it = fill;
       }
 
-    private:
```
NAThompson, line 226: I'm confused; why remove the `private` keyword here?

include/boost/math/quadrature/detail/sinh_sinh_detail.hpp
```diff
@@ -50,7 +50,7 @@ class sinh_sinh_detail
     auto integrate(const F f, Real tolerance, Real* error, Real* L1, std::size_t* levels)->decltype(std::declval<F>()(std::declval<Real>())) const;
 
 private:
-private:
```
NAThompson, line 53: Why remove it? The `get_abscissa_row` function below is not part of the public API.

AtariDreams: Because private is literally above it. It is a duplicate

include/boost/math/special_functions/nonfinite_num_facets.hpp
AtariDreams: Ditto as below

AtariDreams: It is still private. Jus that the private keyword is already above

AtariDreams:
@mborland Ready!

jzmaddock:
@AtariDreams : other than my 2 comments above, I'm happy with this one (but I think those two need to be addressed).

AtariDreams:
#include <boost/math/distributions/detail/inv_discrete_quantile.hpp> was removed and the project failed to build as a result, so I think we need to keep it as is

AtariDreams:
Also the first complaint of yours was addressed @jzmaddock

jzmaddock:
There are two issues mentioned above: the workaround removed from tools/config.hpp properly belongs in tools/workaround.hpp (and yes we really do need float.h and not cfloat when the workaround is active), and the change to tools/precision.hpp will introduce a new compiler warning as it implicitly converts int to bool.

AtariDreams:
Done! @jzmaddock

AtariDreams:
@mborland All ready!

jzmaddock:
Ooops, merged too soon: "constructible" is the correct spelling as least as far as the C++ standard is concerned.  Changed that back in develop.

PR #864, opened 2022-10-31
Fix warnings in special functions

mborland:


include/boost/math/special_functions/detail/hypergeometric_series.hpp
```diff
@@ -239,7 +239,7 @@
            return r;
         }
         int s1, s2;
-        T r = boost::math::lgamma(T(z + n), &s1, pol) - boost::math::lgamma(z, &s2, pol);
+        auto r = static_cast<T>(boost::math::lgamma(T(z + n), &s1, pol) - boost::math::lgamma(z, &s2, pol));
```
jzmaddock, line 242: @mborland this diff all looks good to go to me, but just for future reference, I'm curious why this one warns as the results of the lgamma calls should be T's (even in the float case)?

mborland: I assume `z` is an integer type so the call `boost::math::lgamma(z, &s2, pol)` is promoting to double. If that is a bad assumption I am not sure why.

jzmaddock: No z is a real type.

mborland: > Which compiler/test case was triggering the warnings?

Example is [here](https://github.com/boostorg/math/actions/runs/3439561683/jobs/5737002658#step:11:837) using MSVC 2022 running `git_issue_810` with C++20.

PR #865, opened 2022-10-31
Use 64-bit RNG for types that are 64 bits

AtariDreams:
Do this instead of casting down every time we do a normal distribution.

### DIFF
--- a/include/boost/math/tools/random_vector.hpp
+++ b/include/boost/math/tools/random_vector.hpp
@@ -4,10 +4,10 @@
 //  Boost Software License, Version 1.0. (See accompanying file
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
-#include <vector>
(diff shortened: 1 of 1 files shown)


include/boost/math/tools/random_vector.hpp
```diff
@@ -25,12 +25,12 @@ std::vector<T> generate_random_vector(std::size_t size, std::size_t seed)
     }
     std::vector<T> v(size);
 
-    std::mt19937 gen(seed);
+    std::mt19937_64 gen(seed);
 
     std::normal_distribution<T> dis(0, 1);
-    for(std::size_t i = 0; i < v.size(); ++i)
+    for (auto& i : v)
```
NAThompson, line 31: I might use `auto& x : v` as we generally will assume `dis(gen)` produces a floating point number.

AtariDreams: Done!

Issue #314, opened 2020-02-07
Interpolations with (different) boost units


sebweb3r:
Hey, 

we (a physics group) are using the boost math (Barycentric Rational Interpolation and others) 
with boost units for our code. We have the problem, that the interpolation expects the same 
type for x and y data. In physics we quiet often have different units.

Could you extend the templates for different types/units for x and y.

Thanks in advance.

NAThompson:
Yes, I think that's fine. Could you give me a main I could try to get working?

BTW, I recommend pchip or makima interpolation over barycentric rational for almost all uses (log(N) complexity rather than O(N), with less spurious oscillations).

L0ric0:
i think something like this would be nice: https://github.com/L0ric0/boostorg-mat-issue-314

it could be that i was wrong with the template arguments for ` boost::math::barycentric_rational<quantity<si::energy, double>, quantity<si::electric_charge, double>>` but they should be the types of the axes for the function to interpolate.

NAThompson:
Moved to feature wishlist.

PR #867, opened 2022-10-31
Fix hopefully the last Clang-Tidy suggestions

AtariDreams:
There were some last second things I missed.

Among these warnings were "redundant static casts" and "deprecated C-headers"

There were others too but those warnings were false flags so I kept the ones that actually made sense.

### DIFF
--- a/example/hyperexponential_more_snips.cpp
+++ b/example/hyperexponential_more_snips.cpp
@@ -130,7 +130,7 @@ void print_fitted(ds_info const& ds)
    std::cout << "* Fitted Request Interarrival Time" << std::endl;
    std::cout << " - Mean (SD): " << boost::math::mean(ds.iat_he) << " (" << boost::math::standard_deviation(ds.iat_he) << ") seconds." << std::endl;
    std::cout << " - 99th Percentile: " << boost::math::quantile(ds.iat_he, 0.99) << " seconds." << std::endl;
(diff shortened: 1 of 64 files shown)


NAThompson:
@AtariDreams : I think your talents might be better spent on (say) [this issue](https://github.com/boostorg/math/issues/314). There is a [new units library](https://github.com/mpusz/units) that is being proposed for standardization; if you could change the interpolators to be generically compatible with dimensioned types then you could also give feedback on the proposal and identify pain points. This would be something you could be very proud of and of course put on your resume.

jzmaddock:
@AtariDreams : please can you refer to my previous comments - I'm not against PR's, indeed we love them - but can you:

* Please keep changes to something we can actually manage to review - if we have to sift through and check the logic of changes in 64 files, not only does that take a quite a while, but the risk of missing a mistake is actually pretty high.
* Please don't keep on pushing more and more changes to the same PR.  At a minimum it's super annoying if you're trying to review a PR only for it to change under your feet, but I've also had to go and cancel a whole bunch of CI runs triggered by these continued pushes.  Please remember that CI is a shared resource for the whole of Boost, and that a Math lib run takes up a pretty significant amount of resources.

Given that we've already rejected a PR rather similar to this one, I'm going to close this one down as well.  But please, small focused PR's, and once you've submitted a PR, please leave it alone and don't keep pushing more and more to it!  This of course, is another way of saying "please don't submit the PR until it's actually ready".

PR #868, opened 2022-11-05
Add additional architecture testing to drone

mborland:


jzmaddock:
Cool, I hadn't realised we had that facility!

mborland:
@sdarwin Do you see my error in adding M1? It's mostly copy-paste from the slack channel but I am getting the error `linter: duplicate pipeline names` on [drone side.](https://drone.cpp.al/boostorg/math/978)

sdarwin:
@mborland the first argument 'job name' must be unique.  Change to `"Clang " + cxx`

mborland:
@jzmaddock This seems to be working now, but produced some errors in `special_fun` and `distribution_tests`. Most of them I think are safe to just loosen the tolerance on a hair but several feel more nefarious like this one:

```
Testing: Elliptic Integral Jacobi Zeta: Random Data
953s
6099	jacobi_zeta<double> Max = 2.232e+14 RMS Mean=1.537e+13
953s
6100	    worst case at row: 1921
953s
6101	    { 1.4786438941955566, 3.2146250106279695e-09, 4.7345199477425445e-19 }
953s
6102	Peak error greater than expected value of 1
953s
6103	./handle_test_result.hpp:165: error: in "test_main": check bounds.first >= max_error_found has failed
953s
6104	Mean error greater than expected value of 1
953s
6105	./handle_test_result.hpp:170: error: in "test_main": check bounds.second >= mean_error_found has failed
```

Interestingly, I can't reproduce any of the errors on my M1 machine. @NAThompson are you able to reproduce?

jzmaddock:
@mborland : I'm going to extract the s390x logic out of this to a separate PR as that's the closest to being green, and then we should go from there.

jzmaddock:
@mborland : do we need to revisit running M1 tests?  I'm sure I'm missing something, but I'm not seeing what makes these tests run on M1?

mborland:
> @mborland : do we need to revisit running M1 tests? I'm sure I'm missing something, but I'm not seeing what makes these tests run on M1?

Now that you have ARM working I think it would be easy to add. It is generally this line:
```
      result.append(osx_cxx("M1 Clang " + cxx + " " + suite, "clang++", buildscript="drone", buildtype="boost", xcode_version="14.1", environment={'TOOLSET': 'clang', 'CXXSTD': cxx, 'TEST_SUITE': suite, 'DEFINE': 'BOOST_MATH_NO_REAL_CONCEPT_TESTS,BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS,BOOST_MATH_MULTI_ARCH_CI_RUN', }, globalenv=globalenv))
```

The macOS machines in drone are supposed to be M1.

sdarwin:
While the macOS machines are m1, it seems the OS can simulate x86 or use the native arm architecture.   

The drone process appears to be using x86 so I need to work on fixing that.

mborland:
> 

That makes sense because I was getting failures from CI that I could not reproduce locally on my M1 machine. Thanks.

sdarwin:
Switched m1 drone to arm64. Try it again.

mborland:
@jzmaddock It looks like M1 is good to go as well now. Merging.

PR #869, opened 2022-11-09
Fix hopefully the last Clang-Tidy suggestions

SiliconA-Z:
This has been tested and ran and passes. And has no more issues.


mborland:
@AtariDreams this is extremely similar to the other two PRs we have close with the same title. Please refer to the comments we have previously provided you.

PR #881, opened 2022-11-18
Fix -Wmaybe-uninitialized

mborland:


mborland:
Local GCC-12 doesn't show the warning but CI does. Closing for now.

PR #890, opened 2022-12-03
Fix for binomial distribution quantile edge case

### LINKED ISSUE #889: Binomial Distribution Edge Case Bug
See: https://github.com/scipy/scipy/issues/17388#issuecomment-1336112873

From the OP:
```
print(binom.ppf(q = 0.0, n = 4, p = 0.5)) # prints -1.0 (but should be 0.0?)
print(binom.ppf(q = 0.3, n = 4, p = 1.0)) # prints 0.0 (but should be 4.0?)
```

For the first case we return 0.0 which is the correct behavior, but SciPy [has it's own convention](https://github.com/scipy/scipy/pull/15967#issuecomment-1095817726). For the second case we are incorrect as [confirmed by WolframAlpha ](https://www.wolframalpha.com/input?i=quantile%28binomial+distribution%284%2C1%29%2C0.3%29).

mborland:
Closes #889

### DIFF
--- a/include/boost/math/distributions/binomial.hpp
+++ b/include/boost/math/distributions/binomial.hpp
@@ -224,7 +224,7 @@ namespace boost
            // but zero is the best we can do:
            return 0;
         }
-        if(p == 1)
+        if(p == 1 || success_fraction == 1)
         {  // Probability of n or fewer successes is always one,
            // so n is the most sensible answer here:
            return trials;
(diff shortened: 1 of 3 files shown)


mborland:
@NAThompson can you double check this?

jzmaddock:
That looks fine to me, looks like we adopted a random convention because it wasn't clear what the correct thing to do was, I very much doubt anyone is relying on the old behaviour for such an edge case.

PR #900, opened 2022-12-10
Fix for size_t underflow in tanh_sinh

### LINKED ISSUE #898: tanh_sinh quadrature underflows integers
# Description
It is possible for some integrals (and some scalar types), the tanh_sinh quadrature method underflows an unsigned integer, resulting in undefined behavior.

# Steps to reproduce 
See an example where the https://godbolt.org/z/5seaesETo

The integer underflow occurs on [this line](https://github.com/boostorg/math/blob/develop/include/boost/math/quadrature/detail/tanh_sinh_detail.hpp#L352) when `max_left_position` is 0.

mborland:
Closes #898

### DIFF
--- a/test/git_issue_898.cpp
+++ b/test/git_issue_898.cpp
@@ -1,0 +1,48 @@
+// Copyright Matt Borland, 2022
+// Use, modification and distribution are subject to the
+// Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include <cmath>
+#include <limits>
+#include <boost/math/quadrature/tanh_sinh.hpp>
+#include <boost/math/constants/constants.hpp>
+#include "math_unit_test.hpp"
+
+// numerically evaluate the integral for Stefan-Boltzmann Law from Planck's Law
+// https://en.wikipedia.org/wiki/Stefan%E2%80%93Boltzmann_law#Derivation_from_Planck's_law
+template <typename Real>
+Real StefanBoltzmann()
+{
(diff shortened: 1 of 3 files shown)


test/git_issue_898.cpp
```diff
@@ -0,0 +1,48 @@
...
+template <typename Real>
+Real StefanBoltzmann()
+{
+    constexpr auto h = Real(6.62607015e-34);
+    constexpr auto c = Real(299792458.0);
+    constexpr auto kB = Real(1.380649e-23);
+
+    constexpr auto c1 = Real(2) * boost::math::constants::pi<Real>() * h * c * c ;
+    constexpr auto c2 = h * c / kB;
+    constexpr auto T = 1000.;
```
jzmaddock, line 24: Interestingly, making T a double here *does not* reproduce the issue, making it an int:

```
constexpr auto T = 1000;
```

Not only reproduces the issue, but leads on to an out of bounds array subscript later on.

BTW, I'm not convinced the suggested fix is quite correct: the purpose of `max_left_position` is to protect against evaluation at a singularity (or in this case, an infinite result).  This might need some careful thought...

mborland: If `max_left_position` is 0 you would get underflow to `std::numeric_limits<size_t>::max()`, and then the array index is `max_left_position + 1` which overflows back to 0, but avoids seg fault.

jzmaddock:
@mborland : I've pushed an alternate fix for this in https://github.com/boostorg/math/pull/901, if you have the time, can I get you to sanity check that I've covered all the situations where max_left_index/max_right_index could reach zero?  It is at least asserted now that they never do.

mborland:
Superseded by #901

PR #901, opened 2022-12-11
Fix tanh_sinh integration for edge case,

jzmaddock:
where max_left_index/max_right_index may go to zero if the function being integrated overflows the FP type being used across nearly all of it's range.  Fixes https://github.com/boostorg/math/issues/898


mborland (approved):
Looks good to me. I'll close the other PR.

PR #905, opened 2022-12-27
Require C++14 support

mborland:
Since we have been messaging 1.82 will be the first version that requires C++14 start requiring it in the CI system.


mborland:
@jzmaddock If we are still good to move to C++14 in the next release I would rebase the s390x support on top of this.

PR #907, opened 2022-12-28
Use `size_t` in `size_to_precision`

Flamefire:
As the trait is supposed to be called with `sizeof` the type should be `size_t`

This avoids a conversion warning on e.g. GCC until 10.4. See https://godbolt.org/z/6Pq38zYbW

Reported on Slack: https://cpplang.slack.com/archives/C27KZLB0X/p1672180776054099


jzmaddock:
Looks good to me...

PR #912, opened 2023-01-07
Improve NAN handling in ccmath

mborland:


pulver (commented):
For the `hypot()`, `log()`, and `sqrt()` functions, shouldn't parameters with values of negative infinity result in NaN instead of negative infinity?

E.g. `sqrt(-∞) = nan` rather than `-∞`.

I know this behavior remains unchanged in this PR, but thought this might be an opportune time to fix.

mborland:
> For the `hypot()`, `log()`, and `sqrt()` functions, shouldn't parameters with values of negative infinity result in NaN instead of negative infinity?
> 
> E.g. `sqrt(-∞) = nan` rather than `-∞`.
> 
> I know this behavior remains unchanged in this PR, but thought this might be an opportune time to fix.

The language specification is [+- infinity returns +- infinity](https://en.cppreference.com/w/cpp/numeric/math/hypot). Since we fallback to the STL at runtime I didn't want the edge case handling to differ based on context.

jzmaddock:
> The language specification is [+- infinity returns +- infinity](https://en.cppreference.com/w/cpp/numeric/math/hypot). Since we fallback to the STL at runtime I didn't want the edge case handling to differ based on context.

C99 at least specifies that sqrt of a negative number is a domain error and yields a NaN, as things stand runtime sqrt(-INF) returns NaN and constexpr sqrt(-INF) returns +INF which looks just plain wrong?

mborland:
> > The language specification is [+- infinity returns +- infinity](https://en.cppreference.com/w/cpp/numeric/math/hypot). Since we fallback to the STL at runtime I didn't want the edge case handling to differ based on context.
> 
> C99 at least specifies that sqrt of a negative number is a domain error and yields a NaN, as things stand runtime sqrt(-INF) returns NaN and constexpr sqrt(-INF) returns +INF which looks just plain wrong?

Ok for `sqrt` the standard only specifies +inf. Domain error is implementation defined so NaN would make sense.

pulver:
According to https://en.cppreference.com/w/c/numeric/math/logb

> If arg is ±∞, +∞ is returned

I haven't tested it, but inspecting the code it appears that for the `constexpr` version
```
logb(-∞) = -∞
```
Shouldn't it return `+∞`?

mborland:
> According to https://en.cppreference.com/w/c/numeric/math/logb
> 
> > If arg is ±∞, +∞ is returned
> 
> I haven't tested it, but inspecting the code it appears that for the `constexpr` version
> 
> ```
> logb(-∞) = -∞
> ```
> 
> Shouldn't it return `+∞`?

Yes; the change is wrong.

mborland:
@pulver Can you do one last dummy check for me? I think this is good now and addresses your `sqrt(-∞) = nan`.

pulver (commented):
For `hypot()`, it looks like
```
hypot(-∞,1) = -∞
```
but should return +∞. Same when x and y are swapped, based on [hypot](https://en.cppreference.com/w/cpp/numeric/math/hypot):
 
> if one of the arguments is ±∞, hypot(x,y) returns +∞ even if the other argument is NaN

pulver (approved):
LGTM

PR #914, opened 2023-01-09
Add 3-arg hypot

mborland:


NAThompson:
Pushed commits to finish this off: merged develop; `hypot(0, 0, 0)` now returns 0 instead of dividing by zero; the new overloads are `BOOST_MATH_EXPORT`ed like the 2-arg ones; added tests (zeros, inf beats NaN, no spurious overflow, runtime path) and the signatures to the ccmath docs.

PR #916, opened 2023-01-11
Remove most CI and just test arm64/drone.

jzmaddock:
!!REVERT THIS COMMIT BEFORE MERGE!!


mborland:
I'm surprised this is green without any corrections

jzmaddock:
@mborland : Me too!  But I'm not particularly unhappy about it :) There were a heck of a lot of fixes in s390x testing though which seems to have done the trick.  Testing a full run now....

mborland:
Looking at the drone build logs I think we need to revisit how much we are testing with these emulated architectures. Our median build time prior to merging s390x was 46 minutes. Our recent tests have been taking between 5-15 hours because the number of runners for s390x is limited. I recommend we test these like the sanitizers; pick one language standard (C++17) and that is it. We shouldn't need to test `__float128` using GNU extensions on s390x because long double is 128-bits, and ARM does not support libquadmath.

jzmaddock:
All good thoughts, once this is green I'll cut down the permutations.

jzmaddock:
Failures are spurious, merging.

PR #917, opened 2023-01-11
Add deprecation notice to complex functions

### LINKED ISSUE #911: Complex number functions
Are these still needed?

https://www.boost.org/doc/libs/1_81_0/libs/math/doc/html/inverse_complex.html

I see that all six of these functions are in [std::complex](https://en.cppreference.com/w/cpp/numeric/complex) as of C++11.

mborland:
Closes #911


jzmaddock:
Drone failure is a spurious git-clone fail, merging.  Thanks for this Matt.

Issue #927, opened 2023-01-25
`powm1` Returning Wrong `INF`


mborland:
See https://github.com/scipy/scipy/actions/runs/3880172950/jobs/6618011730#step:6:585
From: https://github.com/scipy/scipy/pull/17432

We should be returning `-INF` instead of `INF`. This is a change between Scipy's last boost release (1.79/1.80) and develop.

mborland:
Scipy had their own handling which should be fixed by: https://github.com/scipy/scipy/pull/17855

PR #928, opened 2023-01-25
Fix for issue #927

mborland:
Return `-INF` or `+INF` if the result is `INF` rather than throwing an overflow exception.


jzmaddock:
I think this is the wrong fix - and actually we are doing exactly what we promised to do here.

If you want the sign of the infinity rather than an exception, then use a Policy which ignores overflow errors and you will get a signed infinity back.

mborland:
What is the appropriate policy to use here to get the signed infinity?

jzmaddock:
If it's project wide then 

```
#define BOOST_MATH_OVERFLOW_ERROR_POLICY ignore_error
```

or

```
#define BOOST_MATH_OVERFLOW_ERROR_POLICY errno_on_error
```

Depending on whether you want to be able to check `errno` for overflow.

If it's just this function, then:

```
using namespace boost::math::policies;

boost::math::specfun(arg, make_policy(overflow_error<errno_on_error>()));
```

PR #934, opened 2023-01-30
Downgrade condition numbers language requirement to C++14

mborland:
@NAThompson


mborland:
Only failure was on CI's end. Merging.

PR #937, opened 2023-01-31
FIX: avoid overflow on overflow check in halley_step on Apple M1

mckib2:
- avoid calculating `[number]*tools::max_value<T>()` to prevent overflows on Apple M1 precessors

### DIFF
--- a/include/boost/math/tools/roots.hpp
+++ b/include/boost/math/tools/roots.hpp
@@ -354,7 +354,12 @@ namespace detail {
          BOOST_MATH_INSTRUMENT_VARIABLE(denom);
          BOOST_MATH_INSTRUMENT_VARIABLE(num);
 
-         if ((fabs(num) < 1) && (fabs(denom) >= fabs(num) * tools::max_value<T>()))
+         // denom/num overflows if:
+         //     |denom| >= |num| * max_value
+         // RHS may overflow on Apple M1, so rearrange:
+         //     |denom| * 1/max_value >= |num|
+         static const T inv_max_value = 1.0 / tools::max_value<T>();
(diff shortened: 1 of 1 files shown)


jzmaddock:
I'm sure I'm missing something obvious here, but if denom >=1 there can be no overflow, and if its < 1 then denom*numeric_limits<>::max() can't overflow either?  Or is a logical branch which should never be taken being speculatively evaluated?

include/boost/math/tools/roots.hpp
```diff
@@ -354,7 +354,12 @@ namespace detail {
          BOOST_MATH_INSTRUMENT_VARIABLE(denom);
          BOOST_MATH_INSTRUMENT_VARIABLE(num);
 
-         if ((fabs(num) < 1) && (fabs(denom) >= fabs(num) * tools::max_value<T>()))
+         // denom/num overflows if:
+         //     |denom| >= |num| * max_value
+         // RHS may overflow on Apple M1, so rearrange:
+         //     |denom| * 1/max_value >= |num|
+         constexpr T inv_max_value = 1.0 / tools::max_value<T>();
```
mborland, line 361:
```suggestion
         const T inv_max_value = 1.0 / tools::max_value<T>();
```

jzmaddock: Even better make that `static const` and then it's evaluated just the once.

mckib2: I wanted `static constexpr`, but wasn't sure that is supported everywhere we want, and indeed, it appears even `constexpr` is problematic.  Will implement this right away

mckib2:

PR #941, opened 2023-02-04
 FIX: avoid overflow on overflow check in toms748_solve safe_div

mckib2:
- Same rationale/fix as https://github.com/boostorg/math/pull/937

### DIFF
--- a/include/boost/math/tools/toms748_solve.hpp
+++ b/include/boost/math/tools/toms748_solve.hpp
@@ -160,7 +160,8 @@ inline T safe_div(T num, T denom, T r)
 
    if(fabs(denom) < 1)
    {
-      if(fabs(denom * tools::max_value<T>()) <= fabs(num))
+      static const T inv_max_value = 1.0 / tools::max_value<T>();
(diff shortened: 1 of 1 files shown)


include/boost/math/tools/toms748_solve.hpp
```diff
@@ -160,7 +160,8 @@ inline T safe_div(T num, T denom, T r)
 
    if(fabs(denom) < 1)
    {
-      if(fabs(denom * tools::max_value<T>()) <= fabs(num))
+      constexpr T inv_max_value = 1.0 / tools::max_value<T>();
```
mborland, line 163:
```suggestion
      const T inv_max_value = 1.0 / tools::max_value<T>();
```

As @jzmaddock noted in https://github.com/boostorg/math/pull/942 we still need to support types that don't have constexpr construction.

jzmaddock: Even better make that `static const` and then it's evaluated just the once.

mckib2: Implemented

mckib2:

PR #942, opened 2023-02-04
Make real concept constexpr

mborland:
With C++14 we can have multi-line constexpr member functions so make `real_concept` support constexpr construction and operations. Should fix the build errors in #937.

### DIFF
--- a/include/boost/math/concepts/real_concept.hpp
+++ b/include/boost/math/concepts/real_concept.hpp
@@ -60,62 +60,62 @@ class real_concept
 {
 public:
    // Constructors:
-   real_concept() : m_value(0){}
-   real_concept(char c) : m_value(c){}
-   real_concept(wchar_t c) : m_value(c){}
-   real_concept(unsigned char c) : m_value(c){}
-   real_concept(signed char c) : m_value(c){}
-   real_concept(unsigned short c) : m_value(c){}
-   real_concept(short c) : m_value(c){}
-   real_concept(unsigned int c) : m_value(c){}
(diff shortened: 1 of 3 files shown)


jzmaddock:
Sorry wrong fix: we want real_concept to model our absolute minimal requirements, and there are number types (for example multiprecision types which reference C library backends such as MPFR) which can never be constepxr.

mborland:
On an unrelated note if we are going to expand constexpr support in C++14 it might be worth dropping MSVC-14.0. It was unhappy in the PRs with tables, is doing the wrong thing here, and [from this pr](https://github.com/boostorg/math/pull/920) is missing support to full C++14. For now I think we are OK, but worth being aware of.

@NAThompson for awareness

NAThompson:
@mborland , @jzmaddock : Does MSVC do autoupdates now? Certainly on Mac it is a task of some difficulty to stay on an old AppleClang version.

mborland:
You can still download old versions on the website https://visualstudio.microsoft.com/vs/older-downloads/. To get an older Apple Clang build I have to login with an Apple Developer account, and it still tries to update every time you launch. This was a problem when Ventura was released but the toolchains hadn't been updated yet...

ckormanyos:
>  if we are going to expand `constexpr` support in C++14 it might be worth dropping MSVC-14.0

MSVC-14.0 cannot handle C++14 `constexpr` reliably. If you would like to use the acutal word `constexpr` without workarounds and within a C++14-compatible environment, the MSVC-14.0 must be dropped.

PR #944, opened 2023-02-05
FIX: avoid overflow on overflow check in gamma_p_derivative_imp

mckib2:
- Same rationale/fix as https://github.com/boostorg/math/pull/937


mckib2:

PR #945, opened 2023-02-05
FIX: avoid overflow on overflow check for Mac M1

mckib2:

### DIFF
--- a/include/boost/math/tools/roots.hpp
+++ b/include/boost/math/tools/roots.hpp
@@ -652,7 +652,9 @@ namespace detail {
          }
          else if (result > max)
          {
-            T diff = ((fabs(max) < 1) && (fabs(result) > 1) && (tools::max_value<T>() / fabs(result) < fabs(max))) ? T(1000) : T(result / max);
+            const T amax = fabs(max);
+            volatile const T aresult = fabs(result); // volatile: force compiler to honor data-dependency in chained bool exprs below
+            T diff = ((amax < 1) && (aresult > 1) && (tools::max_value<T>() / aresult < amax)) ? T(1000) : T(result / max);
             if (fabs(diff) < 1)
                diff = 1 / diff;
             if (!out_of_bounds_sentry && (diff > 0) && (diff < 3))
--- a/include/boost/math/tools/roots.hpp
+++ b/include/boost/math/tools/roots.hpp
@@ -354,7 +354,12 @@ namespace detail {
          BOOST_MATH_INSTRUMENT_VARIABLE(denom);
          BOOST_MATH_INSTRUMENT_VARIABLE(num);
 
-         if ((fabs(num) < 1) && (fabs(denom) >= fabs(num) * tools::max_value<T>()))
+         // denom/num overflows if:
+         //     |denom| >= |num| * max_value
+         // RHS may overflow on Apple M1, so rearrange:
+         //     |denom| * 1/max_value >= |num|
+         static const T inv_max_value = 1.0 / tools::max_value<T>();
+         if ((fabs(num) < 1) && (inv_max_value * fabs(denom) >= fabs(num)))
          {
             // possible overflow, use Newton step:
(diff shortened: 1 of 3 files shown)


mckib2:
@mborland @jzmaddock Didn't realize the changes would be similar for all issues at the start, so consolidating all these small PRs into a single one

include/boost/math/tools/roots.hpp
```diff
@@ -647,7 +652,9 @@ namespace detail {
          }
          else if (result > max)
          {
-            T diff = ((fabs(max) < 1) && (fabs(result) > 1) && (tools::max_value<T>() / fabs(result) < fabs(max))) ? T(1000) : T(result / max);
+            const T amax = fabs(max);
+            volatile const T aresult = fabs(result); // volatile: force compiler to honor data-dependency in chained bool exprs below
```
mborland, line 656: You should really only need `const volatile` in a shared memory environment. Do you still get overflow errors with just `const`?

mckib2: Yes, I tried all the variations I could think of.  What also worked was predclaring the `amax` and `aresult` and switching around `tools::max_value<T>() / fabs(result) < fabs(max)` to `tools::max_value<T>() / amax < aresult`, but I that was only suppressing it in this case for a specific relationship between `result` and `max`, so the `volatile` in my mind makes it a little more generic

mckib2:
I believe SciPy is using [this](https://github.com/cirruslabs/macos-image-templates/pkgs/container/macos-monterey-xcode/22082164?tag=13.3.1) image and seeing the errors.  It's about 9 months old and I wonder if updating to the latest may avoid these (compiler?) issues/bugs

mborland:
> I believe SciPy is using [this](https://github.com/cirruslabs/macos-image-templates/pkgs/container/macos-monterey-xcode/22082164?tag=13.3.1) image and seeing the errors. It's about 9 months old and I wonder if updating to the latest may avoid these (compiler?) issues/bugs

Might be worth updating. I am using macOS Ventura with Clang 14.0.0 on an M1 MacBook Pro and see none of the errors. Anyone on Ventura won't even have the option for the Clang 13 series.

mckib2:
> is this fixed in the latest release as Matt suggests?

Testing this out now

> If not, how come Matt and I are unable to reproduce? There still appears to be something critical missing.

Sounds like neither of you have tried on this specific version of MacOS?  I also notice it only shows up when `-O2` or `-O3` optimizations are enabled.

> What's the Apple/Clang take on this? Is it a known issue upstream?

I wasn't able to find anything upstream, I'll submit an issue

mborland:
> 
> > What's the Apple/Clang take on this? Is it a known issue upstream?
> 
> 
> 
> I wasn't able to find anything upstream, I'll submit an issue

If you are able to make an extremely minimal reproducer like the GCC Bugzilla requires @NAThompson can route it to the right people at Apple.

mckib2:
> > > What's the Apple/Clang take on this? Is it a known issue upstream?
> > 
> > 
> > I wasn't able to find anything upstream, I'll submit an issue
> 
> If you are able to make an extremely minimal reproducer like the GCC Bugzilla requires @NAThompson can route it to the right people at Apple.

Minimal reproducer (not sure how it compares to GCC Bugzilla) is in the upstream llvm-project issue: https://github.com/llvm/llvm-project/issues/60695#issue-1581432780

mckib2:
> If you are able to make an extremely minimal reproducer like the GCC Bugzilla requires @NAThompson can route it to the right people at Apple.

@mborland The issue was rejected by Clang maintainers, I'll try looking up my Apple ID and submitting an issue there

mborland:
Since the issue seems to be fixed in newer macOS I wouldn't be surprised if they don't investigate the bug either.

NAThompson:
Feel free to reopen if this is still a problem-but otherwise closing as stale.

PR #950, opened 2023-02-12
Fix hypergeometric for use with -fno-exceptions.

jzmaddock:
Add test case.
Refs: https://github.com/boostorg/math/pull/947


jzmaddock:
Failures are annoyingly spurious, merging...

Issue #961, opened 2023-03-03
Unexpected exception thrown by ibeta_inv with extremely small x.


WarrenWeckesser:
I'm using the develop branch, commit 82ccb8590723d9a954384e6b453b647129c0799d.  The compiler:
```
$ g++ --version
g++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
```

The code below throws the exception

```
ibeta_inv_tiny_x: /home/warren/repos/git/forks/boost/math/include/boost/math/special_functions/
detail/ibeta_inverse.hpp:122: T boost::math::detail::temme_method_1_ibeta_inverse(T, T, T, const Policy&)
[with T = long double; Policy = 
boost::math::policies::policy<boost::math::policies::promote_float<false>, 
boost::math::policies::promote_double<false>, boost::math::policies::default_policy, 
boost::math::policies::default_policy, boost::math::policies::default_policy, 
boost::math::policies::default_policy, boost::math::policies::default_policy, 
boost::math::policies::default_policy, boost::math::policies::default_policy, 
boost::math::policies::default_policy, boost::math::policies::default_policy>]: 
Assertion `x >= 0' failed.
Aborted (core dumped)
```
At line 122 of `ibeta_inverse.hpp`, the value of `x` that triggers the exception is -5.42101e-20.

There are comments in the code below that give nearby values of the input `x` that exhibit different behaviors.

`ibeta_inv_tiny_x.cpp`:
```
#include <iostream>
#include <iomanip>
#include <cfenv>
#include <boost/math/special_functions/beta.hpp>

using namespace std;
using namespace boost::math;

void show_fp_exception_flags()
{
    if (std::fetestexcept(FE_DIVBYZERO)) {
        cout << " FE_DIVBYZERO";
    }
    // FE_INEXACT is common and not interesting.
    // if (std::fetestexcept(FE_INEXACT)) {
    //     cout << " FE_INEXACT";
    // }
    if (std::fetestexcept(FE_INVALID)) {
        cout << " FE_INVALID";
    }
    if (std::fetestexcept(FE_OVERFLOW)) {
        cout << " FE_OVERFLOW";
    }
    if (std::fetestexcept(FE_UNDERFLOW)) {
        cout << " FE_UNDERFLOW";
    }
    cout << endl;
}

int main(int argc, char *argv[])
{
    double a = 14.208308325339239;
    double b = a;

[...]

PR #962, opened 2023-03-04
Fix ibeta_inv for very small p.

jzmaddock:
Change assert's in temme_method_1_ibeta_inverse to corrections when guess goes out of range. Change handling of non-convergence in second_order_root_finder to use bracketing when the end points are many orders of magnitude apart. Fixes: https://github.com/boostorg/math/issues/961.


WarrenWeckesser:
This fixes the problem with the code that I included in #961. :+1:

PR #965, opened 2023-03-06
Conditionally Disable MP Testing 

mborland:
Should significantly reduce resource usage. Changes a macro used for testing under standalone so now you have to explicitly enable MP testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
         run: ./config_info_travis
         working-directory: ../boost-root/libs/config/test
       - name: Test
-        run: ../../../b2 toolset=$TOOLSET ${{ matrix.suite }} define=CI_SUPPRESS_KNOWN_ISSUES define=SLOW_COMPILER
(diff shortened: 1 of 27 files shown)


jzmaddock:
Could we enable MP-testing for at least ONE compiler in CI?  Otherwise I fear breakages will go unnoticed.

mborland:
> Could we enable MP-testing for at least ONE compiler in CI? Otherwise I fear breakages will go unnoticed.

Yes. Once the drone run finishes I will see what the delta in our resource usage is. If we need to whack testing down more I can continue to sprinkle that macro around, and then define it in a few CI runs.

jzmaddock:
> Yes. Once the drone run finishes I will see what the delta in our resource usage is. If we need to whack testing down more I can continue to sprinkle that macro around, and then define it in a few CI runs.

Got it, cool.  I suspect the multiprc_concept_check_N tests are probably the worst offenders for memory usage.

mborland:
@jzmaddock  I appreciate your comment in `test_1F1` that: `These next 2 tests take effectively "forever" to compile with clang:`

mborland:
The two CI failures in drone are spurious.

PR #968, opened 2023-03-17
Fix rounding for non-representable numbers

mborland:
Partially addresses re-raised #430. Once CI validates this fix works for `llround` will propagate to the rest of the functions.


mborland:
@jzmaddock any idea why this would break the use of pre-compiled headers under MSVC? See: https://github.com/boostorg/math/actions/runs/4447933117/jobs/7810069256?pr=968#step:11:523

jzmaddock:
> any idea why this would break the use of pre-compiled headers under MSVC?

No... I wonder if there are any unrelated changes effecting us - build system?

BTW I can't help thinking there may be a simpler solution, let me have a think.

jzmaddock:
Ah.... the test case is wrong: it assumes that the boundary case on 2^d where there are d digits in the long long is always an integer, as is `float_prior` of that value.  That will hold as long as there are fewer digits in the floating point type than the integer, but not otherwise (long double in this case).

So... I guess either only test floats that have fewer digits than the integer type - since that is the bug case - or else we need to figure out how to fix the test case.  The latter might be tricky, since we're basically duplicating the functionality of llround to carry out the test!  So I'd favour the former for now?

mborland:
The PCH issues are likely related to recently changes in B2. I have opened an issue with them:  https://github.com/bfgroup/b2/issues/229

mborland:
@jzmaddock Can you take a look at this now that it's green? The context switching for constexpr calculation is more verbose than I had hoped.

include/boost/math/special_functions/round.hpp
```diff
@@ -14,6 +15,16 @@
 #include <boost/math/policies/error_handling.hpp>
 #include <boost/math/special_functions/math_fwd.hpp>
 #include <boost/math/special_functions/fpclassify.hpp>
+#include <type_traits>
+#include <limits>
+#include <cmath>
+
+#if __cplusplus >= 201703L || _MSVC_LANG >= 201703L
```
jzmaddock, line 22: Minor point, I get nervous about some direct checks like this, should we use BOOST_NO_CXX17_IF_CONSTEXPR instead, or failing that BOOST_CXX_VERSION?

include/boost/math/special_functions/round.hpp
```diff
@@ -14,6 +15,16 @@
...
+#include <boost/math/ccmath/ldexp.hpp>
+#  if !defined(BOOST_MATH_NO_CONSTEXPR_DETECTION) && !defined(BOOST_LIBS_MATH_TEST_INSTANTIATE_HPP) // Concept tests throw off constexpr ldexp
```
jzmaddock, line 24: The tests pass for me (msvc and cygwin gcc) without the ` && !defined(BOOST_LIBS_MATH_TEST_INSTANTIATE_HPP)` part, and in general the concept checks are there to catch unexpected bugs that are tricky to spot otherwise.  What was failing without this?

jzmaddock:
Thanks for this Matt: other than the two minor comments looks good to me!

include/boost/math/special_functions/round.hpp
mborland: BOOST_NO_CXX17_IF_CONSTEXPR works because there is a definition for standalone mode already

include/boost/math/special_functions/round.hpp
mborland: This was left over from debugging; I'll remove it.

Issue #970, opened 2023-03-28
`simple_continued_fraction`


Tomaqa:
Please provide this information in the [documentation](https://www.boost.org/doc/libs/1_81_0/libs/math/doc/html/math_toolkit/internals/simple_continued_fraction.html), it is available only in the source code:
```
// Deal with non-uniqueness of continued fractions: [a0; a1, ..., an, 1] = a0; a1, ..., an + 1].
// The shorter representation is considered the canonical representation,
```

Also, I am disappointed that I cannot acces particular coefficients of the continued fraction, so the whole class is useless in the case one does not want to use the provided public functions.
I want to use it for [rational approxamations](https://en.wikipedia.org/wiki/Continued_fraction#Best_rational_approximations).
So I also need to convert it to [rational](https://www.boost.org/doc/libs/1_81_0/libs/rational/rational.html) afterwards.

Tomaqa:
I added these member functions as a workaround:
```
    size_t size() const noexcept { return b_.size(); }

    const auto& operator [](int idx) const { return b_[idx]; }
    auto& operator [](int idx) { return b_[idx]; }

    void reserve(size_t size_) { b_.reserve(size_); }
    void push_back(Z c) { b_.push_back(c); }
```

Although the non-const functions may cause inconsistency with the private member variable `x_`, it seems that it is not used anywhere else than in the constructor (which suggests to actually remove the member variable and use it only the parameter itself).

Also, it is a question whether to check index bounds or not.

PR #971, opened 2023-03-29
simple_continued_fraction: added public functions to access and modify the coefficients

Tomaqa:
- All coefficients are accessible *and mutable*, and the class can be used as a container.
- I removed `partial_denominators`, because IMO the name is misleading - front element represents integer part, and is not denominator. It may confuse the user that the function returns sth. where the integer part is omitted.
- I removed member variable `x_` because I consider it useless - it is only necessary in the constructor, and also it would be difficult to make it consistent with the new mutable member function.
- The only exception are lines 154-155 (in the previous file), which I do not understand, e.g. how it uses `.x_.backend()`. But if the precision is somehow needed, only precision should be stored, not the original value.

Partially solves #970, but I did not handle documentation.

I also implemented conversion (non-member) function to `boost::rational`, but I am not sure where to place it, because it needs to include both `simple_continued_fraction.hpp` and `rational.hpp`.


Tomaqa:
Mutability simply allows further operations with continued fractions, not only constructing them from floats.
For example, [rational approximation](https://en.wikipedia.org/wiki/Continued_fraction#Best_rational_approximations) requires to perform binary operations on already built continued fractions, _not with floats_.
Only const accessors to the coefficients are not enough in cases when a continued fraction is supposed to be a result of a binary operation of two continued fractions.
I believe that there are many more use cases.

Modifying particular coefficients does turn the object into an inconsistent state.
The only exception is that one may set a coefficient a_i, i > 0, to a _non-positive_ number. That is true.. This can be easily checked within `push_back`, but mutable `operator []` may be replaced by a setter in order to prevent the inconsistency.

Tomaqa:
Also, I can update the code in order to fix some of the tests that failed. But not all of them, for example in cases where `partial_denominators` are missing, which I argued that should be removed or renamed.

include/boost/math/tools/simple_continued_fraction.hpp
```diff
@@ -29,20 +29,28 @@
 
 namespace boost::math::tools {
 
-template<typename Real, typename Z = int64_t>
+template<typename Real, typename I = int64_t>
```
mborland, line 32: The set of all integers is represented with Z: https://en.wikipedia.org/wiki/Integer. You'll see it throughout the code base.

include/boost/math/tools/simple_continued_fraction.hpp
```diff
@@ -29,20 +29,28 @@
...
 class simple_continued_fraction {
 public:
-    simple_continued_fraction(Real x) : x_{x} {
+    typedef I int_type;
+
+    ~simple_continued_fraction() = default;
+    simple_continued_fraction(const simple_continued_fraction&) = default;
+    simple_continued_fraction& operator =(const simple_continued_fraction&) = default;
+    simple_continued_fraction(simple_continued_fraction&&) = default;
+    simple_continued_fraction& operator =(simple_continued_fraction&&) = default;
```
mborland, line 41: The core guidelines recommends against all of this as the compiler will generate them automatically. See: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero

include/boost/math/tools/simple_continued_fraction.hpp
```diff
@@ -78,29 +85,31 @@ class simple_continued_fraction {
...
-          b_.resize(b_.size() - 1);
+           b_.pop_back();
+           b_.back() += 1;
        }
        b_.shrink_to_fit();
-       
-       for (size_t i = 1; i < b_.size(); ++i) {
+
+       const int size_ = b_.size();
+       for (int i = 1; i < size_; ++i) {
```
mborland, line 94: This will result in a warning for `-Wsign-conversion`.

include/boost/math/tools/simple_continued_fraction.hpp
Tomaqa: OK, my motivation was to use the same identifier as is used in `boost::rational`, but OK, I will stick to the `Z`.

include/boost/math/tools/simple_continued_fraction.hpp
Tomaqa: It won't, I changed it in such a way that both variables are `int`.

mborland: `std::vector<T>::size_type` is an unsigned integer so you would need to cast to `const int size_`. Signed overflow is UB is while unsigned overflow is defined. I think this has been the case since C99.

include/boost/math/tools/simple_continued_fraction.hpp
```diff
@@ -110,53 +119,57 @@ class simple_continued_fraction {
...
+template<typename Real, typename I2>
+std::ostream& operator<<(std::ostream& out, simple_continued_fraction<Real, I2>& scf) {
    constexpr const int p = std::numeric_limits<Real>::max_digits10;
-   if constexpr (p == 2147483647) {
-      out << std::setprecision(scf.x_.backend().precision());
-   } else {
-      out << std::setprecision(p);
-   }
-   
+   static_assert(p != 2147483647, "numeric_limits<Real>::max_digits10 == 2147483647");
```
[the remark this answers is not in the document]
Tomaqa: To resolve this, I added member variable `precision_` and set it within the constructor.
There should be some check like `if constexpr (requires { orig_x.backend(); } ...`, but this is C++20 feature and in C++17 it is quite painful to achieve this ...

Tomaqa:
> I confess I do not see a good reason to allow non-const modifications.

Consider the following example, where I want to compute [best rational approximation within an interval](https://en.wikipedia.org/wiki/Continued_fraction#Best_rational_approximations):
```
using boost::math::tools::simple_continued_fraction;

constexpr RealT eps = 1e-6;

const RealT lhs = x;
const RealT rhs = x + eps;

const auto cont_frac1 = simple_continued_fraction(lhs);
const auto cont_frac2 = simple_continued_fraction(rhs);
auto& coefs1 = cont_frac1.partial_denominators();
auto& coefs2 = cont_frac2.partial_denominators();

const size_t max_size = min(coefs1.size(), coefs2.size());
simple_continued_fraction cont_frac;
auto& coefs = cont_frac.partial_denominators();
coefs.reserve(max_size);
for (size_t i = 0; i < max_size; ++i) {
    const auto c1 = coefs1[i];
    const auto c2 = coefs2[i];
    if (c1 == c2) {
        coefs.push_back(c1);
        continue;
    }

    coefs.push_back(min(c1, c2) + 1);
    break;
}

return cont_frac;
```

It is not possible to construct such `simple_continued_fraction` with only const `partial_denominators`.

include/boost/math/tools/simple_continued_fraction.hpp
```diff
@@ -32,34 +36,59 @@ namespace boost::math::tools {
 template<typename Real, typename Z = int64_t>
 class simple_continued_fraction {
 public:
-    simple_continued_fraction(Real x) : x_{x} {
+    typedef Z int_type;
+
+    simple_continued_fraction(std::vector<Z> data) : b_{std::move(data)} {
```
mborland, line 41:
```suggestion
    simple_continued_fraction(std::vector<Z>&& data) : b_{data} {
```

As written this makes a copy of the data vector and then moves it into `b_`. I expect you want to take an r-value reference instead.

include/boost/math/tools/simple_continued_fraction.hpp
```diff
@@ -171,6 +208,41 @@ std::ostream& operator<<(std::ostream& out, simple_continued_fraction<Real, Z2>&
...
+template<typename Real, typename Z = std::int64_t>
+inline auto simple_continued_fraction_coefficients(Real x)
+{
+    auto temp = simple_continued_fraction<Real, Z>(x);
+    return temp.get_data();
+}
+
+// Can be used with `boost::rational` from <boost/rational.hpp>
+template <typename Rational, typename Real, typename Z = std::int64_t>
+inline Rational to_rational(const simple_continued_fraction<Real, Z>& scf)
```
mborland, line 220: Does this work with other rational types [(e.g. `mpq_t`)](https://gmplib.org/manual/Rational-Number-Functions)?

mborland:
Is this being superseded by #975?

include/boost/math/tools/simple_continued_fraction.hpp
Tomaqa: `simple_continued_fraction(std::vector<Z>)` can be used both with `std::vector<Z>&&` and `const std::vector<Z>&` arguments, whereas `simple_continued_fraction(std::vector<Z>&&)` cannot be used with `const std::vector<Z>&`. In the case of `std::vector<Z>&&` argument, there should no significant difference (`std::vector<Z>` can require one more call to move constructor, which is negligible).

More importantly, I'm pretty sure that your suggestion `simple_continued_fraction(std::vector<Z>&& data) : b_{data}` will call copy constructor on `data`, because you pass it as an lvalue, not as an rvalue.

include/boost/math/tools/simple_continued_fraction.hpp
Tomaqa: I assume that `mpq_t` cannot be used since it is C-style struct which requires to call `mpq_init` etc.
The function can be used with any type `T` that supports `typename T::int_type`, `T(typename T::int_type)` (just integer part) and `T(typename T::int_type, typename T::int_type)` (numerator and denominator).
I assume C++20 is not supported so I do not use concepts or such.

Tomaqa:
> Is this being superseded by https://github.com/boostorg/math/pull/975?

Totally, I did not know that it will connect back with this PR, so I assume #975 is redundant.

NAThompson:
Thanks for this, @Tomaqa, and sorry it sat so long. We're keeping `simple_continued_fraction` immutable, so we won't take the mutable coefficients or construction from a coefficient vector. Several of the problems you found are real, though, and #1483 fixes them: the stored `x_` (which also made the class non-assignable), the precision handling in `operator<<`, the canonical form, and the wrong `#error` message. Closing this as superseded by #1483.

PR #977, opened 2023-04-14
Fix for scipy issue 18302

mborland:
See: https://github.com/scipy/scipy/issues/18302

Adds bounds checking so we don't return a pdf larger than beta, and catches an overflow exception when x is near `std::numeric_limits::min`.

### DIFF
--- a/include/boost/math/distributions/beta.hpp
+++ b/include/boost/math/distributions/beta.hpp
@@ -390,13 +390,47 @@ namespace boost
       }
       using boost::math::beta;
 
-      // Corner case: check_x ensures x element of [0, 1], but PDF is 0 for x = 0 and x = 1. PDF EQN:
+      // Corner cases: check_x ensures x element of [0, 1], but PDF is 0 for x = 0 and x = 1. PDF EQN:
       // https://wikimedia.org/api/rest_v1/media/math/render/svg/125fdaa41844a8703d1a8610ac00fbf3edacc8e7
-      if(x == 0 || x == 1)
+      if(x == 0)
       {
-        return RealType(0);
+        if (a == 1)
+        {
+          return 1 / beta(a, b);
+        }
+        else if (a < 1)
+        {
+          policies::raise_overflow_error<RealType>(function, nullptr, Policy());
+        }
+        else
+        {
+          return RealType(0);
+        }
       }
-      return ibeta_derivative(a, b, x, Policy());
+      else if (x == 1)
+      {
+        if (b == 1)
+        {
+          return 1 / beta(a, b);
+        }
+        else if (b < 1)
+        {
+          policies::raise_overflow_error<RealType>(function, nullptr, Policy());
+        }
+        else
+        {
+          return RealType(0);
+        }
+      }
+      
+      // Bounds checking
+      RealType return_val = ibeta_derivative(a, b, x, Policy());
+      if (return_val > b)
+      {
+        return_val = b;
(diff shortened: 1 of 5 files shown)


WarrenWeckesser:
Thanks for the fast response, @mborland!

It looks like the code in `test/scipy_issue_18302.cpp` uses the default policy--is that correct?  I wrote a short C++ code to check the case `x = 1e-310`, and I didn't get the overflow error with the default policy. I had to use `promote_double<false>`.

mborland:
> Thanks for the fast response, @mborland!
> 
> 
> 
> It looks like the code in `test/scipy_issue_18302.cpp` uses the default policy--is that correct?  I wrote a short C++ code to check the case `x = 1e-310`, and I didn't get the overflow error with the default policy. I had to use `promote_double<false>`.
> 
> 

I did get overflow with the default policy. Now that you bring it up I ran it on an ARM Mac which means long double = double. I think in this case it's worth effectively ignoring the overflow exception because we know what the result is supposed to be when that happens.

WarrenWeckesser:
Ah, I see.  I tested on a Linux machine, where `long double` is extended precision.  I was just worried that the test wasn't actually testing the fix because it would pass even without the fix.  But that is not the case on platforms where `long double` is `double`.

WarrenWeckesser:
Nice.  I think it might be slightly better to use `promote_double<false>` so the test in fact fails on all platforms without the fix, and therefore the test is verifying that the change actually fixes the problem on all platforms, but obviously that's your call.

jzmaddock:
@mborland I think we should fix this updatream in ibeta_derivative here: https://github.com/boostorg/math/blob/109a814e89f77ff8a3fc8f0391f6b35a12640669/include/boost/math/special_functions/beta.hpp#L1455  It's the 1/y that fails and we need to decide whether to return the derivative at zero (which may well be infinite) or at `tools::min_value<T>()` which may well not be.  I'm edging towards the latter as the easier fix.

mborland:
@jzmaddock This is green and I think hits your comments from both threads.

include/boost/math/distributions/beta.hpp
```diff
@@ -390,13 +390,47 @@ namespace boost
...
+        {
+          return RealType(0);
+        }
+      }
+      
+      // Bounds checking
+      RealType return_val = ibeta_derivative(a, b, x, Policy());
+      if (return_val > b)
+      {
+        return_val = b;
```
jzmaddock, line 430: This doesn't look right to me - surely the PDF can be larger than the shape parameter?  Indeed can it not go to infinity at the endpoints?

mborland: You're right. I was trying to catch the instance where we overshoot the endpoint (like in the linked issue). I can't think of any particularly good ways to do so.

mborland: `test_beta_dist` uses 100 eps as the tolerance in the spot checks. I get under 40 eps above the endpoint from the linked issue locally.

jzmaddock: Nod.  The only endpoint-overshoot that can occur is when one of alpha and beta is exactly one, and then the endpoint is `1 / beta(alpha, beta)` and we currently return values very close to but either side of that.  Whether that's actually an issue I'm not sure.

mborland:
@jzmaddock Any other changes for this one, or good to merge?

Issue #982, opened 2023-04-28
hypergeometric_1F1 with non-positive parameters is wrong


StatsMan53:
hypergeometric_1F1(-2, -4, 2.5) returns 0.3609023, when the correct result is 2.770833 = 2+37/48
hypergeometric_1F1(-2, -4, -2.5) returns 4.39669, when the correct result is 0.2708333 = 37/48

In general, the result of hypergeometric_1F1(a,b,x) is suspect when b<=a<-1 are non-positive integers.  The code introduces a point discontinuity under these conditions, so a workaround is to average
hypergeometric_1F1(a,b-epsilon,x) and hypergeometric_1F1(a,b+epsilon,x)

PR #983, opened 2023-05-02
Fix case where b is a negative integer and z is also negative.

jzmaddock:
Add tests etc.
Fixes: https://github.com/boostorg/math/issues/982.


jzmaddock:
Failures are network timeouts, merging...

Issue #992, opened 2023-06-06
float128 with gcc master


sloriot:
Hi,

with gcc@ce2188e4320cbb46d6246bd3f478ba20440c62f3, I have the following error during compilation:

```
boost_1_82_0/include/boost/math/cstdfloat/cstdfloat_limits.hpp:46:13: error: redefinition of ‘class std::numeric_limits<__float128>’
   46 |       class numeric_limits<boost::math::cstdfloat::detail::float_internal128_t>
```

I guess it has to do with `BOOST_CSTDFLOAT_HAS_INTERNAL_FLOAT128_T` being not correctly defined after gcc's update or something like that. I only tested boost 1.82 (sorry if it has already been fixed in HEAD).

cc @afabri

sloriot:
minimal example:
```
#include <boost/math/cstdfloat/cstdfloat_limits.hpp>

int main()
{}
```

mborland:
> minimal example:
> 
> ```
> #include <boost/math/cstdfloat/cstdfloat_limits.hpp>
> 
> int main()
> {}
> ```

I don't see it with GCC 13.1 on godbolt: https://godbolt.org/z/h39GTMz16.

sloriot:
that's gcc master targeting gcc 14

mborland:
> that's gcc master targeting gcc 14

That's fun. Thanks for the heads up.

afabri:
x86-64 gcc (trunk)   in godbolt, right?

mborland:
> x86-64 gcc (trunk) in godbolt, right?

Yes, and it does reproduce the error.

PR #993, opened 2023-06-07
Disable numeric_limits specialization for GCC-14

mborland:
Closes: #992

### DIFF
--- a/include/boost/math/cstdfloat/cstdfloat_limits.hpp
+++ b/include/boost/math/cstdfloat/cstdfloat_limits.hpp
@@ -24,7 +24,7 @@
 #pragma GCC system_header
 #endif
 
-  #if defined(BOOST_CSTDFLOAT_HAS_INTERNAL_FLOAT128_T) && defined(BOOST_MATH_USE_FLOAT128) && !defined(BOOST_CSTDFLOAT_NO_LIBQUADMATH_SUPPORT)
+  #if defined(BOOST_CSTDFLOAT_HAS_INTERNAL_FLOAT128_T) && defined(BOOST_MATH_USE_FLOAT128) && !defined(BOOST_CSTDFLOAT_NO_LIBQUADMATH_SUPPORT) && (!defined(__GNUC__) || (defined(__GNUC__) && __GNUC__ < 14))
 
     #include <limits>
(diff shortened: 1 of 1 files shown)


jzmaddock:
Just checking.... does libstdc++3 <limits> encase `numeric_limits<__float128>` inside a #if?  They often do, in which case it might be better to check for whatever internal macro they are using rather than assuming gcc-14 unconditionally supports this.

mborland:
It does have an `#if` block, and the diff [is here](https://gcc.gnu.org/git/?p=gcc.git;a=blobdiff;f=libstdc%2B%2B-v3/include/std/limits;h=5f341e63b931693022d847c17baff3214dee3269;hp=8bafd6fb9724aefe43cdf7a840dac2bb8c6d7e73;hb=f150a084e25;hpb=51cf0b3949b88b5e622872e4dac019ceb27400e1.). The two macros they use are already wrapped into defining `BOOST_MATH_USE_FLOAT128`.

jzmaddock:
Ah, nothing useful there then, this all looks good to me then.

include/boost/math/cstdfloat/cstdfloat_limits.hpp
```diff
@@ -24,7 +24,7 @@
 #pragma GCC system_header
 #endif
 
-  #if defined(BOOST_CSTDFLOAT_HAS_INTERNAL_FLOAT128_T) && defined(BOOST_MATH_USE_FLOAT128) && !defined(BOOST_CSTDFLOAT_NO_LIBQUADMATH_SUPPORT)
+  #if defined(BOOST_CSTDFLOAT_HAS_INTERNAL_FLOAT128_T) && defined(BOOST_MATH_USE_FLOAT128) && !defined(BOOST_CSTDFLOAT_NO_LIBQUADMATH_SUPPORT) && (!defined(__GNUC__) || (defined(__GNUC__) && __GNUC__ < 14))
```
jwakely, line 27: This check seems wrong. The new `numeric_limits<__float128>` specialization will be enabled when using new libstdc++ headers with Clang, which does not define `__GNUC__` to 14 (it can be told to use any value, but by default it defines it to 4).

PR #995, opened 2023-06-15
Small github actions updates

sdarwin:
Removes a warning about `apt`, use `apt-get` . Fixes an error that appeared when testing on self-hosted runners, the package `libtbb-dev` was missing for one job.


mborland (approved):
Thanks Sam

PR #997, opened 2023-06-16
median_absolute_deviation bug fix for non-zero center

rasmushenningsson:
This fixes a bug in `median_absolute_deviation` causing incorrect answers when a non-zero center is used. (Note that the default center is the median, not zero, so this bug should affect most users of `median_absolute_deviation`.)

This pull request also provides unit tests for the non-zero center case.


mborland (approved):
LGTM. Thanks for this. I approved the CI run; you can ignore the CircleCI failure.

rasmushenningsson:
We all make mistakes. :slightly_smiling_face: 
Thanks for the quick responses and all your work on boost::math!

PR #1001, opened 2023-07-24
Use Dijkstra identities for 0(0th) Fibonacci convention

ryanelandt:
Dijkstra's note "In honour of Fibonacci" provides identities for the convention where 0 is the 1st Fibonacci number. Boost Math uses the convention that 0 is the 0th Fibonacci number. This PR updates these identities to follow the 0 is the 0th convention.

### DIFF
--- a/include/boost/math/special_functions/fibonacci.hpp
+++ b/include/boost/math/special_functions/fibonacci.hpp
@@ -32,8 +32,8 @@ inline BOOST_CXX14_CONSTEXPR T unchecked_fibonacci(unsigned long long n) noexcep
     if (n <= 2) return n == 0 ? 0 : 1;
     /* 
      * This is based on the following identities by Dijkstra:
-     *   F(2*n)   = F(n)^2 + F(n+1)^2
-     *   F(2*n+1) = (2 * F(n) + F(n+1)) * F(n+1)
+     *   F(2*n-1) = F(n-1)^2 + F(n)^2
+     *   F(2*n)   = (2*F(n-1) + F(n)) * F(n)
      * The implementation is iterative and is unrolled version of trivial recursive implementation.
      */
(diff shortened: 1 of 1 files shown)


ryanelandt:
The identity depends on the convention.

Using the 0(0th) convention the first formula is:   F(2n-1) = F(n-1)^2 + F(n)^2
Using the 0(1st) convention the first formula is:   F(2n)   = F(n)^2 + F(n+1)^2

To see why, let's assign a letter to the Fibonacci numbers as follows: a=0, b=1, c=1, d=2, e=3, f=5, g=8, h=13, ... . Now consider the formula:
d^2 + e^2 = h

Using the **0(0th)** convention d=3rd, e=4th, h=7th, so the **0(0th)** formula is **right**, but the 0(1st) formula is wrong.
Using the **0(1st)** convention d=4th, e=5th, h=8th, so the 0(0th) formula is wrong, but the **0(1st)** formula is **right**.

ryanelandt:
The code is correct. But, for comments to be sensible, they need to explain what the code is doing. The original comments do not do this. Consider an analogous situation...

The commented function below converts temperature:
```
double kelvin_to_rankine(double temp_in) {
  // Uses the formula F = 9/5 * C + 32
  return temp_in * 1.8;
}
```
The comment explaining what this function does is not sensible because the formula is wrong, it's the formula for something else. Instead the commented function should read:
```
double kelvin_to_rankine(double temp_in) {
  // Uses the formula R = 9/5 * K as both are zero at absolute zero
  return temp_in * 1.8;
}
```
In the Fibonacci case here, the formulas in the comments need to match the convention being used otherwise they don't make any sense and are super confusing.

ryanelandt:
Awesome, thanks.

PR #1003, opened 2023-07-25
update beta error messages and checks

ryanelandt:
This is chunk 2 for #1000.

**Runtime errors**
One approach to run-time errors takes the form:
```
if(!(boost::math::isfinite)(a))
   return policies::raise_domain_error<T>(function, "a is bad because reason (got a=%1%).", a, pol);
```
When using this approach, it's understood that the error message should agree with the check that is actually being performed.

**This PR**
This PR fixes 6 places where the check and stated error message lacked agreement in the `beta.hpp` file. These incorrect error messages make debugging the Newton solver take longer than it would have otherwise.

**Ideas about future solutions**
Disagreement between the check and the error message are too easy to make. I think a better solution on a high level would to be able to use the BOOST_TEST types macros for run-time checks. I don't know if this kind of thing is possible. But either way, that's out of the scope of this PR.

### DIFF
--- a/include/boost/math/special_functions/beta.hpp
+++ b/include/boost/math/special_functions/beta.hpp
@@ -1021,17 +1021,14 @@ T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, bool normalised, T* p_de
    BOOST_MATH_ASSERT((p_derivative == 0) || normalised);
 
    if(!(boost::math::isfinite)(a))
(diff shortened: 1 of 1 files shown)


mborland (approved):

mborland:
LGTM. You are correct that there are probably other places where other incorrect error messages are lingering. Thanks.

ryanelandt:
Thanks!

PR #1004, opened 2023-07-25
only solve problems if have enough precision

ryanelandt:
In the `ibeta_inverse.hpp` file, there exists root-finding problems of the form:
```
log(x)  + a * log(1 - x) + t
```
This function has an extrema at 1 / (1 + a). 

**Newton solver finds dubious roots**

For extreme values of a, the extrema evaluates to 0 or 1 due to floating point limitations. For these cases, the objectives function evaluation _always has the same sign, and is never zero_. The Newton solver finds a root anyway.

**This PR**

- Uses `if (cross == 0 || cross == 1) { return cross; }` to return early for values that lack the numerical precision to construct a meaningful objective function
- Adds a `BOOST_MATH_ASSERT` to prevent the construction of root finding problems that are problematic.

### DIFF
--- a/include/boost/math/special_functions/detail/ibeta_inverse.hpp
+++ b/include/boost/math/special_functions/detail/ibeta_inverse.hpp
@@ -25,23 +25,16 @@ namespace boost{ namespace math{ namespace detail{
 template <class T>
 struct temme_root_finder
 {
-   temme_root_finder(const T t_, const T a_) : t(t_), a(a_) {}
+   temme_root_finder(const T t_, const T a_) : t(t_), a(a_) {
+      const T x_extrema = 1 / (1 + a);
+      BOOST_MATH_ASSERT(0 < x_extrema && x_extrema < 1);
+   }
 
    boost::math::tuple<T, T> operator()(T x)
    {
       BOOST_MATH_STD_USING // ADL of std names
 
       T y = 1 - x;
-      if(y == 0)
-      {
-         T big = tools::max_value<T>() / 4;
-         return boost::math::make_tuple(static_cast<T>(-big), static_cast<T>(-big));
-      }
-      if(x == 0)
-      {
-         T big = tools::max_value<T>() / 4;
-         return boost::math::make_tuple(static_cast<T>(-big), big);
-      }
       T f = log(x) + a * log(y) + t;
       T f1 = (1 / x) - (a / (y));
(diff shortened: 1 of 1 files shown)


mborland:
I'll look into the CI system to see what's going on with the TSAN failures across all of your PRs, but it is clearly unrelated.

include/boost/math/special_functions/detail/ibeta_inverse.hpp
```diff
@@ -25,23 +25,16 @@ namespace boost{ namespace math{ namespace detail{
...
+      const T x_extrema = 1 / (1 + a);
+      BOOST_MATH_ASSERT(0 < x_extrema && x_extrema < 1);
+   }
 
    boost::math::tuple<T, T> operator()(T x)
    {
       BOOST_MATH_STD_USING // ADL of std names
 
       T y = 1 - x;
-      if(y == 0)
```
jzmaddock, line 35: I'm not sure we can guarantee that this code is no longer needed?

It's purpose was to return "sane" values that the root finder can still use, even in the case that T has no infinity causing the log functions below to "explode".

ryanelandt: The purpose of this code appeared to be to avoid domain errors in the `log` functions when evaluating `x` values less than 0 or greater than `. In #1002, the upper Kolmogorov quantile search bound was changed to 1 from `boost::math::tools::max_value<RealType>()` so it doesn't appear to be necessary to be. Removing it doesn't appear to cause test failures.

jzmaddock: Trivial test case is `ibetac_inv(1e50, 10, 1./10)` which is from test_ibeta_inv.hpp:284 and does cause the test to fail here with msvc.

ryanelandt: This case never makes it to the Newton Solver because as it gets special cased out. Are you sure the branch you are testing is up to date?

https://github.com/boostorg/math/blob/e682c1be3326a7b4ead43ff1c50149a4ce226e2e/include/boost/math/special_functions/detail/ibeta_inverse.hpp#L407-L408

jzmaddock: You're right - my bad!

PR #1007, opened 2023-07-31
Fix inverse_discrete_quantile for large guess

Yuhta:
If `guess` passed to `inverse_discrete_quantile` cannot be represented as floating point number, it is possible that `guess + 1` or `guess - 1` does not change the value at all and we are stuck in infinite loop inside `round_to_floor` or `round_to_ceil`.  Fix this by increase/decrease more than 1 in these cases.

Example code to reproduce this:
```c++
boost::math::binomial_distribution<> dist(9079765771874083840, 0.561815);
boost::math::quantile(dist, 0.0365346);
```

### DIFF
--- a/test/test_binomial.cpp
+++ b/test/test_binomial.cpp
@@ -716,6 +716,18 @@ void test_spots(RealType T)
 
    check_out_of_range<boost::math::binomial_distribution<RealType> >(1, 1); // (All) valid constructor parameter values.
 
+   // TODO: Generic ibeta_power_terms has accuracy issue when long
+   // double is not precise enough, causing overflow in this case.
+   if(!std::is_same<RealType, real_concept>::value ||
+      sizeof(boost::math::concepts::real_concept_base_type) > 8)
+   {
+      using namespace boost::math::policies;
+      typedef policy<discrete_quantile<integer_round_outwards> > Policy;
+      binomial_distribution<RealType, Policy> dist(9079765771874083840, 0.561815);
+      // Accuracy is not too important here; the main purpose is to
+      // make sure it is not stuck.
+      BOOST_CHECK_CLOSE(quantile(dist, 0.0365346), 5101148604445670400, 1e12);
+   }
 
 } // template <class RealType>void test_spots(RealType)
 
--- a/include/boost/math/distributions/detail/inv_discrete_quantile.hpp
+++ b/include/boost/math/distributions/detail/inv_discrete_quantile.hpp
@@ -302,15 +302,13 @@ inline typename Dist::value_type round_to_floor(const Dist& d, typename Dist::va
    //
    while(result != 0)
    {
-      cc = result - 1;
+      cc = floor(float_prior(result));
       if(cc < support(d).first)
          break;
       pp = c ? cdf(complement(d, cc)) : cdf(d, cc);
-      if(pp == p)
-         result = cc;
-      else if(c ? pp > p : pp < p)
+      if(c ? pp > p : pp < p)
          break;
-      result -= 1;
+      result = cc;
    }
 
    return result;
@@ -336,15 +334,13 @@ inline typename Dist::value_type round_to_ceil(const Dist& d, typename Dist::val
    //
    while(true)
    {
-      cc = result + 1;
+      cc = ceil(float_next(result));
       if(cc > support(d).second)
          break;
       pp = c ? cdf(complement(d, cc)) : cdf(d, cc);
-      if(pp == p)
-         result = cc;
-      else if(c ? pp < p : pp > p)
+      if(c ? pp < p : pp > p)
          break;
-      result += 1;
+      result = cc;
    }
 
    return result;


Yuhta:
I can reproduce the issue with a standalone executable built with ` g++ test.cc -I$PWD/include -std=c++17 -g -DBOOST_MATH_STANDALONE`.  However, when I try to put the test case into `test_binomial.cpp`, the original version just passed without stuck.  Is there any compiler flags that make it behave differently?

test/test_binomial.cpp
```diff
@@ -716,6 +716,19 @@ void test_spots(RealType T)
...
+#ifdef TEST_FLOAT
+     RealType expected = 5101149012694663168;
+#endif
+#ifdef TEST_DOUBLE
+     RealType expected = 5101148604445670400;
+#endif
+#ifdef TEST_LDOUBLE
+     RealType expected = 5101148604445669693;
+#endif
+     BOOST_CHECK_EQUAL(quantile(dist, 0.0365346), expected);
```
Yuhta, line 730: I can reproduce the issue with a standalone executable built with ` g++ test.cc -I$PWD/include -std=c++17 -g -DBOOST_MATH_STANDALONE`.  However, when I try to put the test case into `test_binomial.cpp`, the original version just passed without stuck.  Is there any compiler flags that make it behave differently?

Yuhta: I see it's `BOOST_MATH_DISCRETE_QUANTILE_POLICY`.  How can I test under the default value?

mborland: The default is `#define BOOST_MATH_DISCRETE_QUANTILE_POLICY integer_round_outwards`, but at the top of this file it is defined as `real`. You can pass a different policy to the distribution as a template parameter.

test/test_binomial.cpp
```diff
@@ -716,6 +716,14 @@ void test_spots(RealType T)
 
    check_out_of_range<boost::math::binomial_distribution<RealType> >(1, 1); // (All) valid constructor parameter values.
 
+  {
+     using namespace boost::math::policies;
+     typedef policy<discrete_quantile<integer_round_outwards> > Policy;
+     binomial_distribution<RealType, Policy> dist(9079765771874083840, 0.561815);
+     // Accuracy is not too important here; the main purpose is to make
+     // sure it is not stuck.
+     BOOST_CHECK_CLOSE(quantile(dist, 0.0365346), 5101148604445670400, 1e12);
```
mborland, line 725: On MSVC platforms you are getting an overflow here: https://github.com/boostorg/math/actions/runs/5721216709/job/15503232562#step:11:2377. `boost::math::concepts::real_concept` is a long double wrapped in a class to check conceptual requirements for support of user defined types.

mborland (changes requested):
Outside of the overflow this makes sense to me. We have been dealing with non-representable numbers in other places recently too: https://github.com/boostorg/math/pull/968

test/test_binomial.cpp
Yuhta: @mborland It's an unrelated bug in https://github.com/boostorg/math/blob/e682c1be3326a7b4ead43ff1c50149a4ce226e2e/include/boost/math/special_functions/beta.hpp#L489, more specifically `log(x * c / a) + b * log(y * c / b))` is less accurate on MSVC because `long double` is just `double` there.

Do you think it's acceptable to skip `real_concept` test on MSVC for this test case?

mborland: I ran you branch locally on an M1 Mac (double = long double) and that works fine. I should have time this week to investigate why MSVC wants to be different.

Yuhta: @mborland Note that in the newest version I skip the test for `real_concept` if `long double` is not precise enough.  I can reproduce the overflow on Linux if I force `real_concept_base_type` to be `double`.

yzhang1991:
@mborland Gentle ping :)

mborland (approved):
Sorry about having this sit. Once the CI cycles I can merge it in. Beta is having other overflows documented here: https://github.com/boostorg/math/pull/1009. Like your test case it takes some extreme inputs to hit these so I am not overly concerned.

test/test_binomial.cpp
```diff
@@ -716,6 +716,18 @@ void test_spots(RealType T)
 
    check_out_of_range<boost::math::binomial_distribution<RealType> >(1, 1); // (All) valid constructor parameter values.
 
+   // TODO: Generic ibeta_power_terms has accuracy issue when long
+   // double is not precise enough, causing overflow in this case.
+   if(!std::is_same<RealType, real_concept>::value ||
```
jzmaddock, line 721: What happens here when T is a double, does that still work?  Ideally we shouldn't be having spurious internal overflow as a) it breaks types with no infinity (this is potentially true for long double too) and b) it sets the hardware overflow flag which can trip up downstream consumers.

jzmaddock: The problem is in beta.hpp#478 onwards.  Don't see a trivial fix just yet :(

inv_distrete_quantile.hpp#150 also needs adjusting to `else if((adder != 0) && (a + adder != a))` to speed up this new test case.

Yuhta: With built-in `double` and `long double` it works fine because Lanczos approximation is enabled.

Yuhta:
@mborland It seems one CI job has environmental issue: https://drone.cpp.al/boostorg/math/1435/242/1
Can you rerun it or I need to push again?

mborland:
> @mborland It seems one CI job has environmental issue: https://drone.cpp.al/boostorg/math/1435/242/1
> 
> Can you rerun it or I need to push again?

I restarted it. The s390x runners can be finicky.

Yuhta:
@mborland It fails with similar thing but on a different job: https://drone.cpp.al/boostorg/math/1450/66/1

jzmaddock:
@Yuhta : sorry to revisit this... but this, or at least this new test case is causing failures on MacOS/Clang, and after some investigation, it seems that Clang is not so much _wrong_ to fail, as other compilers are _lucky_ to pass.

The issue is this: the binomial distribution for large parameters calls the incomplete Beta, and the incomplete Beta for parameters greater than about 1/epsilon essentially returns garbage.  There is no way around this that I can see - we go through all sorts sorts of contortions to try and avoid using logarithmic calculations and to preserve digits when we do - but in this case all the digits in the log calculation cancel out and we're left not even really knowing the order of magnitude of the result.  All the alternate methods of calculation I tried are essentially worse, with interval arithmetic yielding `[0, +INF]` as the bounds on the result.  Ouch.

So... what was the use case here?  And what would happen if instead returning garbage we just gave up and threw an `evaluation_error` for large parameters?  I'm not saying that that is the correct way to go, just can't think of anything better right now...

See also discussion inhttps://github.com/boostorg/math/pull/1125.

CC @mborland

Yuhta:
@jzmaddock We are ok with `evaluation_error`, before my fix it was stuck there forever, we just wanted to avoid that situation, the result in this case is not usable anyway.

jzmaddock:
Thanks @Yuhta , I've decided to just disable the test for apple/clang for now - it does return, but returns garbage.

PR #814, opened 2022-08-05
Faster floating point comparisons

mborland:

PR #1012, opened 2023-08-05
More efficient bisection for 1D Newton root finder

### LINKED ISSUE #1008: 1D Newton iteration count issue(s)
The documentation for `newton_raphson_iterate` says:
```
uintmax_t& max_iter
    An optional maximum number of iterations to perform.
    On exit, this is updated to the actual number of iterations performed. 
```
There are two issues relating to `max_iter`:
1. `max_iter` is copied to a variable `count`. After each Newton iteration, one is subtracted from `count`. Because of the `do {...} while (count && ...)` structure of the Newton iteration, if `max_iter = 0`, rollover occurs, making `max_iter` functionally infinite.
4. If `0 < max_iter`, running out of iterations exits the Newton loop as if the tolerance was met.

These two observations are demonstrated in the code below.

```c++
const auto fn = [](double x) { return std::make_pair(x * x - 3, 2 * x); };
std::uintmax_t iters;
std::cout << std::setprecision(16);

for (int count = 0; count < 10; count++) {
   iters = count;
   const double x0 = boost::math::tools::newton_raphson_iterate(fn, 10.0, 0.0, 100.0, 52, iters);
   std::cout << "max_iters: " << count << " -- used_iters: " << iters << " -- x: " << x0 << std::endl;
}
```

```
max_iters: 0 -- used_iters: 8 -- x: 1.732050807568877
max_iters: 1 -- used_iters: 1 -- x: 5.15
max_iters: 2 -- used_iters: 2 -- x: 2.86626213592233
max_iters: 3 -- used_iters: 3 -- x: 1.956460731776899
max_iters: 4 -- used_iters: 4 -- x: 1.74492093914502
max_iters: 5 -- used_iters: 5 -- x: 1.732098271119538
max_iters: 6 -- used_iters: 6 -- x: 1.732050808219183
max_iters: 7 -- used_iters: 7 -- x: 1.732050807568877
max_iters: 8 -- used_iters: 8 -- x: 1.732050807568877
max_iters: 9 -- used_iters: 8 -- x: 1.732050807568877
```

ryanelandt:
This is chunk #4 for #1000. Also closes #1008.

This PR adds tools that enable efficient bisection for the 1D Newton root finder.

### The Situation

The 1D Newton root finder has a bisection fallback mechanism. If one of the bounds is large (i.e., `std::numeric_limits::max()`), naively bisecting the bounds is slow. For example, for type `double`, bisecting from `1.79769e+308` to `1` takes `1024` iterations. The situation is even worse for types with more precision. When the Newton solver runs out of (e.g., 200) iterations during bisection, the solver declares success even if far from a root. Again, see issue #1008.

### Functionality

What's the maximum number of bisections needed to root find a `double`? A `double` is 64 bits. So the maximum number of iterations needed is 64. Efficient bisection is bisection in bit space. An interesting consequence of bisection in bitspace, is that infinity is a bisectable bound.

For types that do not conform to the standard (i.e., IEEE 754), an approximate version of the bisection algorithm above is used. The approximate version supports bisection with infinity if the type is specialized (i.e., if `std::numeric_limits<T>::is_specialized = true`).

### Fun Facts

The mean of `0` and `infinity` is:
- `infinity` (arithmetic mean)
- `NaN` (geometric mean)
- 1.5 (bitspace mean for `double` and `float`)
- 1.0 (bitspace mean for `long double` using the approximate algorithm)

### DIFF
--- a/include/boost/math/tools/ieee754_linear.hpp
+++ b/include/boost/math/tools/ieee754_linear.hpp
@@ -1,0 +1,366 @@
+#ifndef BOOST_MATH_TOOLS_IEEE754_BITSPACE_HPP
+#define BOOST_MATH_TOOLS_IEEE754_BITSPACE_HPP
+
+#ifdef _MSC_VER
+#pragma once
+#endif
+
+#include <type_traits>
+
+
+#ifdef BOOST_HAS_FLOAT128
+#include <boost/multiprecision/float128.hpp>
+#endif
+
+
+namespace boost {
+namespace math {
+namespace tools {
+namespace detail {
+namespace ieee754_linear {
+
+   // The `IsFloat32` class contains a static constexpr method `value()` that
+   // returns true if the type T is a 32 bit floating point type. This duck type
+   // test for 32 bit float types returns true for the following types:
+   //   - `float`
+   //   - `_Float32` (NOTE: not a type alias for `float`)
+   //   - `std_real_concept` when emulating a 32 bit float with EMULATE32
+   //   - other types that seem to be 32 bit floats
+   class IsFloat32 {
+   public:
+      template <typename T>
+      static constexpr bool value() {
+         return std::numeric_limits<T>::is_iec559 &&
+                std::numeric_limits<T>::radix == 2 &&
+                std::numeric_limits<T>::digits == 24 &&  // Mantissa has 23 bits + 1 implicit bit
+                sizeof(T) == 4;
+      }
+   };
+
+   // The `IsFloat64` class contains a static constexpr method `value()` that
+   // returns true if the type T is a 64 bit floating point type. This duck type
+   // test for 64 bit float types returns true for the following types:
+   //   - `double`
+   //   - `_Float64` (NOTE: not a type alias for `double`)
+   //   - `std_real_concept` when emulating a 64 bit float with EMULATE64
+   //   - other types that seem to be 64 bit floats
+   class IsFloat64 {
+   public:
+      template <typename T>
+      static constexpr bool value() {
+         return std::numeric_limits<T>::is_iec559 &&
+                std::numeric_limits<T>::radix == 2 &&
+                std::numeric_limits<T>::digits == 53 &&  // Mantissa has 52 bits + 1 implicit bit
+                sizeof(T) == 8;
+      }
+   };
+
+   // The `IsFloat128` class contains a static constexpr method `value()` that
+   // returns true if the type T is a 128 bit floating point type. This duck type
+   // test for 128 bit float types returns true for the following types:
+   //   - `__float128`
+   //   - `_Float128` (NOTE: not a type alias for `__float128`)
+   //   - `std_real_concept` when emulating a 128 bit float with EMULATE128
+   //   - other types that seem to be 128 bit floats
+   class IsFloat128 {
+   public:
+      template <typename T>
+      static constexpr bool value() {
+         return std::numeric_limits<T>::is_iec559 &&
+                std::numeric_limits<T>::radix == 2 &&
+                std::numeric_limits<T>::digits == 113 &&  // Mantissa has 112 bits + 1 implicit bit
+                sizeof(T) == 16;
+      }
+   };
+
+   // The `Layout_IEEE754Linear` class represents IEEE 754 floating point types
+   // for which increasing float values (with the same sign) have increasing 
+   // values in bit space. And for which there is a corresponding unsigned integer
+   // type U that has the same number of bits as T. Types that satisfy these
+   //  conditions include 32, 64, and 128 bit floats. The table below shows select
+   // numbers for `float` (single precision).
+   //
+   //   0            |  0 00000000 00000000000000000000000  |  positive zero
+   //   1.4013e-45   |  0 00000000 00000000000000000000001  |  std::numeric_limits<float>::denorm_min()
+   //   1.17549e-38  |  0 00000001 00000000000000000000000  |  std::numeric_limits<float>::min()
+   //   1.19209e-07  |  0 01101000 00000000000000000000000  |  std::numeric_limits<float>::epsilon()
+   //   1            |  0 01111111 00000000000000000000000  |  positive one
+   //   3.40282e+38  |  0 11111110 11111111111111111111111  |  std::numeric_limits<float>::max()
+   //   inf          |  0 11111111 00000000000000000000000  |  std::numeric_limits<float>::infinity()
+   //   nan          |  0 11111111 10000000000000000000000  |  std::numeric_limits<float>::quiet_NaN() 
+   //
+   // NOTE: the 80 bit `long double` is not "linear" due to the "integer part". See:
+   //       https://en.wikipedia.org/wiki/Extended_precision#x86_extended_precision_format
+   //
+   template <typename T, typename U>
+   class Layout_IEEE754Linear {
+      static_assert(std::numeric_limits<T>::is_iec559, "Type must be IEEE 754 floating point.");
+      static_assert(std::is_unsigned<U>::value, "U must be an unsigned integer type.");
+      static_assert(sizeof(T) == sizeof(U), "Type and uint size must be the same.");
+
+   public:
+      using type_float = T;  // The linear floating point type
+   };
+
+   // The `Layout_Unspecified` class represents floating point types for which
+   // are not supported by the `Layout_IEEE754Linear` class.
+   template <typename T>
+   class Layout_Unspecified {
+   public:
+      using type_float = T;  // The linear floating point type
+   };
+
+   // The `LayoutIdentifier` class identifies the layout type for a floating
+   // point type T. The layout type is one of the following:
+   //   - `Layout_IEEE754Linear<T, U>` for 32, 64, and 128 bit floats
+   //   - `Layout_Unspecified<T>` for other types
+   class LayoutIdentifier {
+   public:
+      // Layout: 32 bit linear
+      template <typename T>
+      static typename std::enable_if<IsFloat32::value<T>(), Layout_IEEE754Linear<T, std::uint32_t>>::type
+      get_layout() { return Layout_IEEE754Linear<T, std::uint32_t>(); }
+
+      // Layout: 64 bit linear
+      template <typename T>
+      static typename std::enable_if<IsFloat64::value<T>(), Layout_IEEE754Linear<T, std::uint64_t>>::type
+      get_layout() { return Layout_IEEE754Linear<T, std::uint64_t>(); }
+
+      // Layout: 128 bit linear
+#if defined(BOOST_HAS_INT128) && defined(BOOST_HAS_FLOAT128)
+      // NOTE: returns Layout_IEEE754Linear<__float128, ...> instead of Layout_IEEE754Linear<T, ...>
+      //       in order to work with boost::multiprecision::float128, which is a wrapper
+      //       around __float128.
+      template <typename T>
+      static typename std::enable_if<IsFloat128::value<T>(), Layout_IEEE754Linear<__float128, boost::uint128_type>>::type
+      get_layout() { return Layout_IEEE754Linear<__float128, boost::uint128_type>(); }
+
+      template <typename T>
+      static constexpr bool is_layout_nonlinear() {
+         return !IsFloat32::value<T>() &&
+                !IsFloat64::value<T>() &&
+                !IsFloat128::value<T>();
+      }
+#else
+      template <typename T>
+      static constexpr bool is_layout_nonlinear() {
+         return !IsFloat32::value<T>() &&
+                !IsFloat64::value<T>();
+      }
+#endif
+
+      // Layout: unspecified
+      template <typename T>
+      static typename std::enable_if<is_layout_nonlinear<T>(), Layout_Unspecified<T>>::type
+      get_layout() { return Layout_Unspecified<T>(); }
+   };
+
+   // Base class for the `BitsInflated` and `BitsDeflated` classes.
+   //
+   // This class stores the bit values of a floating point type `T` as a
+   // sign bit as a unsigned integer magnitude. For example, a floating
+   // point type with 50 discrete values (zero is counted twice) between
+   // -Inf and +Inf is shown below.
+   //
+   //    -24 -20 -16 -12 -8  -4   0   4   8   12  16  20  24
+   //     .   .   .   .   .   .   .   .   .   .   .   .   .
+   //     |-----------------------|-----------------------|
+   //   -Inf                      0                     +Inf
+   //
+   template <typename T, typename U>
+   class BitsFromZero: public Layout_IEEE754Linear<T, U> {
+   public:
+      bool sign_bit() const { return sign_bit_; }
+      const U& mag() const { return mag_; }
+      U& mag() { return mag_; }
+
+   protected:
+      BitsFromZero(const bool sign, const U mag) : sign_bit_(sign), mag_(mag) {}
+
+      BitsFromZero(const T x) { 
+         // The sign bit is 1, all other bits are 0
+         constexpr U bits_sign_mask = U(1) << (sizeof(U) * 8 - 1);
+
+         U bits_;
+         std::memcpy(&bits_, &x, sizeof(U));
+         sign_bit_ = bits_ & bits_sign_mask;
+         mag_ = bits_ & ~bits_sign_mask;
+      }
+
+      void flip_sign_bit() { sign_bit_ = !sign_bit_; }
+
+   private:
+      bool sign_bit_;
+      U mag_;
+   };
+
+   // The inflated bitspace representation of a floating point type `T`.
+   //
+   // This representation always includes denormal numbers, even if denorm
+   // support is turned off. For example, a floating point type with 50
+   // discrete values (zero is counted twice) between -Inf and +Inf is shown
+   // below with *** marking denormal numbers.
+   //
+   //    -24 -20 -16 -12 -8  -4   0   4   8   12  16  20  24
+   //     .   .   .   .   .   .   .   .   .   .   .   .   .
+   //     |-------------------|***|***|-------------------|
+   //   -Inf                      0                     +Inf
+   //
+   template <typename T, typename U>
+   class BitsInflated : public BitsFromZero<T, U> {
+   public:
+      BitsInflated(const T x) : BitsFromZero<T, U>(x) {}
+      BitsInflated(const bool sign, const U mag) : BitsFromZero<T, U>(sign, mag) {}
+      
+      T reinterpret_as_float() const {
+         T f_out;
+         std::memcpy(&f_out, &this->mag(), sizeof(T));
+         return this->sign_bit() ? -f_out : f_out;
+      }
+
+      void divide_by_2() { this->mag() >>= 1; }
+   };
+
+   // The deflated bitspace representation of a floating point type `T`. 
+   //
+   // Denorm Support ON:
+   //   This representation is identical to the inflated representation.
+   //
+   // Denorm Support OFF:
+   //    This representation shifts the bitspace representation toward `0`
+   //    to remove gaps in the bitspace caused by denormal numbers. For 
+   //    example, consider a floating point type with 50 discrete values
+   //    (zero is counted twice) between -Inf and +Inf is shown below with
+   //    *** marking denormal numbers shown below.
+   // 
+   //                  Inflated representation:
+   //    -24 -20 -16 -12 -8  -4   0   4   8   12  16  20  24
+   //     .   .   .   .   .   .   .   .   .   .   .   .   .
+   //     |-------------------|***|***|-------------------|
+   //   -Inf                      0                     +Inf
+   // _________________________________________________________________
+   //
+   //                  Deflated representation:
+   //    -24 -20 -16 -12 -8  -4   0   4   8   12  16  20  24
+   //     .   .   .   .   .   .   .   .   .   .   .   .   .
+   //        |-------------------|||-------------------|
+   //      -Inf                   0                  +Inf
+   //
+   template <typename T, typename U>
+   class BitsDeflated : public BitsFromZero<T, U> {
+   public:
+      BitsDeflated(const bool sign, const U mag) : BitsFromZero<T, U>(sign, mag) {}
+
+      T static_cast_int_value_to_float() const {
+         T mag_float = static_cast<T>(this->mag());
+         return this->sign_bit() ? -mag_float : mag_float;
+      }
+
+      // Move over `n` in bitspace
+      BitsDeflated<T,U> operator+(int n) const {
+         return BitsDeflated<T,U>(this->sign_bit(), this->mag() + n);
+      }
+
+      BitsDeflated<T,U> operator-(const BitsDeflated<T,U>& y) const {
+         auto y_copy = y;
+         y_copy.flip_sign_bit();
+         return *this + y_copy;
+      }
+
+      BitsDeflated<T,U> operator+(const BitsDeflated<T,U>& y) const {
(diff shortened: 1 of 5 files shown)


test/test_roots.cpp
```diff
@@ -649,11 +651,121 @@ void test_failures()
...
+         const T x_lo = (std::min)(x_i, x_j);
+         const T x_hi = (std::max)(x_i, x_j);
+
+         BOOST_CHECK_LE(x_lo, x_mid_ij);
+         BOOST_CHECK_LE(x_mid_ij, x_hi);
+      }
+   }
+}
+
+void test_bisect_all_cases() {
```
mborland, line 723: Can you add `std::float32_t` and `std::float64_t` from C++23's `<stdfloat>`. You can test `__has_include(<stdfloat>)` and then the macros that each of the types define. We have g++13 in the CI system to cover those.

test/test_roots.cpp
```diff
@@ -649,11 +651,121 @@ void test_failures()
...
+}
+
 BOOST_AUTO_TEST_CASE( test_main )
 {
 
    test_beta(0.1, "double");
 
+   test_count();
+   test_bisect_all_cases();
+#if 0
```
mborland, line 768: Do the following tests fail with this new method?

mborland (changes requested):

include/boost/math/tools/roots.hpp
```diff
@@ -213,6 +213,245 @@ inline std::pair<T, T> bisect(F f, T min, T max, Tol tol) noexcept(policies::is_
...
+
+      // Convert bits to float
+      static T uint_to_float(U bits) {
+         T x;
+         std::memcpy(&x, &bits, sizeof(T));
+         return x;
+      }
+
+   public:
+      static T solve(T x, T X) {
```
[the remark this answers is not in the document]
ryanelandt: I changed `X` to `y`.

include/boost/math/tools/roots.hpp
```diff
@@ -213,6 +213,245 @@ inline std::pair<T, T> bisect(F f, T min, T max, Tol tol) noexcept(policies::is_
...
+         return uint_to_float(bits_mag) * sign_X;
+      }
+   };  // class Midpoint754
+
+
+   template <typename T>
+   class MidpointNon754 {
+   private:
+      static_assert(!std::is_same<T, float>::value, "Need to use Midpoint754 solver when T is float");
+      static_assert(!std::is_same<T, double>::value, "Need to use Midpoint754 solver when T is double");
```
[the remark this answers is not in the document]
ryanelandt: I added an explanation as to why `long double` does not use the `Midpoint754` solver.

include/boost/math/tools/roots.hpp
```diff
@@ -213,6 +213,245 @@ inline std::pair<T, T> bisect(F f, T min, T max, Tol tol) noexcept(policies::is_
...
+         return -do_solve(-x, -X);
+      }
+
+   private:      
+      struct EqZero {
+         EqZero(T x) { BOOST_MATH_ASSERT(x == 0 && "x must be zero."); }
+      };
+
+      struct EqInf {
+         EqInf(T x) { BOOST_MATH_ASSERT(x == static_cast<T>(std::numeric_limits<double>::infinity()) && "x must be infinity."); }
```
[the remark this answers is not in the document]
ryanelandt: This awkwardness was there to support types (e.g., boost::math::concepts::real_concept) for which infinity exists, but std::numeric_limits<T>::infinity() returns 0. I switched to using `boost::math::isfinite` when checking for infinity at most places.

test/test_roots.cpp
ryanelandt: I think I was able to do this. Let me know if I didn't do the right thing.

test/test_roots.cpp
ryanelandt: I accidentally left this in. Good catch.

ryanelandt:
@mborland thank you for the review. Please let me know if there are additional changes.

include/boost/math/tools/roots.hpp
```diff
@@ -213,6 +213,245 @@ inline std::pair<T, T> bisect(F f, T min, T max, Tol tol) noexcept(policies::is_
...
+      }
+      template <typename U = T>
+      static typename std::enable_if<!std::numeric_limits<U>::is_specialized, T>::type
+      do_solve(PosFinite x, EqInf X) {
+          BOOST_MATH_ASSERT(false && "infinite bounds support requires specialization.");
+          return static_cast<T>(std::numeric_limits<T>::signaling_NaN());
+      }
+
+      template <typename U = T>
+      static typename std::enable_if<std::numeric_limits<U>::is_specialized, U>::type
```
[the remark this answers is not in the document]
ryanelandt: We want to be able to bisect with denormals because the solution to the root finding problem could be a denormal number.

Reading the [depreciation document](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2614r2.pdf), it seems like the depreciation of `has_denorm` has a lot to do with it being a `static constexpr`. This is problematic because the actual behavior can change at runtime with e.g., `_MM_SET_FLUSH_ZERO_MODE`.

I think we can actually delete `std::numeric_limits::min()` because `denorm_min` returns `numeric_limits::min()` if denormals are off. Does that seen sensible?

include/boost/math/tools/roots.hpp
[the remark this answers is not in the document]
mborland: A guess there are a couple cases:

1. On some systems (e.g. Windows and MacOS) `long doubles` are just aliases to `double`. When long doubles are 64-bits we could just cast to double and use `Midpoint754` to speed things up.

2. On x86_64 linux systems you have the 80-bit `long double`. Normally you have `unsigned __int128` but you'd have to shift the bits around to cast it back properly. Likely not worth the magic required.

3. On systems with IEEE 128-bit long doubles (e.g. ARM and S390x) you have access to `unsigned __int128` which should generally work. We have both of the aforementioned architectures running natively in the CI system.

[Here](https://github.com/boostorg/math/blob/develop/include/boost/math/ccmath/signbit.hpp#L94) is where I have defined macros before for determining the size of long double.

include/boost/math/tools/roots.hpp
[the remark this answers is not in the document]
mborland: > 
> I think we can actually delete `std::numeric_limits::min()` because `denorm_min` returns `numeric_limits::min()` if denormals are off. Does that seen sensible?

I think so. Since you are already check for `is_specialized` you should avoid ever getting 0 instead of a finite minimum value.

include/boost/math/tools/roots.hpp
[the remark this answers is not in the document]
ryanelandt: > On some systems (e.g. Windows and MacOS) `long doubles` are just aliases to `double`. When long doubles are 64-bits we could just cast to double and use `Midpoint754` to speed things up.

These systems will use the specialization for `double`.
 
> On x86_64 linux systems you have the 80-bit `long double`. Normally you have `unsigned __int128` but you'd have to shift the bits around to cast it back properly. Likely not worth the magic required.

I tried to make this work. Starting with `numeric_limits<long double>::max()` and adding 1 bit gives `NaN`, not `Inf` as it does for `float`, `double` and `__float128`. It's just not meant to be.

> On systems with IEEE 128-bit long doubles (e.g. ARM and S390x) you have access to `unsigned __int128` which should generally work. We have both of the aforementioned architectures running natively in the CI system.

I got this to work for both `_float128` and `boost::multiprecision::float128` (wrapper for `__float128`). The implementation does not require `libquadmath`.

include/boost/math/tools/roots.hpp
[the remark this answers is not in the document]
ryanelandt: > I think so. Since you are already check for `is_specialized` you should avoid ever getting 0 instead of a finite minimum value.

https://github.com/boostorg/multiprecision/pull/562

ryanelandt:
I added IEEE754 bisection support for `__float128`. I also updated the reason why the 80 bit `long double` cannot use the IEEE754 solver. Please take another look when you can.

ryanelandt:
I realized that the code overlap between what is needed for find the midpoint between two IEEE754 floats and the code needed to find the fast float distance between two IEEE754 floats is high. I'm in the process of refactoring this functionality into a few classes that when operated on allow the following operations to be performed in a few lines of code: `midpoint`, `float_distance`, and `float_advance`. I'm putting this PR on hold until I can get this done.

ryanelandt:
The introduced tools in `ieee754_linear` lower the programming effort required to perform bit-themed operations on IEEE754 types that include: `float`, `double`, and `__float128`. Runtime-detected value of denorm support are used in calculations. These tools were used to calculate the midpoint in bitspace for the above mentioned types. The tools can be applied in a slightly different way to calculate fast float distance:
```c++
      template <typename T, typename U>
      static T fast_float_distance(T x, T y, boost::math::tools::detail::ieee754_linear::Layout_IEEE754Linear<T, U>) {
         using boost::math::tools::detail::ieee754_linear::BitsInflated;
         using boost::math::tools::detail::ieee754_linear::Denormflator;
         const auto df = Denormflator<T, U>();  // Calculates if `has_denorm` is true at this instant

         // Step 1 -- Calculate inflated representations
         const BitsInflated<T, U> y_inflated(y);
         const BitsInflated<T, U> x_inflated(x);

         // Step 2 -- Calculate deflated representations
         const auto y_def = df.deflate(y_inflated);
         const auto x_def = df.deflate(x_inflated);

         // Step 3 -- Subtract the two deflated representations
         const auto xy_def_distance = (y_def - x_def)

         // Step 4 -- The distance between the deflated points is the fast float distance
         return xy_def_distance.static_cast_int_value_to_float();
      }
```
Or it can be used to perform the float advance operation:
```c++
      template <typename T, typename U>
      static T float_advance(T x, int n, boost::math::tools::detail::ieee754_linear::Layout_IEEE754Linear<T, U>) {
         using boost::math::tools::detail::ieee754_linear::BitsInflated;
         using boost::math::tools::detail::ieee754_linear::Denormflator;
         const auto df = Denormflator<T, U>();  // Calculates if `has_denorm` is true at this instant

         // Step 1 -- Calculate inflated representation
         const BitsInflated<T, U> x_inflated(x);

         // Step 2 -- Calculate deflated representation
         const auto x_def = df.deflate(x_inflated);

         // Step 3 -- Move over n bits
         const auto x_plus_n_def = x_def + n;

         // Step 4 -- Inflate
         const auto x_plus_n = df.inflate(x_plus_n_def);

         // Step 5 -- Convert to float
         return x_plus_n.reinterpret_as_float();
      }
```

mborland:
If you are looking for a different quick way to calculate float distance @NAThompson showed me this trick: https://github.com/boostorg/math/pull/814/files#diff-45f56077fb4aa15ca1aca5025321e5c275fd20b3e161585a25442eaa2c7952a3R724

ryanelandt:
The same idea of viewing floats in bitspace is used here. Seeing that trick used in #814 inspired this refactor. In addition, this  implementation attempts to:
- augment calculations for denorm support as detected at runtime
- operate on numbers with different signs
- use template parameters to allow it to work for a variety of types (e.g., `std::float32_t` and `std::float64_t`, or even `boost::math::std_real_concept` if it's set up to emulate a float type).

include/boost/math/tools/ieee754_linear.hpp
```diff
@@ -0,0 +1,371 @@
...
+
+#ifdef _MSC_VER
+#pragma once
+#endif
+
+#include <type_traits>
+
+
+#ifdef BOOST_HAS_FLOAT128
+#include <boost/multiprecision/float128.hpp>
```
jzmaddock, line 17: I think this breaks standalone mode as BOOST_HAS_FLOAR128 may be set but the multiprecision headers will not be available.  We also try to avoid mutual/cyclic dependencies like these.  What is the include actually used for?

ryanelandt: Good catch with the dependency. I've removed this include and replaced it with the following includes:
```c++
#include <boost/math/tools/config.hpp>  // For BOOST_MATH_FLOAT128_TYPE
#include <boost/cstdfloat.hpp>  // For numeric_limits support for 128 bit floats
```
There may be a better way of including the needed files that I'm unaware of.

include/boost/math/tools/roots.hpp
```diff
@@ -18,12 +18,19 @@
...
 
 #include <boost/math/special_functions/sign.hpp>
 #include <boost/math/special_functions/next.hpp>
+#include <boost/math/tools/ieee754_linear.hpp>
 #include <boost/math/tools/toms748_solve.hpp>
 #include <boost/math/policies/error_handling.hpp>
 
+#ifdef BOOST_HAS_FLOAT128
+#include <boost/multiprecision/float128.hpp>
+using boost::multiprecision::float128;
```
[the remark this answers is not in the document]
ryanelandt: This header and using declaration was unneeded. Good catch.

include/boost/math/tools/ieee754_linear.hpp
jzmaddock: Unless you've changed something in that test, it uses nothing from multiprecision at all so the include should not be needed.

ryanelandt: As stated in a previous comment, I've removed the header `#include <boost/multiprecision/float128.hpp>` and replaced it with `#include <boost/cstdfloat.hpp>  // For numeric_limits support for 128 bit floats`. The `std_real_concept_check_128` test needs for `std::numeric_limits` specializations to be defined for the `BOOST_MATH_FLOAT128_TYPE`. If this header is removed, then test will **not** compile. (I tried this)

Why this occurs: the `std_real_concept_check_128` test makes the `std_real_concept` emulate a 128 bit floating point type. The tools in `ieee754_linear` recognize this as an IEEE 754 floating point type. The comment on line 133 goes on to say: 
```c++
// NOTE: returns Layout_IEEE754Linear<BOOST_MATH_FLOAT128_TYPE, ...> instead of Layout_IEEE754Linear<T, ...>
//       in order to work with non-trivial types that wrap trivial 128 bit floating point types.
```

jzmaddock: OK, I see the issue, but I think that we're looking at this the wrong way:

1) Not depending on boost/cstdfloat.hpp is a "good thing" because while that header is genuinely very useful, it's also rather fragile to changes in GCC releases.
2) I think the assumptions in `IsFloat128` (and maybe elsewhere) are wrong: just because a type has the correct size and properties, does NOT mean it's a trivial wrapper around a native type.  It could for example be a pointer (or two) to the actual implementation which just happens to have the same size.  mpfr_t is 24 bytes on my system so doesn't quite qualify, but if I'm reading it correctly, MPFR could be configured in a such a way that it was a 16 bye / 128 bit type which was set up to emulate a 128 bit float.  Another way of looking at this is that std_real_concept_type and the other concept checking types should be viewed as black boxes - their purpose is to emulate to some degree or another, some other random type about which we know nothing that the library may be instantiated with in the future.  You may be able to actually see their internals, but you really shouldn't look ;)

So I would change

```
   class IsFloat128 {
   public:
      template <typename T>
      static constexpr bool value() {
         return std::numeric_limits<T>::is_iec559 &&
                std::numeric_limits<T>::radix == 2 &&
                std::numeric_limits<T>::digits == 113 &&  // Mantissa has 112 bits + 1 implicit bit
                sizeof(T) == 16;
      }
   };
```

To maybe:

```
   class IsFloat128 {
   public:
      template <typename T>
      static constexpr bool value() {
         return std::numeric_limits<T>::is_iec559 &&
                std::numeric_limits<T>::radix == 2 &&
                std::numeric_limits<T>::digits == 113 &&  // Mantissa has 112 bits + 1 implicit bit
#ifdef BOOST_HAS_FLOAT128
                (std::is_floating_Point<T>::value || std::is_same<__float128, T>::value)
#else
                std::is_floating_Point<T>::value
#endif
      }
   };
```

Likewise for the other IsFloatXX classes.

Then change:

```
   template <typename T, typename U>
   class Layout_IEEE754Linear {
      static_assert(std::numeric_limits<T>::is_iec559, "Type must be IEEE 754 floating point.");
      static_assert(std::is_unsigned<U>::value, "U must be an unsigned integer type.");
      static_assert(sizeof(T) == sizeof(U), "Type and uint size must be the same.");

   public:
      using type_float = T;  // The linear floating point type
   };
```

To:

```
   template <typename T, typename U>
   class Layout_IEEE754Linear {
      static_assert(std::numeric_limits<T>::is_iec559 
#ifdef BOOST_HAS_FLOAT128
       || std::is_same<T, __float128>::value
#endif 
       , "Type must be IEEE 754 floating point.");
      static_assert(std::is_unsigned<U>::value, "U must be an unsigned integer type.");
      static_assert(sizeof(T) == sizeof(U), "Type and uint size must be the same.");

   public:
      using type_float = T;  // The linear floating point type
   };
```

Which would allow the header dependency to be removed as well.  I do appreciate it's less clean, but I suspect it's a safer option in the long run!

ryanelandt: > Not depending on boost/cstdfloat.hpp is a "good thing" because while that header is genuinely very useful, it's also rather fragile to changes in GCC releases.

Does the following change make things less fragile? Changing the include from:
```c++
```
to 
```c++
#if defined(BOOST_HAS_INT128) && defined(BOOST_HAS_FLOAT128)
#if !defined(BOOST_CSTDFLOAT_NO_LIBQUADMATH_LIMITS)
  #include <boost/math/cstdfloat/cstdfloat_limits.hpp>
#endif
#endif
```

>I think the assumptions in IsFloat128 (and maybe elsewhere) are wrong: just because a type has the correct size and properties, does NOT mean it's a trivial wrapper around a native type. It could for example be a pointer (or two) to the actual implementation which just happens to have the same size. [...] MPFR could be configured in a such a way [...] to emulate a 128 bit float. 

For the `value()` function of any of the `IsFloat` classes to return true, the following statement needs to be true:
```
std::numeric_limits<T>::is_iec559
```
The 32, 64, and 128 bit IEEE floats have a specific layout consisting of a sign bit, an exponent, and a fraction. A number type containing two pointers will not meet this standard. Skimming the MPFR codebase, this library is clearly set up with an awareness of IEEE 754 arithmetic. They don't appear to specialize numeric limits, and therefore don't set `std::numeric_limits<T>::is_iec559 = true`. I think the check that `std::numeric_limits<T>::is_iec559 = true` provides reasonable protection.

jzmaddock: Yup apologies, you're correct that `std::numeric_limits<T>::is_iec559` should provide a reasonable check given that that std defines a binary layout.

I would still prefer to avoid the cstdfloat includes if possible though - and we *know* that __float128 is an iec559 compatible type, so including it for that one case is overkill when we can just use `is_same<T, __float128>::value` ?

Other than that and one minor nit about <stdfloat> inclusion this looks good to go.

test/ieee754_linear_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <boost/multiprecision/cpp_bin_float.hpp>
+#include <boost/multiprecision/cpp_dec_float.hpp>
+#include <boost/multiprecision/cpp_complex.hpp>
+#include <boost/math/concepts/real_concept.hpp>
+
+#ifdef BOOST_HAS_FLOAT128
+#include <boost/multiprecision/float128.hpp>
+#endif
+
+#if __has_include(<stdfloat>)
```
jzmaddock, line 20: It's unfortunate, but __has_include is an insufficient check, because it may pass even when the translation mode is not C++23, so we will need to check the value of BOOST_CXX_VERSION as well (unfortunately __cplusplus is non-reliable for MSVC).  I yes, must get around to adding proper Boost.Config macros and tests for these new C++23 headers sometime!

ryanelandt: The include pattern:
```c++
#if __has_include(<stdfloat>)
#  include <stdfloat>
#endif
```
occurs in `include/boost/math/tools/promotion.hpp`, `include/boost/math/concepts/real_concept.hpp`, and about 30 test files including `test/jacobi_test.cpp`. I think it makes sense to allow it here for consistency. If required it should be fixed across the codebase in a separate PR.

jzmaddock: This came up here: https://github.com/boostorg/math/issues/1023 and the headers were fixed at that time.  I hadn't realised this had been sprinkled across all the tests as well :(

The only remaining header I can see which is incorrect is here: https://github.com/boostorg/math/blob/09d82da6e9884da86495279b331ff038dd1330b8/include/boost/math/ccmath/signbit.hpp#L28  @mborland ?

include/boost/math/tools/ieee754_linear.hpp
ryanelandt: I don't see a good way to do this. The code operates by duck typing a float type `T`. For example, something is determined to be a 32 bit binary floating point type based on the value of `IsFloat32::value<T>()`. Something is determined to be such a type if: 
```c++
std::numeric_limits<T>::is_iec559 &&
std::numeric_limits<T>::radix == 2 &&
std::numeric_limits<T>::digits == 24 &&  // Mantissa has 23 bits + 1 implicit bit
sizeof(T) == 4;
```
evaluates to true. If any of the four conditions above cannot be evaluated, then the code will fail to compile. The include is needed for it to compile, even if the method were to be unused.

----------

The alternative way of structuring this code is to manually match up binary floating point types with the appropriate methods, but this is messy and tends to miss valid use cases. `is_same<T, __float128>::value` might work on GCC, but what about an Intel compiler?

-----------

I'd like to avoid this include here too as it shouldn't be necessary. Perhaps the issue occurs here:
https://github.com/boostorg/math/blob/09d82da6e9884da86495279b331ff038dd1330b8/include/boost/math/tools/config.hpp#L399-L415
where a 128 bit floating point type is defined without defining `std::numeric_limits` for that type.

test/ieee754_linear_test.cpp
ryanelandt: It might be desirable for the pattern:
```c++
#if defined __has_include
#  if __cplusplus > 202002L || _MSVC_LANG > 202002L 
#    if __has_include (<X>)
#    include <X>
#    endif
#  endif
#endif
```
to me made into a macro similar to
```c++
INCLUDE_CPP23_HEADER(X)
```

jzmaddock: Yes, but you still need to be able to determine whether the header has been included or not (ie whether you can use those contents or not), so the normal method is:

```
#ifndef BOOST_NO_CXX23_STDFLOAT
#include <stdfloat>
#endif 



#ifndef BOOST_NO_CXX23_STDFLOAT

// do something that uses the header

#endif 
```

I'm working on the C++23 macros for Boost.Config now...

ryanelandt: The approach described above typically applies when one is manually defining methods. A manually defined method may compile if certain headers have been included. For example, one might manually define a method for type `__float128`, another for `_Quad`, and another for `std::float128_t`. Careful preprocessor use is required because certain methods will only compile if the correct headers have been included.

In contrast to this approach, this approach relies on duck typing. In C++ duck typing happens at compile time and is implemented with templates and SFINAE. It also allows a single templatized method to be used for various types. For example, the 128 bit type is used by the following types: `__float128`, `_Quad`, `std::float128_t`, `multiprecision::float128`, as well as `std_real_concept` when it's set up to emulate a 128 bit float. I may be missing some, but the compiler will not.

The benefit of the duck type approach that _the code only needs to compile for the template parameters that are actually used_. For this reason, the approach does not require the careful preprocessor use of the manual approach. 

Also, thank you for working on the C++23 macros.

NAThompson:
@ryanelandt : My apologies for never getting this past the finish line-3 years later . . . it's now fixed.

PR #1014, opened 2023-08-18
fix unused abterm1 compile warnings

ryanelandt:
The file `exp_sinh_detail.hpp` contains an unconditionally unused variable `abterm1` that causes compile warnings. This issue was fixed for C++17 and above in #884 with the addition of the `BOOST_MATH_MAYBE_UNUSED` macro. 

**This PR** implements a direct fix that removes the warning for all C++ versions. It also removes the `BOOST_MATH_MAYBE_UNUSED` macro as this is the only place it's used in the Boost (all libraries).


jzmaddock:
Looks good to me, not sure how that crept in there.... merging.

ryanelandt:
Thanks!

Issue #1016, opened 2023-08-22
Findings from CI in Multiprecision on MSVC 14_0


ckormanyos:
In Multiprecision Issues like [this](https://github.com/boostorg/multiprecision/issues/564) we find stuff maybe relevant for Math.

- MSVC 14.0 does not comprehend `/std:c++17` and the CI run(s) should use `[14, latest]` in the _standard_ matrix.
- MSVC 14.0 seems to not comprehend the preprocessor _jargon_ for testing if it _has_ the header `<stdfloat>`. We need to do some preprocessor exclusions around [here](https://github.com/boostorg/math/blob/dc10fc04b29a356c3cf12c2052c45b0c3b3da5f2/include/boost/math/tools/promotion.hpp#L30). This one is probably running wild in the field so maybe we get to it for 1.84.

On the second topic, something like below?

```cpp
#if !(defined(_MSC_VER) && (_MSC_VER <= 1900))
#if __has_include(<stdfloat>)
#  include <stdfloat>
#endif
```

But I don't know 1f 14.1 comprehends `__has_include` so we need to check that too.

ckormanyos:
I can take a crack at these items.

Cc: @jzmaddock and @mborland

jzmaddock:
You should use #ifdef __has_Include to verify it's existence.

ckormanyos:
> use #ifdef __has_Include to verify it's existence

```cpp
#if (defined(__has_include) && __has_include(<stdfloat>))
#  include <stdfloat>
#endif
```

Like that. Thx John.
I will try. But it was a problem with lexical parsing so I'm not sure it works. I will report after trying it out...

ckormanyos:
See also #1017

mborland:
Since MSVC 14.0 is not really C++14 compliant this is another vote from me to drop support for it entirely. I don't think its worth the effort to add workarounds for something that we're going to continue to break as the code is modernized/features added.

ckormanyos:
> Since MSVC 14.0 is not really C++14 compliant this is another vote from me to drop support for it entirely. I don't think its worth the effort to add workarounds for something that we're going to continue to break as the code is modernized/features added.

Yea. I'd be ready to drop it any time. It#s going to be a nuicanse with `constexpr`anyway, ...

Yet, this workaround at the moment is just about finished. I'd just go ahead and finish it. Then we could continue the _drop_ _vc140_ discussion at another place/time?

PR #1017, opened 2023-08-22
MSVC 14_0 stdndards and has-include corrections

ckormanyos:
The purpose of this PR is to handle trivial syntax improvements mentioned in #1016.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,12 +180,14 @@ jobs:
       run:
         shell: cmd
     env:
-      ARGS: toolset=${{ matrix.toolset }} address-model=64 cxxstd=${{ matrix.standard }}
+      ARGS0: toolset=${{ matrix.toolset }} address-model=64 cxxstd=${{ matrix.standard0 }}
+      ARGS2: toolset=${{ matrix.toolset }} address-model=64 cxxstd=${{ matrix.standard2 }}
     strategy:
       fail-fast: false
       matrix:
         toolset: [ msvc-14.0, msvc-14.2 ]
-        standard: [ 14, 17 ]
+        standard0: [ 14, latest ]
+        standard2: [ 14, 17 ]
         suite: [ github_ci_block_1, github_ci_block_2 ]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


ckormanyos:
The lexical parser isn't into it.

```
fatal error C1012: unmatched parenthesis: missing ')'
```

I need to go with the former preprocessor stuff, or at least try it out first.

ckormanyos:
Well the code logic is OK and things are going green but i messed up the GHA logic a bit and a few too many runners are being run. So this will need one more logic fix on the YAML layer. Coming right up, ...

jzmaddock:
>The lexical parser isn't into it.

It does work: Boost.Config is using it for all compilers including vintage msvc: https://github.com/boostorg/config/blob/d483059795f8bf77120ee423d1660532223e5795/include/boost/config/detail/select_stdlib_config.hpp#L18

ckormanyos:
>> The lexical parser isn't into it.

> Boost.Config is using it for all ...

Ahhh good save, John. I tried the line:

```cpp
#if (defined(__has_include) && (__has_include(<stdfloat>)))
```

But the variation should be:

```cpp
#if (defined(__has_include)
#if (__has_include(<stdfloat>))
#include <stdfloat>
#endif
#endif
```

ckormanyos:
> good save, John

John (@jzmaddock) could we agree on?

```cpp
#if defined(__has_include)
#if __has_include(<stdfloat>)
#include <stdfloat>
#endif
#endif
```

Cc: @mborland

mborland:
> ```c++
> #if defined(__has_include)
> #if __has_include(<stdfloat>)
> #include <stdfloat>
> #endif
> #endif
> ```

That is the format that GCC recommends: https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

ckormanyos:
> the format that GCC recommends

Great thank you Matt (@mborland). I will cycle this through, then:

```cpp
#if defined __has_include
#  if __has_include (<stdfloat>)
#    include <stdfloat>
#  endif
#endif
```

ckormanyos:
Matt (@mborland) while we are at it, I used super-rudimentary logic on the YAML-ing. Would you please take a look and see if you can agree to that? Or can we do better?

mborland:
> Matt (@mborland) while we are at it, I used super-rudimentary logic on the YAML-ing. Would you please take a look and see if you can agree to that? Or can we do better?

It looks good to me. Definitely more compact than making it own config for one compiler.

ckormanyos:
> t looks good to me.

Thanks Matt. OK let's let this thing cacle through. It is cycling green even with recent changes in develop to `denorm_min` in Multiprecision. These are in develop now.

I'm not sure if this is or is not expected. But I'm happy to observe that.

ckormanyos:
Single failure on drone to clone. Merging.

Issue #1018, opened 2023-08-22
big_constant.hpp:63:52: error: unused parameter 's'


ckormanyos:
This may seem really overly pedantic, but I do, in fact, hit a snag on [this](https://github.com/boostorg/math/blob/dc10fc04b29a356c3cf12c2052c45b0c3b3da5f2/include/boost/math/tools/big_constant.hpp#L63).

The subroutine in question is intended to fail a `static_assert` anyway, so there's not really a point using or not using the parameter `s`.

But believe it or not, I get a LINT-er not liking it.

ckormanyos:
See PR #1019

PR #1019, opened 2023-08-22
Eliminate unused parameter

ckormanyos:
Here we have an extremely trivial removal of an unused parameter `s` that had been flagged by a compiler or LINT-er in 1.83.


ckormanyos:
Intended to handle #1018

Issue #1023, opened 2023-08-28
Warning because stdfloat is included with C++17


ChristophStrehle:
When I include boost/math/concepts/real_concept.hpp or boost/math/tools/promotion.hpp from boost 1.83 I get the following compiler warning with Visual Studio 17.7 and the v143 version 14.37 toolset compiling for C++17:

```warning STL4038: The contents of <stdfloat> are available only with C++23 or later.```

Looks like stdfload is there but should not be included with C++17. The two includes are here:
https://github.com/boostorg/math/blob/develop/include/boost/math/tools/promotion.hpp#L31
https://github.com/boostorg/math/blob/develop/include/boost/math/concepts/real_concept.hpp#L49

Btw I don't get this warning when I include boost/math/tools/config.hpp:
https://github.com/boostorg/math/blob/develop/include/boost/math/tools/config.hpp#L157

ckormanyos:
We were just tuning the exact header doing the recognition of this header. We used the macro `__has_include` exclusively.

John, do you think we should combine thas PP querry with a check on `__cplusplus` to ensure that it is also _high_ enough?

jzmaddock:
>We were just tuning the exact header doing the recognition of this header. We used the macro __has_include exclusively.
>
>John, do you think we should combine thas PP querry with a check on __cplusplus to ensure that it is also high enough?

You always have to do that: __has_Include only tells you whether the file exists on disk, not whether you can actually include it in the current -std= mode.  That's why we have Boost.Config macros to handle this and make it less error prone, I just haven't added those for C++23 yet!

mborland:
That's my mistake. I'll just keep going with greater than `202002L` because there's no macro definition for C++23 yet, and `<stdfloat>` does not have a feature test macro either: https://en.cppreference.com/w/cpp/feature_test

ckormanyos:
>  keep going with greater than `202002L` because there's no macro definition for C++23 yet, and `<stdfloat>` does not have a feature test macro either

Awesome plan. Before I wrack my brain with the PP logic, could you please do this Matt, as you're more practiced at it than I?

ckormanyos:
Ahhh you already beat me to it with #1023 sorry about the noise. Thanks guys and thanks Christophe (@ChristophStrehle)

PR #1024, opened 2023-08-28
Fix issue #1023

mborland:


mborland:
Failures are from CI issues.

Issue #1006, opened 2023-07-30
bug in boost libraries used by scipy.stats.beta


Gabriel-Kissin:
I've found a bug in a couple of python `scipy` probability functions which leads to the kernel crashing. In the case of the `beta` function, the error originates in the source in boost which appears to be [at boost.org](https://www.boost.org/doc/libs/1_82_0/boost/math/special_functions/beta.hpp) and [on github](https://github.com/boostorg/math/blob/develop/include/boost/math/special_functions/beta.hpp) in one of the lines reading `BOOST_MATH_ASSERT(*p_derivative >= 0)`.   

I originally posted this issue at https://github.com/scipy/scipy/issues/18906. But am now reposting the python code here as requested there by @jzmaddock.   

```
import scipy.stats
print([shape_param.domain for shape_param in scipy.stats.beta._shape_info()])
# [[5e-324, inf], [5e-324, inf]]
print(scipy.stats.beta(a=1e-308, b=5).ppf(.2))
# 0.0
print(scipy.stats.beta(a=1e-309, b=5).ppf(.2))
# OverflowError: Error in function boost::math::tgamma<d>(d): Overflow Error
# (raises error, but kernel doesn't crash)
print(scipy.stats.beta(a=1e-323, b=5).ppf(.2))
# KERNEL CRASHES with message: `Assertion failed: *p_derivative >= 0, file  
 ..\..\scipy\_lib\boost/boost/math/special_functions/beta.hpp, line 739`
```

PR #1009, opened 2023-07-31
Fix assertions and fenv errors in beta distribution

mborland:
I was able to reproduce the error in python using the script from #1006. The assertion was not triggered using the same template parameters in C++, but the fenv flags are being set. Below is the diagnostic data from `BOOST_MATH_INSTRUMENT`

```
====== BEGIN OUTPUT ======
lgamma_imp called with e N5boost4math7lanczos12lanczos17m64E
../../../boost/math/special_functions/erf.hpp:1226 result_type = e
../../../boost/math/special_functions/erf.hpp:1227 value_type = e
../../../boost/math/special_functions/erf.hpp:1228 precision_type = N5boost4math8policies7digits2ILi64EEE
../../../boost/math/special_functions/erf.hpp:1237 tag_type = St17integral_constantIiLi64EE
../../../boost/math/special_functions/erf.hpp:427 64-bit precision erf_imp called
../../../boost/math/special_functions/erf.hpp:1191 result_type = e
../../../boost/math/special_functions/erf.hpp:1192 value_type = e
../../../boost/math/special_functions/erf.hpp:1193 precision_type = N5boost4math8policies7digits2ILi64EEE
../../../boost/math/special_functions/erf.hpp:1202 tag_type = St17integral_constantIiLi64EE
../../../boost/math/special_functions/erf.hpp:427 64-bit precision erf_imp called
../../../boost/math/special_functions/erf.hpp:1191 result_type = e
../../../boost/math/special_functions/erf.hpp:1192 value_type = e
../../../boost/math/special_functions/erf.hpp:1193 precision_type = N5boost4math8policies7digits2ILi64EEE
../../../boost/math/special_functions/erf.hpp:1202 tag_type = St17integral_constantIiLi64EE
../../../boost/math/special_functions/erf.hpp:427 64-bit precision erf_imp called
../../../boost/math/special_functions/erf.hpp:1191 result_type = e
../../../boost/math/special_functions/erf.hpp:1192 value_type = e
../../../boost/math/special_functions/erf.hpp:1193 precision_type = N5boost4math8policies7digits2ILi64EEE
../../../boost/math/special_functions/erf.hpp:1202 tag_type = St17integral_constantIiLi64EE
../../../boost/math/special_functions/erf.hpp:427 64-bit precision erf_imp called
../../../boost/math/special_functions/erf.hpp:1191 result_type = e
../../../boost/math/special_functions/erf.hpp:1192 value_type = e
[...]

PR #1026, opened 2023-09-01
Avoid spurious overflow and divide by zero in ibeta.

jzmaddock:
Add tests extracted from SciPy bug report.
Fixes https://github.com/boostorg/math/issues/1006. 
See also https://github.com/boostorg/math/pull/1009.


mdhaber:
Thanks @jzmaddock! This seems to do the trick on my M1 mac. 
```python3
import scipy.stats
scipy.stats.beta(a=1e-323, b=5).ppf(.2)
```
Before this patch, I get an error; after this patch, I get `1.8189894035458565e-12`.

Issue #1028, opened 2023-09-08
C++23 std::numeric_limits<T>::has_denorm warning


mborland:
The latest visual studio marked `has_denorm` deprecated so we get a new warning:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.37.32822\include\limits(40,18): note: 'float_denorm_style' has been explicitly marked deprecated here
_EXPORT_STD enum _CXX23_DEPRECATE_DENORM float_denorm_style {
                 ^
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.37.32822\include\yvals_core.h(1440,7): note: expanded from macro '_CXX23_DEPRECATE_DENORM'
    [[deprecated("warning STL4042: "                                                                                   \
      ^
In file included from libs\decimal\test\test_decimal32_cmath.cpp:8:
.\boost/math/special_functions/next.hpp(112,129): error: 'has_denorm' is deprecated: warning STL4042: std::float_denorm_style, std::numeric_limits::has_denorm, and std::numeric_limits::has_denorm_loss are deprecated in C++23. You can define _SILENCE_CXX23_DENORM_DEPRECATION_WARNING or _SILENCE_ALL_CXX23_DEPRECATION_WARNINGS to suppress this warning. [-Werror,-Wdeprecated-declarations]
   return get_smallest_value<T>(std::integral_constant<bool, std::numeric_limits<T>::is_specialized && (std::numeric_limits<T>::has_denorm == std::denorm_present)>());
                                                                                                                                ^
.\boost/math/special_functions/next.hpp(492,74): note: in instantiation of function template specialization 'boost::math::detail::get_smallest_value<float>' requested here
      return 1 + fabs(float_distance(static_cast<T>((b < 0) ? T(-detail::get_smallest_value<T>()) : detail::get_smallest_value<T>()), b, pol));
                                                                         ^
.\boost/math/special_functions/next.hpp(710,22): note: in instantiation of function template specialization 'boost::math::detail::float_distance_imp<float, boost::math::policies::policy<>>' requested here
[...]

PR #1029, opened 2023-09-11
has_denorm_now

ryanelandt:
As pointed out in #1028, `std::numeric_limits<T>::has_denorm` has been deprecated in C++23.

This PR refactors `get_smallest_value` to no longer use `has_denorm`. A new method `has_denorm_now` is created that checks if denorms are supported at runtime by comparing the output of `boost::math::detail::get_smallest_value` with that of `boost::math::tools::min_value`. Most other instances of the compile time constant `has_denorm` are replaced by the runtime calculated version `has_denorm_now`.


jzmaddock:
Thanks Matt, this looks good to me!  Any reason not to merge?

PR #1033, opened 2023-10-02
ten should be a constexpr as a float

AtariDreams:
Its int counterpart is a constexpr, and the ten variable never changes, so for consistency, we should make float_type ten constexpr as welll.


jzmaddock:
CI failures are network issues, merging...

Issue #1035, opened 2023-10-10
`non_central_t` distribution pdf left tail continuity


mborland:
Example can be found here: https://github.com/scipy/scipy/issues/19348#issuecomment-1749250804

PR #1038, opened 2023-10-13
Correct non-central-t series convergence bug.

jzmaddock:
Fixes https://github.com/boostorg/math/issues/1035. See also https://github.com/scipy/scipy/issues/19348. Accuracy in left tail is still poor, and the reflection formula appears to be to blame as it's use causes the series to cancel out the first term, but it appears we have no real choice in the matter here.  At least we do now get a few digits correct.


mborland:
Thanks for getting this one John.

PR #1039, opened 2023-10-17
Incorporate cast simplifications from PR880

jzmaddock:
Add tests and make sure everything is covered, and fix resulting errors.


jzmaddock:
One failure is a network error, merging...

PR #1041, opened 2023-10-31
Fix clang-14 run in CI

mborland:
The updated GHA runners broke Clang-14 in C++20 mode. G++12 in C++20 mode is covered in our drone runs which are fine..

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
       matrix:
         compiler: [ g++-12, clang++-14 ]
-        standard: [ c++14, c++17, c++20 ]
+        standard: [ c++14, c++17 ]
         suite: [ github_ci_block_1, github_ci_block_2 ]
     steps:
       - uses: actions/checkout@v3


jzmaddock:
Just out of curiosity, what was the breakage, it seems a shame we can't run clang in C++20.

mborland:
> Just out of curiosity, what was the breakage, it seems a shame we can't run clang in C++20.

There are two to watch for now. This will hit any Clang < 16 in 2b mode:

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/stl_iterator.h:2618:35: error: missing 'typename' prior to dependent type name 'iterator_traits<_It>::iterator_category'
      { using iterator_category = iterator_traits<_It>::iterator_category; };
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


And this will happen in C++20 mode:

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: error: call to consteval function 'std::chrono::hh_mm_ss::_S_fractional_width' is not a constant expression
        static constexpr unsigned fractional_width = {_S_fractional_width()};
                                                      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: note: undefined function '_S_fractional_width' cannot be used in a constant expression
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2275:2: note: declared here
        _S_fractional_width()
        ^
1 error generated.
```

The GHA 22.04 image updated from GCC12 to 13 as the default version a few days ago.

ckormanyos:
> The GHA 22.04 image updated from GCC12 to 13 as the default version a few days ago.

In broke parts of `iterator_traits` and parts of `chrono`. I spent all weekend working around in order to keep those compilers. But you need to have super-simple uses of iterators and chrono-functions. I don't think Boost will have such simplistic cases...

Tough break on this one.

PR #1043, opened 2023-11-02
Qualify log1p calls.

jzmaddock:
This fixes ambiguous overloads in the multiprecision case between boost::multiprecision and boost::math versions of log1p. See https://github.com/boostorg/multiprecision/pull/568#issuecomment-1790195318


jzmaddock:
Failure is unrelated (cohen_accelleration_test.cpp fails on cygwin - tolerance too tight?).

PR #1044, opened 2023-11-03
Fix detection of ccmath usage in trunc

mborland:
As reported by @robertramey on the ML

### DIFF
--- a/include/boost/math/special_functions/trunc.hpp
+++ b/include/boost/math/special_functions/trunc.hpp
@@ -16,10 +16,11 @@
 #include <boost/math/tools/config.hpp>
 #include <boost/math/policies/error_handling.hpp>
 #include <boost/math/special_functions/fpclassify.hpp>
+#include <boost/math/tools/is_constant_evaluated.hpp>
 
 #ifndef BOOST_NO_CXX17_IF_CONSTEXPR
-#include <boost/math/ccmath/ldexp.hpp>
 #  if !defined(BOOST_MATH_NO_CONSTEXPR_DETECTION)
+#    include <boost/math/ccmath/ldexp.hpp>
 #    define BOOST_MATH_HAS_CONSTEXPR_LDEXP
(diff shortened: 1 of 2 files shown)


mborland:
Errored on clone in drone. Merging.

jzmaddock:
@mborland sorry to be late to this, I don't think this quite gets it - for one thing we have the same logic in round.hpp which also needs fixing, but the problem is more subtle - we have modern clang sitting on top of an ancient (pre-C++17) std library, so from the compilers perspective, everything that we're currently checking here is supported (or potentially so) where as the std library's <type_traits> has no C++17 `_v` suffixed constants.  Other than checking for a random C++17 header as a proxy in addition to the existing compiler checks I'm not sure what to suggest.  In the long run we could yet another macro to Boost.Config I guess... ah wait, I think this is `__cpp_lib_bool_constant >= 201505L`?  The latter requires inclusion of either type_traits or version, so the logic had probably better somewhere central.  

I'll see if I can put together a patch shortly, might be too late for this release though as I can see quite a few headers getting touched.

mborland:
I thought clang didn't get a version of `__builtin_constant_evalutated` until 9? If the type_traits had compiled the ccmath ldexp would have errored on initializing a constexpr variable with a non-constexpr expression since it uses the STL at runtime. Unless someone is running say Clang-10 with libstdc++-5 shouldn't properly checking for `BOOST_MATH_NO_CONSTEXPR_DETECTION ` be sufficient? I think there's another week until RC release so you should have time.

jzmaddock:
This is what I had in mind: https://github.com/boostorg/math/pull/1045

And yes, from a practical point of view, you may be correct that no one would use clang-10 on top of gcc-5, though I'm constantly surprised by what what people do do!

robertramey:
FYI. It's not just  mere "people".  It's Boost Developers.  Its from the test matrix which is our thing.

PR #1042, opened 2023-11-02
Merge for 1.84

mborland:

PR #1045, opened 2023-11-03
Update/improve ccmath configuration.

jzmaddock:


jzmaddock:
@mborland most of the changes here are to ccmath: does this look OK to you?  The drone failures are due to github falling over yesterday BTW, I don't see an easy way to re-run that?

mborland:
> @mborland most of the changes here are to ccmath: does this look OK to you? The drone failures are due to github falling over yesterday BTW, I don't see an easy way to re-run that?

It looks good to me. I have organizational access to drone so I restarted the build. The status should be updated here on re-run.

mborland:
By the way once you merge the PR to merge to master for next release (https://github.com/boostorg/math/pull/1042) will also automatically pull your commits in.

jzmaddock:
Thanks Matt, one last failure is a network timeout, merging...

ckormanyos:
I don't know _why_ this was never a problem before, but nested namespaces are C++17 and I just had a MSVC compiler report this when set to C++14 at lines like [this](https://github.com/boostorg/math/blob/11e2348372088f0ede36ff6cbf2d70c8e1783a3b/include/boost/math/ccmath/isnan.hpp#L16)

ckormanyos:
Ah OK. It _should_ be caught by the preprocessor. Now how on earth did I set a MSVC compiler to C++14 and still reach that line of code? I will investigate...

ckormanyos:
OK. John and Matt. Sorry to report, ... but There seems to be some kind of very obscure problem. Sadly I must report it.
@jzmaddock and @mborland 

- I take VS2019 and make a test project.
- Set language standard to C++14
- Then `#include <boost/math/ccmath/isinf.hpp>

And I get this:

grafik

ckormanyos:
Oh well, you know, MSVC doesn't really adhere to the well-known `__cplusplus` values. So maybe something there?

Anyway, here is the complete code:

```cpp
#include <boost/multiprecision/number.hpp>
#include <boost/math/ccmath/isinf.hpp>

auto main() -> int
{
}
```

ckormanyos:
Sorry guys this is really garbled since it literally just crossed my path and I need to isolate better.

This!!! is the offending code:

```cpp
#define BOOST_MP_STANDALONE
#define BOOST_MATH_STANDALONE

#include <boost/config.hpp>
#include <boost/multiprecision/number.hpp>
#include <boost/math/ccmath/isinf.hpp>

auto main() -> int
{
}
```

grafik

jzmaddock:
It's the standalone bit that does this - it means we have no Boost.Config, so either we need to completely copy all of it's detection logic, or accept that the helpful #error message doesn't get triggered in this case, and rely on the compiler rejecting C++17 ism's which it does perfectly well IMO.

I might be able to improve things a little for latest MSVC releases which have the C++20 feature detection macros, but not for older releases which do not.  Not sure it's completely worth it anyway, the existing error message is pretty clear that this needs C++17 to compile?

jzmaddock:
Ah... except that breaks client code that relies on BOOST_MATH_NO_CCMATH being set when the header isn't supported.... there may be an easy fix too, give me a moment!

jzmaddock:
Improvement in https://github.com/boostorg/math/pull/1046.

ckormanyos:
> Ah... except that breaks client code that relies on BOOST_MATH_NO_CCMATH being set when the header isn't supported.... there may be an easy fix too, give me a moment!

Thank you John. I can confirm for my client code, it is fixed in your new PR #1046

PR #1046, opened 2023-11-05
Improve ccmath support detection.

jzmaddock:
See discussion in https://github.com/boostorg/math/pull/1045.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
         run: cp -r $GITHUB_WORKSPACE/* libs/math
         working-directory: ../boost-root
       - name: Install deps
-        run: python tools/boostdep/depinst/depinst.py math
+        run: python tools/boostdep/depinst/depinst.py math -I libs/math/example -I libs/math/tools
         working-directory: ../boost-root
       - name: Bootstrap
         run: ./bootstrap.sh
@@ -116,7 +116,7 @@ jobs:
         run: cp -r $GITHUB_WORKSPACE/* libs/math
         working-directory: ../boost-root
       - name: Install deps
-        run: python tools/boostdep/depinst/depinst.py math
+        run: python tools/boostdep/depinst/depinst.py math -I libs/math/example -I libs/math/tools
         working-directory: ../boost-root
       - name: Bootstrap
         run: ./bootstrap.sh
@@ -157,7 +157,7 @@ jobs:
         run: cp -r $GITHUB_WORKSPACE/* libs/math
         working-directory: ../boost-root
       - name: Install deps
-        run: python tools/boostdep/depinst/depinst.py math
+        run: python tools/boostdep/depinst/depinst.py math -I libs/math/example -I libs/math/tools
         working-directory: ../boost-root
       - name: Bootstrap
         run: ./bootstrap.sh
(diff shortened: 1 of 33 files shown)


ckormanyos:
Confirm that this fixes the client code which first detected the issue discussed toward to bottom of #1045.

Thank you so much John (@jzmaddock) for such a quick response to even this trivial issue!

mborland:
Should we apply this logic to all of the headers? Isfinite has the [old logic](https://github.com/boostorg/math/blob/develop/include/boost/math/ccmath/isfinite.hpp#L19), and some have none at all.

ckormanyos:
> Should we apply this logic to all of the headers?

This is, indeed, a slim area, ... Clients using MSVC language standard 14 or lower. Anyway, I think that seems sensible both to repair it and to repair it consistently.

For more information, by sheer random chance, I was revitalizing a very old legacy project recently and just so happened to have MSVC 141 C++ standard 14 running on CI (otherwise i would not have seen this at all...).

Anyway, the error message(s) I got were [here](https://github.com/ckormanyos/mp_cpp/actions/runs/6761699209/job/18376842057#step:7:94).

There are errors on `isnan`, `isinf`, `abs` and `ldexp`. But these might be the ones that were simply _picked_ _up_ by MSVC in my particular instantiations.

ckormanyos:
Hi Matt (@mborland) I was thinking about this issue and noticed C++17 is not so far away from C++14 (the language standard supported by all of Math7Multiprecision).

I suppose you won't really like this question but, ...

- How far is `ccmath` let's say, _away_ _from_ C++14?
- Could/should we backport it?
- I could heklp backport it for 1.85 if this makes things easier or more clear...

mborland:
> Hi Matt (@mborland) I was thinking about this issue and noticed C++17 is not so far away from C++14 (the language standard supported by all of Math/Multiprecision).
> 
> I suppose you won't really like this question but, ...
> 
> * How far is `ccmath` let's say, _away_ _from_ C++14?
> * Could/should we backport it?
> * I could help backport it for 1.85 if this makes things easier or more clear...

It's not particularly far away from C++14, but that is not the pressing issue. For it to work we require proper `__builtin_is_constant_evaluated` which is GCC >= 9, clang >= 9, or MSVC > 1925. All those compilers support C++17 just fine.

ckormanyos:
> It's not particularly far away from C++14, but that is not the pressing issue. For it to work we require proper `__builtin_is_constant_evaluated`

Understood. Challenging, getting through the maze of compilers...

ckormanyos:
```
2023-11-06T09:50:49.0791974Z ../tools/process_perf_results.cpp:12:10: fatal error: boost/format.hpp: No such file or directory
2023-11-06T09:50:49.0793363Z    12 | #include <boost/format.hpp>
2023-11-06T09:50:49.0794313Z       |          ^~~~~~~~~~~~~~~~~~
2023-11-06T09:50:49.0795175Z compilation terminated.
```

jzmaddock:
WTF?  Nothing has changed, either here, or in boostdep or Boost.Format... rerunning just to see what happens!

ckormanyos:
> WTF? Nothing has changed, either here, or in boostdep or Boost.Format... rerunning just to see what happens!

If it runs, you might have hit something that I've always wondered abount, non-atomic checkins in the modular-boost. I wonder if something like that _might_ have happened?

Let's see.

I don't even know if commits to the master project or develop are atomic. That would surprise me...

ckormanyos:
The thing is still looking for `<boost/format.hpp>` for several exemplary CPP-Files.

This file is present in 1.83.

I wonder if this header has been either knowingly or mistakenly removed from Boost on Develop in preparation for 1.84?

pdimov:
In short, add `-I example -I tools` to the `depinst` invocation to tell it to scan these directories for dependencies (in addition to the default `include `src` `test`).

https://github.com/boostorg/boostdep/issues/20

ckormanyos:
Awesome! Thanks John and Matt

ckormanyos:
It is running perfectly _in_ _the_ _field_ on my unrelated project that actually uses it on MSVC `std=c++14`. I just check on the runners for my unrelated project. Great!

PR #1057, opened 2023-12-13
Fix cstdfloat numeric limits detection macro for libstdc++14

mborland:
Reported by @jwakely here: https://github.com/boostorg/math/pull/993

### DIFF
--- a/include/boost/math/cstdfloat/cstdfloat_limits.hpp
+++ b/include/boost/math/cstdfloat/cstdfloat_limits.hpp
@@ -24,7 +24,7 @@
 #pragma GCC system_header
 #endif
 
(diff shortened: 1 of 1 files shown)


mborland:
Failures are internal CI errors. Merging

jwakely:
Thanks! We'll get this backported to the Fedora packages.

jwakely:
Feel free to @ me here if you want me to check gcc-config-related changes like this, I'd rather review them _before_ they make it into a release than when we're packaging it for Fedora :-)

mborland:
> Feel free to @ me here if you want me to check gcc-config-related changes like this, I'd rather review them _before_ they make it into a release than when we're packaging it for Fedora :-)

Will do. Thanks.

PR #1059, opened 2023-12-20
Remove last usage of boost::array

NAThompson:
Now that boost.math is C++14, remove boost::array in order to improve orthogonality of the library.


mborland:
> @mborland , @jzmaddock : Pretty trivial but hopefully helpful.
> 
> 
> 
> Should we do the same in `odeint`?

Everything C++03 should be ripped out of odeint. I think the biggest offender to compile time in there is MPL.

jzmaddock:
Thanks Matt!

PR #1060, opened 2023-12-21
Fix for msvc defining _Complex_I in complex.h

mborland:
@jzmaddock. I think pushing and popping the value of "I" seems to be the safest workaround unless you have a better idea.

### DIFF
--- a/doc/internals/polynomial.qbk
+++ b/doc/internals/polynomial.qbk
@@ -21,7 +21,7 @@
       polynomial(){}
       template <class U>
       polynomial(const U* data, unsigned order);
-      template <class I>
+      template <class Iterator>
       polynomial(I first, I last);
       template <class U>
       explicit polynomial(const U& point, 
--- a/doc/internals/polynomial.qbk
+++ b/doc/internals/polynomial.qbk
@@ -22,7 +22,7 @@
       template <class U>
       polynomial(const U* data, unsigned order);
       template <class Iterator>
-      polynomial(I first, I last);
+      polynomial(Iterator first, Iterator last);
       template <class U>
       explicit polynomial(const U& point, 
             typename std::enable_if<std::is_convertible<U, T> >::type* = nullptr);
--- a/include/boost/math/special_functions/detail/bessel_ik.hpp
+++ b/include/boost/math/special_functions/detail/bessel_ik.hpp
@@ -77,7 +77,7 @@ inline T bessel_i_small_z_series(T v, T x, const Policy& pol)
    return prefix * result;
 }
 
-// Calculate result_K(v, x) and result_K(v+1, x) by method analogous to
+// Calculate K(v, x) and K(v+1, x) by method analogous to
 // Temme, Journal of Computational Physics, vol 21, 343 (1976)
 template <typename T, typename Policy>
 int temme_ik(T v, T x, T* result_K, T* K1, const Policy& pol)
@@ -200,7 +200,7 @@ int CF1_ik(T v, T x, T* fv, const Policy& pol)
     return 0;
 }
 
-// Calculate result_K(v, x) and result_K(v+1, x) by evaluating continued fraction
+// Calculate K(v, x) and K(v+1, x) by evaluating continued fraction
 // z1 / z0 = U(v+1.5, 2v+1, 2x) / U(v+0.5, 2v+1, 2x), see
 // Thompson and Barnett, Computer Physics Communications, vol 47, 245 (1987)
 template <typename T, typename Policy>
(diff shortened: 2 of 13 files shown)


jzmaddock:
I wonder what the best way to fix this is?  A quick grep shows the following files effected:

tools/polynomial.hpp
tools/mp.hpp
special_functions/detail?bessel_ik.hpp (uses variables named I, J, K)
special_functsions/bessel.hpp
bindings/rr.hpp
bindings/mpreal.hpp
bindings/mpfr.hpp

Plus lots of occurrences of "I" in comments which we can ignore.

BTW I used special_functions.hpp purely as a sledge hammer to try and get a quick reproducer, it would be better to fix the individual files.

Also your fix is msvc specific, and I can reproduce with gcc, which outputs thousands of errors along the lines of:

```
In file included from /usr/lib/gcc/x86_64-pc-cygwin/11/include/c++/complex.h:38,
                 from t.cpp:1:
./boost/math/tools/mp.hpp:171:34: error: expected ‘>’ before numeric constant
  171 | template<typename L, std::size_t I>
      |                                  ^
In file included from ./boost/math/policies/policy.hpp:11,
                 from ./boost/math/special_functions/math_fwd.hpp:31,
                 from ./boost/math/special_functions/airy.hpp:11,
                 from ./boost/math/special_functions.hpp:15,
                 from t.cpp:2:
./boost/math/tools/mp.hpp:172:46: error: could not convert ‘(__complex__ float){0.0f, 1.0e+0f}’ from ‘__complex__ float’ to ‘long unsigned int’
  172 | using mp_at_c = typename detail::mp_at_c<L, I>::type;
      |                                              ^
      |                                              |
      |                                              __complex__ float
./boost/math/tools/mp.hpp:172:47: error: ‘<declaration error>’ is not a template [-fpermissive]
  172 | using mp_at_c = typename detail::mp_at_c<L, I>::type;
      |                                               ^~
In file included from /usr/lib/gcc/x86_64-pc-cygwin/11/include/c++/complex.h:38,
                 from t.cpp:1:
./boost/math/tools/mp.hpp:174:31: error: expected nested-name-specifier before numeric constant
  174 | template<typename L, typename I>
```

So my temptation is to replace `I` with `Integer` for template type names, that just leaves the bessel functions which use the mathematically correct symbols I, J and K for variable names (lower case names mean something else in this context).  It's not pretty but maybe change to `result_I`, `result_J` and `result_K` ?

mborland:
Yeah CI shows that Cygwin and GCC running in GNU mode all do the wrong thing. Renaming everything works. The meta tools can have Index instead of I.

doc/internals/polynomial.qbk
```diff
@@ -21,7 +21,7 @@
       polynomial(){}
       template <class U>
       polynomial(const U* data, unsigned order);
-      template <class I>
+      template <class Iterator>
       polynomial(I first, I last);
```
jzmaddock, line 25: Doesn't this line also need changing?

include/boost/math/special_functions/detail/bessel_ik.hpp
```diff
@@ -77,10 +77,10 @@ inline T bessel_i_small_z_series(T v, T x, const Policy& pol)
    return prefix * result;
 }
 
-// Calculate K(v, x) and K(v+1, x) by method analogous to
+// Calculate result_K(v, x) and result_K(v+1, x) by method analogous to
```
jzmaddock, line 80: Nit pick: K here refers to the Bessel function K, not to our variable, so I think we should leave the comment unchanged.

include/boost/math/special_functions/detail/bessel_ik.hpp
```diff
@@ -200,7 +200,7 @@ int CF1_ik(T v, T x, T* fv, const Policy& pol)
     return 0;
 }
 
-// Calculate K(v, x) and K(v+1, x) by evaluating continued fraction
+// Calculate result_K(v, x) and result_K(v+1, x) by evaluating continued fraction
```
jzmaddock, line 203: Likewise, comment refers to the special function K.

include/boost/math/special_functions/detail/bessel_ik.hpp
```diff
@@ -374,7 +374,7 @@ int bessel_ik(T v, T x, T* I, T* K, int kind, const Policy& pol)
     current = Ku1;
     T scale = 1;
     T scale_sign = 1;
-    for (k = 1; k <= n; k++)                   // forward recurrence for K
+    for (k = 1; k <= n; k++)                   // forward recurrence for result_K
```
jzmaddock, line 377: Again K refers to the mathematical function not our variable... life would be easier if mathematicians didn't keep reusing the same names over and over!

jzmaddock:
Many thanks Matt for taking this on, some mostly nit picking comments above!

PR #1061, opened 2023-12-22
Add codecov run to CI

mborland:


mborland:
@Flamefire This is the same codecov yaml from boost-ci I have put in a few libraries now. Any idea why the setup boost step is failing with missing library? It looks like an error in `get_libname`? https://github.com/boostorg/math/actions/runs/7299976550/job/19894765017#step:10:126. Thanks for the assistance.

mborland:
Looks like our libraries.json was not formatted as expected. Everything is working now.

Flamefire:
> Looks like our libraries.json was not formatted as expected. Everything is working now.

To be specific: You put your description into a list which seems to be valid and used by libraries containing sub-libraries. I don't see a reason to use this for single-library libraries where you can just use a dict (as you did in your change)   
I'm not sure how that JSON file is defined but seemingly that list has to be supported by Boost.CI -> https://github.com/boostorg/boost-ci/pull/225

Currently your PR fails due to a timeout on GHA. You can possibly split it into multiple jobs instead of building release/debug and static/shared in one job to reduce the time. https://github.com/boostorg/math/pull/1061/files#diff-512f30627f95d7fcdc76dc362d896351e9a4539994b7ab00b41979e653a6d0a9R170 should give you an idea how. Otherwise feel free to ask.   
I guess it would even be enough to build just the static OR the shared variant for coverage which already cuts down the time in half. But you can also have one job each, codecov combines all uploads for a commit.

mborland:
> > Looks like our libraries.json was not formatted as expected. Everything is working now.
> 
> To be specific: You put your description into a list which seems to be valid and used by libraries containing sub-libraries. I don't see a reason to use this for single-library libraries where you can just use a dict (as you did in your change) I'm not sure how that JSON file is defined but seemingly that list has to be supported by Boost.CI -> [boostorg/boost-ci#225](https://github.com/boostorg/boost-ci/pull/225)

A release or two ago we went from having sub-libraries listed in the meta to just overall Boost.Math so that makes sense.

> 
> Currently your PR fails due to a timeout on GHA. You can possibly split it into multiple jobs instead of building release/debug and static/shared in one job to reduce the time. https://github.com/boostorg/math/pull/1061/files#diff-512f30627f95d7fcdc76dc362d896351e9a4539994b7ab00b41979e653a6d0a9R170 should give you an idea how. Otherwise feel free to ask. I guess it would even be enough to build just the static OR the shared variant for coverage which already cuts down the time in half. But you can also have one job each, codecov combines all uploads for a commit.

Thanks for the tips.

mborland:
@Flamefire so my tests are finishing in about an hour running the full test suite, but the upload step is timing out at 5 hours. In our normal CI we cut it in half github_ci_block_1 and 2. Is there a way to run these two test suites individually instead of together? Hopefully that'll reduce the upload time by half.

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,189 @@
...
+        run: source ci/github/install.sh
+
+      - name: Setup coverage collection
+        if: matrix.coverage
+        run: ci/github/codecov.sh "setup"
+
+      - name: Run tests
+        if: '!matrix.coverity'
+        run: ../../../b2 toolset=${{ matrix.compiler }} cxxstd=${{ matrix.cxxstd }} ${{ matrix.suite }} define=CI_SUPPRESS_KNOWN_ISSUES define=SLOW_COMPILER cxxflags=-fkeep-static-functions --coverage' linkflags=--coverage address-model=64 link=shared variant=debug -j5
+        working-directory: ../boost-root/libs/math/test
```
Flamefire, line 185:
```suggestion
        run: ci/build.sh
```

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,189 @@
...
+
+      - name: Setup Boost
+        env:
+          B2_ADDRESS_MODEL: ${{matrix.address-model}}
+          B2_COMPILER: ${{matrix.compiler}}
+          B2_CXXSTD: ${{matrix.cxxstd}}
+          B2_SANITIZE: ${{matrix.sanitize}}
+          B2_STDLIB: ${{matrix.stdlib}}
+          # More entries can be added in the same way, see the B2_ARGS assignment in ci/enforce.sh for the possible keys.
+          # B2_DEFINES: ${{matrix.defines}}
```
Flamefire, line 170:
```suggestion
          B2_DEFINES: CI_SUPPRESS_KNOWN_ISSUES=1 SLOW_COMPILER=1
```

Flamefire (commented):
> @Flamefire so my tests are finishing in about an hour running the full test suite, but the upload step is timing out at 5 hours. In our normal CI we cut it in half github_ci_block_1 and 2. Is there a way to run these two test suites individually instead of together? Hopefully that'll reduce the upload time by half.

5 hours... Never thought that collecting the coverage data could take so long.

Yes splitting the 2 suites might help, although it seems to be the amount of files included which is the issue, but let's try.

https://github.com/boostorg/math/pull/1061/commits/d1a5ae0902b5f4459b314f6a0e2b6565d3aab7aa was very close already. It just forgot that b2 runs in the boost root which enables some more "magic"/avoids known issues.

Check if the suggested changes work

PR #1062, opened 2023-12-22
Differential evolution

NAThompson:


include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace boost::math::tools {
+
+namespace detail {
+template <class T>
+concept subscriptable = requires(const T &c) { c[0]; };
+}
```
mborland, line 30: Can this be replaced with enable_if to allow for C++14?

include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+           ArgumentContainer* initial_guess= nullptr,
+           std::invoke_result_t<Func, ArgumentContainer> target_value = std::numeric_limits<std::invoke_result_t<Func, ArgumentContainer>>::quiet_NaN(),
+           std::atomic<bool> *cancellation = nullptr,
+           std::vector<std::pair<ArgumentContainer, std::invoke_result_t<Func, ArgumentContainer>>>
+               *queries = nullptr,
+          std::atomic<std::invoke_result_t<Func, ArgumentContainer>>* current_minimum_cost = nullptr
+         ) {
+    constexpr bool has_resize = requires(ArgumentContainer& t) {
+        t.resize(std::declval<typename ArgumentContainer::size_type>());
+    };
```
mborland, line 122: I don't think you truly need the requirement here.

include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+    using ResultType = std::invoke_result_t<Func, ArgumentContainer>;
+    using std::clamp;
+    using std::round;
+    using std::isnan;
+    using Size_t = ArgumentContainer::size_type;
```
mborland, line 127: I would rename this to avoid the confusion of just differing by one capital letter. Maybe `container_size_t`?

include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+    }
+    // Why don't we provide an option to initialize with (say) a Gaussian distribution?
+    // > If the optimum's location is fairly well known,
+    // > a Gaussian distribution may prove somewhat faster, although it
+    // > may also increase the probability that the population will converge prematurely.
+    // > In general, uniform distributions are preferred, since they best reflect
+    // > the lack of knowledge about the optimum's location.
+    //  - Differential Evolution: A Practical Approach to Global Optimization
+    using std::uniform_real_distribution;
+    uniform_real_distribution<Real> dis(Real(0), Real(1));
```
mborland, line 153:
```suggestion
    std::uniform_real_distribution<Real> dis(Real(0), Real(1));
```

There should be no legal replacements for std::uniform_real_distribution

include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+    for (size_t i = 0; i < cost.size(); ++i) {
+      cost[i] = cost_function(population[i]);
+      if (!isnan(target_value) && cost[i] <= target_value) {
+         target_attained = true;
+      }
+      if (current_minimum_cost && cost[i] < *current_minimum_cost) {
+          *current_minimum_cost = cost[i];
+      }
+      if (queries) {
+        #pragma omp critical
```
mborland, line 175: There does not seem to be an associated parallel region?

include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+      for (Size_t j = 0; j < bounds_.size(); ++j) {
+        auto const &bound = bounds_[j];
+        population[i][j] = bound[0] + dis(g) * (bound[1] - bound[0]);
+      }
+    }
+    if (initial_guess) {
+        population[0] = *initial_guess;
+    }
+
+    std::atomic<bool> target_attained = false;
```
mborland, line 164: Is there a way to signal or use a barrier? I think you're going to trash caches with this if it's not a builtin to OMP.

include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+              // We'll just have to wait until the end of this generation.
+          }
+          if (current_minimum_cost && cost[i] < *current_minimum_cost) {
+              *current_minimum_cost = cost[i];
+          }
+        }
+      }
+    }
+
+    auto it = std::min_element(cost.begin(), cost.end());
```
mborland, line 247: If you're committing to C++20, std::min_element can take an execution policy argument. It should be able to use the same thread pool as OMP if implemented with TBB.

include/boost/math/tools/differential_evolution.hpp
```diff
@@ -0,0 +1,261 @@
...
+// See:
+// https://www.cp.eng.chula.ac.th/~prabhas//teaching/ec/ec2012/storn_price_de.pdf
+template <typename BoundsContainer>
+  requires detail::subscriptable<BoundsContainer>
+class differential_evolution {
+public:
+  using BoundType = BoundsContainer::value_type;
+  using Real = BoundType::value_type;
+  differential_evolution(BoundsContainer bounds, Real F = 0.65, double crossover_ratio = 0.5, size_t NP = 200,
+                         size_t max_generations = 1000, size_t threads = std::thread::hardware_concurrency())
```
mborland, line 43:
```suggestion
                         size_t max_generations = 1000, int threads = std::thread::hardware_concurrency())
```

This would warn of implicit unsigned to signed conversion.

mborland: You're right. I thought it was int but it's actually unsigned: https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency

jzmaddock:
I'm not sure I know enough about the problem domain to comment in detail (except that we do need this!), but I worry somewhat about the interface to `argmin`, specifically the need to explicitly specify the `ArgumentContainer` type via the ugly `obj.template argmin<ArgumentContainer>()` syntax (one bit of C++ I've never liked!).  

I wonder, is it reasonable to suppose that `ArgumentContainer` is the same type as the bounds?  That's not currently possible with the current interface, but if the type were templated on `ArgumentContainer` then the bounds could be two `ArgumentContainer` values: one for the minimum of each parameter, and one for the maximum.

Second question: does this need to be an object?  Which is to say, does it get reused and/or cache values?  Or should it be a free function?  I appreciate there are a lot of parameters though!

On the docs front, it would be useful to have at least one usage example in the docs, rather than just pointing to the test cases.

mborland:
What about both a free function that takes a structure that unpacks into one that just takes the params list like the scipy signature?

jzmaddock:
Some good suggestions in the review there... it's possible to take the parameter structure a step further, and create a function with named arguments:  https://pdimov.github.io/blog/2020/09/07/named-parameters-in-c20/  The drawback is that the parameters have to appear in the correct order, but defaulted ones can be omitted.

> Is it impossible? I wonder if we could say BoundsContainer = std::vector<std::array<Real,2>> then implies ArgumentContainer is std::vector<Real>, or if BoundsContainer= std::array<std::array<Real,2>, n> then ArgumentContainer must be std::array<Real, n>. Admittedly my template-fu is not sufficient to achieve this but it does seem doable.

Yes definitely doable I think, requires a traits class with a template-type template parameter to extract the contained type:

```
template <class Container> struct rebind_container;

template <template <typename...> class Container, class T1, class T2>
struct rebind_container<Container<T1, T2>>
{
   using type = Container<typename T1::value_type, T2>;
};
```

PR #1063, opened 2024-01-09
Fix race conditions in differential evolution

NAThompson:
Through a combination of silly mistakes, I missed a pile of race conditions in the OpenMP threading.

Switch to C++ threading. Note that this change requires serial generation of trial vectors.

Hopefully I can figure out to parallelize the generation of trial vectors to reduce the serial section a la Ahmdahl's law,
while simultaneously keeping thread sanitizer happy.


mborland:
> @jzmaddock : I wonder if we could revisit the "no unicode in comments" rule-I had to remove all the accents from the author's names which I found somewhat inelegant. Going a bit further, maybe something like:
> 
> ```c++
> template<typename ℝ>
> ```
> 
> could really improve readability. I know we can use `// boost-no-inspect`, but the other inspection rules are quite valuable.

My two cents is I think revisiting the no unicode in comments would be good, but adding unicode into the actual code is not. SO says it varies from formerly illegal to just unsupported now: https://stackoverflow.com/questions/26660180/are-unicode-and-special-characters-in-variable-names-in-clang-not-allowed.

jzmaddock:
I don't know if it's still the case (but I assume it is) that users on Windows in far eastern countries do not use Unicode character sets, and Unicode in source files (even in comments) renders them non-compilable.  If you search hard enough you'll find some old bug reports on this which crop up whenever Unicode has crept in accidentally.  Usually been my fault in the past, and usually for authors names as in your case here.  Whatever we should run this past the mailing list rather than making a unilateral change?

mborland (commented):
TSAN and Helgrind both seem happy now. My one concern is that codecov shows that all the regions that involve synchronization are not covered by tests. Is this due to GHA not having enough threads available? It would probably be good to cover all of your param checks, and testing of validating your initial guess.

Issue #1064, opened 2024-01-13
Loading XML file with float128 in it is broken.


cosurgi:
Hi,

Using boost 1.74 here. I cannot load an XML file with float128 in it. Saving works. Try this example:

```
#include <boost/cstdfloat.hpp> // must be the first include
#include <boost/multiprecision/float128.hpp>
#include <complex>
using Real    = boost::float128_t;
using Complex = std::complex<Real>;

#include <boost/algorithm/string.hpp>
#include <boost/archive/codecvt_null.hpp>
#include <boost/archive/xml_iarchive.hpp>
#include <boost/archive/xml_oarchive.hpp>
#include <boost/iostreams/device/file.hpp>
#include <boost/iostreams/filtering_stream.hpp>
#include <boost/serialization/complex.hpp>
#include <locale>
#include <string>

#if BOOST_VERSION >= 104700 // To correctly recognize serialization of Inf and NaN numbers, there are different includes for different boost versions
#include <boost/math/special_functions/nonfinite_num_facets.hpp>
#else
#include <boost/math/nonfinite_num_facets.hpp>
#endif
namespace IO {
template <class T, class oarchive> void save(std::ostream& ofs, const std::string& objectTag, T& object)
{
	std::locale default_locale(std::locale::classic(), new boost::archive::codecvt_null<char>);
	std::locale locale2(default_locale, new boost::math::nonfinite_num_put<char>);
	ofs.imbue(locale2);
	oarchive oa(ofs, boost::archive::no_codecvt);
	oa << boost::serialization::make_nvp(objectTag.c_str(), object);
	ofs.flush();
}
template <class T, class iarchive> void load(std::istream& ifs, const std::string& objectTag, T& object)
{
	std::locale default_locale(std::locale::classic(), new boost::archive::codecvt_null<char>);
	std::locale locale2(default_locale, new boost::math::nonfinite_num_get<char>);
	ifs.imbue(locale2);
	iarchive ia(ifs, boost::archive::no_codecvt);
	ia >> boost::serialization::make_nvp(objectTag.c_str(), object);
}
template <class T> void save(const std::string fileName, const std::string& objectTag, T& object)
{
	boost::iostreams::filtering_ostream out;
	out.push(boost::iostreams::file_sink(fileName));
	if (!out.good()) throw std::runtime_error("Error opening file " + fileName + " for writing.");
	save<T, boost::archive::xml_oarchive>(out, objectTag, object);
}
[...]

PR #1066, opened 2024-01-20
Improve cstdfloat iostream code.

jzmaddock:
Copies input stream parsing from multiprecision.
Fixes https://github.com/boostorg/math/issues/1064 This is still not "proper" iostream code, but it's much better than it was.


jzmaddock:
Failures are unrelated (will fix separately), merging...

PR #1067, opened 2024-01-21
Fix missing newline.

jzmaddock:


jzmaddock:
>Quick question: Do you have a .clang-format you like so we can automate some of these problems away?

Sorry no, I have found it can make a right mess of things if you're not careful, haven't tried it recently though...

PR #1071, opened 2024-01-23
Random search

NAThompson:


ckormanyos:
Hi @NAThompson I just found a random instance of something random that looked very close to expectation and might somehow be related to some of this recent work.

In a completely unrelated CI that had no association with this work, I think some runner found a random outlier [here](https://drone.cpp.al/boostorg/math/1777/240/2).

PR #1074, opened 2024-01-25
CMA-ES

NAThompson:


example/cma_es_example.cpp
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright Nick Thompson, 2024
+ * Use, modification and distribution are subject to the
+ * Boost Software License, Version 1.0. (See accompanying file
+ * LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+ */
+#if __APPLE__ || __linux__
+#include <signal.h>
```
mborland, line 8: This should be every POSIX system right? You could `#if __has_include(<unistd.h>)` which is a requirement of every POSIX system.

include/boost/math/optimization/cma_es.hpp
```diff
@@ -0,0 +1,387 @@
...
+    detail::validate_initial_guess(*params.initial_guess, params.lower_bounds, params.upper_bounds);
+  }
+  const size_t n = params.upper_bounds.size();
+  // Equation 48 of the arxiv review:
+  if (params.population_size == 0) {
+    //auto tmp = 4.0 + floor(3*log(n));
+    // But round to the nearest multiple of the thread count:
+    //auto k = static_cast<size_t>(std::ceil(tmp/params.threads));
+    //params.population_size = k*params.threads;
+    params.population_size = static_cast<size_t>(4.0 + floor(3*log(n)));
```
mborland, line 78:
```suggestion
    params.population_size = static_cast<size_t>(4 + floor(3*log(n)));
```

Some of the `<stdfloat>` types will warn/error on trying to add a double.

include/boost/math/optimization/cma_es.hpp
```diff
@@ -0,0 +1,387 @@
...
+  // mu:= number of parents, lambda := number of offspring.
+  Eigen::Matrix<Real, Eigen::Dynamic, Eigen::Dynamic> C = Eigen::Matrix<Real, Eigen::Dynamic, Eigen::Dynamic>::Identity(n, n);
+  ArgumentContainer mean_vector;
+  Real sigma = 1;
+  if (params.initial_guess) {
+    mean_vector = *params.initial_guess;
+  }
+  else {
+    // See the footnote in Table 1 of the arxiv review:
+    sigma = 0.3*(params.upper_bounds[0] - params.lower_bounds[0]);
```
mborland, line 192:
```suggestion
    sigma = Real(0.3)*(params.upper_bounds[0] - params.lower_bounds[0]);
```

mborland (commented):
Without having read the paper here are my trivial comments. The only other thing is there's like zero white space which makes it kind of hard to read. Maybe empty lines before and after outer loops, ifs, equation references, etc would improve readability.

PR #1078, opened 2024-02-01
Avoid unused variable warning in ibeta_inverse

andrjohns:
Using the `ibeta_inverse` function currently gives an unused variable warning at compile time since there is a variable that is only used in an assert:

```cpp
      const T x_extrema = 1 / (1 + a);
      BOOST_MATH_ASSERT(0 < x_extrema && x_extrema < 1);
```

This can be avoided (and a few operations saved) by checking whether `a` is bounded by the values which would result in `x_extrema` failing the assert:

```cpp
      BOOST_MATH_ASSERT(
         std::numeric_limits<T>::epsilon() < a
         && a < std::numeric_limits<T>::infinity());
```

### DIFF
--- a/include/boost/math/special_functions/detail/ibeta_inverse.hpp
+++ b/include/boost/math/special_functions/detail/ibeta_inverse.hpp
@@ -15,6 +15,8 @@
 #include <boost/math/special_functions/erf.hpp>
 #include <boost/math/tools/roots.hpp>
 #include <boost/math/special_functions/detail/t_distribution_inv.hpp>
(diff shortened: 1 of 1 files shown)


jzmaddock:
It will need to be `(boost::math::tools<T>::epsilon)()` to avoid the errors inside `assert` I think.

andrjohns:
@jzmaddock apologies for the false starts, this should be ready for review now

jzmaddock:
Thanks @andrjohns - looks good to me, merging...

Issue #1077, opened 2024-01-31
gauss_kronrod does not allow for non-literal types


SteveBronder:
Using `gauss_kronrod` with a non-literal type causes a compiler error.

```
lib/boost_1.81.0/boost/math/quadrature/gauss_kronrod.hpp:139:41: error: the type ‘const std::array<stan::math::var_value<double>, 8>’ of ‘constexpr’ variable ‘data’ is not literal
  139 |       static constexpr std::array<T, 8> data = {
```

Example godbolt here

https://godbolt.org/z/K4Mxa85Mz

I think a fix would be to detect if the type is literal for the gauss kronrod detail and have one implimentation that uses static constexpr arrays and the other uses arrays constructed at run time

SteveBronder:
It's a scalar autodiff type used in the stan math library. Looking at the definitions for [LiteralType](https://en.cppreference.com/w/cpp/named_req/LiteralType) it fails 

> a type with at least one constexpr (possibly template) constructor that is not a copy or move constructor,

I think for our purposes if `gauss_kronrod` allowed you to choose types for the weights/abscicca and a seperate type for the given functions return type then it would work for us and also fix the problem

jzmaddock:
> Any ideas of how to support this?

Maybe, I think some experimentation might be needed though...

SteveBronder:
> But the arena allocator . . . that seems more difficult to support

The arena allocator is at global scope so it's not an issue. 

I think just making a gauss kronrod detail that has something like `std::enable_if_t<is_literal<T>::value>` and another with the opposite would fix it

jzmaddock:
> I think just making a gauss kronrod detail that has something like std::enable_if_t<is_literal<T>::value> and another with the opposite would fix it

That involves a huge amount of duplication: I have a simpler solution working here, but I need to see if I can reduce bloat even further...

PR #1079, opened 2024-02-01
Update pre-computed constexpr Gauss and Gauss-Kronrod constants

jzmaddock:
to always store literal types.
Fixes: https://github.com/boostorg/math/issues/1077.


SteveBronder:
How would you feel about something like the below? I wrote a `is_constexpr_constructible()` function that is used to deduce which `gauss_detail` to use. If `Real`'s constant type is constructible from a long double at compile time (which I think is true for `float`, `double`, and `long double`)  then it uses the one that is constexpr. Otherwise it uses one that constructs the values at runtime. 

Looking at the version here I think something bad would happen across if we ran this once, cleaned up the arena, and tried running again. I'm always bad at the intricacies of static types in functions, but if they are declared once for the life of the program and we clean up the memory arena then I think the next time we run this the program would do UB accessing memory that has been cleaned up or reallocated.

https://godbolt.org/z/hxWMGG4rG

include/boost/math/quadrature/gauss_kronrod.hpp
```diff
@@ -226,7 +161,7 @@ class gauss_kronrod_detail<T, 15, 4>
 public:
    static  std::array<T, 8> const & abscissa()
    {
-      static  std::array<T, 8> data = {
+      static  std::array<T, 8> data = { // LCOV_EXCL_START
```
SteveBronder, line 164: Is `LCOV_EXCL_START` documented somewhere?

include/boost/math/quadrature/gauss_kronrod.hpp
```diff
@@ -82,138 +82,73 @@ class gauss_kronrod_detail;
 
 #endif
 
+#ifndef BOOST_HAS_FLOAT128
 template <class T>
 class gauss_kronrod_detail<T, 15, 0>
```
SteveBronder, line 87: Do you need to have the final template parameter be an integer or can you just use the `gauss_constant_category<T>::storage_type` as the final template parameter?

include/boost/math/quadrature/gauss.hpp
```diff
@@ -29,14 +29,37 @@ struct gauss_constant_category
...
       (std::numeric_limits<T>::radix == 2) ?
       (
-         (std::numeric_limits<T>::digits <= std::numeric_limits<float>::digits) && std::is_convertible<float, T>::value ? 0 :
-         (std::numeric_limits<T>::digits <= std::numeric_limits<double>::digits) && std::is_convertible<double, T>::value ? 1 :
-         (std::numeric_limits<T>::digits <= std::numeric_limits<long double>::digits) && std::is_convertible<long double, T>::value ? 2 :
 #ifdef BOOST_HAS_FLOAT128
-         (std::numeric_limits<T>::digits <= 113) && std::is_constructible<__float128, T>::value ? 3 :
+         (std::numeric_limits<T>::digits <= 113) && std::is_constructible<__float128, T>::value ? 0 :
+#else
+         (std::numeric_limits<T>::digits <= std::numeric_limits<long double>::digits) && std::is_convertible<long double, T>::value ? 0 :
```
SteveBronder, line 35: Is there a reason to have the value now that you deduce the storage type here?

include/boost/math/quadrature/gauss.hpp
```diff
@@ -74,106 +97,56 @@ template <class Real, unsigned N, unsigned Category>
...
-      static constexpr std::array<T, 4> data = {
-         0.000000000e+00f,
-         4.058451514e-01f,
-         7.415311856e-01f,
-         9.491079123e-01f,
+      static constexpr std::array<storage_type, 4> data = {
+         static_cast<storage_type>(0.00000000000000000000000000000000000e+00L),
+         static_cast<storage_type>(4.05845151377397166906606412076961463e-01L),
+         static_cast<storage_type>(7.41531185599394439863864773280788407e-01L),
+         static_cast<storage_type>(9.49107912342758524526189684047851262e-01L),
```
SteveBronder, line 112: What is the reasoning behind putting have the function return a const& to a constexpr type? i.e. why can't the signature just be `constexpr`?

include/boost/math/quadrature/gauss.hpp
```diff
@@ -29,14 +29,37 @@ struct gauss_constant_category
...
+#else
+         (std::numeric_limits<T>::digits <= std::numeric_limits<long double>::digits) && std::is_convertible<long double, T>::value ? 0 :
 #endif
          (std::numeric_limits<T>::digits10 <= 110) ? 4 : 999
       ) : (std::numeric_limits<T>::digits10 <= 110) ? 4 : 999;
+   
+   using storage_type =
+      typename std::conditional<(std::numeric_limits<T>::is_specialized == 0), T,
+         typename std::conditional<(std::numeric_limits<T>::radix == 2),
+            typename std::conditional< ((std::numeric_limits<T>::digits <= std::numeric_limits<float>::digits) && std::is_convertible<float, T>::value),
```
SteveBronder, line 43: The math library requires a c++14 compiler now so you should be able to use `std::conditional_t` now to remove the typename and ::type

SteveBronder (commented):
Just a few small comments. I'm not really following how this fixes https://github.com/boostorg/math/issues/1077

include/boost/math/quadrature/gauss_kronrod.hpp
jzmaddock: It's used by the lcov tool when collecting code coverage data (ie it's documented by the lcov tool).  For some reason GCC has trouble instrumenting arrays which are dynamically initialized, with only the final value being marked as covered, so we have to exclude those lines from coverage analysis.

include/boost/math/quadrature/gauss_kronrod.hpp
jzmaddock: Not quite, we now have 3 cases, two of which both use T as storage_type.

include/boost/math/quadrature/gauss.hpp
jzmaddock: I don't think functions returning arrays are allowed?  But also just because the function is marked as constexpr, does not mean that it will be evaluated as such, only that it can be.  We could probably mark the existing signatures constexpr, but I'm not sure it actually makes any practical difference - either it's inlined and optimised away or not, either which way, we can never actually call the function in a constexpr context since some overloads (for the multiprecision cases) cannot be constexpr.

include/boost/math/quadrature/gauss.hpp
jzmaddock: As above.

include/boost/math/quadrature/gauss.hpp
jzmaddock: Done.

jzmaddock:
I hope this should be good to go: @SteveBronder can you try this PR against your actual code?

SteveBronder:
Thanks I'll give this a whirl on Monday!

include/boost/math/quadrature/gauss.hpp
SteveBronder: > I don't think functions returning arrays are allowed?

In c++14? That should be fine

> But also just because the function is marked as constexpr, does not mean that it will be evaluated as such, only that it can be.

If a function can be constexpr idt the compiler gets a choice on whether to make it constexpr or not. At least reading over the constexpr function spec from cppreference I don't see anything that says that constexpr function invocations are compiler defined.

> either it's inlined and optimised away or not, either which way, 

Making those constexpr removes the static here which imo does matter for user defined types

> we can never actually call the function in a constexpr context since some overloads (for the multiprecision cases) cannot be constexpr.

See the constexpr function docs section [here](https://en.cppreference.com/w/cpp/language/constexpr). For any of the overloads that can be constexpr this will be.

SteveBronder:
This works!

include/boost/math/quadrature/gauss.hpp
jzmaddock: >Making those constexpr removes the static here which imo does matter for user defined types

Right, we do actually want the functions static in the general case (since there is no need for a `this` pointer).  In any case, unless someone demonstrates that substantially better code can be produced with a different signature, I'll leave as is for now.

jzmaddock:
> Looks like the build failure is unrelated-want to merge?

Yup, merging...

PR #1085, opened 2024-02-11
Try LCOV Exclude on BIG_CONSTANT

ckormanyos:
The purpose of this PR is to see if `LCOV_EXCL_LINE` can/should simply be applied to big-constants and if this simple change _catches_ all of these lines. Also we ponder if this is philosophically the right fix for big-constants?


ckormanyos:
This is not the right way. Close PR and perform further local investigations

PR #1086, opened 2024-02-12
Linear BIG_CONST macros and try lcov-excl

ckormanyos:
The purpose of this PR is to streamline/linearize macros sich as `BOOST_MATH_BIG_CONSTANT` and subsequently check if line-by-line lcov-exclude makes sense.


ckormanyos:
Hi @jzmaddock and @mborland and @NAThompson this work might seem ugly/controversial/hurried. But I looked into simplifying the big-constant macros such that they can be neatly expressed as single-line macros. This allows for simple line-by-line `LCOV_EXCL_LINE` on lines that (should be) rather well-known to be covered but which `lcov` seems to be missing (maybe due to optimization settings).

Anyway, it is a method, but might be controversial. Comments? Thoughts?

ckormanyos:
I've tried (but not yet succeeded) to directly put the code annotation into the macros themselves.

ckormanyos:
One thing I did notice is that when the couple hundred annotations are in place, it really draws focus to the several hundred remaining specfun missed lines that are actually much more, let's say, interesting and content-rich. So feedback would be appreciated, ...

Cc: @jzmaddock and @mborland  and @NAThompson

mborland (approved):
Looks good to me. 200+ lines removed does help.

ckormanyos:
> Looks good to me. 200+ lines removed does help.

Thanks Matt. It ended up being closer to 1,000 lines. I could not find any other large, missed tables in specfun. But some were, in fact, covered. I annotated the ones I found must have been covered but were seemingly mistakenly marked by `lcov` as uncovered.

ckormanyos:
Here are the remaining trouble spots:
- specfun
- distributions
- octonion.hpp

This is a really great achievement so far in Math!

grafik

mborland:
> Here are the remaining trouble spots:
> * octonion.hpp

The one test file doesn't have real testing in it: https://github.com/boostorg/math/blob/develop/test/octonion_test.cpp.

ckormanyos:
> The one test file doesn't have real testing in it:

I suppose we might as well just exclude it from the `lcov`analysis at the command_line-level?

ckormanyos:
>> The one test file doesn't have real testing in it:

> I suppose we might as well just exclude it from the `lcov` analysis at the command_line-level?

On second thought, that is too hasty. This is our one-time chance to actually get some octernion testing. Or notice that they fail. I think we should try for tests. I can give this one a crack in another branch later, ...

jzmaddock:
Forgive me, late night last night... but BOOST_MATH_BIG_CONSTANT is now being called with 2 arguments, but still defined with 3?  I know I'm missing something... !

BTW the basic issue is that multi-line statements are badly handled by lcov, so there are quite a few function calls that look like:

```
foo(
   a, 
   b,
   c);
```

especially when the arguments are long error messages... and these are marked as uncovered except for the last line.  The issue with the arrays are the same, it's not the macros as such that cause the issue, just that we have dynamic initialization spread over several lines.  For the functions I've been collapsing down to one line where possible, and using LCOV_EXCL_LINE when not.  For the arrays, I've topped and tailed the list of initializers with LCOV_EXCL_START/LCOV_EXCL_STOP.

BTW I have just started on coverage for beta, hoping to get to erf or so in the next tranche, I am going to be super busy this week though unfortunately :(

ckormanyos:
Hi John, thanks for those explanations.

>  but `BOOST_MATH_BIG_CONSTANT` is now being called with 2 arguments, but still defined with 3?

Not exaclty, but there are such changes in this PR. That's me, messing with stuff.

I left the number of parameters of `BOOST_MATH_BIG_CONSTANT`to be 3. And it is still called in the PR with 3.

The other mega-macro, `BOOST_MATH_HUGE_CONSTANT` originally had 3 parameters, but the digits parameter was used nowhere. So I removed it.

> For the arrays, I've topped and tailed the list of initializers with `LCOV_EXCL_START`/`LCOV_EXCL_STOP`

I can do that and eliminate the thousand or so single-line annotations in this PR. I did, however, diligently look if I felt that they were covered prior to x-ing them out. So I'm good with that.

Also, I kind of like the way I refactored `BOOST_MATH_BIG_CONSTANT` and `BOOST_MATH_HUGE_CONSTANT`, as I think the intermediary inline function add clarity. Sorry about silently going from 3 to 2 params on `_HUGE`, but the unused param was in my mind simply confusing.

ckormanyos:
oops. I did mess up some of the calls to `BOOST_MATH_BIG_CONSTANT`. My bad. I will repair all these.

ckormanyos:
> oops. I did mess up some of the calls to `BOOST_MATH_BIG_CONSTANT`. My bad. I will repair all these.

I'll finish this up (whatever I started) but just let it sit there, as it seems like it was a bit _hurried_.

@jzmaddock if you want some, none, or all of it, pls let me know. It was/is a good exercise.

ckormanyos:
> I'll finish this up (whatever I started) but just let it sit there, as it seems like it was a bit _hurried_.

I do, in fact, like the single-line expression(s) of `BOOST_MATH_BIG_CONSTANT`/`BOOST_MATH_HUGE_CONSTANT`. This allows these to be `lcov`-excluded on single-line-annotations.

The three-parameters of `_HUGE` are controversial. But if any part of this is cool, it's that work. Or any part of it.

Anyway, I'll let this one _bake_ for a while.

ckormanyos:
I converted this PR to a draft. I'm not going to really dive into this any time soon, but I synchronized with Develop and it now shows some of the tabble/macro work that subjectively seemed to make sense to me.

PR #1087, opened 2024-02-12
Get something of octonion tested

ckormanyos:


ckormanyos:
So what I'm finding out is that octonions are fascination. I believe there is a relation to them and order-2 Lie groups. Long on my TODO list and i have no clue about them.

Anyway, the original file [octonion_test.cpp](https://github.com/boostorg/math/blob/develop/test/octonion_test.cpp) in my opinion has more of an exemplary nature. it is described in Math's docs also as such. We can annotate it and potentially rename it as an example.

For actual testing, ... I have in the branch started a new file called [octonion_test_simple.cpp](https://github.com/boostorg/math/blob/try_testing_octonion/test/octonion_test_simple.cpp). It is reminiscent of the original file but its focus will be on testing the actual octonion implementation more clearly and achieving coverage.

I need to study up and get more test data. Python does a lot of this stuff. I'd like some data poinnts for a few basic elementary functions too like `exp()`, etc. So this will be a cool little hobby...

ckormanyos:
> Wasn't there a discussion about deprecating that file since there is now a qvm library in boost?

I'd like that. It seems like legacy code with old style.

I could in this PR:
- Formulate a depriction notice (till 1.86).
- Deprecate octernion and quaternion? Or just octernion?
- Eliminate negative influence from this stuff in CodeCov.

I would need help with the docs upon actual deprecation. But gone seems good on this one.

I just need all-party agreement.

Cc: @jzmaddock and @mborland abd @NAThompson

mborland:
I don't think QVM supports Octonion so we shouldn't deprecate this without replacement. We should add testing because it's already here.

ckormanyos:
> I don't think QVM supports Octonion so we shouldn't deprecate this without replacement. We should add testing because it's already here.

I was afraid you'd say that. GSoC 2030 make octonion great again

ckormanyos:
Ok this is better than it was.

- But we still need a couple hundred lines of octonian coverage.
- I also disabled `lcov`for Input-Streaming of octonians.

We can sort this octonian thing out in a few later rounds.

For now I'll merge, as it's better than earlier.

PR #1090, opened 2024-02-13
Achieve more octonion coverage

ckormanyos:


ckormanyos:
OK folks we hit about $70\\%$ of octonion.hpp. I will return to this when priorities get higher.

PR #1092, opened 2024-02-13
Run and cover compile-test instantiation

ckormanyos:
The purpose of this PR is to cover the file [instantiate.hpp](https://github.com/boostorg/math/blob/develop/test/compile_test/instantiate.hpp).

The file [instantiate.hpp](https://github.com/boostorg/math/blob/develop/test/compile_test/instantiate.hpp) has been extremely useful for compilation-syntactical correctness.

It is/was not all that far away from actually running with `boost::float64_t`, which is exercised (with a few tens of modifications in the header) in this PR.

### DIFF
--- a/include/boost/math/distributions/cauchy.hpp
+++ b/include/boost/math/distributions/cauchy.hpp
@@ -17,8 +17,9 @@
 #include <boost/math/constants/constants.hpp>
 #include <boost/math/distributions/complement.hpp>
 #include <boost/math/distributions/detail/common_error_handling.hpp>
-#include <utility>
+
 #include <cmath>
+#include <utility>
 
 namespace boost{ namespace math
 {
--- a/include/boost/math/distributions/cauchy.hpp
+++ b/include/boost/math/distributions/cauchy.hpp
@@ -276,29 +276,30 @@ inline RealType quantile(const complemented2_type<cauchy_distribution<RealType, 
 
 template <class RealType, class Policy>
 inline RealType mean(const cauchy_distribution<RealType, Policy>&)
-{  // There is no mean:
-   typedef typename Policy::assert_undefined_type assert_type;
-   static_assert(assert_type::value == 0, "assert type is undefined");
-
-   return policies::raise_domain_error<RealType>(
-      "boost::math::mean(cauchy<%1%>&)",
-      "The Cauchy distribution does not have a mean: "
-      "the only possible return value is %1%.",
-      std::numeric_limits<RealType>::quiet_NaN(), Policy());
+{
+   // There is no mean:
+   return
+      policies::raise_domain_error<RealType>
+      (
+         "boost::math::mean(cauchy<%1%>&)",
+         "The Cauchy distribution does not have a mean: "
+         "the only possible return value is %1%.",
+         std::numeric_limits<RealType>::quiet_NaN(), Policy()
+      );
 }
 
 template <class RealType, class Policy>
 inline RealType variance(const cauchy_distribution<RealType, Policy>& /*dist*/)
 {
    // There is no variance:
-   typedef typename Policy::assert_undefined_type assert_type;
-   static_assert(assert_type::value == 0, "assert type is undefined");
-
-   return policies::raise_domain_error<RealType>(
-      "boost::math::variance(cauchy<%1%>&)",
-      "The Cauchy distribution does not have a variance: "
-      "the only possible return value is %1%.",
-      std::numeric_limits<RealType>::quiet_NaN(), Policy());
+   return
+      policies::raise_domain_error<RealType>
+      (
+         "boost::math::variance(cauchy<%1%>&)",
+         "The Cauchy distribution does not have a variance: "
+         "the only possible return value is %1%.",
+         std::numeric_limits<RealType>::quiet_NaN(), Policy()
+      );
 }
 
 template <class RealType, class Policy>
@@ -316,42 +317,42 @@ template <class RealType, class Policy>
 inline RealType skewness(const cauchy_distribution<RealType, Policy>& /*dist*/)
 {
    // There is no skewness:
-   typedef typename Policy::assert_undefined_type assert_type;
-   static_assert(assert_type::value == 0, "assert type is undefined");
-
-   return policies::raise_domain_error<RealType>(
-      "boost::math::skewness(cauchy<%1%>&)",
-      "The Cauchy distribution does not have a skewness: "
-      "the only possible return value is %1%.",
-      std::numeric_limits<RealType>::quiet_NaN(), Policy()); // infinity?
+   return
+      policies::raise_domain_error<RealType>
+      (
+         "boost::math::skewness(cauchy<%1%>&)",
+         "The Cauchy distribution does not have a skewness: "
+         "the only possible return value is %1%.",
+         std::numeric_limits<RealType>::quiet_NaN(), Policy()
+      );
 }
 
 template <class RealType, class Policy>
 inline RealType kurtosis(const cauchy_distribution<RealType, Policy>& /*dist*/)
 {
    // There is no kurtosis:
-   typedef typename Policy::assert_undefined_type assert_type;
-   static_assert(assert_type::value == 0, "assert type is undefined");
-
-   return policies::raise_domain_error<RealType>(
-      "boost::math::kurtosis(cauchy<%1%>&)",
-      "The Cauchy distribution does not have a kurtosis: "
-      "the only possible return value is %1%.",
-      std::numeric_limits<RealType>::quiet_NaN(), Policy());
+   return
+      policies::raise_domain_error<RealType>
+      (
+         "boost::math::kurtosis(cauchy<%1%>&)",
+         "The Cauchy distribution does not have a kurtosis: "
+         "the only possible return value is %1%.",
+         std::numeric_limits<RealType>::quiet_NaN(), Policy()
+      );
 }
 
 template <class RealType, class Policy>
 inline RealType kurtosis_excess(const cauchy_distribution<RealType, Policy>& /*dist*/)
 {
    // There is no kurtosis excess:
-   typedef typename Policy::assert_undefined_type assert_type;
-   static_assert(assert_type::value == 0, "assert type is undefined");
-
-   return policies::raise_domain_error<RealType>(
-      "boost::math::kurtosis_excess(cauchy<%1%>&)",
-      "The Cauchy distribution does not have a kurtosis: "
-      "the only possible return value is %1%.",
-      std::numeric_limits<RealType>::quiet_NaN(), Policy());
+   return
+      policies::raise_domain_error<RealType>
+      (
+         "boost::math::kurtosis_excess(cauchy<%1%>&)",
+         "The Cauchy distribution does not have a kurtosis: "
+         "the only possible return value is %1%.",
+         std::numeric_limits<RealType>::quiet_NaN(), Policy()
+      );
 }
 
 template <class RealType, class Policy>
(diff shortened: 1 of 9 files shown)


ckormanyos:
OK this one was very difficult for me.

It clears up a lot of chaff.
- Start running the many lines of [`instantiate.hpp`](https://github.com/boostorg/math/blob/develop/test/compile_test/instantiate.hpp). This required a bit of refactoring and range/memory checking.
- Find and solve problems with `distributions`. If a diestibution's `policy` does not have certain attributes, then these can't be checkrd for.

I left about 500 more line in [`instantiate.hpp`](https://github.com/boostorg/math/blob/develop/test/compile_test/instantiate.hpp) still needing coverage. These are in the so-called _mixed_ instantiations. But now that I figured out how this thing ticks and where the silly problems were, I#ll handle those any time in the future with ease. Leave thos 500 lines to me.

Together with the minimalistic [`octonion measures`](https://github.com/boostorg/math/pull/1090) we are poinsed to get into the nineties of coverage, revealing the real and actual things-of-content that we need to address, instead of the chaff, which I've almost cleared out by now...

This thingy needs to cycle green, then party.

ckormanyos:
My personal best (or worst) 14 failed CI runs 'till I dialed it.

ckormanyos:
By the way, this drone run is getting so bad that it's almost worse-than-neutral. It fails so often, that I disregard it almost by self-preservatoin, thereby nearly missing the few things-of-substance it finds. I'll think on it, if there is a better way to _reach_ some of those old/arcane compilers. Or we just have to seriously parse those outputs, which I do not like. I like green being green and red being red.

ckormanyos:
Failures in drone are failures to obtain compiler

include/boost/math/distributions/cauchy.hpp
```diff
@@ -275,29 +276,30 @@ inline RealType quantile(const complemented2_type<cauchy_distribution<RealType,
 
 template <class RealType, class Policy>
 inline RealType mean(const cauchy_distribution<RealType, Policy>&)
-{  // There is no mean:
```
jzmaddock, line 278: Apologies for being absent while this was being developed, just trying to catch up now....

We really shouldn't remove the static_asserts here: the point is that the cauchy distribution has no mean, so calling `mean` on such a distribution is almost certainly a programming error that should be caught early (rather than at runtime).  There are a small number of situations where we're writing generic software which "works with any distribution" where we can change the Policy as a "get out of jail free card" to convert the compile time error to a runtime one.

ckormanyos: I'm really glad to hear from you John. I was extremelty unsure about this. I came to the conclusion that (via default template parameters) the exact `policy` in question that was being used did not actually have any of the attributes needed for the `static_assert`.

These `static_assert`s are being asserted on [`defualt_policy`](https://github.com/boostorg/math/blob/d4a42f930e92cc2ad4ca0d99d323f1159493c4ae/include/boost/math/policies/policy.hpp#L223), which is empty.

So that thing either needs content, or the static assertions are bogus.

I'm not entirely sure of this, but seem to have gleaned this from the code, ...

jzmaddock: To be clear, the default is that these functions *do* static_assert, we should have added a test to that effect.... somewhere lost in the depths of time, I remember Paul and I having a long discussion about this, the "obvious" thing to do was to simply not provide functions for mean etc when those are undefined, but that then prevents the option of writing generic code.  So these need either a policy that contains `assert_undefined<false>`, or we need a `#define BOOST_MATH_ASSERT_UNDEFINED_POLICY false`.

include/boost/math/special_functions/detail/bessel_y0.hpp
```diff
@@ -145,6 +145,8 @@ T bessel_y0(T x, const Policy&)
 
     static const char* function = "boost::math::bessel_y0<%1%>(%1%,%1%)";
 
+    static_cast<void>(function[0U]);
```
jzmaddock, line 148: This can just be removed I think?

include/boost/math/distributions/cauchy.hpp
jzmaddock: Ahh, instantate.hpp includes the define at it's start, but that will not be triggered if policy.hpp has been included first... so I'm guilty of some fragile programming there :(

ckormanyos: We could make the default policy have a bunch of constexpr stuff to pass static asserts. Or we could handle on the top layer with compiler switches.

I expect we will find several things of this nature as we press on toward ninety five, more percent.

This code base is awesome. Rarely, have I covereed stuff with such depth and detail and established nature and found such deep thought going into it.

But we will find a few thingy-ies needing tweaks and patches toward, 93, 95, 98, 99

include/boost/math/special_functions/detail/bessel_y0.hpp
ckormanyos: I'm ok with syntax stuff. I tend to over-cast and handle warnings like a maniac. And thus (as some would say) obfuscate. I'll get back to this later after we hash out these basics.

include/boost/math/distributions/cauchy.hpp
ckormanyos: > Ahh, `instantate.hpp` includes the define at it's start, but that will not be triggered if `policy.hpp` has been included first, ...

Yeah! That is what I was missing. I studied that code for a while but did not dial that nuance.

Can we repair this in this running PR? Or should i take another run at it. I really got a good start at this thing...

ckormanyos: In other words @jzmaddock I'm looking to get about 1,500 lines covered if we can get these istantiations compiling and running in all situations. We are close. Maybe I need to include that policy and then restore those `static_asserts`. We are close to that in my latest PR.

test/compile_test/instantiate.hpp
```diff
@@ -308,13 +332,26 @@ void instantiate(RealType)
    boost::math::sinc_pi(v1);
    boost::math::sinhc_pi(v1);
    boost::math::asinh(v1);
-   boost::math::acosh(v1);
+   {
+      using std::fabs;
+
+      auto v_special { v1 };
```
jzmaddock, line 338: Why not just set v_special to something specific that's in domain for acosh?  The fabs is unused as well.

test/compile_test/instantiate.hpp
```diff
@@ -343,7 +380,20 @@ void instantiate(RealType)
...
    boost::math::lambert_w0(v1);
-   boost::math::lambert_wm1(v1);
+   {
+      using std::fabs;
+
+      auto v_special { v1 };
+
+      using std::exp;
+
+      while(v_special > exp(RealType(-1)))
```
jzmaddock, line 390: As above, why not just pick a value?

test/compile_test/instantiate.hpp
```diff
@@ -409,9 +467,21 @@ void instantiate(RealType)
 
    boost::math::unchecked_bernoulli_b2n<RealType>(i);
    boost::math::bernoulli_b2n<RealType>(i);
-   boost::math::bernoulli_b2n<RealType>(i, i, &v1);
+   {
+      RealType* v1_array = new RealType[i];
```
jzmaddock, line 471: Some scoped storage would be good ;)

test/compile_test/instantiate.hpp
```diff
@@ -409,9 +467,21 @@ void instantiate(RealType)
...
+
+      boost::math::bernoulli_b2n<RealType>(i, i, &v1_array[0U]);
+
+      delete [] v1_array;
+   }
    boost::math::tangent_t2n<RealType>(i);
-   boost::math::tangent_t2n<RealType>(i, i, &v1);
+   {
+      RealType* v1_array = new RealType[i];
```
jzmaddock, line 479: Ditto.

test/compile_test/instantiate.hpp
ckormanyos: Sure. I appreciate that

test/compile_test/instantiate.hpp
ckormanyos: What do you mean? Like std::vector?

jzmaddock: I was going to suggest a Boost solution, but these days it's spelled std::unique_ptr: https://en.cppreference.com/w/cpp/memory/unique_ptr

test/compile_test/instantiate.hpp
```diff
@@ -538,7 +618,18 @@ void instantiate(RealType)
    boost::math::sinc_pi(v1 * 1);
    boost::math::sinhc_pi(v1 * 1);
    boost::math::asinh(v1 * 1);
-   boost::math::acosh(v1 * 1);
+   {
+      using std::fabs;
+
+      auto v_special { v1 };
```
jzmaddock, line 624: As above, specific value.

ckormanyos: There are a bunch of those. I'm more concerned with how to get the exceptions and the instantiations running and compiling.

jzmaddock:
Before I get too far into specific details - what is the purpose of this change?

Let me explain: the file instantiate.hpp was designed for one purpose only - concept checking.  As a side effect, it also catches a few other things like missing `using std::whatever;` declarations, and (non-template) functions accidentally not marked as inline.  It was never really designed to be run, although that's not necessarily a bad thing in itself except if...

Does this increase code coverage?  If so, that's arguably a bad thing... let me try and explain, instantiate.hpp does just that, it instantiates things, it doesn't anywhere actually check that they do the right thing!  In other words if there's missing coverage we shouldn't hide it behind something that's simply not designed for runtime testing, we should add some actual tests that validate behaviour.

test/test_compile_test_and_run.cpp
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+auto main() -> int
+{
+  local::instantiate_runner();
+
+  const auto result_is_ok = (boost::report_errors() == 0);
+
+  return (result_is_ok ? 0 : -1);
```
jzmaddock, line 38: Silly question: why not just return `boost::report_errors()` like everyone else? ;)

ckormanyos:
> Before I get too far into specific details - what is the purpose of this change?

That's a good question. When I started, I wanted the code to actually run, not just compile. I don't exactly remember when CI started failing. But when I got the code running in some situations, it started failing to compile in others.

This was confusing to me. So I started a series of efforts to get the code compiling and running in all situations. THis ultimately led to an actual change needed in the standalone configuration.

Then I kind of lost track of it. honestly I do not know or remember the actual change which led to the `static_assert`s failing. But they seem to fail at times when I do not expect this. So that is unrelated to the running of the code.

I think the code should compile and run. And my latest PR has this, except for one thing, the static asserts in `cauchy` and `non-central` do not work.

ckormanyos:
I really do think the code in all situations should compile _and_ run. This might be a philosophical difference, but I don't see why they should exclude each other.

And this is where I got stuck.

If I kick out those 7 static asserts in Cauch and non-central beta, I get both. I can not find the needed syntax to get compilation working for those distributions missing such things as skewness.

ckormanyos:
There must be a way to get both. I have not yet found it. But I'm close in 3801e07c9b684f010e4650a6851a0eb667467644

test/test_compile_test_and_run.cpp
ckormanyos: > Silly question: why not just return `boost::report_errors()`

We can work out these syntactic details. Thanks.

> Does this increase code coverage?

Yes it increases code coverage.

> If so, that's arguably a bad thing.

To me this doesn't make sense. I get that there can be compile checks. But when the arguments are fixed, the code should run. Maybe I missed some syntactic details when I got the code running. Then I made a few mistakes (maybe) and the code does not compile. But i think it should do both.

We can hash out the syntactic details later, but I think the code in instantiate.hpp should both compile and run. Sure this might be a bit of a shocker in legacy code, but I would have a difficult time understanding why compileing exlcudes running (as long as the arguments are in range).

PR #1094, opened 2024-02-15
More coverage of instantiation

ckormanyos:
The purpose of this PR is to cover most of the remaining [`instantiate.hpp`](https://github.com/boostorg/math/blob/develop/test/compile_test/instantiate.hpp).

As a first step, however, remove `BOOST_MATH_NO_DISTRIBUTION_CONCEPT_TESTS` and let's see if instantiation sinply now compiles and runs in standalone (it was previously excluded from standalone).

### DIFF
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -31,6 +31,7 @@
 
 #define BOOST_PREVENT_MACRO_SUBSTITUTION
 #define BOOST_MATH_NO_REAL_CONCEPT_TESTS
+#define BOOST_MATH_NO_DISTRIBUTION_CONCEPT_TESTS
 #define BOOST_MATH_NO_LEXICAL_CAST
 
 // Since Boost.Multiprecision is in active development some tests do not fully cooperate yet.
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -31,7 +31,6 @@
 
 #define BOOST_PREVENT_MACRO_SUBSTITUTION
 #define BOOST_MATH_NO_REAL_CONCEPT_TESTS
-#define BOOST_MATH_NO_DISTRIBUTION_CONCEPT_TESTS
 #define BOOST_MATH_NO_LEXICAL_CAST
 
 // Since Boost.Multiprecision is in active development some tests do not fully cooperate yet.
--- a/include/boost/math/tools/throw_exception.hpp
+++ b/include/boost/math/tools/throw_exception.hpp
@@ -24,7 +24,19 @@
 
 #else // Standalone mode - use standard library facilities
 
-#define BOOST_MATH_THROW_EXCEPTION(expr) throw expr;
+#ifdef _MSC_VER
+#  ifdef _CPPUNWIND
+#    define BOOST_MATH_THROW_EXCEPTION(expr) throw expr;
+#  else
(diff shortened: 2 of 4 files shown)


ckormanyos:
As it turns out, standalone can not tolerate distribution concept checks (due to necessity of `function_requires`). But standalone can (probably) handle all the instantiations/runs of the special functions. Special function instantiations were previously excluded from standalone (along for the ride) with distribution concept checks.

ckormanyos:
Hi @mborland I'm a bit confused by some failures in standalone CI-block 1 (where this PR now attempts to run more concept checks on special functions).

I get some errors on exception handling in hypergeometric. Also I see something like missing threads in Bernoulli, claiming I need `BOOST_MATH_BERNOULLI_UNTHREADED`. I thought I could use most special functions in standalone.

The file with new features is [`instantiate.hpp`](https://github.com/boostorg/math/blob/cover_instantiate_more/test/compile_test/instantiate.hpp). The big difference in this PR is that _only_ distribution concept checks are disabled whereas special function concept checks are active. I think we are close.

Can you see anything obvious I need to configure? I did not know if I needed specialties for standalone that I maybe forgot?

mborland:
> 
> Can you see anything obvious I need to configure? I did not know if I needed specialties for standalone that I maybe forgot?

Your actual failures are coming from: https://github.com/boostorg/math/actions/runs/7910989740/job/21594474668?pr=1094#step:17:1016. I didn't disable exception handling in the standalone testing but there may be a colliding configuration somewhere else? The Bernoulli thing is marked as failed as expected which is fine.

ckormanyos:
Hi @mborland it perplexes me. In the Jam file there are places around lines 505-506 that compile test_instantiate numbers 1 and 2. This is done both with and without exceptions. And it has worked for a long time and still does in my branch for non-standalone.

In the failures we get now these are from the latter line 506 which disables exceptions via command line argument to the compiler. And it is here that it fails for standalone.

I’m not super experienced with standalone but I don’t see why I shouldn’t be able to disable exceptions on command line and fail compiling certain special functions. Sure I could throw in BOOST_DISABLE_EXCEPTIONS for the standalone case, but I don’t think I should have to?

mborland:
> Hi @mborland it perplexes me. In the Jam file there are places around lines 505-506 that compile test_instantiate numbers 1 and 2. This is done both with and without exceptions. And it has worked for a long time and still does in my branch for non-standalone.
> 
> In the failures we get now these are from the latter line 506 which disables exceptions via command line argument to the compiler. And it is here that it fails for standalone.
> 
> I’m not super experienced with standalone but I don’t see why I shouldn’t be able to disable exceptions on command line and fail compiling certain special functions. Sure I could throw in BOOST_DISABLE_EXCEPTIONS for the standalone case, but I don’t think I should have to?

I think something that was lost is non-standalone mode boost.throw_exception was dealing with the no-except flag whereas in standalone it wasn't. I added a commit that should handle this now.

ckormanyos:
Great find @mborland I’ll let this green up. Then I’ll activate the remaining 500 uncovered lines. I’ll then let all the instantiation stuff in all test suites then not only compile but also run. Should be done in a couple days.

ckormanyos:
This is actually a good intermediate result with an actual configuration improvement. For the sake of clarity, I’ll merge this one now.

then I’ll cover the remaining mixed instantiation lines in a soon-to-come future PR

cc: @mborland and @jzmaddock and @NAThompson

PR #1095, opened 2024-02-15
Start finishing instantiate-tests (more to come)

ckormanyos:
The purpose of this PR is to let many instantiation tests also run. Standalone/non-Standalone symmetry is achieved. Time to finish this and cover the last 500 lines in `mixed`-whatever.


ckormanyos:
First step does not include th full 500 missing lines. These will be present shortly.

ckormanyos:
Hi John (@jzmaddock) How about this preliminary fix on 1 of the seven places to fix that domain assertion. Such powerful language, ... _domain_ _assertion_.

ckormanyos:
Let's see if it cycles green...

test/test_compile_test_and_run.cpp
```diff
@@ -4,6 +4,8 @@
 //  Boost Software License, Version 1.0. (See accompanying file
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+#define TEST_COMPLEX
+
 #include <boost/core/lightweight_test.hpp>
 #include <boost/cstdfloat.hpp>
```
jzmaddock, line 10: I have a hunch that cstdfloat.hpp is pulling in something that sets the policy defaults before instantiate.hpp gets included.

ckormanyos: Ok. We are moving fast in two threads, that is probably because I started two threads. So be it. 

We do, nonetheless, need to fix this properly. I've seen also your comments on style in the other thread. We can combine all these. But for some reason, I can't find the right syntax on the policies to make every compiler _happy_.

ckormanyos:
If we look [here](https://github.com/boostorg/math/actions/runs/7914656922) the code is compiling and running.

And (you might call em odd), but I want that code both compiling and running. This is not usually considered strange. That's why I wrote the code. So I'd like to find a way to get both.

If that proves impossible, or I'm pressing too hard or fast, we can back off totally on this. But it will be hard to convince me that the code can't compile and run.

ckormanyos:
Hi John (@jzmaddock),

Thanks so much for your inputs on this PR.

I thought a lot about your comments. And some or (but not all) of the stuff I implemented does not make sense. So what I did in this recent push was handle most of your review comments and reduce the one single mystery I had remaining (hopefully) to one single new file.

The name of the original test file I had selected was somewhat non-intuitive and the new renamed file is called [`test_instantiate3.cpp`](https://github.com/boostorg/math/blob/finish_instantiate/test/test_instantiate3.cpp).

It's cycling now. The `static_assert`s are back in place in the Cauchy and non-central-beta distributions.

If we find a mystery I'd like to solve it. If you would strongly prefer to _stop_ attempting to execute the instantiation code, that's as easy as two single lines, ... `LCOV_EXCL_START`/`LCOV_EXCL_STOP`.

But I'd like to see if the mystery compile failures are still present and find out (if so) why?

Cc: @mborland and @NAThompson

ckormanyos:
Hi @jzmaddock regardless of whether we run or don't run the code in instantiate.hpp, I have a mystery which simply confuses me.

In the most recent drone run all the time `test_instantiate1.cpp` and `test_instantiate2.cpp` pass. But `test_instantiate3.cpp`, even though it seems very similar to me fails all over the place on older GCC compilers.

If we can clear up the why? on this, then we can negotiate on the question, ... to run or not to run, ... the code in instantiate.hpp.

Do you see any clear reason why the code in [`test_instantiate3.cpp`](https://github.com/boostorg/math/blob/finish_instantiate/test/test_instantiate3.cpp) should fail _compilation_? This has perplexed and stymied me for days.

ckormanyos:
The perplexing part is in [drone 1769](https://drone.cpp.al/boostorg/math/1769).

Concern was earlier raised that `<boost/cstdfloat.hpp>` which has a lot of depth might be bringing in some stuff to occlude.

I'll now try a run that avoids using `<boost/cstdfloat.hpp>`.

ckormanyos:
> try a run that avoids using `<boost/cstdfloat.hpp>`

Well, I'm off to hit this day, but it seems like drone is cycling green without `<boost/cstdfloat.hpp>`. Now I know where to look.

@jzmaddock to run or not to run code in instantiate.hpp, that is the last open question.

ckormanyos:
So it seems like <boost/cstdfloat.hpp> introduces something at compile time that occludes a default policy template parameter. What a cool search topic. I'll start searching

jzmaddock:
@ckormanyos you need to remember that the default is a global thing, once it's been set you can't re-set it to something else.  So once policy.hpp has been #included, that's it, the defaults are set, and nothing will change that.  So either the #define needs to be at the top of the .cpp file before any #includes, or if you're relying on instantiate.hpp to set the defaults it needs, then that header needs to be the **first** #include in the list.

ckormanyos:
Thanks john. This is great. I’ll try that and see.

In your honest opinion, how tough is the tussle going to be to run or not actually execute that code?

ckormanyos:
>> In your honest opinion, how tough is the tussle going to be to run or not actually execute that code?

> It shouldn't really be that hard, the question is whether we learn that much from it - as I said, the code was only designed as a concept check - it doesn't verify behaviour or check that the function calls do anything meaningful, those tests are elsewhere.

First off I apologize if my introduction of this was awkward or my wording unfortunately stark. That was not so cool.

Thanks for going on with this.

I ended up going back-and-forth (and back-and-forth) twice, two entire cycles on this in my mind.

My result: I really like these as unit tests. These lines were written with great attention to detail. When I first made them runable, I ran into exceptions, stack problems and a whole host of other things. This is a true gem where hundreds of special functions are called back-to-back, providing a wonderful top-level sanity check (in great addition to the instantiation check).

My wish:

- Address the remaining review issues (such as `scoped_ptr`.
- Cover the `_mixed` lines.
- And party with it.

I really appreciate the instantiations as a unit test.

Having said that, this post brought me into other issues.
- `<boost/cstdfloat>` has a few simple simplifications possible. We talked in the past regarding this, but the complex template puuls in `boost/math/constants/constants.hpp`, which today i would not have in the top-level of boost just for $\pi$ and ${\log}(10)$.
- We did find a configuration weakness for STANDALONE. This is good. So now standalone tests _instantiate_.

I'm going to go on with this and try for 1.85.

ckormanyos:
This thing, being now better, is actually coming in now.

The sense is that it now deos _not_ change the distributions in any way and we can include STANDALONE tests. So I'll chase down the review issues and `_mixed` in a future PR. But I have to jam this one in since the distributions were, in fact, a bit unjudiciously modified in develop. And that was not right.

Issue #1068, opened 2024-01-21
Better coverage?


jzmaddock:
I've just briefly looked at the codecov reports (many thanks for doing this @mborland !) to see what needs improving, and I'm seeing a lot of false flags, for example in https://app.codecov.io/gh/boostorg/math/pull/1067/blob/include/boost/math/special_functions/lanczos.hpp line 870 only the final initializer in the list is marked as covered, even though they must have all been executed if one has!  Is there anything we can do about this?

ckormanyos:
Numerous `return` lines are, in fact, marked as covered. The tool seems to be experiencing problems with `BOOST_MATH_BIG_CONSTANT`. We could see what happens on some explicit local tests as time allows.

Personally, I have found coverage on highly-templated C++ code to be somewhat challenging.

I also recommend to cut the optimization level way down to `-O1` or even `-O0` when running gciv/lcov. I have not looked at the make system. But that is also something can be tuned. I almost never change code for the sake of covering templates, but try to tune the environment or code-annotations to better reflect the honest situation.

ckormanyos:
That being said, my personal experience is that it can take a comparable amount of time to attain $100\\%$ line coverage as it took or takes to write the code in the first place. I find that testing effort is often underestimated on the road to quality. Somewhere a while back I mentioned increasing line coverage very briefly and suspect we need a dedicated effort to reach (via specially contrived test-cases) more hard-to-reach lines.

In the case mentioned here, however, i do believe that lcov might have representation problems with some of those big macros.

jzmaddock:
It took a bit of detective work, but yes, it's using lcov, and lcov has exclusion macros we can use.  I did spot a few genuine cases not covered while browsing through, so maybe I'll see what I can squash.

ckormanyos:
Sometimes, well we will ultimately just _know_ _better_ than `lcov`, which my opinion is, cant fully handle modern C++.

Then I use (in justified cases (whatever that is decided to mean ofr our projects)

> lcov has exclusion macros we can use

For multipple lines:
```
// LCOV_EXCL_START
My_uncovered_line();
// LCOV_EXCL_STOP
```

Or for single line(s)
```
my_uncovered_line(); // LCOV_EXCL_LINE
my_other_uncovered_line(); // LCOV_EXCL_LINE
```

ckormanyos:
See also:
- #1084 `tgamma()`/`lgamma()` edge cases for $250$ decimal digits.
[...]

PR #1096, opened 2024-02-17
Finish instantiation runners

ckormanyos:
Fishish instantiation runner work started in #1094 and #1095

### DIFF
--- a/test/compile_test/instantiate.hpp
+++ b/test/compile_test/instantiate.hpp
@@ -196,9 +196,7 @@ void instantiate(RealType)
 #endif
 #endif
    int i { 1 };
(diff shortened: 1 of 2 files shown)


ckormanyos:
OK. This preliminarily wraps up the topic of getting _instantiations_ actually running (in addition to simply compiling).

I'll probably ramp down with the trouble-making for the ramainder of the road to 1.85 (unless something broken or mysterious with my recent work comes up).

If anyone is followint the action on CodeCov, we now have a solid $92{\ldots}93\\%$.

ckormanyos:
See also #1068

PR #1099, opened 2024-02-20
Improve coverage 2

jzmaddock:


ckormanyos:
Hi John,

I've been following this work with great interest and admiration. In recent [red CI](https://github.com/boostorg/math/actions/runs/8046852073/job/21974931131?pr=1099#step:17:4273) there's some of this kind of stuff related to `BOOST_MATH_BERNOULLI_UNTHREADED`

Sorry I did not chatch all of this in my own bare-metal standalone tests.

But there are several compiler switches which might seemingly do the same thing.
  - `BOOST_MATH_BERNOULLI_UNTHREADED` is this really needed for Bernoulli exclusively?
  - `BOOST_MATH_HAS_THREADS` OK, Boost.Math _has_ threads.
  - `BOOST_MATH_DISABLE_THREADS` hmmm... Does Boost.Math also _have_ threads and them disable them?
  - `BOOST_MATH_NO_ATOMIC_INT` if Boost.Math does not have threads is is sane/save to assume it also lacks atomics?

I could not wade through all this, but it seems like we could simplify some things as we move forward? Thoughts?

Cc: @mborland

ckormanyos:
I guess in my (maybe naive world), Boost.Math has threads by default. One sinlge option to disable threads could replace all redundant of the other three options?

Maybe I'm over-simplifying?

Cc: @jzmaddock and @mborland

jzmaddock:
> I could not wade through all this, but it seems like we could simplify some things as we move forward? Thoughts?

Unfortunately I fear probably not.

The Bernoulli numbers only have to deal with threads in the sense that the cache has to be generated in a thread safe way.  And this only applies to the multiprecision case I believe.  The most efficient way is to use atomics, but while <atomic> is guaranteed, there is no guarantee that it contains anything useful (or efficient).  Hence BOOST_MATH_NO_ATOMIC_INT does NOT imply no threads, just no efficient atomics.

Then we have platforms like Mingw/Cygwin which have no working std::thread or mutexes, from memory you can't even #include a thread related header without getting an error :(

So the options are:

1) We have threads and atomic int's, so use these.
2) We have threads, but no atomics, use mutexes (less efficient).
3) We appear to have no threads - wohoo, all nice and simple!  However, this may be very unsafe (see below).
4) We have threads, but the user knows up front that they're not using them, so disables support with BOOST_DISABLE_THREADS for the fastest/simplest possible code.  Then as (3).

I had to look up BOOST_MATH_BERNOULLI_UNTHREADED: when we have option (3) above, the default behaviour is to static_assert on the grounds that we don't know how to make things safe, but BOOST_MATH_BERNOULLI_UNTHREADED can be defined to suppress the error if the user knows what they're doing.

And finally, if I can remember to type BOOST_MATH_ASSERT instead of BOOST_ASSERT, I might actually get some of this green!

ckormanyos:
>> I could not wade through all this, but it seems like we could simplify some things as we move forward? Thoughts?

> Unfortunately I fear probably not.

... and a whole bunch of very sensible information...

Yes John (@jzmaddock), I get it now. Somehow I was vaguely familiar with your points, that you have clarified perfeclty. Please disregard my noise and leave everything alone, with your comment that we all need to use `BOOST_MATH_whatever` consistently, as required (I had to deal with this one myself also on my metal benchmarks).

What a great result. Yet sorry about the noise.

Thx again.

Cc: @mborland

jzmaddock:
OK, this is good to go I hope, one failure is a network timeout.  Will merge and then see if I can re-trigger multiprecision develop.

Issue #1098, opened 2024-02-19
Conflicting checks for BOOST_CXX14_CONSTEXPR


bartoli:
I built boost 1.84 using VS2015 U3 compiler.
Then i am trying to use it from VS2022 IDE (using VS2015 compiler), on code that works with 1.63.

There are compile errors because of conflicting BOOST_CXX14_CONSTEXPR definitions.

First, a redefinition warning says that it is first defined in boost/config/details/suffix.hpp.
Here, BOOST_NO_CXX14_CONSTEXPR is defined for this toolset (because _MSC_VER=1900 for vs2015 i presume), so BOOST_CXX14_CONSTEXPR is defined to nothing.
But then later, boost has included boost/math/tools/config.hpp, which defines BOOST_CXX14_CONSTEXPR depending on the condition 
#if (__cplusplus > 201400L || _MSVC_LANG > 201400L)
Which succceeds, then redefines BOOST_CXX14_CONSTEXPR to constexpr

Then the code does not build because of those constexpr

Shouldn't the check for the macro definition be the same in both places (the first one would seem the valid one?), or even , shouldn't the macro be defined in a single place?

glenfe:
Transferred to boostorg/math. I really don't like the idea of Math defining macros defined in Config. They should only define BOOST_MATH_CXX14_CONSTEXPR and only use that.

jzmaddock:
You appear to be mixing including other Boost headers, with defining BOOST_MATH_STANDALONE - that's really not a good idea, if the rest of Boost is present, then don't define BOOST_MATH_STANDALONE and let Boost.Math use Boost.Config for configuration.  We might need to look at VS2015's configuration in standalone mode, but to be honest, I'm not sure we really support that old compiler in that usage mode anyway...  @mborland ?

mborland:
> I didn’t check but do the docs provide guidance on both these points? Don’t mix standalone with the rest of boost, and which compilers actually support knowingly standalone

I guess I never wrote anywhere that explicitly states don't mix standalone with the rest of boost. Closest thing is that standalone Multiprecision is compatible with standalone math. Refactoring all the config macro names should allow interoperability but that's not really the use case. On the second point there is "Currently a compiler that is fully compliant with C++14" but then we list VS2015 as supported which is conflicting information.

ckormanyos:
Something like, “it is best to use either purely standalone mode or math within the context of a complete Boost. But it is not intended to mix standalone mode with the rest of the Boost headers.”

[...]

PR #1100, opened 2024-02-22
Replace redefinition of Boost.Config macros in Standalone mode

mborland:
Create our own definition of everything instead. Also add note on standalone mode usage, and update warning banner about C++14 and MSVC 14.0. 

Closes: #1098


mborland:
@jzmaddock and @ckormanyos Any issues with this PR? It's mostly bulk replacement of macros to alleviate Glen's concerns about support two different consumption modes. Once support for MSVC 14.0 is dropped in 2 releases there's a fair bit of simplification that can be made regarding all the NO_CXX14_CONSTEXPR macros.

ckormanyos:
Hi @mborland it is great. Remarkable speed getting this into 1.85. Thanks. I was only able to take a cursory glance due to high time pressure in other areas. From the things I saw and CI green, looks great.

I'll run standalone stuff in a bunch of other projects on Modular-Boost-develop in my nightly tonight in other places. So that'll be an independent check...

Cc: @jzmaddock

ckormanyos:
Hi Matt (@mborland)

Bit of a long story here, so please bear with me...

I did run into [problems](https://github.com/ckormanyos/real-time-cpp/actions/runs/8012557843/job/21888061959) with this new standalone config.

The first set was probglems of my own. I had been mixing standalone with non-standalone macros and also including the regular `<boost/config.hpp>`

So now if I define `BOOST_MATH_STANDALONE`, then I need to exclusivelx use `BOOST_MATH_something` macros and **not** include `<boost/config.hpp>`. This is fine with me, but some clients may have also been wrognly mixing these and there might be activity on lists or GH clearing this up.

The final hit was mixing math/multiprecision standalone, where you run into the fact that multiprecision standalone [requires](https://github.com/boostorg/multiprecision/blob/27666619b85f014e6724c0343923b77b10ddb704/include/boost/multiprecision/detail/standalone_config.hpp#L21) inclusion of `<boost/config.hpp>`. So now you can't mix math/multiprecision both standalone. That does not go. And we can't roll out like this.

The problem is, I was unable to fix it quickly. I am only able to report on it.

Cc: @jzmaddock

ckormanyos:
The story then gets a bit more complicated. If I then include `<boost/config.hpp>` to maks multiprecision standalone happy, then there are conflicting macros with the new config ways of math. It's all solvable, but it looks like a bit of unwinding and sync is needed between the two entities math/multiprecision.

Cc: @mborland and @jzmaddock and @NAThompson

mborland:
Let me take a look. It seems as though previously combining standalone math and standalone multiprecision would have the same macro redefinition errors as the original issue since config is bundled into the standalone multiprecision package. Now that math does not use any of the same macro names as config this should not be an issue.

ckormanyos:
Hi Matt (@mborland) when I straightened out the compiler options to use exclusively `BOOST_MATH_whatever` and _also_ has `<boost/config.hpp>` present for detection when building Multiprecision standalone, all my metal standalone benchmarks [work again](https://github.com/ckormanyos/real-time-cpp/actions/runs/8029792844).

There seems to be no issue and everything is OK. It was my mistake.

Thanks

Cc: @jzmaddock and @NAThompson

PR #1102, opened 2024-02-23
Add dual-standalone test case

mborland:
@ckormanyos Locally I am not seeing any errors/warnings from this test case. Do you have an example as reported here: https://github.com/boostorg/math/pull/1100#issuecomment-1960805834?


ckormanyos:
Hi Matt, great question. I am away from office at this moment. I have a test case running on GCC-arm-none-eabi exhibiting this. I’ll produce that link for you tonight

ckormanyos:
Hi Matt (@mborland) there seems to be no issue and everything is OK in my standalone metal benchmarks. I did, however, need to use the macros properly, which had not previously been the case. See also notes in #1100 in the post [here](https://github.com/boostorg/math/pull/1100#issuecomment-1962319460).

PR #1111, opened 2024-03-06
Improve erf/expm1/expint coverage.

jzmaddock:
In the expm1 case, tighten up error handling and testing.


jzmaddock:
@mborland @ckormanyos @NAThompson This PR is now good to go and completes code coverage improvements as far as G for gamma.hpp.  There are a lot of accumulated changes here, so I'll hold off for a few days at least in case there are any comments...

Also not sure what's happened to the drone CI runner: it seems to be terminally broken right now?

mborland:
@sdarwin have something gone sideways with our drone config? It looks like we have quite a few runs over the past several days that aren't picking up? There are some like this that seem to have completely crashed before the retry step in the clone phase: https://drone.cpp.al/boostorg/math/2225/193/1

jzmaddock:
>This all looks good to me John. Once it's in I can run it through the actual CUDA testing to see if there's any changes that need to be made there. I think the only thing I saw with the replacement of BOOST_MATH_FP_XXX with the standard FP_XXX macros. No matter since Scipy is now pinned against specific versions instead of tracking develop.

Where was this?  I didn't think I'd touched tools/config.hpp?

include/boost/math/special_functions/gamma.hpp
```diff
@@ -944,13 +972,15 @@ BOOST_MATH_GPU_ENABLED T full_igamma_prefix(T a, T z, const Policy& pol)
...
+      // LCOV_EXCL_STOP
    }
    //
    // This error handling isn't very good: it happens after the fact
    // rather than before it...
+   // Typically though this method is used when the result is small, we should probably not overflow here...
    //
-   if((boost::math::fpclassify)(prefix) == (int)BOOST_MATH_FP_INFINITE)
-      return policies::raise_overflow_error<T>("boost::math::detail::full_igamma_prefix<%1%>(%1%, %1%)", "Result of incomplete gamma function is too large to represent.", pol);
+   if((boost::math::fpclassify)(prefix) == (int)FP_INFINITE)
```
mborland, line 982:
```suggestion
   if((boost::math::fpclassify)(prefix) == (int)BOOST_MATH_FP_INFINITE)
```

jzmaddock: Good catch!

I'll do that before merging.

Unrelated question: should BOOST_MATH_FP_* map to the std values of FP_* ?

jzmaddock:
OK, drone is back up and running again, just the few known about sycl tests failing, so merging...

Issue #1115, opened 2024-03-27
Name clash with macro from X11/X.h


iv-m:
Xorg protocol definitions include `Complex` macro, which is defined to 0 (see e.g. https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/xorgproto-2024.1/include/X11/X.h#L552).

It turns out that there is code out there that includes both `X11/X.h` (through wxWidgets) and `boost/math/tools/roots.hpp`, which gives the following error:

```
[...]
/usr/include/boost/math/tools/roots.hpp: At global scope:
/usr/include/boost/math/tools/roots.hpp:787:16: error: expected identifier before numeric constant
  787 | template<class Complex, class F>
      |                ^~~~~~~
/usr/include/boost/math/tools/roots.hpp:787:16: error: expected '>' before numeric constant
/usr/include/boost/math/tools/roots.hpp:788:1: error: expected unqualified-id before numeric constant
  788 | Complex complex_newton(F g, Complex guess, int max_iterations = std::numeric_limits<typename Complex::value_type>::digits)
      | ^~~~~~~
[... and so on ...]
```

While it's not a bug in Boost.Math as it is, it makes sense to me to rename the template parameter to a less common word,  to avoid this and, potentially, other similar clashes.

iv-m:
Thank you.

mborland:
> Thank you.

No problem. I asked permission to merge into Boost 1.85 since its mid-release cycle (release date 10 April). I expect it'll make it.

mborland:
Merged into master for 1.85

PR #1116, opened 2024-03-27
Fix X11 naming clash in roots

mborland:
Closes: #1115


jzmaddock:
Thanks Matt, it's a tiny point, but template names are generally CamelCase by convention, so maybe `ComplexType` instead?  Just makes it easier to Grok the code, as `complex_type` would generally be a local typedef.

mborland:
> Thanks Matt, it's a tiny point, but template names are generally CamelCase by convention, so maybe `ComplexType` instead? Just makes it easier to Grok the code, as `complex_type` would generally be a local typedef.

For some reason I was thinking that ComplexType would still get pasted over since it has the sub-string.

PR #1121, opened 2024-04-18
Fix for skew_normal quantile non-convergence

mborland:
@jzmaddock This may not be _the_ fix but it certainly gets things working. I just tested different shapes of the issue distribution to find decent places to make the change in allowable steps and how many we might need. Im not quite search how to narrow down the search range from (-inf, inf) to speed up convergence organically.


jzmaddock:
This should have been fixed elsewhere now, can we close?

PR #1122, opened 2024-04-19
Don't use obsolete predef/checks/predef.jam B2 module.

grafikrobot:


mborland (approved):
There's a ton of CI failures on this but they all seem to be issue with ubuntu's mirrors:

```
Failed to fetch http://archive.ubuntu.com/ubuntu/dists/jammy-updates/main/binary-amd64/Packages.gz  File has unexpected size (2018947 != 2018948). Mirror sync in progress? [IP: 91.189.91.83 80]
```

Merging because it should be safe.

Issue #1126, opened 2024-04-29
Problem in geometric.hpp for MacOs Arm64


kenarab:
This line should be corrected
At /math/distributions/geometric.hpp at line 404
```
  return std::log(p); // success_fraction
      
```
Now the code is
```
  return log(p); // success_fraction
      
```

When building that library. Throws this error

```
In file included from /opt/homebrew/include/boost/math/distributions.hpp:26:
/opt/homebrew/include/boost/math/distributions/geometric.hpp:404:16: error: unexpected namespace name 'log': expected expression
        return log(p); // success_fraction
               ^
1 error generated.
make[2]: *** [CMakeFiles/PonisIqCoreNLV.dir/Unity/unity_0_cxx.cxx.o] Error 1
make[1]: *** [CMakeFiles/PonisIqCoreNLV.dir/all] Error 2
make: *** [all] Error 2
```

kenarab:
I will try to correct by my self on the repo, but don't know if it is possible

PR #1127, opened 2024-04-29
Update geometric.hpp

kenarab:
The proposal is derived from this issue

https://github.com/boostorg/math/issues/1126

This line should be corrected
At /math/distributions/geometric.hpp at line 404
```
  return std::log(p); // success_fraction
      
```
Now the code is
```
  return log(p); // success_fraction
      
```

When building that library. Throws this error

```
In file included from /opt/homebrew/include/boost/math/distributions.hpp:26:
/opt/homebrew/include/boost/math/distributions/geometric.hpp:404:16: error: unexpected namespace name 'log': expected expression
        return log(p); // success_fraction
               ^
1 error generated.
make[2]: *** [CMakeFiles/PonisIqCoreNLV.dir/Unity/unity_0_cxx.cxx.o] Error 1
make[1]: *** [CMakeFiles/PonisIqCoreNLV.dir/all] Error 2
make: *** [all] Error 2
```

### DIFF
--- a/include/boost/math/distributions/geometric.hpp
+++ b/include/boost/math/distributions/geometric.hpp
@@ -401,7 +401,7 @@ namespace boost
       }
       if(k == 0)
       {
-        return log(p); // success_fraction
+        return std::log(p); // success_fraction
       }
       //RealType q = 1 - p;  // Bad for small p
       //RealType probability = 1 - std::pow(q, k+1);
--- a/include/boost/math/distributions/geometric.hpp
+++ b/include/boost/math/distributions/geometric.hpp
@@ -384,6 +384,7 @@ namespace boost
     inline RealType logcdf(const geometric_distribution<RealType, Policy>& dist, const RealType& k)
     { // Cumulative Distribution Function of geometric.
       using std::pow;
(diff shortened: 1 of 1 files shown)


mborland (changes requested):
`log` should be found by ADL for any `RealType`. The correct fix would be adding `using std::log;` under this line: https://github.com/boostorg/math/blob/develop/include/boost/math/distributions/geometric.hpp#L386. I think it also needs a `using std::exp;`.

include/boost/math/distributions/geometric.hpp
```diff
@@ -384,6 +384,7 @@ namespace boost
     inline RealType logcdf(const geometric_distribution<RealType, Policy>& dist, const RealType& k)
     { // Cumulative Distribution Function of geometric.
       using std::pow;
+      using std::log;
```
mborland, line 387:
```suggestion
      using std::log;
      using std::exp;
```

mborland (approved):

mborland:
macOS clang failures are being addressed elsewhere

PR #1131, opened 2024-05-04
Polynomial roots via eigenvalues of the companion matrix

NAThompson:


jzmaddock:
>Quick question: Given a floating point type T, how to I form the associated complex type in the multiprecision case? To wit, std::complex<T> will fail for float128, cpp_bin_float_100, etc.

Ugh, seems to be a bridge we haven't crossed yet.  

However, `typeof(polar(T(), T())`  would do it, since we never use expression templates for that.

Note that in order to use multiprecision types with Eigen we need to include <boost/multiprecision/eigen.hpp> which specializes a bunch of Eigen traits classes.  I guess to avoid nasty surprises this header should include that when available, though it adds an often not needed dependency :(

The static_assert on is_floating_point would have to go as well...

jzmaddock:
> #include <boost/math/cstdfloat/cstdfloat_complex_std.hpp>

We really shouldn't need that, and it might get rid of the error?

And of course I should have said `decltype` since `typeof` is rather last century (like me!!)  :)

jzmaddock:
Well, I was going to say we should file a bug report again Eigen... and tried but gitlab tied me up in circles and I gave up :(

cantonios:
> > Well, I was going to say we should file a bug report again Eigen... and tried but gitlab tied me up in circles and I gave up :(
> 
> Done: https://gitlab.com/libeigen/eigen/-/issues/2818; I do have some worry that they'll set it to WONTFIX until we have some canonical way of providing `complex<T>` for multiprecision `T`,

We won't WONTFIX ;).  But you'll need to provide the merge request.

>  but maybe they know some tricks.

Provided a suggestion on the Eigen bug.  Not really a "trick", but we have some traits you can use.

Otherwise, you could explicitly specialize `std::complex<your_custom_type>` - that should be _legal_ as long as your type satisfies the [`NumericType`](https://en.cppreference.com/w/cpp/named_req/NumericType) requirement.

jzmaddock:
Thanks @cantonios for the prompt reply.

> Otherwise, you could explicitly specialize std::complex<your_custom_type> - that should be legal as long as your type satisfies the [NumericType](https://en.cppreference.com/w/cpp/named_req/NumericType) requirement.

We're looking at that now, as you say specializing `complex` appears to be legal, however overloading all the non-member functions (in namespace std) is definitely NOT legal and may well have unintended consequences, but I don't see a better way for now.

cantonios:
> however overloading all the non-member functions (in namespace std) is definitely NOT legal

Why do you need to?  You should be able to rig it up so that non-members are found via ADL.  e.g. https://godbolt.org/z/fxTb8c8cf

> but I don't see a better way for now.

The better way (at least for Eigen) is to add an `Eigen::NumTraits<...>::Complex` member and use that.

jzmaddock:
> Why do you need to? You should be able to rig it up so that non-members are found via ADL.

For sure, those overloads can be found via ADL, my expectation is that most users using `std::complex`, will be calling `std::exp`.  Perhaps I'm wrong?

> The better way (at least for Eigen) is to add an Eigen::NumTraits<...>::Complex member and use that.

Yes, that's what we were hoping for ;)

cantonios:
> For sure, those overloads can be found via ADL, my expectation is that most users using `std::complex`, will be calling `std::exp`. Perhaps I'm wrong?

Yes, this is usually true, but arguably a bug on the users' end if they expect it to work with custom types.  Besides, you still have this issue if you _don't_ specialize `std::complex`, opting for your own custom `CustomComplex` implementation, whatever `decltype(polar(...))` is.

Issue #1132, opened 2024-05-08
float_next/prior should not raise domain errors when the result is clearly infinite


jzmaddock:
For example float_next(+INF) should be +INF and not a domain_error.

See https://github.com/boostorg/multiprecision/pull/618.

PR #1135, opened 2024-05-17
Correct float_next(+INF) and float_prior(-INF)

jzmaddock:
Fixes https://github.com/boostorg/math/issues/1132

### DIFF
--- a/include/boost/math/special_functions/next.hpp
+++ b/include/boost/math/special_functions/next.hpp
@@ -194,10 +194,14 @@ T float_next_imp(const T& val, const std::true_type&, const Policy& pol)
 
    int fpclass = (boost::math::fpclassify)(val);
 
-   if((fpclass == (int)FP_NAN) || (fpclass == (int)FP_INFINITE))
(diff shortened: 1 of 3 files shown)


tchaikov:
thank you. IIUC, once this change lands, we should revert https://github.com/boostorg/multiprecision/pull/618, which is basically a work around of this issue.

jzmaddock:
One failure is a network timeout... merging.

include/boost/math/special_functions/next.hpp
```diff
@@ -194,10 +194,14 @@ T float_next_imp(const T& val, const std::true_type&, const Policy& pol)
...
-      if(val < 0)
+      if (val < 0)
          return -tools::max_value<T>();
+      return val;  // +INF
+   }
+   else if (fpclass == (int)FP_NAN)
+   {
       return policies::raise_domain_error<T>(
          function,
          "Argument must be finite, but got %1%", val, pol);
```
jwakely, line 207: These error messages are no longer really accurate, the argument can be an infinity which is of course not finite.

Issue #1129, opened 2024-05-03
Audiff test failures for Clang/MacOS


jzmaddock:
In order to get us green again, I've disabled 2 tests in test_autodiff_8.cpp for Clang/MacOS:

https://github.com/boostorg/math/blob/a243640baddfaa318859d622fddb7abc3ba8a978/test/test_autodiff_8.cpp#L12

and

https://github.com/boostorg/math/blob/a243640baddfaa318859d622fddb7abc3ba8a978/test/test_autodiff_8.cpp#L135

One for @pulver.

PR #1136, opened 2024-05-21
Update CI

mborland:
Removes MSVC 14.0 and adds clangs 15, 16, 17, 18 as well as GCC 14. Update all drone sanitizers to latest and greatest.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -32,10 +32,10 @@ def main(ctx):
     #
     # Sanitizers:
     #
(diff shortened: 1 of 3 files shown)


mborland:
@jzmaddock is [this failure](https://github.com/boostorg/math/actions/runs/9171166088/job/25215137805?pr=1136#step:11:619) on macOS unexpected? I didn't see it listed in https://github.com/boostorg/math/issues/1129

Issue #1143, opened 2024-06-02
Incorrect `cyl_bessel_j` value at infinity


matwey:
Hello,

I use the following code with boost 1.83.0
```c++
std::cout << boost::math::cyl_bessel_j(0, std::numeric_limits<double>::infinity()) << std::endl;
std::cout << boost::math::cyl_bessel_j(1, std::numeric_limits<double>::infinity()) << std::endl;
```
and then I see
```
-nan
-nan
```
while Bessel J function of any order is known to be exactly zero at infinity.

jzmaddock:
Confirmed, I'll check the other Bessel functions while I'm at it.  Thanks for the report!

matwey:
Thanks. I've found that at least `sinc_pi` is also not defined correctly for infinity.

PR #1144, opened 2024-06-07
Update Bessel functions at infinity.

jzmaddock:
Also sinc functions, and update tests.
Fixes https://github.com/boostorg/math/issues/1143.


jzmaddock:
One remaining failure is a network issue, windows image failures have now been fixed by github.

PR #1146, opened 2024-06-14
Make boost::math::ccmath::fminf and boost::math::ccmath::fminl constexpr

### LINKED ISSUE #1145: Potentially missing inline/constexpr specifier in boost/math/ccmath/fmin.hpp
The functions `boost::math::ccmath::fminf` and `boost::math::ccmath::fminl` in [`include/boost/math/ccmath/fmin.hpp`](https://github.com/boostorg/math/blob/master/include/boost/math/ccmath/fmin.hpp) are neither marked `inline` nor `constexpr`. 
This leads to duplicate symbol definitions if `boost/math/ccmath/fmin.hpp` gets included in more than one TU in a binary. 
```
multiple definition of `boost::math::ccmath::fminf(float, float)';
multiple definition of `boost::math::ccmath::fminl(long double, long double)'
```
Am I using said header wrong or should these functions be marked `constexpr`?

Here's a minimal Godbolt example that shows the issue:
https://godbolt.org/z/9Yoz7PKoa

In case this turns out to be a bug, a PR has been prepared to fix this issue: #1146

quxflux:
The publicly visible declarations of `boost::math::ccmath::fminf` and `boost::math::ccmath::fminfl` are neither marked `inline` nor `constexpr`, which causes "multiple definition" linker errors when the `boost/math/ccmath/fmin.hpp` header gets included in multiple translation units of a binary.

Therefore this PR marks said functions `constexpr`.

Fixes #1145

### DIFF
--- a/include/boost/math/ccmath/fmin.hpp
+++ b/include/boost/math/ccmath/fmin.hpp
@@ -72,13 +72,13 @@ constexpr auto fmin(T1 x, T2 y) noexcept
     }
 }
 
-float fminf(float x, float y) noexcept
+constexpr float fminf(float x, float y) noexcept
 {
     return boost::math::ccmath::fmin(x, y);
 }
 
 #ifndef BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
-long double fminl(long double x, long double y) noexcept
+constexpr long double fminl(long double x, long double y) noexcept
 {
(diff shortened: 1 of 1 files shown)


mborland (approved):
LGTM. Approved the CI run. Thanks!

quxflux:
Thanks for the quick response 🙂

PR #1148, opened 2024-07-02
Merge for 1.86

mborland:


mborland:
@sdarwin Can you please enable priviliged access in drone for this repo? TSAN seems to have issues with memory mapping

sdarwin:
- boostorg/math was already set to "Trusted: Enables privileged container settings."
- In .drone.star or .drone.jsonnet, the step also needs to use `privileged=True` or the equivalent. However that seems to be set.
  
Try this in drone.sh:

```
if [[ $(uname) == "Linux" ]]; then
    echo 0 | sudo tee /proc/sys/kernel/randomize_va_space
fi

```

sdarwin:
And https://stackoverflow.com/questions/77850769/fatal-threadsanitizer-unexpected-memory-mapping-when-running-on-linux-kernels

`sudo sysctl vm.mmap_rnd_bits=28` or `sudo sysctl vm.mmap_rnd_bits=30`

PR #1149, opened 2024-07-02
Add randomize_va_space to drone

mborland:


mborland:
@sdarwin both `sudo sysctl vm.mmap_rnd_bits=28` and 30 give the same memory mapping error. Any other ideas, or does it need to go at an earlier point in the script?

sdarwin:
It should be 28: `sudo sysctl vm.mmap_rnd_bits=28` , and inside the` if [[ $(uname) == "Linux" ]]; then` , however if that is not working, try to use clang 18.1 

https://github.com/google/sanitizers/issues/1716 

Newer versions of TSan (LLVM 18.1.0 onwards: https://github.com/llvm/llvm-project/commit/0784b1eefa36d4acbb0dacd2d18796e26313b6c5) will automatically re-exec without ASLR, if the layout is incompatible. Additionally, TSan in LLVM 18.1.0 onwards can support 30 bits of ASLR entropy without disabling ASLR (https://github.com/llvm/llvm-project/commit/7d039effc4930be9240446a4241d268a39960e0b).

mborland:
Clang-18 seems to be working fine. Thanks for the help!

PR #1157, opened 2024-07-16
Add CUDA workflow and proof of concept

mborland:


mborland:
@sdarwin I added a workflow with `runs-on: gpu-runner-1`, but the job doesn't seem to be getting picked up. Is there something else that I need to add?

sdarwin:
There may be more than one blocker.  

The first step:  enabling a larger GitHub actions runner requires being an Owner of an organization, and then configuring billing and spending limits.   I'm not able to do this on `boostorg` so gpu-runner-1 exists in `cppalliance`.   

We might explore the options of setting up billing in `boostorg`.   However, since the effectiveness of the gpu tests is still unknown, how about creating a repository in `cppalliance` ?

mborland:
> There may be more than one blocker.
> 
> The first step: enabling a larger GitHub actions runner requires being an Owner of an organization, and then configuring billing and spending limits. I'm not able to do this on `boostorg` so gpu-runner-1 exists in `cppalliance`.
> 
> We might explore the options of setting up billing in `boostorg`. However, since the effectiveness of the gpu tests is still unknown, how about creating a repository in `cppalliance` ?

I'll try a similar PR on `cppalliance/decimal` since the concept is the same.

PR #1158, opened 2024-07-19
Add intel compiler to CI

mborland:


mborland:
@jzmaddock and @NAThompson Now that this is running fine most of these intel failures just look like little tolerance adjustments. The hypergeometric testing has something horrible wrong going on with a number of gross errors or infinities.

mborland:
Tested via SYCL now

PR #1160, opened 2024-07-20
Add support for modular build structure.

grafikrobot:
This is part of the effort to make the Boost libraries "modular" for build and consumption. See https://lists.boost.org/Archives/boost/2024/01/255704.php and https://github.com/grafikrobot/boost-b2-modular/blob/b2-modular/README.adoc for more information.

This PR depends on the following other PRs being merged to both develop and master branches of the respective repos:

- https://github.com/boostorg/boost/pull/854

This PR will be changed to ready for review, i.e. not draft, when the above are merged. Do not merge this one until that time.


jzmaddock:
@grafikrobot There's a lot of failures here, for example:

```
gcc.compile.c++ ../../../bin.v2/libs/math/test/test_zeta_floatmax_t.test/gcc-14/debug/x86_64/link-static/threading-multi/visibility-hidden/float128/test_zeta.o
In file included from float128/test_zeta.cpp:6:
float128/setup.hpp:19:62: error: static assertion failed: These tests should only be run for 128-bit floating point types.
   19 | static_assert(std::numeric_limits<boost::floatmax_t>::digits == 113, "These tests should only be run for 128-bit floating point types.");
      |               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
float128/setup.hpp:19:62: note: the comparison reduces to ‘(64 == 113)’
```

That test, and all the float128 tests are predicated on:

```
            [ check-target-builds ../config//has_128bit_floatmax_t "128-bit floatmax_t" : : <build>no ]
```

And that configure step is showing "no" and so should NOT be building?

jzmaddock:
@grafikrobot thanks for the fixes, however there are still configuration related failures, for example if we stick to this job: https://drone.cpp.al/boostorg/math/2055/66/2 then the first failure is a linker error for test_2F0 - it's looking for libquadmath which isn't present.  The Jamfile for that target has:

```
[ check-target-builds ../config//has_float128 "GCC libquadmath and __float128 support" : <define>BOOST_MATH_TEST_FLOAT128 <linkflags>"-Bstatic -lquadmath -Bdynamic" ]
```

and the configuration output says:

```
   - GCC libquadmath and __float128 support : no [3]
```

But it's still trying to link against it.

There are other configuration failings as well, that's just the first one in the log.

Thanks!

jzmaddock:
Apologies Rene, no they are not consistent, and no they are not really meant to be, each test has it's own set of requirements, some of them are "if this is available link to it because a dependency needs it", some are "if this is available enable testing" and so on.

Plus it's not just the float128 configurations that are failing: I highlighted that because it was the first and most numerous thing to show up, but there are others as well as I think.

jzmaddock:
> Plus it's not just the float128 configurations that are failing: I highlighted that because it was the first and most numerous thing to show up, but there are others as well as I think.

Or perhaps not as this appears to be green now.

In principle though, shouldn't each conditional usage be able to define it's own set of requirements depending on what that target needs?

mborland:
> Question is.. Did you intend to make all those different? Or should they be uniform and it's a historical mistake that they differ?
> 

`has_float128` and `has_intel_quad` are separate because they represent two different types that are platform specific. `has_128bit_floatmax_t` would be if either of the previous are defined. Can `b2` differentiate between intel compiler being `icpc` or `icpx`? `icpx` is the new clang based intel compiler, and it does not support all the compiler flags that `icpc` uses.

mborland:
This is effectively green now with only Codecov failing. Based on the massive diff stat shown in the codecov comment it looks like something went wrong parsing an upload since all of those runs are green. Anything else @jzmaddock before I squash and merge this?

PR #1161, opened 2024-07-23
Initial CUDA and SYCL support

mborland:
Adds CUDA support for binomial distribution
Adds CUDA and SYCL support for beta function and arcsine distribution.

 This is just the changes to the include directory cherry picked from: https://github.com/cppalliance/cuda-math/pull/2.


include/boost/math/special_functions/beta.hpp
```diff
@@ -31,14 +32,16 @@ namespace detail{
 // Implementation of Beta(a,b) using the Lanczos approximation:
 //
 template <class T, class Lanczos, class Policy>
-T beta_imp(T a, T b, const Lanczos&, const Policy& pol)
+BOOST_MATH_GPU_ENABLED T beta_imp(T a, T b, const Lanczos&, const Policy& pol)
 {
    BOOST_MATH_STD_USING  // for ADL of std names
 
+   #ifndef BOOST_MATH_NO_EXCEPTIONS
```
jzmaddock, line 39: Shouldn't we keep this block, even with no exceptions, and let the error handlers return NaN?  Otherwise for invalid arguments, pretty much anything could happen, including infinite loops or whatever...

include/boost/math/special_functions/beta.hpp
```diff
@@ -911,13 +916,17 @@ template <>
 struct Pn_size<float>
 {
    static constexpr unsigned value = 15; // ~8-15 digit accuracy
+#ifndef BOOST_MATH_HAS_GPU_SUPPORT
```
jzmaddock, line 919: Do we still have no static_assert?  That's a shame :(

include/boost/math/tools/polynomial.hpp
```diff
@@ -418,7 +418,7 @@ class polynomial
 #endif
    }
 
-   polynomial<T> integrate() const
+   BOOST_MATH_GPU_ENABLED polynomial<T> integrate() const
```
jzmaddock, line 421: Question: can we actually use this on a device - with std::vector?

jzmaddock:
Thanks Matt - this is looking really cool!

I guess testing wise, my main worry is that we see a random PR go green and merging it accidentally breaks CUDA/Sycl builds because something isn't correctly marked up.  If you're able to test building and linking for CUDA (even if not running) on a regular Github runner that would be amazing!

include/boost/math/tools/polynomial.hpp
mborland: Looks like no: https://github.com/NVIDIA/cccl/discussions/1032

include/boost/math/policies/error_handling.hpp
```diff
@@ -348,32 +376,36 @@ inline constexpr T raise_overflow_error(
 }
 
 template <class T>
-inline T raise_overflow_error(
+BOOST_MATH_GPU_ENABLED inline T raise_overflow_error(
            const char* ,
            const char* ,
            const  ::boost::math::policies::overflow_error< ::boost::math::policies::errno_on_error>&) BOOST_MATH_NOEXCEPT(T)
 {
+   #ifndef BOOST_MATH_SYCL_ENABLED
```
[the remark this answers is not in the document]
mborland: I was thinking adding a GPU default policy of:

```
using gpu_policy = policy<
      promote_float<false>,
      promote_double<false>,
      overflow_error<ignore_error>
      >;
```

I guess this could also just overwrite the value of `policies::policy<>()` when `BOOST_MATH_HAS_GPU_SUPPORT` is defined?

include/boost/math/special_functions/beta.hpp
mborland: The `static_assert` should be fine, but the issue here is `operator[]` on std::array can't be used.

mborland:
@jzmaddock I added a complete SYCL run, and just the compilation step for CUDA like you have for Config.

mborland:
Pending any other questions or comments I think this initial push is good to merge. I will continue to work on this in https://github.com/cppalliance/cuda-math and then port things over once they're tested

include/boost/math/special_functions/beta.hpp
jzmaddock: Doh!

include/boost/math/policies/error_handling.hpp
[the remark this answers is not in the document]
jzmaddock: Right, when building for the device, set the default error handling to ignore_error for everything, ie:

```
#    define BOOST_MATH_DOMAIN_ERROR_POLICY ignore_error
#     define BOOST_MATH_POLE_ERROR_POLICY ignore_error
#     define BOOST_MATH_OVERFLOW_ERROR_POLICY ignore_error
#     define BOOST_MATH_EVALUATION_ERROR_POLICY ignore_error
#     define BOOST_MATH_ROUNDING_ERROR_POLICY ignore_error
```

Currently the default when exceptions are disabled is `errno_on_error`, but we have no access to `::errno` so don't want that here.

jzmaddock:
> Pending any other questions or comments I think this initial push is good to merge.

Go for it when you're ready! :)

PR #1162, opened 2024-07-24
Batch number 2 of GPU support

mborland:
Adds CUDA and SYCL support to the following: Cauchy distribution, cbrt, sign manipulation functions, trunc, round, modf.

Changes the default policy of CUDA to match that of SYCL and adds GPU specific handling to the overflow ignore error handler (i.e. gets rid of use of `std::numeric_limits`)

### DIFF
--- a/include/boost/math/distributions/cauchy.hpp
+++ b/include/boost/math/distributions/cauchy.hpp
@@ -13,6 +13,7 @@
 #pragma warning(disable : 4127) // conditional expression is constant
 #endif
 
+#include <boost/math/tools/config.hpp>
 #include <boost/math/distributions/fwd.hpp>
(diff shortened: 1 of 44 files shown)


mborland:
The drone build on this is actually green: https://drone.cpp.al/boostorg/math/1975. Not sure why it's still listed as a pending workflow. @jzmaddock let me know if you have any comments. This one has fewer substantial changes than the previous batch.

mborland:
I think this is safe to merge. In the next batch I wrote my own complete version of `limits` to get around the fact that libcu++ does not provide it which significantly simplifies some of the changes. Once it's here in math I'll ask libcu++ if they want me to port it to them.

PR #1164, opened 2024-07-26
GPU batch 3

mborland:
Adds CUDA and SYCL support for `log1p`, `log1pmx`, `digamma`, `trigamma`, `sin_pi`, `cos_pi`, `pow`, and the bernoulli distribution.

Significant changes: Adds a complete implementation of `<limits>` in tools/numeric_limits.hpp. This allows us to work around it missing from libcu++ without giant chains of `std::is_same_v` and `if constexpr`. The other big change is the method of dispatching for `sin_pi` and `trigamma` to avoid recursion that is disallowed by SYCL. `trigamma` also now uses `std::integral_constant` directly rather than through a pointer because that confused NVCC quite badly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -606,7 +606,7 @@ jobs:
       - name: Build tests
         run: |
           cd ../boost-root/__build__
-          cmake --build . --target tests
+          cmake --build . --target tests -j $(nproc)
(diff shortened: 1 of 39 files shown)


mborland:
Only failure was the out of memory on drone that is now fixed.

PR #1165, opened 2024-07-29
GPU Batch 4

mborland:
Adds support for: exponential dist, landau dist, mapairy dist, holtsmark dist, saspoint5 dist, extreme value dist, laplace dist, and logistic dist.

Adds new `boost::math::tools::pair` which is just an alias of `std::pair` or `thrust::pair` depending on context. Same goes for tuple and all the additional helper functions (e.g. make_pair). This fixes `range()` and `support()` functions for distributions on CUDA devices. 

Edit: @tk-yoshimura you may be interested in this since it adds both CUDA and SYCL support to your newly contributed distributions. The changes were fairly small with the biggest being recursion was eliminated from the `quantile` function to support SYCL.


mborland:
Only CI failure is a clone fail. The disk memory errors have been cleaned up as expected

PR #1167, opened 2024-07-30
GPU Batch 5

mborland:
Adds GPU support to: sqrt1pm1, erf, erfc, erf_inv, erfc_inv, powm1, lgamma, tgamma, and associated gamma functions.

Removes recursion from lgamma and tgamma by adding dispatch functions to handle the reflection case.


mborland:
@jzmaddock Can you see any reason why Windows platforms would not like the diffs for `erfc_inv` and `tgamma`? The former just changes the dispatching method from pointers to `std::integral_constant` to references and the latter adds a dispatching function. I would assume if there was something fundamentally wrong with the way dispatching is occuring it would hit other platforms. Both are throwing on numeric overflow

dschmitz89:
@mborland : this looks awesome. Could you give a little high level overview of what the goals of these GPU efforts in Boost.Math are? SciPy is doing something similar at the moment and we are also using some special functions from you guys. CC @steppi in case he did not see these recent GPU PRs

mborland:
> @mborland : this looks awesome. Could you give a little high level overview of what the goals of these GPU efforts in Boost.Math are? SciPy is doing something similar at the moment and we are also using some special functions from you guys. 

I think Albert's summary is pretty good. @jzmaddock has a quite old PR adding CUDA support here: https://github.com/boostorg/math/pull/127 so I have been extracting from and expanding on that. The development on this repo: https://github.com/cppalliance/cuda-math since my employer can only directly add GPU enabled CI there, and then cherry picking onto boost proper. Two questions for you two @dschmitz89 and @steppi 

1) Is there a specific prioritization of functions or distributions that would benefit you? 

2) Are you looking for CUDA only or are there other platforms that you need/want covered? We have been inserting both CUDA and SYCL support so far. SYCL has some more significant limitations like no recursion. So far it's been relatively easy to work around by adding dispatching functions but there will be functions that need to be completely re-written or will remain unsupported such as anything involving root finding so I'd rather punt rewrites until there is real demand.

izaid:
I think this is great too! I'll let @steppi answer any specific prioritization of functions, but the ones you've done so far look good to me. I think we really only care about CUDA for now, but others may think differently.

@mborland I wanted to throw in one point though. Are you testing these using just nvcc, or nvrtc as well? nvrtc is CUDA's runtime (JIT) compiler, and is what CuPy uses for instance. It has more stringent requirements...

steppi:
> 2\. anything involving root finding

I’m trying to figure out root finding that will work on GPU now. I’ll keep you posted on my progress.

mborland:
> @mborland I wanted to throw in one point though. Are you testing these using just nvcc, or nvrtc as well? nvrtc is CUDA's runtime (JIT) compiler, and is what CuPy uses for instance. It has more stringent requirements...

Right now it's just NVCC but we can look into adding NVRTC as well. Tracked here: https://github.com/cppalliance/cuda-math/issues/7.

> I’m trying to figure out root finding that will work well on GPU now. I’ll keep you posted on my progress.

Cool. Thank you

dschmitz89:
Nice that the conversation is starting. :) On the SciPy side I would defer to @steppi for all technical details as I don't have expertise in CUDA or special function computations.

mborland:
> Nice that the conversation is starting. :) On the SciPy side I would defer to @steppi for all technical details as I don't have expertise in CUDA or special function computations.

Feel free to open further issues/discussion here or downstream at: https://github.com/cppalliance/cuda-math. Either way it will get seen. 

Only CI failure is from codecov so I am going to merge this one in.

PR #1170, opened 2024-08-06
Add proof of concept for NVRTC support

mborland:
Development and completed CUDA runs can be found here: https://github.com/cppalliance/cuda-math/pull/8

This one was requested by @izaid for CuPy usage. To me it seems beneficial to just alias libcu++ anywhere we can for this level of support because NVRTC has extremely strict rules. We're going to have to create aliases to everything in `std::` to either point to `cuda::std::`, `thrust::`, or roll our own implementation for things that don't exist such as the already completed implementation of `<limits>`. I'm also not quite sure how to pass relative paths to `nvrtcCompileProgram` so right now I have everything as hard-coded full paths. This is also for runtime compilation so we get less useful results out of just compilation than we do with NVCC here, but everything is still run first through the cuda-math repo. 

Any thoughts: @jzmaddock, @ckormanyos, @NAThompson?

### DIFF
--- a/include/boost/math/special_functions/gamma.hpp
+++ b/include/boost/math/special_functions/gamma.hpp
@@ -14,6 +14,22 @@
 #pragma once
 #endif
 
+#ifdef __CUDACC_RTC__
+
+namespace boost {
+namespace math {
+
+template <typename T>
+__host__ __device__ T tgamma(T x)
+{
+   return ::tgamma(x);
+}
+
+} // namespace math
+} // namespace boost
+
+#else
+
 #include <boost/math/tools/config.hpp>
 #include <boost/math/tools/series.hpp>
 #include <boost/math/tools/fraction.hpp>
@@ -2279,5 +2295,7 @@ BOOST_MATH_GPU_ENABLED inline tools::promote_args_t<T1, T2>
 #include <boost/math/special_functions/detail/igamma_inverse.hpp>
 #include <boost/math/special_functions/detail/gamma_inva.hpp>
 #include <boost/math/special_functions/erf.hpp>
+
+#endif // __CUDACC_RTC__
 
 #endif // BOOST_MATH_SF_GAMMA_HPP
--- a/include/boost/math/special_functions/gamma.hpp
+++ b/include/boost/math/special_functions/gamma.hpp
@@ -14,21 +14,7 @@
 #pragma once
 #endif
 
-#ifdef __CUDACC_RTC__
-
-namespace boost {
-namespace math {
-
-template <typename T>
-__host__ __device__ T tgamma(T x)
-{
-   return ::tgamma(x);
-}
-
-} // namespace math
-} // namespace boost
-
-#else
+#ifndef __CUDACC_RTC__
 
 #include <boost/math/tools/config.hpp>
 #include <boost/math/tools/series.hpp>
@@ -2295,6 +2281,23 @@ BOOST_MATH_GPU_ENABLED inline tools::promote_args_t<T1, T2>
 #include <boost/math/special_functions/detail/igamma_inverse.hpp>
 #include <boost/math/special_functions/detail/gamma_inva.hpp>
 #include <boost/math/special_functions/erf.hpp>
+
+#else
+
+namespace boost {
+namespace math {
+
+inline __host__ __device__ float tgamma(float x) { return ::tgammaf(x); }
+inline __host__ __device__ double tgamma(double x) { return ::tgamma(x); }
+
+template <typename T, typename Policy>
+inline __host__ __device__ T tgamma(T x, const Policy&)
+{
+   return boost::math::tgamma(x);
+}
+
+} // namespace math
+} // namespace boost
 
 #endif // __CUDACC_RTC__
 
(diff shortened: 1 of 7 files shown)


izaid:
Thanks for setting this up @mborland! And ccing @steppi here.

Yes, NVRTC is quite strict, but unfortunately it's used in various places so we would need to support it. We went through this as well, and our solution indeed was just to alias the relevant parts of `std`.

mborland:
@izaid, @steppi, @dschmitz89 Do any of you want or need ROCm support as long as we are going back through this?

izaid:
I don't think we need ROCm support, CUDA is the big win.

include/boost/math/special_functions/gamma.hpp
```diff
@@ -14,6 +14,22 @@
 #pragma once
 #endif
 
+#ifdef __CUDACC_RTC__
+
+namespace boost {
+namespace math {
+
+template <typename T>
+__host__ __device__ T tgamma(T x)
```
jzmaddock, line 23: It's somewhat a matter of taste, but I would probably go with:

```
inline __host__ __device__ float tgamma(float x) { return ::tgammaf(x); }
inline __host__ __device__ double tgamma(double x) { return ::tgamma(x); }
```

Certainly it should be inline at least.  I would also add overloads that accept a Policy (and which ignore it), so that when other code calls tgamma it has something to forward to.

jzmaddock:
> We're going to have to create aliases to everything in std:: to either point to cuda::std::, thrust::

Since our calls are all unqualified, can we not just change BOOST_MATH_STD_USING to point to the appropriate namespace?

steppi:
> Since our calls are all unqualified, can we not just change BOOST_MATH_STD_USING to point to the appropriate namespace?

@izaid and I wanted to do the equivalent of this in SciPy. The hitch was that to get things working in CuPy (which uses NVRTC) we needed a patchwork of different namespaces.  `thrust` for complex arithmetic, mostly `cuda::std`, sometimes just `::` from `cuda_runtime.h`, and I think even had to copy and paste some stuff from the standard library and add the CUDA markup.

include/boost/math/special_functions/gamma.hpp
steppi: I think the template functions are automatically inline, but I agree about having separate float and double versions, if that’s what you mean here. Usually you see things like having minimax polynomial expansions which are different for different precision levels. (I know you know this @jzmaddock because I’ve seen this done in the boost math codebase, just pointing it out for everyone else.)

mborland:
> > We're going to have to create aliases to everything in std:: to either point to cuda::std::, thrust::
> 
> Since our calls are all unqualified, can we not just change BOOST_MATH_STD_USING to point to the appropriate namespace?

The equivalent of `BOOST_MATH_STD_USING` is easy since all the mathematical functions require no special headers and are all in the global namespace. The things we will need from `thrust::` are containers, and we have aliases already like tuple: https://github.com/boostorg/math/blob/develop/include/boost/math/tools/tuple.hpp#L14. Utilities are in `cuda::std::` such as a version of `<type_traits>`, etc.: https://nvidia.github.io/cccl/libcudacxx/standard_api/utility_library/type_traits.html. My thought is to create a bunch of alias headers in `boost/math/tools` like `boost::math::tuple` so if you call say `boost::math::is_same_v` it is either `std::is_same_v` or `cuda::std::is_same_v` depending on context. For posterity these could all be extracted and put into a central place. @jzmaddock what do you think about landing all these kind of context aware aliases in Config? I thought it might be out of scope, but wanted your thoughts. I'd rather not propose a Boost.GPU_Compat library,

jzmaddock:
> My thought is to create a bunch of alias headers in boost/math/tools like boost::math::tuple so if you call say boost::math::is_same_v it is either std::is_same_v or cuda::std::is_same_v depending on context. For posterity these could all be extracted and put into a central place. @jzmaddock what do you think about landing all these kind of context aware aliases in Config? I thought it might be out of scope, but wanted your thoughts. I'd rather not propose a Boost.GPU_Compat library,

It's not unreasonable to push this to Config: I'm sure other libraries could make use of this stuff too... but do we want to support this in standalone mode?  It might be worth at least raising this on the mailing list?

mborland:
> It's not unreasonable to push this to Config: I'm sure other libraries could make use of this stuff too... but do we want to support this in standalone mode? It might be worth at least raising this on the mailing list?

I would keep the original copies here in math so we can continue supporting standalone mode, because that is what the Python packages depend on. I'll post on the ML.

mborland:
> > It's not unreasonable to push this to Config: I'm sure other libraries could make use of this stuff too... but do we want to support this in standalone mode? It might be worth at least raising this on the mailing list?
> 
> I would keep the original copies here in math so we can continue supporting standalone mode, because that is what the Python packages depend on. I'll post on the ML.

Looks like the ML is in favor of everything going in it's own library. Pending any other comments I think this is a good limited proof that we can make it work. I'll keep building out the basic functions e.g. `fpclassify` not provided by CUDA and see if we can get a simple distribution working.

PR #1172, opened 2024-08-08
NVRTC Batch 2

mborland:
Adds support to: sign functions, `lgamma`, and all the fpclassify functions.

Internally this adds a complete implementation of `<type_traits>` that is usable in both CUDA and non-CUDA environments. I also added the C++17 overloads because they are ergonomic. I also had to add definitions of the macros that fpclassify checks against (e.g. `FP_NAN` is now `BOOST_MATH_FP_NAN`) since they are not provided in the NVRCT environment. 

Downstream run with completed CI can be found here: https://github.com/cppalliance/cuda-math/pull/9

Based on the conversation from the boost dev mailing list ports of all the utilities that are getting built in path will be added here: https://github.com/cppalliance/GPUtils so that the wheel does not need to be reinvented again. @steppi that repo may or may not be useful to you in the future.

### DIFF
--- a/include/boost/math/special_functions/cbrt.hpp
+++ b/include/boost/math/special_functions/cbrt.hpp
@@ -1,4 +1,5 @@
 //  (C) Copyright John Maddock 2006.
+//  (C) Copyright Matt Borland 2024.
 //  Use, modification and distribution are subject to the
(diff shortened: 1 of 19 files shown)


mborland:
Only failure is from Drone failure to clone so this should be safe to merge.

PR #1176, opened 2024-08-12
Document logpdf, logcdf, and specializations

mborland:
@jzmaddock and @dschmitz89 let me know what you think of these additions.

### DIFF
--- a/doc/distributions/dist_tutorial.qbk
+++ b/doc/distributions/dist_tutorial.qbk
@@ -128,11 +128,48 @@ And quantiles are just the same:
    quantile(my_dist, p);  // Returns the value of the random variable x
                           // such that cdf(my_dist, x) == p.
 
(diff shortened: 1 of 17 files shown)


jzmaddock:
Thanks @mborland that reads great, we also need some reference docs in non_members.qbk, the `__usual_accessors` macro needs updating, and it would be good to have some some boilerplate text to add to each individual distribution's docs after the __usual_accessors macro to indicate whether the logcdf/pdf accessors add anything that log(cdf) doesn't.

dschmitz89:
This looks great thanks. Could you also add computational details on the distribution specific pages for the high precision implementations? For the other methods these are often provided.

mborland:
> This looks great thanks.

Formulas have been added in https://github.com/boostorg/math/pull/1176/commits/65096bee14b66788eba3694fc7d317aa8854d845.

PR #1180, opened 2024-08-15
NVRTC Batch 5

mborland:
Finally, we have enough foundation on NVRTC that we can now add distributions.

Adds NVRTC support to: arcsine dist, bernoulli dist, cauchy dist, exponential dist, extreme value dist, holtsmark dist, landau dist, laplace dist, logistic dist, mapairy dist, and saspoint5 dist. This is the same support that currently exists for NVCC

Replaces the use of `thurst::pair` with `cuda::std::pair` so that it works with NVRTC. Implement our own versions of `cuda::std::forward` and `cuda::std::make_tuple` since they are missing from libcu++.

NVRTC support is almost in line with what is supported in NVCC and SYCL minus a few outstanding special functions that are not in the libcu++ library. Those will be in the next batch so that development moving forward will apply to all platforms at once rather than piecemeal.

Completed CI runs on device can be found at: https://github.com/cppalliance/cuda-math/pull/13

CC: @steppi, @izaid, @dschmitz89

### DIFF
--- a/include/boost/math/distributions/arcsine.hpp
+++ b/include/boost/math/distributions/arcsine.hpp
@@ -30,24 +30,27 @@
 #ifndef BOOST_MATH_DIST_ARCSINE_HPP
 #define BOOST_MATH_DIST_ARCSINE_HPP
 
-#include <cmath>
 #include <boost/math/tools/config.hpp>
 #include <boost/math/tools/tuple.hpp>
-#include <boost/math/distributions/fwd.hpp>
 #include <boost/math/distributions/complement.hpp> // complements.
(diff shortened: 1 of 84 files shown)


dschmitz89:
Same here, on behalf of the Python community, thank you very much, this whole undertaking will be an important step for reusable special functions throughout the ecosystem!

mborland:
Only CI failure is a bad connect downloading packages. Should be safe to merge.

mborland:
> Excellent! I'll start looking into getting this working in CuPy.

Let us know how it goes and feel free to ping us with issues you find.

PR #1181, opened 2024-08-16
Fix warning C4701: "potentially uninitialized local variable used"

fsb4000:
Hello.
I tried to use your library (1.86.0)  and I faced with an issue. I didn't have the issue with 1.85.0.
I use VS 2022 and I get an error.
```
boost-math\include\boost\math\special_functions\detail\bessel_ik.hpp(417) : error C2220: the following warning is treated as an error
boost-math\include\boost\math\special_functions\detail\bessel_ik.hpp(417) : warning C4701: potentially uninitialized local variable 'n' used
boost-math\include\boost\math\special_functions\detail\bessel_ik.hpp(417) : warning C4701: potentially uninitialized local variable 'u' used
```
As you can see we initialize `n` and `u` only in the `else` branch and later we use them in `if (reflect)`. We might not get into this `if` if we haven't been in that `else` but the compiler don't know about it. I tried to fix the issue.

### DIFF
--- a/include/boost/math/special_functions/detail/bessel_ik.hpp
+++ b/include/boost/math/special_functions/detail/bessel_ik.hpp
@@ -18,7 +18,6 @@
 #include <boost/math/constants/constants.hpp>
 #include <boost/math/policies/error_handling.hpp>
 #include <boost/math/tools/config.hpp>
-#include <iostream>
 
 // Modified Bessel functions of the first and second kind of fractional order
 
--- a/include/boost/math/special_functions/detail/bessel_ik.hpp
+++ b/include/boost/math/special_functions/detail/bessel_ik.hpp
@@ -18,6 +18,7 @@
 #include <boost/math/constants/constants.hpp>
 #include <boost/math/policies/error_handling.hpp>
 #include <boost/math/tools/config.hpp>
+#include <iostream>
 
 // Modified Bessel functions of the first and second kind of fractional order
 
@@ -334,8 +335,6 @@ int bessel_ik(T v, T x, T* result_I, T* result_K, int kind, const Policy& pol)
        T eight_z = 8 * x;
        Kv = 1 + (mu - 1) / eight_z + (mu - 1) * (mu - 9) / (2 * eight_z * eight_z) + (mu - 1) * (mu - 9) * (mu - 25) / (6 * eight_z * eight_z * eight_z);
        Kv *= exp(-x) * constants::root_pi<T>() / sqrt(2 * x);
-       n = 0;
-       u = 0;
     }
     else
     {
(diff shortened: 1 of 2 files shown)


ckormanyos:
Cool find. I'll try to maybe find/add a test case that hits the lines?

ckormanyos:
> Cool find. I'll try to maybe find/add a test case that hits the lines?

Wait a sec, ...

According to the [covrage report](https://app.codecov.io/gh/boostorg/math/blob/develop/include%2Fboost%2Fmath%2Fspecial_functions%2Fdetail%2Fbessel_ik.hpp), the lines _are_ being hit. Is there maybe a better fix? Delay declaration of the variables or `static_cast<void>`?

jzmaddock:
I would need to double check, but I seem to remember fixing this while doing some code-coverage work.  Ah, yes it is, see: https://github.com/boostorg/math/blob/develop/include/boost/math/special_functions/detail/bessel_ik.hpp

ckormanyos:
I think you can declare n and u at lines 340'and 341? Isn't that the point of the warning?

fsb4000:
>I think you can declare n and u at lines 340'and 341? Isn't that the point of the warning?

They are used not only in the else but later in another if.
https://github.com/boostorg/math/blob/5a4f8bab07aae3792614cf49e845cdf7f85c6e2d/include/boost/math/special_functions/detail/bessel_ik.hpp#L338-L341

https://github.com/boostorg/math/blob/5a4f8bab07aae3792614cf49e845cdf7f85c6e2d/include/boost/math/special_functions/detail/bessel_ik.hpp#L415-L417

Ok, I will try to create a test case...

fsb4000:
I found a test case for that:
```
boost::math::cyl_bessel_k(-2.25, std::numeric_limits<double>::infinity());
```
Where can I find what result should be for this?
I added 
```
std::cout << "v = " << v<< " x = " << x << " kind = " << kind << " reflect = " << reflect <<std::endl;
std::cout << "u = " << u << " n = " << n << std::endl;
```
before `T z = (u + n % 2);` and I got:
```
v = 2.25 x = inf kind = 2 reflect = 1
u = 3.18226e-312 n = 4147116544
```
the call result was `0`
What is the real result should be?

fsb4000:
I found a way to fail for that case. And I think I found a more correct fix: calculate `n` and `u` before the `if`.
I would also add the warning C4701 to your CI but I don't know what to add to https://github.com/boostorg/math/blob/develop/.github/workflows/ci.yml for `toolset: [ msvc-14.2 ]`

jzmaddock:
My bad, you are correct that this is NOT fixed in develop, but IS in this branch/PR: https://github.com/boostorg/math/pull/1111 that is still at least somewhat work in progress.  I would prefer not to create two competing fixes if possible, as the coverage test branch will be merged fairly soon anyway.  Please leave this open and I'll take a look at your test case once I get some time (which will be a couple of weeks).

StephanTLavavej:
@jzmaddock These warnings are still blocking MSVC's STL from updating its version of Boost.Math. Might you have a chance to look at this soon?

jzmaddock:
Apologies @StephanTLavavej, since I haven't merged the other branch, lets just go ahead and merge this one, and I'll deal with the conflicts later.

PR #1183, opened 2024-08-19
Add the `pow1p` function

mborland:
This implements the `pow1p` function as discussed and implemented in https://github.com/scipy/scipy/issues/21404 and https://github.com/scipy/scipy/pull/21410. Simplifies a couple of the branches and adds in policy based design for the error paths. This version is also tested with `real_concept` and on the GPU platforms (NVCC, NVRTC, and SYCL). 

@fancidev let me know if you see any errors in simplification. The one to call out is here: https://github.com/boostorg/math/compare/pow1p?expand=1#diff-1a6b6446fd424370a1fe6e67b8166e41ab4805c1ecfc4ca15ceae7492f4023a5R137. I fall back to `pow` instead of throwing on non-integer power.

@NAThompson and @jzmaddock  your review would be appreciated as always.

CC: @WarrenWeckesser

### DIFF
--- a/doc/math.qbk
+++ b/doc/math.qbk
@@ -313,6 +313,7 @@ and use the function's name as the link text.]
 [def __sqrt1pm1 [link math_toolkit.powers.sqrt1pm1 sqrt1pm1]]
 [def __rsqrt [link math_toolkit.powers.rsqrt rsqrt]]
 [def __powm1 [link math_toolkit.powers.powm1 powm1]]
+[def __pow1p [link math_toolkit.powers.pow1p pow1p]]
 [def __hypot [link math_toolkit.powers.hypot hypot]]
 [def __pow [link math_toolkit.powers.ct_pow pow]]
 
--- a/doc/sf/pow1p.qbk
+++ b/doc/sf/pow1p.qbk
@@ -1,0 +1,34 @@
+[/
+  Copyright Matt Borland 2024
+  Distributed under the Boost Software License, Version 1.0.
+  (See accompanying file LICENSE_1_0.txt or copy at
+  http://www.boost.org/LICENSE_1_0.txt).
+]
+
+[section:pow1p pow1p]
+
+[h4 Synopsis]
+
+    #include <boost/math/special_functions/pow1p.hpp>
+
+    namespace boost {
+    namespace math {
+
+    template <typename T1, typename T2, typename Policy>
+    BOOST_MATH_GPU_ENABLED tools::promote_args_t<T1, T2>
+    pow1p(const T1 x, const T2 y, const Policy& pol)
+
+    template <typename T1, typename T2>
+    BOOST_MATH_GPU_ENABLED tools::promote_args_t<T1, T2>
+    pow1p(const T1 x, const T2 y)
+
+    }} // namespaces
+
+
+The function `pow1p` computes (1 + x)[super y ] where x and y are real numbers. 
+This function is particularly useful for scenarios where adding 1 to x before raising it to the power of y is required. 
+It provides a more numerically stable and efficient way to perform this computation, especially for small values of x.
+When x is very small, directly computing (1 + x)[super y ] using standard arithmetic operations can lead to a loss of precision due to floating-point arithmetic limitations. 
+The pow1p function helps mitigate this issue by internally handling such cases more accurately.
+
+[endsect]
--- a/include/boost/math/special_functions.hpp
+++ b/include/boost/math/special_functions.hpp
@@ -79,6 +79,7 @@
 #include <boost/math/special_functions/lambert_w.hpp>
 #include <boost/math/special_functions/gegenbauer.hpp>
 #include <boost/math/special_functions/jacobi.hpp>
+#include <boost/math/special_functions/pow1p.hpp>
 #ifndef BOOST_MATH_NO_EXCEPTIONS
 #include <boost/math/special_functions/legendre_stieltjes.hpp>
 #endif
--- a/include/boost/math/special_functions/pow1p.hpp
+++ b/include/boost/math/special_functions/pow1p.hpp
@@ -17,13 +17,10 @@
 #include <boost/math/policies/error_handling.hpp>
 
 // For cuda we would rather use builtin nextafter than unsupported boost::math::nextafter
-// NVRTC does not support the forward declarations header
 #ifndef BOOST_MATH_HAS_GPU_SUPPORT
 #  include <boost/math/special_functions/next.hpp>
-#  ifndef BOOST_MATH_HAS_NVRTC
-#    include <utility>
-#    include <boost/math/special_functions/math_fwd.hpp>
-#  endif // BOOST_MATH_HAS_NVRTC
+#  include <boost/math/special_functions/math_fwd.hpp>
+#  include <utility>
 #endif // BOOST_MATH_ENABLE_CUDA
 
 namespace boost {
@@ -52,6 +49,20 @@ BOOST_MATH_FORCEINLINE BOOST_MATH_GPU_ENABLED T local_fma(const T x, const T y, 
     return x * y + z;
 }
 
+#else
+
+template <typename T>
+BOOST_MATH_FORCEINLINE BOOST_MATH_GPU_ENABLED T local_fma(const T x, const T y, const T z)
+{
+    return ::fma(x, y, z);
+}
+
+template <>
+BOOST_MATH_FORCEINLINE BOOST_MATH_GPU_ENABLED float local_fma(const float x, const float y, const float z)
+{
+    return ::fmaf(x, y, z);
+}
+
 #endif // BOOST_MATH_HAS_GPU_SUPPORT
 
 template <typename T, typename Policy>
@@ -71,7 +82,7 @@ BOOST_MATH_GPU_ENABLED T pow1p_imp(const T x, const T y, const Policy& pol)
     if (x == T(-1))
     { 
         // pow(+/-0, y)
-        if (boost::math::isfinite(y) && y < 0) 
+        if ((boost::math::isfinite)(y) && y < 0) 
         {
             return boost::math::policies::raise_domain_error<T>(function, "Division by 0", x, pol);
         }
@@ -80,7 +91,7 @@ BOOST_MATH_GPU_ENABLED T pow1p_imp(const T x, const T y, const Policy& pol)
         return pow(T(0), y);
     }
 
-    if (x == T(-2) && boost::math::isinf(y)) 
+    if (x == T(-2) && (boost::math::isinf)(y)) 
     { 
         // pow(-1, +/-inf)
         return T(1);
@@ -92,9 +103,9 @@ BOOST_MATH_GPU_ENABLED T pow1p_imp(const T x, const T y, const Policy& pol)
         return T(1);
     }
 
-    if (boost::math::isinf(y))
+    if ((boost::math::isinf)(y))
     {
-        if (boost::math::signbit(y) == 1)
+        if ((boost::math::signbit)(y) == 1)
         {
             // pow(x, -inf)
             return T(0);
@@ -106,7 +117,7 @@ BOOST_MATH_GPU_ENABLED T pow1p_imp(const T x, const T y, const Policy& pol)
         }
     }
 
-    if (boost::math::isinf(x)) 
+    if ((boost::math::isinf)(x)) 
     { 
         // pow(+/-inf, y)
         return pow(x, y);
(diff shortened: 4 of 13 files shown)


fancidev:
It’s really cool to have this in boost! Thanks a ton!

There is some refinement I’m looking to apply: (1) represent the result of $s/t$ using double-double, and (2) expand $\ln(1+u)$ to second order. I think with these refinements 
the implementation is (more or less) provably accurate to a few machine epsilon. The current code still have some input that leads to relative error over `1e-14`.

I’m going to work out the (textual) proof first, and then suggest the change here.

Another optimization that could be applied is that if a floating point type is available whose mantissa has more bits than the total bits of $T$, such as double vs float or x87 long double vs double, then it should be accurate enough (for T) to compute $(1+x)^y$ using `exp(y*log1p(x))`, provided `*`, `log1p`, and `exp` are accurate for the extended precision type. I’m going to verify this on paper and through numerical experiments too.

Finally, I’m going to prepare some edge input as test cases, to complement your randomized test cases, which are really great!

P.S. The (floating point) exception signaling (via the policy object) might have some corner cases to refine, but I’d leave that after the computational part is finalized.

Thanks again for bring this into boost!!

fancidev:
I made an update to the algorithm (in the referenced PR) that makes the evaluation more accurate for tiny $x$ and huge $y$:

https://github.com/scipy/scipy/pull/21410/commits/a3f89c50a20a6f5c1506b8e2c0756a499190d667

I think the method is now accurate to a few epsilon relative error for all $x > -1$ and all $y$, and I don't intend to make any further change to this part. A bit more update may come for the case $x < -1$. I'm still in the progress of adding more test cases and completing the algorithm description.

@mborland  If you could update line 193-219 of `pow1p.hpp` with the changes it would be great! Otherwise I'll suggest a patch off your PR after I finish the remaining points.

mborland:
> I made an update to the algorithm (in the referenced PR) that makes the evaluation more accurate for tiny x and huge y :
> 
> [scipy/scipy@a3f89c5](https://github.com/scipy/scipy/commit/a3f89c50a20a6f5c1506b8e2c0756a499190d667)
> 
> I think the method is now accurate to a few epsilon relative error for all x > − 1 and all y , and I expect minimal changes to this part in the future. A bit more update may come for the case x < − 1 . I'm still in the progress of adding more test cases and completing the algorithm description.
> 
> @mborland If you could update line 193-219 of `pow1p.hpp` with the changes it would be great! Otherwise I'll suggest a patch off your PR after I finish the remaining points.

Changes have been applied.

fancidev:
Thanks for checking this out, @JacobHass8.

> I'm not sure if the ULP distance being 1000 is acceptable. For this example the ULP distance is 667 (for types `long double` and `boost::math::concepts::real_concept`). Perhaps Taylor expanding to higher order would improve this?

The ULP should be in the single digit ideally. A likely cause of the loss of accuracy is that the compiler aggressively optimises the "fast_two_sum" construction such that it is no longer exact. In follow-up commits to the original scipy PR, I spent quite a bit of time to hack the [fast_two_sum](https://github.com/scipy/scipy/pull/21410/files) function with various preprocessor macros to try to avoid premature optimisation, but I can't say it works reliably for every compiler/system combination. The lack of a reliable fast_two_sum is a reason I failed to make progress on pushing the PR to production.

A second reason is that I did not get 100% confident about the error analysis, and even if I did, the analysis seems a bit overly complicated to verify. (I wouldn't like to repeat the error analysis for the function $(1+x)^y-1$, which I find useful in a few places too.) The reason for going through this complex route was to make use of the system's `pow`, `exp` and `log` functions. But there are reasons against it too: Firstly, the accuracy of system's `pow` function varies quite a bit across systems, making it hard to provide a guarantee of the accuracy of `pow1p`. Secondly, a lot of the (tedious) special case handling are actually duplicated between `pow1p` and `pow`. And lastly but importantly, it seems common implementations of `pow` already does the exp-log trick to double-double precision, so it would seem more efficient at runtime to build upon that route, which would also simply the error analysis a lot.

> Additionally, I let x be positive with x << 1 and y = 1 / x again. Mathematica is saying that ( 1 + x ) y ≈ 1 but the current implementation of `pow1p` is returning e . In this case we are not hitting the double-T range code.

If I'm not mistaken, $\lim_{x \rightarrow 0} (1+x)^{1/x} = \lim_{n \rightarrow \infty} (1+1/n)^n = e$.

ckormanyos:
>> I'm not sure if the ULP distance being 1000 is acceptable.

> The ULP should be in the single digit ideally. A likely cause of the loss of accuracy is that the compiler aggressively optimises the `fast_two_sum` construction such that it is no longer exact.

Writing any kind of portable `fast_two_sum` is challenging in C/C++. Basically, you need to guarantee a precise floating-point model so that sums and differences, even if minute, remain sums and differences. In some compiler jargon, this would mean something like you can't use `-ffast-math` on GCC/clang or you need `/fp:precise` on MSVC. But these constraints don't work in Math since we can't ask clients to do any of these settings.

You _might_ try making sure that the operands of sums and differences in `fast_two_sum` have the `volatile` quality. Then just empirically find out if it passes all of CI with the tolerances needed. CI picks up the compilers that are known-supported. And if any compiler is found later that fails, it could be treated as a special case.

NAThompson:
Pushed commits to finish this off:
- Merged develop; `fma` is now required (no fallback), and `real_concept` gets one.
- Fixed infinite `y` with |1+x| < 1, a NaN from `inf * 0` for tiny `x` and huge `y`, and compilation with expression-template types.
- Tests now check against a high-precision reference: at most 3 ULP for float and double, including tiny `x` with huge `y`.
- Docs page is now actually included, with ULPs plots; added a Google Benchmark file; the NVRTC float test really tests float.

Multiprecision types inherit an inaccuracy in `pow` for huge `y`: boostorg/multiprecision#774.

mborland:
The SYCL testing is very red after this merge:

```
      Start 64: run-boost_math-test_pow1p
64/64 Test #64: run-boost_math-test_pow1p ..............................***Failed    0.01 sec
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:76
  Condition '1 == inf' is not satisfied:
  Expected =  1.000000 = 1.000000e+00 = 0x1p+0
  Computed =  inf = inf = inf
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:77
  Condition '1 == 0' is not satisfied:
  Expected =  1.000000 = 1.000000e+00 = 0x1p+0
  Computed =  0.000000 = 0.000000e+00 = 0x0p+0
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:78
  Condition '1 == 0' is not satisfied:
  Expected =  1.000000 = 1.000000e+00 = 0x1p+0
  Computed =  0.000000 = 0.000000e+00 = 0x0p+0
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:79
  Condition '1 == inf' is not satisfied:
  Expected =  1.000000 = 1.000000e+00 = 0x1p+0
  Computed =  inf = inf = inf
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:106:
  ULP distance in float precision is 7.25e+08, which exceeds 10, error/ulps  = 7.25e+07.
  Expected: +114200739624191635123863552.000000000 = +0x1.79dbcap+86
  Computed: +1.000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:107:
  ULP distance in float precision is 7.27e+08, which exceeds 10, error/ulps  = 7.27e+07.
  Expected: +0.000000000 = +0x1.5ae192p-87
  Computed: +1.000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:108:
  ULP distance in float precision is 7.27e+08, which exceeds 10, error/ulps  = 7.27e+07.
  Expected: +0.000000000 = +0x1.5ae192p-87
  Computed: +1.000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:109:
  ULP distance in float precision is 7.25e+08, which exceeds 10, error/ulps  = 7.25e+07.
  Expected: +114200739624191635123863552.000000000 = +0x1.79dbcap+86
  Computed: +1.000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:76
  Condition '1 == inf' is not satisfied:
  Expected =  1.000000000 = 1.000000000e+00 = 0x1p+0
  Computed =  inf = inf = inf
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:77
  Condition '1 == 0' is not satisfied:
  Expected =  1.000000000 = 1.000000000e+00 = 0x1p+0
  Computed =  0.000000000 = 0.000000000e+00 = 0x0p+0
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:78
  Condition '1 == 0' is not satisfied:
  Expected =  1.000000000 = 1.000000000e+00 = 0x1p+0
  Computed =  0.000000000 = 0.000000000e+00 = 0x0p+0
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:79
  Condition '1 == inf' is not satisfied:
  Expected =  1.000000000 = 1.000000000e+00 = 0x1p+0
  Computed =  inf = inf = inf
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:116:
  ULP distance in double precision is 3.12e+18, which exceeds 10, error/ulps  = 3.12e+17.
  Expected: +28930191842539452754151977627739590089446424584682436941278548188818344930656420058019313018415117757204444868185051263212256659788282401812620042700792598066089704042564989062003086775586422168396903256948736.00000000000000000 = +0x1.6870ea75e682dp+692
  Computed: +1.00000000000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:117:
  ULP distance in double precision is 3.12e+18, which exceeds 10, error/ulps  = 3.12e+17.
  Expected: +0.00000000000000000 = +0x1.6ba4b26a9cd8ap-693
  Computed: +1.00000000000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:118:
  ULP distance in double precision is 3.12e+18, which exceeds 10, error/ulps  = 3.12e+17.
  Expected: +0.00000000000000000 = +0x1.6ba4b26a9cd8ap-693
  Computed: +1.00000000000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:119:
  ULP distance in double precision is 3.12e+18, which exceeds 10, error/ulps  = 3.12e+17.
  Expected: +28930191842539452754151977627739590089446424584682436941278548188818344930656420058019313018415117757204444868185051263212256659788282401812620042700792598066089704042564989062003086775586422168396903256948736.00000000000000000 = +0x1.6870ea75e682dp+692
  Computed: +1.00000000000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:125:
  ULP distance in double precision is 3.92e+18, which exceeds 10, error/ulps  = 3.92e+17.
  Expected: +7665372684472616600788028792826963467677834954930590899916837789646868735087351663154028364760772317743295208338795346008564169096650904119695512496206772766749859783776068140433858593568275822153579440719815935707161421640289520312804189814758661232334303395840.00000000000000000 = +0x1.f28c4bf6bb77ap+869
  Computed: +1.00000000000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:126:
  ULP distance in double precision is 3.9e+18, which exceeds 10, error/ulps  = 3.9e+17.
  Expected: +0.00000000000000000 = +0x1.e232c8f68116cp-867
  Computed: +1.00000000000000000 = +0x1p+0
  Mollified relative error: +1
Error at /home/runner/work/math/boost-root/libs/math/test/test_pow1p.cpp:test:127:
  ULP distance in double precision is 4.3e+18, which exceeds 10, error/ulps  = 4.3e+17.
  Expected: +151080175567658523717897244298888278647996963721296842461326150509180585931444210226113829561334982782673834527891449761063784711872664908714704948362772270354234305517499478994116894342686148859330332292890624032967081408749219065291702879464800979674654726767565225930022148580736761856.00000000000000000 = +0x1.fbff5c6b3b44p+953
  Computed: +1.00000000000000000 = +0x1p+0
  Mollified relative error: +1
Error count: 19, total ulp distance = 6143014438304712832
```

NAThompson:
@mborland : Fixing.

PR #1193, opened 2024-09-06
GPU Batch 10

mborland:
The focus of this PR was to round out support for the distribution now that many of the fundamental special functions have GPU support. Adds support to the following distributions: inverse gaussian, lognormal, non central beta, non central f, non central chi squared, negative binomial, normal, pareto, poisson, rayleigh, students t, triangular and uniform.

The three remaining unsupported distributions are the hyperexponential, hypergeometric, and non-central t. The first two need  related special functions support. Non-central T needs GPU support for quadrature built out, specifically exp-sinh. My initial thoughts are restricting quadrature to NVCC and NVRTC based on available APIs for memory allocations, but we will see.

On device CI runs can be found: https://github.com/cppalliance/cuda-math/pull/21.

CC: @dschmitz89, @izaid, @steppi.

### DIFF
--- a/doc/distributions/inverse_gaussian.qbk
+++ b/doc/distributions/inverse_gaussian.qbk
@@ -12,11 +12,11 @@
       typedef RealType value_type;
       typedef Policy   policy_type;
 
-      inverse_gaussian_distribution(RealType mean = 1, RealType scale = 1);
+      BOOST_MATH_GPU_ENABLED inverse_gaussian_distribution(RealType mean = 1, RealType scale = 1);
 
-      RealType mean()const; // mean default 1.
-      RealType scale()const; // Optional scale, default 1 (unscaled).
-      RealType shape()const; // Shape = scale/mean.
+      BOOST_MATH_GPU_ENABLED RealType mean()const; // mean default 1.
+      BOOST_MATH_GPU_ENABLED RealType scale()const; // Optional scale, default 1 (unscaled).
+      BOOST_MATH_GPU_ENABLED RealType shape()const; // Shape = scale/mean.
    };
    typedef inverse_gaussian_distribution<double> inverse_gaussian;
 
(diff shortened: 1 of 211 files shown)


jzmaddock:
The use of quadrature in the non central T is a new addition to improve accuracy in the tails: IMO it would not be unreasonable to disable this on the GPU.  Likewise there may be other functions which need slimming down for GPU usage, even though that may mean a lower quality of implementation.  The trick is deciding where to draw the line ;)

mborland:
> The use of quadrature in the non central T is a new addition to improve accuracy in the tails: IMO it would not be unreasonable to disable this on the GPU. Likewise there may be other functions which need slimming down for GPU usage, even though that may mean a lower quality of implementation. The trick is deciding where to draw the line ;)

That makes sense. Now that we have more substantial testing there are some clear outliers appearing in testing run time: https://github.com/cppalliance/cuda-math/actions/runs/10741618717/job/29792448169?pr=21#step:9:262. Once things are correct we can take optimization steps. @NAThompson is also interested in the prospect of the quadrature algorithms being used by others on device.

jzmaddock:
>  @NAThompson is also interested in the prospect of the quadrature algorithms being used by others on device.

Nod, that's actually a good use case given that the algorithms are in principle massively parallel.  In that case it probably makes more sense for the Integrand to be on the device, and the quadrature routine the glue that pulls the results together?

mborland:
> > @NAThompson is also interested in the prospect of the quadrature algorithms being used by others on device.
> 
> Nod, that's actually a good use case given that the algorithms are in principle massively parallel. In that case it probably makes more sense for the Integrand to be on the device, and the quadrature routine the glue that pulls the results together?

I think so. I have to work through how to make a shared memory region on the device that won't be leaked. Right now nick uses a shared_ptr to vectors with all the data. Neither of those facilities are available on device. CUDA has a version if malloc and free which make it easy enough to implement vector, but a shared pointer is more involved.

PR #1199, opened 2024-09-15
cmake: directly define project version

lucascolley:
See explanation and discussion at https://lists.boost.org/Archives/boost/2024/09/257793.php. cc @pdimov, x-ref https://github.com/scipy/scipy/pull/21270

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@
 
 cmake_minimum_required(VERSION 3.5...3.16)
 
-project(boost_math VERSION "${BOOST_SUPERPROJECT_VERSION}" LANGUAGES CXX)
+project(boost_math VERSION 1.87.0 LANGUAGES CXX)
 
 add_library(boost_math INTERFACE)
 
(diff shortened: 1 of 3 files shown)


mborland:
These changes look fine to me. The real question is will this be better for you? About 2 years ago when Scipy moved to using the Boost.Math standalone submodule (https://github.com/scipy/scipy/pull/17432) you were able to start tracking the head commit of math instead of having to wait for the tri-annual release. Won't this put you back in a position where you are waiting potentially months for fixes or features? Example. I am pretty sure your implementation of the Landau distribution relies on ours right now (through `_ufuncs`), but the Landau distribution has not made any official release for us yet and will not until 11 December at the earliest.

mborland:
> Indeed, there is a trade-off to be made there. We haven't made that decision yet, but these changes will enable us to go either way.

Sounds good. Since this adds a maintenance burden of like 30 seconds every 4 months I am not opposed.

lucascolley:
Thanks!

PR #1207, opened 2024-10-10
Make sure we don't activate float128 use when explicitly disabled

mborland:
@kursatyurt can you please give this patch a go and see if it solves your issue? If not can you please add a minimal reproducer and toolchain used to your issue. Thanks


kursatyurt:
@mborland Thanks, this solves the issue. I will write a reproducer if I can find time. 
I am using `nvcc 12.2`. It was in a more complex project with Kokkos; hopefully, I can write a minimal example.

jzmaddock:
Thanks guys, looks good to me!

PR #1208, opened 2024-10-10
More fixes for path of compile tests

mborland:


pdimov:
Wouldn't it be easier to use `INCLUDE_DIRECTORIES ../include_private` in `test/CMakeLists.txt`? That's what `test/Jamfile` does:

https://github.com/boostorg/math/blob/522989011aca84e375e71b3a571007c2fbb9bd6d/test/Jamfile.v2#L75

Also, `boost_test[_jamfile]` supports global variables such as `BOOST_TEST_LINK_LIBRARIES` (and `BOOST_TEST_INCLUDE_DIRECTORIES`) that can be used to factor out the common parts. These are the equivalent of putting things in Jamfile's project requirements.

PR #1212, opened 2024-10-15
Fix many, many obvious test bugs

rdoeffinger:
Needed for tests to pass with 128 bit long double types


mborland (approved):
These changes all look reasonable to me. I approved the CI run.

mborland:
Only CI failure is fail to clone on an s390x device (happens semi-regularly). Merging.

test/test_autodiff_2.cpp
```diff
@@ -536,7 +547,7 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(atan2_function, T, all_float_types) {
     auto y = y_sampler.next();
 
     auto autodiff_v = atan2(make_fvar<T, m>(x), make_fvar<T, m>(y));
-    auto anchor_v = atan2(x, y);
+    auto anchor_v = atan2_wrap(x, y);
```
jzmaddock, line 550: Just curious: what was the issue here?  Surely the `using std::atan2` should take care of things?

PR #1214, opened 2024-10-15
Update tests to take promote_double policy into account

rdoeffinger:
Makes it easier to change its default to false in the future. The remaining failing tests are test_nc_chi_squared test_bessel_j_prime and test_bessel_y_prime.
They probably have incorrect logic for the double == long double case already now.


rdoeffinger:
I am unable to see a connection between the test failures and this change...

PR #1216, opened 2024-10-23
fix: define BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS for emscripten

### LINKED ISSUE #1215: emscripten builds need BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
Using boost::math when compiling to wasm32 via emscripten is not possible out of the box right now because `BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS` is not defined.

It seems like this is a regression that was introduced by the following PR #753 which seems to be related to #728 and #729.

@jzmaddock to me it looks like uncommenting the `define` in `config.hpp`

```
#if (defined(__NetBSD__) || defined(__EMSCRIPTEN__)\
   || (defined(__hppa) && !defined(__OpenBSD__)) || (defined(__NO_LONG_DOUBLE_MATH) && (DBL_MANT_DIG != LDBL_MANT_DIG))) \
   && !defined(BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS)
//#  define BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
#endif`
```

would solve the problem and i'd happily prepare a PR if you see no negative consequences for `NetBSD` and `hppa` since they would also get affected by this change.

NickelWenzel:
Extracted check for `emscripten` and define `BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS` in case it wasn't previously defined.

This should fix #1215.

### DIFF
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -220,10 +220,14 @@
 
 #include <boost/math/tools/user.hpp>
 
-#if (defined(__NetBSD__) || defined(__EMSCRIPTEN__)\
+#if (defined(__NetBSD__)\
    || (defined(__hppa) && !defined(__OpenBSD__)) || (defined(__NO_LONG_DOUBLE_MATH) && (DBL_MANT_DIG != LDBL_MANT_DIG))) \
    && !defined(BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS)
 //#  define BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS
+#endif
+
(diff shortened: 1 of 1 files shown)


NickelWenzel:
@mborland could you maybe trigger the drone pipeline again, the clone step did not succeed for some configuration: https://drone.cpp.al/boostorg/math/2113/286/1

mborland:
> @mborland could you maybe trigger the drone pipeline again, the clone step did not succeed for some configuration: https://drone.cpp.al/boostorg/math/2113/286/1

That happens periodically with on that specific platform. I think we're safe to merge.

PR #1219, opened 2024-10-29
Merge to master for 1.87

mborland:


mborland:
@jzmaddock, @NAThompson, and @ckormanyos is there anything else that you'd like to see in this release? I'm merging a bit early since tomorrow is the deadline for major changes. Since our diff stat sitting at about 145K LOC because all of the CUDA-fication I would classify these changes as major.

PR #1222, opened 2024-11-28
Fix clang warning when compiling with -ffast-math: use of infinity is undefined behavior due to the currently enabled floating-point options

pps83:
Fixes the following warning:
```
C:\boost_1_86_0\boost/math/ccmath/isinf.hpp(29,69): warning : use of infinity is undefined behavior due to the currently enabled floating-point options [-Wnan-infinity-disabled]
C:\boost_1_86_0\boost/math/ccmath/isinf.hpp(29,25): warning : use of infinity is undefined behavior due to the currently enabled floating-point options [-Wnan-infinity-disabled]
C:\boost_1_86_0\boost/math/ccmath/ldexp.hpp(50,38): note: in instantiation of function template specialization 'boost::math::ccmath::isinf<float>' requested here
C:\boost_1_86_0\boost/math/ccmath/ldexp.hpp(69,33): note: in instantiation of function template specialization 'boost::math::ccmath::ldexp<float, true>' requested here
```

### DIFF
--- a/include/boost/math/ccmath/isinf.hpp
+++ b/include/boost/math/ccmath/isinf.hpp
@@ -25,6 +25,7 @@ constexpr bool isinf BOOST_MATH_PREVENT_MACRO_SUBSTITUTION(T x) noexcept
 #if defined(__clang_major__) && __clang_major__ >= 6
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wtautological-constant-compare"
+#pragma clang diagnostic ignored "-Wnan-infinity-disabled"
 #endif
             return x == std::numeric_limits<T>::infinity() || -x == std::numeric_limits<T>::infinity();
 #if defined(__clang_major__) && __clang_major__ >= 6


pps83:
the question is: should this be suppressed or not? with `-fast-math` I get tons of warnings from this code.

jzmaddock:
> the question is: should this be suppressed or not? with -fast-math I get tons of warnings from this code.

As long as -ffast-math doesn't break the code, then yes it should be suppressed I would say.

pps83:
@mborland can you please check the PR?

mborland (approved):

ckormanyos:
Hi folks, this thing just broke my CI in an unrelated project, in which I have `-Werror` activated. I think or maybe it seems like It seems that some `__clang_major` versions are caught by the test on greater than `6`.

See also this [broken CI](https://github.com/ckormanyos/wide-decimal/actions/runs/12365013148/job/34509258897#step:5:33).

I haven't quite figured out the exact way to weed through all the compilers yet, but I think a higher version check _might_ be needed. But I also got red on version `15`.

Cc: @jzmaddock and @mborland

PR #1223, opened 2024-12-06
Suppress unused parameter warnings

Flamefire:


mborland (approved):
LGTM. There's no reason this should impact linear regression on windows. I'll check that out post-merge.

PR #1225, opened 2024-12-17
Change minimum clang version for -Wnan-infinity-disabled

mborland:
Fixes breakage introduced in #1222 

CC: @ckormanyos


ckormanyos:
Hi Matt thank you for looking into this. It appears as though an unrelated slight change in runners is freaking out on our old friend Bernoulli.

Cc: @mborland

ckormanyos:
Hi Matt (@mborland) the reason for failing here was addressed in #1226, as discussed there.

So that is fixed now and in develop.

Cc: @jzmaddock

PR #1226, opened 2024-12-19
First batch changes double-fp-backend

ckormanyos:
I need to make some tiny changes in the Math tests when testing the candidate `double_fp_backend<double>` as a multiprecision backend.

The min./max. exponents are really small (comparatively) for a multiple-precision type. I'm running the tests for _double-double_, which has `min_exponent10` of $-291$ and `max_exponent10` of $308$. So this causes a few over/underflows unexpectedly for a multiprecision type under test.

A second batch of similarly small, but more difficult-to-make changes will be needed. But I will do these in a second step, pending approval of the first, simpler batch.

Cc: @jzmaddock and @mborland


ckormanyos:
See also [PR154 in GSoC repo](https://github.com/BoostGSoC21/multiprecision/pull/154)

Cc: @sinandredemption

ckormanyos:
One example of a change needed comes from the test point of the irregular Bessel function $K_{\nu}$.

We need to compute

$$
K_{-1000}\left(700\right)\text{,}
$$

which internally needs to compute the value of

$$
e^{-700}{\approx}10^{-305}
$$

This number, however, underflows `cpp_double_fp_backend<double>` for 64-bit double, having `min_exponent10` of $-291$.

ckormanyos:
Hi John (@jzmaddock) and Matt (@mborland) there seem to be some tolerance issues in files like `linear_regression_test.cpp` on the clang-port to MSVC 2022.

I'm not qualified to "up" the relevant tolerances. Does anyone know if these failures are relevant? Is it just a tolerance issue? Or does something actually need to be done?

I'm not entirely sure, but I do not think my changes in areas like Bessel functions are related. But you never know in Math. Anyway, could you guys take a look? Thanks.

mborland:
Github updated the version of MSVC2022 and it caused three tests to fail consistently. I can nudge up the tolerance a bit because it's been consistent across runs instead of sporadic like normal tolerance problems.

ckormanyos:
> I can nudge up the tolerance a bit

Should I just do that here while I'm on such a roll? If that's what's to be done, I can handle it. Thx Matt (@mborland)

mborland:
> > I can nudge up the tolerance a bit
> 
> Should I just do that here while I'm on such a roll? If that's what's to be done, I can handle it. Thx Matt (@mborland)

Please do

cosurgi:
Wow, @ckormanyos thank you very much for working on this. I am quite busy at the moment, so I only had a brief look, but please keep working. I don't remember much, but if it will be needed I will try to remember some of my code, when you will need it.

reporting/performance/test_kn.cpp
```diff
@@ -48,7 +48,9 @@ static const std::array<std::array<T, 3>, 9> kn_data = { {
    { { SC_(-10.0), SC_(1.0), SC_(1.80713289901029454691597861302340015908245782948536080022119e8) } },
    { { SC_(100.0), SC_(5.0), SC_(7.03986019306167654653386616796116726248616158936088056952477e115) } },
    { { SC_(100.0), SC_(80.0), SC_(8.39287107246490782848985384895907681748152272748337807033319e-12) } },
-   { { SC_(-1000.0), SC_(700.0), SC_(6.51561979144735818903553852606383312984409361984128221539405e-31) } },
```
jzmaddock, line 51: Given that the answer is within a sensible range this should work and give the correct answer for all T.  Do you know where in the code this is failing?

test/test_bessel_j.hpp
```diff
@@ -281,7 +281,10 @@ void test_bessel(T, const char* name)
     BOOST_CHECK_EQUAL(boost::math::cyl_bessel_j(T(0), T(2.5)), boost::math::cyl_bessel_j(T(0), T(-2.5)));
     BOOST_CHECK_EQUAL(boost::math::cyl_bessel_j(T(1), T(2.5)), -boost::math::cyl_bessel_j(T(1), T(-2.5)));
     #ifndef SYCL_LANGUAGE_VERSION
-    BOOST_CHECK_CLOSE_FRACTION(boost::math::cyl_bessel_j(364, T(38.5)), SC_(1.793940496519190500748409872348034004417458734118663909894e-309), tolerance);
+    BOOST_IF_CONSTEXPR (std::numeric_limits<T>::min_exponent10 <= -307)
```
jzmaddock, line 284: Denormals not withstanding, this should work for all T as well - it should underflow to zero when the exponent does not have enough range?

test/test_bessel_k.hpp
```diff
@@ -133,7 +133,9 @@ void test_bessel(T, const char* name)
         {{ SC_(-10.0), SC_(1.0), SC_(1.80713289901029454691597861302340015908245782948536080022119e8) }},
         {{ SC_(100.0), SC_(5.0), SC_(7.03986019306167654653386616796116726248616158936088056952477e115) }},
         {{ SC_(100.0), SC_(80.0), SC_(8.39287107246490782848985384895907681748152272748337807033319e-12) }},
-        {{ SC_(-1000.0), SC_(700.0), SC_(6.51561979144735818903553852606383312984409361984128221539405e-31) }},
+        (std::numeric_limits<T>::is_specialized && (std::numeric_limits<T>::min_exponent10 > -300))
```
jzmaddock, line 136: As with the first comment, given that the result is in range, this should work for all T, do you know where things go wrong?

reporting/performance/test_kn.cpp
ckormanyos: > where in the code this is failing?

Yes John (@jzmaddock) that was a tough one. Internally the calculation must compute $e^{-700}$ which underflows the type.

test/test_bessel_k.hpp
ckormanyos: Same as above.

ckormanyos:
> thank you very much for working on this

Thanks @cosurgi we are actually getting close. Conservatively, I think we can get the double-fp backend into summer 1.89. But I might even hope for April's 1.88. Let's see if we can get all the edge cases to _behave_. It's not that far off the mark now. I'd like to first get double-double stable and _in_, then in a second push charge on for quad.

ckormanyos:
>>> I can nudge up the tolerance a bit

>> Should I just do that here while I'm on such a roll? If that's what's to be done, I can handle it. Thx Matt (@mborland)

> Please do

Cycling now

reporting/performance/test_kn.cpp
jzmaddock: Sure, but it works for type double which also can't compute e^-700.  Ah wait, are these calling the 113-bit precision rational approximations?

ckormanyos: > Ah wait, are these calling the 113-bit precision rational approximations

Yes John. This is a conversation I was kind of waiting to have. My exponent range is EVEN less than that of `double`. But I end up in the multiprecision section of the special functions.

This worked surprisingly well and there seem to be very few errors. There are, however, a few we will need to hash out.

ckormanyos: This somewhat novel type has $106$ digits, yet is limited to `min_exponent10` of $-291$ and `max_exponent10` of $+308$

ckormanyos:
OK guys, CI is green. Yet the first batch of changes is not quite ready.

For the type `cpp_double_double`, we still have to negotiate on:

- the test of `cyl_bessel_j(364, 35.5)`
- and the test of `cyl_bessel_k(-1000, 700)`

 I consider the second point to be done. The sub-normal case is a bit more tricky and i need to check if my zero-handling os correct and what's up on that test case.

Cc: @jzmaddock 

Also @mborland the adjusted tolerances on the cases we mentioned are (perhaps liberal) but green.

jzmaddock:
OK, I think I see the issue, min and max values are asymetric, so in bessel_k0.hpp:473 the line:

```
if(x < tools::log_max_value<T>())
```

Should actually be:

```
if(-x > tools::log_min_value<T>())
```

There will be other places as well... you caught me be lazy!!  But that (along with any other similar mistakes) should fix things.

ckormanyos:
> OK, I think I see the issue, min and max values are asymetric,

OK amazing catch John, let me run that thing through CI and see how it shakes out.

> There will be other places as well... 

It's really rather easy-going. I have a lits with one single point of `ellint_3`, a whole bunch of `inv_gamma()` with GIANT exponents (I can deal with these), a few things more on $K_{\nu}$. But that's pretty much it. So all in all, the generic-ness of Math/Multiprecision is simply awesome.

Let me see if we can dial in this double-trouble type ASAP.

ckormanyos:
Ah OK. John with the change in `bessel_k0.hpp` it is very close. I do, however, still need to

```
#define BOOST_NO_LIMITS_COMPILE_TIME_CONSTANTS
```

otherwise I get the vanilla hard-coded values for logarithmc of max/min values. And I'm fine, perfectly fine defining this (at least preliminarily, in the `cpp_double_fp_backend` main header file until we hash out the rest of the wrinkles.

Let me back out the changes in $K_{\nu}$ tests.

Then the only open point is:

$$
J_{364)\left(35\frac{1}{2}\right)
$$

ckormanyos:
> the only open point is

I mean in this PR. There will be batch2 of specfun changes needed to negotiate coming when we finish batch1.

ckormanyos:
OK I needed to correct two wrongly-placed `constexpr` attributes.

I need to move forward with this thing. So I just referted the test-case-change on `cyl_bessel_j`. So the negotiation is in my view closed. If/when this thing cycles, I'd like to consider to merge it. Then i can plow on with `cpp_double_fp_backend`...

jzmaddock:
> OK amazing catch John, let me run that thing through CI and see how it shakes out.

What I meant to say was - "I tend to make the same mistake over and over" so since all the Bessel rational approximations are pretty similar, there are probably other places where we have this bug whether or not we happen to have (un)fortunate test values in the (un)sweet spot that triggers the issue.

> Ah OK. John with the change in bessel_k0.hpp it is very close. I do, however, still need to
>#define BOOST_NO_LIMITS_COMPILE_TIME_CONSTANTS

That should definitely NOT be needed unless there's something wrong with your numeric_limits specialization?

>Then the only open point is:
> J 364 ( 35 1 2 ) 

The result in that case is a denorm at double precision, is that the issue?

ckormanyos:
> What I meant to say was - "I tend to make the same mistake over and over"

Me too.

>> Ah OK. John with the change in bessel_k0.hpp it is very close. I do, however, still need to
>> '#define BOOST_NO_LIMITS_COMPILE_TIME_CONSTANTS'

> That should definitely NOT be needed unless there's something wrong with your numeric_limits specialization?

Maybe. Or with proper handling of underflow(s) and/or zeros. I'll take a bit more time to figure out what I really need to do. But that is my assignment.

In this PR, there is basically now just one change of an exponent-check constant, two removals of the word `constexpr`, and the tolerance changes on the new runners. So I'm likely to push it, since there are no controversial changes there. But I believe the backend I'm testing might not quite be right (yet) in some of these edge cases. Let me figure out how to move forward. Then when I have some more well-formulated info, I'll pick up some of the harder points in a newer PR...

ckormanyos:
OK, so s390 is hanging on drone. Everything else is green. This thing will fix one single failure in the onslaught of `cpp_double_fp_backend` tests. There will be more work needed.

There is nothing more controversial in this PR. I can play around with it since the erroneous `constexpr` attributes are corrected.

Also the tolerances are corrected on some of those distributions mentioned above.

So I'll push this thing.

Then I need to study up on the backend details and find out exactly what is causing some of the final edge case failures. So this will be ocntinued in future PRs.

Cc: @jzmaddock and @mborland

PR #1233, opened 2025-01-12
Fix constexpr cmath floor of 1 returning 0 #1232

### LINKED ISSUE #1232: Constexpr cmath floor of 1 is returning 0
`boost::math::ccmath::floor(1.0)` is returning `0.0` instead of `1.0`.
Looks like the issue is [here](https://github.com/boostorg/math/blob/c5d36664fd9a15293ae3b319f71e51138c926b7c/include/boost/math/ccmath/floor.hpp#L36), when it compares the `arg` with `result` (which is `1`) and returns `0` if the `arg >= result`.

yousseflakhal:
### Problem
The `boost::math::ccmath::floor` function incorrectly returns `0.0` for `1.0` in a `constexpr` context.

### Solution
- Fixed the implementation to correctly handle `1.0`.
- Added test cases in `ccmath_floor_test.cpp`.

### Linked Issue
Fixes #1232

### DIFF
--- a/include/boost/math/ccmath/floor.hpp
+++ b/include/boost/math/ccmath/floor.hpp
@@ -33,7 +33,7 @@ inline constexpr T floor_pos_impl(T arg) noexcept
 
     T result = 1;
 
-    if(result < arg)
+    if(result <= arg)
     {
         while(result < arg)
         {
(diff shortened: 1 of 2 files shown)


ckormanyos:
Thanks for pressing on with this.

I just approved this workflow run. We must approve workflow runs for first-time contributions @yousseflakhal.

Let's let this thing run through CI now.

Cc: @mborland and @jzmaddock

mborland (approved):
LGTM Thanks!

PR #1234, opened 2025-01-17
Simplify the formula for the CDF of the Cauchy distribution.

WarrenWeckesser:
The Cauchy CDF function may be expressed as atan2(1, -x)/pi.

### DIFF
--- a/include/boost/math/distributions/cauchy.hpp
+++ b/include/boost/math/distributions/cauchy.hpp
@@ -45,23 +45,11 @@ BOOST_MATH_GPU_ENABLED RealType cdf_imp(const cauchy_distribution<RealType, Poli
    //
    // This calculates the cdf of the Cauchy distribution and/or its complement.
    //
-   // The usual formula for the Cauchy cdf is:
+   // This implementation uses the formula
    //
-   // cdf = 0.5 + atan(x)/pi
+   //     cdf = atan2(1, -x)/pi
    //
-   // But that suffers from cancellation error as x -> -INF.
-   //
-   // Recall that for x < 0:
-   //
-   // atan(x) = -pi/2 - atan(1/x)
-   //
-   // Substituting into the above we get:
-   //
-   // CDF = -atan(1/x)/pi  ; x < 0
-   //
-   // So the procedure is to calculate the cdf for -fabs(x)
-   // using the above formula, and then subtract from 1 when required
-   // to get the result.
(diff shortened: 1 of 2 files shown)


mborland (approved):
This looks good to me. Thanks!. The only CI failure is an S390x runner failing to clone (which the CI manager is aware this has been occurring more frequently).

WarrenWeckesser:
I should have also updated [`cauchy.qbk`](https://github.com/boostorg/math/blob/develop/doc/distributions/cauchy.qbk).  I'll create a new PR to update the CDF "Implementation Notes" section.

Is there a guide or recipe somewhere for how to run quickbook to build the docs?

mborland:
> I should have also updated [`cauchy.qbk`](https://github.com/boostorg/math/blob/develop/doc/distributions/cauchy.qbk). I'll create a new PR to update the CDF "Implementation Notes" section.
> 
> Is there a guide or recipe somewhere for how to run quickbook to build the docs?

There's a synopsis here: https://github.com/boostorg/math?tab=readme-ov-file#building-documentation. In practice you need some pretty old libraries and binaries to make it work so the CI will tell you if something is formatted incorrectly.

jzmaddock:
>There's a synopsis here: https://github.com/boostorg/math?tab=readme-ov-file#building-documentation. In practice you need some pretty old libraries and binaries to make it work so the CI will tell you if something is formatted incorrectly.

The markup is reasonably human readable - so what @mborland says, just commit, and we can always fix it up.

BTW you shouldn't actually need old binaries though - the only thing that is version-fixed is the document DTD - though I confess I haven't updated anything in a good while ;)

WarrenWeckesser:
> There's a synopsis here: ...

Argh, right there in the README file!  Not sure how I missed that.

On my Linux machine, I had to install xsltproc (I guess I had all the other dependencies already installed), and I changed `darwin` to `gcc` in `user-config.jam`.

The HTML built successfully, and the change that I made looks OK.  For a change like this, do I also commit all the updated HTML files as part of the PR, or should I only submit the changed to `cauchy.qbk`?

jzmaddock:
Just the .qbk file for the minute, we'll regenerate the docs prior to release (and the CI builds it too).  Thanks!

WarrenWeckesser:
If the only change is in `cauchy.qbk`, should I used `[skip ci]` in the commit message?

jzmaddock:
No might as well let CI build as it checks the docs for validity.

PR #1235, opened 2025-01-21
Update library version on Master

mborland:
CC: @pdimov


jzmaddock:
Is there any way to avoid the need to update this with each release?

mborland:
> Is there any way to avoid the need to update this with each release?

Not to my knowledge. Scipy needed it for their new integration, and it didn't seem like a huge cost item especially when I forget and Peter tells me to fix it.

PR #1236, opened 2025-01-21
Update gamma.hpp

kenarab:
Missing name for Lanczos parameter in gamma_imp_final at line 114

It caused an error when compiling at line 124

     std::cout << "tgamma_imp called with " << typeid(z).name() << " " << typeid(l).name() << std::endl;
 
 
Checked on old code and the parameter has l name

### DIFF
--- a/include/boost/math/special_functions/gamma.hpp
+++ b/include/boost/math/special_functions/gamma.hpp
@@ -111,7 +111,7 @@ BOOST_MATH_GPU_ENABLED T sinpx(T z)
 // tgamma(z), with Lanczos support:
 //
 template <class T, class Policy, class Lanczos>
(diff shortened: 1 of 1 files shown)


mborland (changes requested):
Can you add a `(void)l` somewhere? This will cause unused parameter warnings except when printing is enabled.

kenarab:
Added   the requested code inside the conditional compilation before debugging code

(void)l;

mborland:
> Added the requested code inside the conditional compilation before debugging code
> 
> (void)l;

I added a commit moving it outside of the #if block since thats where it'll cause issues. Sorry if I was not sufficiently clear.

mborland (approved):

kenarab:
I'm sorry I'm not familiar with that practice so placed it where as little disruption could cause.

kenarab:
Next time I see this kind of issue I will do it better

Issue #1240, opened 2025-01-29
Extend hypergeometric distribution PMF for non-integral arguments


mborland:
Discussion in Scipy review can be found starting here: https://github.com/scipy/scipy/pull/22312#discussion_r1912619742

CC: @mdhaber, @fpelliccioni, @steppi

PR #1244, opened 2025-02-04
feat: Extend hypergeometric distribution PMF for non-integral arguments

fpelliccioni:
# Summary

This PR extends the hypergeometric distribution's PMF to support non-integer values of `x` using cubic Hermite interpolation. If `x` is not an integer, the implementation selects at least three valid integer points for interpolation. If fewer than three points are available, it raises a domain error.

# Known Issue

This change introduces a dependency on `cubic_hermite`, which throws exceptions instead of using Boost.Math’s `policies::raise_domain_error`. As a result, some compile-time tests fail due to exceptions being disabled in those tests.

# Next Steps

- A separate PR will update `cubic_hermite` (and potentially other interpolators) to conform to Boost.Math’s policy-based error handling.
- This PR is left in draft so reviewers can provide feedback on the approach.

# Question: Best way to test this?

Hypergeometric test data in `test/hypergeometric_test_data.ipp` appears to be auto-generated from Mathematica, with some values removed due to absolute vs relative error concerns.
Since we now support non-integer values, what is the best approach to generate a new test dataset? Does anyone know of a reliable source for non-integer hypergeometric PMF values?

Fixes https://github.com/boostorg/math/issues/1240

### DIFF
--- a/include/boost/math/distributions/hypergeometric.hpp
+++ b/include/boost/math/distributions/hypergeometric.hpp
@@ -15,7 +15,9 @@
 #include <boost/math/distributions/detail/hypergeometric_pdf.hpp>
 #include <boost/math/distributions/detail/hypergeometric_cdf.hpp>
 #include <boost/math/distributions/detail/hypergeometric_quantile.hpp>
+#include <boost/math/interpolators/cubic_hermite.hpp>
 #include <boost/math/special_functions/fpclassify.hpp>
+#include <boost/math/special_functions/round.hpp>
 #include <cstdint>
 
 namespace boost { namespace math {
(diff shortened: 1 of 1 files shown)


mdhaber:
Thanks for this @fpelliccioni!

@steppi would interpolation be ok, or were you hoping for the definition to match that of evaluation in terms of `binom`?

@fpelliccioni it might solve some of the testing and policy issues to implement this using binomial coefficients for small arguments and (the existing?) Lanczos approximation for large arguments; e.g. see https://github.com/scipy/scipy/pull/22312#discussion_r1932918404. Then reference data could be computed in Mathematica using `Binomial`.

steppi:
> @steppi would interpolation be ok, or were you hoping for the definition to match that of evaluation in terms of `binom`?

Yeah, I was hoping this would evaluate

$$\frac{{r \choose k}{N -r \choose k}}{{N \choose n}$$

for non-integral arguments. Interpolating with Hermite polynomials would actually work for our application in stats though, but is kind of arbitrary, and thus like you and @fpelliccioni said, there would be no natural reference implementation to compare to. I'm pretty sure the Lanczos code path in Boost's hypergeometric pmf implementation will just work out of the box for non-integral arguments.

jzmaddock:
Before we get too carried away with interpolation, why not just use the binomial definition (and our existing code) over the real domain?  That would seem to be more accurate, and probably rather more efficient as well.

A more pressing (and perhaps difficult) question is what we **want** to have happen?  Is there any precedence for evaluating the hypergeometric over a non-discrete domain?  Or to put this another way: is anyone relying on the hypergeometric PMF generating an error when a non-integer is passed?

mdhaber:
@jzmaddock This request ultimately stems from https://github.com/scipy/scipy/pull/22312#discussion_r1912619742. We asked to be able to evaluate the function at non-integer arguments (like some other Boost discrete disributions) so that we can approximate infinite sums involving the PMF in a monotonically decreasing summand with an integral. Any shape-preserving interpolation would work (so PCHIP might be preferred to avoid ringing), but the original suggestion (repeated above https://github.com/boostorg/math/pull/1244#issuecomment-2658494655) was to use the binomial definition and the existing code where possible.

JacobHass8:
> Before we get too carried away with interpolation, why not just use the binomial definition (and our existing code) over the real domain? That would seem to be more accurate, and probably rather more efficient as well.

I've implemented the Lanczos code path and Gamma function code path as mentioned in https://github.com/boostorg/math/pull/1244#issuecomment-2659510817. However, the accuracy of the results aren't impressive for doubles. I've tried $N=170$ and $N=500$, and was only able to get about 10 significant digits to match Mathematica. 

I can make a new PR with this implementation, but it may require an overhaul depending on how precise the results are expected to be.

PR #1245, opened 2025-02-12
feat: use policy-based error handling in interpolators

fpelliccioni:
This PR updates `cubic_hermite`, `makima`, and `pchip` interpolators to use `boost::math::policies::raise_domain_error` instead of throwing exceptions. 
This aligns their error handling with Boost.Math’s policy-based approach.

# Related PR
#1244 depends on this PR.

# Summary
Replaces exceptions with policy-based error handling.
Ensures compatibility with compile-time tests where exceptions are disabled.

# Next Steps
Once merged, #1244 will be unblocked and ready for review.


mborland:
I'm not a fan of changing the API to add policy support unless you have a compelling reason. I think this could simply have replaced the current `throw std::whatever` to using the boost policy for signaling errors. 

I feel like this is un-ergonomic:
```
   auto ret = boost::math::interpolators::pchip<std::vector<double>>::create(std::move(data_x), std::move(data_y), 3, 2);
   check_result<double>(ret.first->operator()(1.0));
```

compared to:
```
boost::math::interpolators::pchip<std::vector<double>> s(std::move(data_x), std::move(data_y), 3, 2);
   check_result<double>(s(1.0));
```

CC: @NAThompson

include/boost/math/interpolators/detail/cubic_hermite_detail.hpp
```diff
@@ -31,15 +33,21 @@ class cubic_hermite_detail {
         using std::isnan;
         if (x_.size() != y_.size())
         {
-            throw std::domain_error("There must be the same number of ordinates as abscissas.");
+            error_msg_ = "There must be the same number of ordinates as abscissas.";
+            valid_ = false;
+            return;
```
mborland, line 38: So that we don't change the existing behavior what about something like:

```
boost::math::policies::raise_domain_error(...)
error_msg_ = ...;
valid_ = false;
return;
```

The raise domain error will throw when exceptions are enabled, and when exceptions are disabled we use track the validity of the state. This is the implementation of raise_domain_error: https://github.com/boostorg/math/blob/develop/include/boost/math/policies/error_handling.hpp#L208-L222. Like the comment says you shouldn't be able to hit the actual return value, but the continued implementations below will write to errno, NOP, etc. depending on the policy.

test/test_hypergeometric_dist.cpp
```diff
@@ -493,6 +493,19 @@ void test_spots(RealType /*T*/, const char* type_name)
...
+// BOOST_AUTO_TEST_CASE(test_non_integer_k)
+// {
+//    boost::math::hypergeometric_distribution<double> dist(30, 10, 15);  // N=30, r=10, n=15
+
+//    BOOST_CHECK_CLOSE(
+//       pdf(dist, 4.5),
+//       pdf(dist, 4.0) * 0.5 + pdf(dist, 5.0) * 0.5,
+//       0.1);  // Interpolación lineal
+
+//    // Valores fuera del soporte deben dar 0
```
mborland, line 505: If you leave this comments in can you please translate them to English.

mborland (commented):
I like this a lot more than the first iteration.

include/boost/math/interpolators/detail/cubic_hermite_detail.hpp
jzmaddock: +1 on that.

test/test_hypergeometric_dist.cpp
fpelliccioni: Sorry, that was commited by mistake, reverting...

mborland (approved):
LGTM now. I kicked off the CI run to make sure it agrees.

NAThompson:
@fpelliccioni : Apologies for not merging this after we approved it . . . . looks like this fell through the cracks.

PR #1254, opened 2025-04-08
Replace deprecated runners in CI

mborland:


mborland:
I'm not quite sure why we have a bloodbath from Boost.Exception now. The file that's popping up as an error hasn't been modified in over 2 years.

jzmaddock:
We must have a test that has <exception-handling>off and is - possibly indirectly - pulling in Boost.Exception.  I don't see it at present though.

mborland:
I think it's coming up as a secondary dependency from Boost.Container since the failure looks like:

```
gcc.compile.c++ ../../../bin.v2/libs/container/build/gcc-14/debug/x86_64/exception-handling-off/link-static/threading-multi/visibility-hidden/pool_resource.o
gcc.compile.c++ ../../../bin.v2/libs/system/build/gcc-14/debug/x86_64/exception-handling-off/link-static/threading-multi/visibility-hidden/error_code.o
gcc.archive ../../../bin.v2/libs/container/build/gcc-14/debug/x86_64/exception-handling-off/link-static/threading-multi/visibility-hidden/libboost_container.a
gcc.archive ../../../bin.v2/libs/system/build/gcc-14/debug/x86_64/exception-handling-off/link-static/threading-multi/visibility-hidden/libboost_system.a
gcc.compile.c++ ../../../bin.v2/libs/exception/build/gcc-14/debug/x86_64/exception-handling-off/link-static/threading-multi/visibility-hidden/clone_current_exception_non_intrusive.o
In file included from ../../../libs/exception/src/clone_current_exception_non_intrusive.cpp:14:
../../../boost/exception/detail/clone_current_exception.hpp:22:6: error: #error This header requires exception handling to be enabled.
   22 | #    error This header requires exception handling to be enabled.
      |      ^~~~~

    "g++-14"   -fvisibility-inlines-hidden -fno-exceptions -std=c++20 -m64 -pthread -O0 -fno-inline -Wall -g -fvisibility=hidden  -DBOOST_ALL_NO_LIB=1 -DBOOST_MATH_RUN_MP_TESTS -DCI_SUPPRESS_KNOWN_ISSUES -DSLOW_COMPILER   -I"../../.."  -c -o "../../../bin.v2/libs/exception/build/gcc-14/debug/x86_64/exception-handling-off/link-static/threading-multi/visibility-hidden/clone_current_exception_non_intrusive.o" "../../../libs/exception/src/clone_current_exception_non_intrusive.cpp"
```

`test_instantiate` has some very old dependencies in it so maybe breaking those would help?

jzmaddock:
test_instantiate is a compile-only test, it shouldn't be building *any* dependencies IMO.

mborland:
> test_instantiate is a compile-only test, it shouldn't be building _any_ dependencies IMO.

Should this be changed to compile then?

```
[ run test_instantiate1.cpp test_instantiate2.cpp  : : : <exception-handling>off : test_instantiate_no_eh ]
```

In the no-exception handling runs everything fails so that only kind of helps?

jzmaddock:
You're correct: the build log clearly shows that test_instantiate is the culprit, BUT it passes locally for me, and the Jamfile clearly shows no dependencies at all for that target, and certainly no mention of Boost.Exception being required.  For that matter Boost.Exception (the library) is absolutely NOT a dependency, only the stub header.  So I'm confused right now.

jzmaddock:
Reproduced locally.

jzmaddock:
The dependency is introduced by Boost.Test which isn't used by that target, I nearly have a fix that I'm testing locally...

Issue #1255, opened 2025-04-10
Bessel yn isn't applying reflection factor correctly in all cases (small z, negative n)


gajjanag:
Hi.

Noticed internally at Meta while trying to integrate Boost yn into some of our code.

Can be reproduced as:
Yn (negative n arguments) Take n == -1
x= 0x1.03ebbp-128 , boost returns val = -0x1.41085ep+127

While Mathematica gives an extremely large positive value here (roughly 0x1.41085e5d24b94p+127), showing that Boost gets the sign incorrect here, though the magnitude seems right.

I believe the issue arises from https://github.com/boostorg/math/blob/529f3a759d83aa9437613666ea6293c9336d4069/include/boost/math/special_functions/detail/bessel_yn.hpp#L56, essentially the "small z" path does not incorporate the reflection "factor" before returning.

jzmaddock:
Thanks for the report: I'm trying to work through some Math issues now, so I'll add this one to the list ;)

ckormanyos:
>> trying to integrate Boost yn

>> While Mathematica gives an extremely large positive value here (roughly 0x1.41085e5d24b94p+127), showing that Boost gets the sign incorrect here, though the magnitude seems right.

> work through some Math issues now, so I'll add this one to the list

Thanks for the report.

If it is just reflection, I might be capaable of handling and fixing it and adding test(s). Let me take a look.

ckormanyos:
The issue has been reproduced locally. I'm providing test code.

```cpp
#include <boost/math/special_functions/bessel.hpp>

#include <iomanip>
#include <iostream>
#include <sstream>

auto main() -> int
{
  using float_type = float;

  const float_type x { 0x1.03ebbp-128F };

  const float_type ctrl { 0x1.41085ep+127F };

  const float_type result = ::boost::math::cyl_neumann(-1, x);

  std::stringstream strm { };

  strm << std::setprecision(std::numeric_limits<float_type>::digits10) << "result: " << result << "\nctrl  : " << ctrl;

  std::cout << strm.str() << std::endl;
}
```

ckormanyos:
Error reproduced at [https://godbolt.org/z/nMjrhMdvs](https://godbolt.org/z/nMjrhMdvs).

ckormanyos:
OK #1256 has cycled. The new tests pass and the lines are covered in CI on CodeCov. See image below.

[...]

PR #1256, opened 2025-04-11
Handle issue 1255 and update docs

ckormanyos:


ckormanyos:
See also #1255

ckormanyos:
Something weird is happening in CI, in this PR and also in the CI that ran when @mborland simply incremented the library version [here](https://github.com/boostorg/math/actions/runs/14402716458).

It seems like something is influencing files  that were not touched in the changes.

Cc: @jzmaddock

ckormanyos:
For some reason, when exceptions are off at compilation time, we are hitting [this line](https://github.com/boostorg/exception/blob/9238ef8e7f97e619ed3e58bc7becf16b4966ccf0/include/boost/exception/detail/clone_current_exception.hpp#L22). Then certain tests in the Math test suite fail when Boost has no exceptions yet this file is still  included.

That line is, however, two years old. I wonder what tiny change and where causes this phenomenon today?

ckormanyos:
Hmmm I don't know so much about the dependencies in the Jamfile. But why is Math's test suite trying to build the libboost_exception.a archive [here](https://github.com/boostorg/math/actions/runs/14402716458/job/40391903290#step:17:2940)?

This is in the recent version increment from @mborland

jzmaddock:
There is some cruft in the Jamfile, but there are other issues too.  I'm working through them in https://github.com/boostorg/math/pull/1254, it's getting closer to green, but there are no cigars being lit just yet!

ckormanyos:
> some cruft in the Jamfile, but there are other issues too. I'm working through them in https://github.com/boostorg/math/pull/1254, it's getting closer to green,

Thanks John. OK cool. I'll hold my horses a bit and wait till that goes green and gets in develop.

In the meantime, I'll try to ensure that the new tests for #1255 are all passing.

ckormanyos:
It it is good enough for #1254, it's good enough for #1255, with the repeated comment, OK looks good to go, only failures are from cycl Cauchy tests.

We'll let @mborland look at those.

PR #1257, opened 2025-04-13
Collected fixes for incomplete beta in the asymptotic large a/b region

### LINKED ISSUE #1198: Noncentral F distribution CDF edge cases
SciPy is thanks to Boost finally able to remove parts of the ancient cdflib Fortran library. Recently, the noncentral F distribution CDF was  fully replaced by the boost implementation in https://github.com/scipy/scipy/pull/21505, resulting in a far more accurate results for a wide range of inputs. Still, edge cases were uncovered which we would like to share with you. They occur for very large and very tiny values for the degrees of freedom. Note that for the second case even Mathematica is not able to compute a result, so maybe we do not need to thrive for a feasible value but raise an error or return nan. @mdhaber found a limiting formula that might be helpful: https://github.com/scipy/scipy/pull/21505#issuecomment-2334793846

| v1| v2| l | x | Boost result |
|--------|--------|--------|--------|--------|
| 1e-100| 3 | 1.5 | 1e100 | nan |
| 1e20| 1e20 | 1 | 1 | -124.40303872682864 |

jzmaddock:
Fixes #1198.

This change introduces some rather extreme new test cases, expect CI carnage!


jzmaddock:
Just the sycl tests failing so merging...

dschmitz89:
Thanks @jzmaddock ! Do you expect this to also close https://github.com/scipy/scipy/issues/21725 ? Another case of large a and b.

dschmitz89:
> Thanks @jzmaddock ! Do you expect this to also close [scipy/scipy#21725](https://github.com/scipy/scipy/issues/21725) ? Another case of large a and b.

Looks like it does not.

PR #1261, opened 2025-04-27
Handle various conversion warnings

ckormanyos:
Hi folks,

I'm really happy with this one. After literally years of improvements in our combined work, it is now possible to run parts of Math + Multiprecision _unchanged_ off-the-rack on both PC as well as embedded targets.

The purpose of this PR is to handle a few last warnings found when cross-compiling and running Math + Multiprecision on both PC-based GCC as well as 32-bit embedded ARM(R) bare-metal `gcc-arrm-none-eabi`.

The microcontroller compiler found several warnings that seemed quite reasonable to deal with.

I used warnings:

```bash
WARN_FLAGS  :=  -Wall                             \
                -Wextra                           \
                -Wpedantic                        \
                -Wmain                            \
                -Wundef                           \
                -Wconversion                      \
                -Wsign-conversion                 \
                -Wunused-parameter                \
                -Wuninitialized                   \
                -Wshadow
```

This PR does not cover warnings in _all_ of Math, nor in _all_ of Multiprecision. But quite a bit of specfun and both `cpp_bin_float` (which also pulls in `cpp_int`) as well as `cpp_dec_float` at $101$ decimal digits + Math have been successfully executed on PC and microcontroller and become relatively warning-free.

For `cpp_bin_float` and `cpp_dec_float` small amounts of dynamic RAM were needed (primarily for `std::allocator` in the `std::string` template class). So I wrote a tiny one-shot allocator for the global `new()` function and dispensed with the heap.

The Multirpecision benchmark computes the $101$ decimal digit cube root value of

$$
\left(\frac{123456}{100}\right)^{1/3}~\approx~10.72763694322831704548693173735276478\ldots
$$

The Math benchmark computes a small series of cylindrical Bessel function values for 64-bit `::boost::float64_t`

$$
J_{1.23}\left(1/n\right), n=1{\ldots}10
$$

The FLASH/RAM consumptions and timings for the Multiprecision benchmark are shown below (target ARM(R) Cortex(R) M4F STM32F446 from ST-Microelectronics(R) clocked at $168~MHz$). The multiprecision numbers were instantiated with $101$ decimal digits. I also used stubbed functions for some C-language posix primitives such as those pulled in by I/O streaming (needed by Math's exception handling even when exceptions are turned off). I also used RTTI disabled, optimization `-O2`.

| type                 | FLASH                    | RAM                      | time [ms]    |
|-----------------|---------------------|----------------------|--------------|
| `cpp_bin_float` | 97k                        | 6.3k                        |  1.0              |
| `cpp_dec_float` | 80k                        | 26k                        |  2.8              |

Cc: @jzmaddock and @mborland

### DIFF
--- a/include/boost/math/policies/policy.hpp
+++ b/include/boost/math/policies/policy.hpp
@@ -132,7 +132,7 @@ namespace policies{
 
 #define BOOST_MATH_META_INT(Type, name, Default)                                                \
    template <Type N = Default>                                                                  \
-   class name : public boost::math::integral_constant<int, N>{};                                \
+   class name : public boost::math::integral_constant<int, static_cast<int>(N)>{};              \
                                                                                                 \
    namespace detail{                                                                            \
    template <Type N>                                                                            \
--- a/include/boost/math/policies/policy.hpp
+++ b/include/boost/math/policies/policy.hpp
@@ -132,7 +132,7 @@ namespace policies{
 
 #define BOOST_MATH_META_INT(Type, name, Default)                                                \
    template <Type N = Default>                                                                  \
-   class name : public boost::math::integral_constant<int, static_cast<int>(N)>{};              \
+   class name : public boost::math::integral_constant<Type, N> { };                             \
                                                                                                 \
    namespace detail{                                                                            \
    template <Type N>                                                                            \
--- a/include/boost/math/special_functions/detail/bernoulli_details.hpp
+++ b/include/boost/math/special_functions/detail/bernoulli_details.hpp
@@ -108,7 +108,7 @@ inline std::size_t find_bernoulli_overflow_limit(const std::false_type&)
    // Set a limit on how large the result can ever be:
    static const auto max_result = static_cast<double>((std::numeric_limits<std::size_t>::max)() - 1000u);
 
-   unsigned long long t = lltrunc(boost::math::tools::log_max_value<T>());
+   unsigned long long t = static_cast<unsigned long long>(lltrunc(boost::math::tools::log_max_value<T>()));
    max_bernoulli_root_functor fun(t);
    boost::math::tools::equal_floor tol;
    std::uintmax_t max_iter = boost::math::policies::get_max_root_iterations<Policy>();
@@ -372,7 +372,7 @@ public:
          {
             for(; n; ++start, --n)
             {
-               *out = b2n_asymptotic<T, Policy>(static_cast<typename container_type::size_type>(start * 2U));
+               *out = b2n_asymptotic<T, Policy>(static_cast<int>(start * 2U));
                ++out;
             }
          }
--- a/include/boost/math/constants/info.hpp
+++ b/include/boost/math/constants/info.hpp
@@ -13,7 +13,9 @@
 #include <boost/math/constants/constants.hpp>
 #include <iostream>
 #include <iomanip>
+#ifndef BOOST_MATH_NO_RTTI
 #include <typeinfo>
+#endif
 
 namespace boost{ namespace math{ namespace constants{
 
@@ -22,7 +24,11 @@ namespace boost{ namespace math{ namespace constants{
       template <class T>
       const char* nameof(BOOST_MATH_EXPLICIT_TEMPLATE_TYPE_SPEC(T))
       {
+         #ifndef BOOST_MATH_NO_RTTI
          return typeid(T).name();
+         #else
+         return "unknown";
+         #endif
       }
       template <>
       const char* nameof<float>(BOOST_MATH_EXPLICIT_TEMPLATE_TYPE_SPEC(float))
--- a/include/boost/math/distributions/landau.hpp
+++ b/include/boost/math/distributions/landau.hpp
@@ -3382,7 +3382,7 @@ BOOST_MATH_GPU_ENABLED inline RealType landau_quantile_lower_imp_prec(const Real
 
         // Rational Approximation
         // Maximum Relative Error: 5.3064e-35
-        //LCOV_EXCL_START
+        // LCOV_EXCL_START
         BOOST_MATH_STATIC const RealType P[14] = {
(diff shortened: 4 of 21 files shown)


ckormanyos:
Oh yeah, you'll also see some warnings handled in other stuff like `toms748_solve`. These are in there since I also did some warning hunting on Ubuntu using my own [`extended_complex`](https://github.com/ckormanyos/extended_complex) adaption to comupte $500$ decimal digit non-trivial roots of the complex-valued Riemann-zeta function, another cool project I've benched on for the past ten years or so...

include/boost/math/policies/policy.hpp
```diff
@@ -132,7 +132,7 @@ namespace policies{
 
 #define BOOST_MATH_META_INT(Type, name, Default)                                                \
    template <Type N = Default>                                                                  \
-   class name : public boost::math::integral_constant<int, N>{};                                \
+   class name : public boost::math::integral_constant<int, static_cast<int>(N)>{};              \
```
jzmaddock, line 135: Should we cast to `Type` here since the argument could be a short or something (I'm sure it never is in practice, but it could be).

include/boost/math/special_functions/detail/bernoulli_details.hpp
```diff
@@ -372,7 +372,7 @@ class bernoulli_numbers_cache
          {
             for(; n; ++start, --n)
             {
-               *out = b2n_asymptotic<T, Policy>(static_cast<typename container_type::size_type>(start * 2U));
+               *out = b2n_asymptotic<T, Policy>(static_cast<int>(start * 2U));
```
jzmaddock, line 375: I have a hunch this is the wrong fix: a Bernoulli number may be too large for type `int` (overflow) and/or the wrong type for the iterator we are copying too - as the OutputIterator is a template parameter.  

Ideally, we would figure out what the `value_type` of the iterator is, and cast to that, but I don't think we can do that in C++20?  Failing that I think you need to look back up the instantiation stack to see what OutputIterator caused the warning, and possibly you may have to live with it (ie it's a genuine warning that you are doing something which may overflow).

jzmaddock:
Thanks Chris, other than my couple of comments, this all looks good to me.

include/boost/math/policies/policy.hpp
ckormanyos: I think you're right. But to be clear, you mean changing to something like below, right? That would make the most sense to me upon reading your comment.

```cpp
class name : public boost::math::integral_constant<Type, N>{};
```

include/boost/math/special_functions/detail/bernoulli_details.hpp
ckormanyos: But the function `b2n_asymptotic(int)` itself accepts an input parameter of type `int`. Would it make sense to refactor that entire thing to accept `std::size_t` or even `std::ptrdiff_t`? The asme is true for `t2n_asymptotic(int)`.

include/boost/math/policies/policy.hpp
jzmaddock: Yes.  But thinking about it, I'm not sure if that change might have unintended ripple effects - you might need to try it and see!

include/boost/math/special_functions/detail/bernoulli_details.hpp
jzmaddock: Sorry Chris, I completely mis-read that, please ignore my comment, I thought you were casting the result to an int not the argument.  You were quite right!

include/boost/math/policies/policy.hpp
ckormanyos: Hi John. It worked fine. I had a few typos underway. I will now merge this.

Cc: @jzmaddock and @mborland

ckormanyos:
OK so I see some `LCOV` warnings and (in this issue not even dealing with that) I did some corrections.

But I'm not quite sure that we have the code annotations exactly right. I looked 3 times but could not find the root cause of inconsistencies. When `specfun` job in `LCOV` uploads the results there are line/annotation inconsistencies. I repaired some of them and got more coverage and consistency, but we are still probably better than we thought.

I'll track this down in another issue. This is not the proper place to perfect our cover annotations.

Cc: @jzmaddock and @mborland

jzmaddock:
>OK so I see some LCOV warnings and (in this issue not even dealing with that) I did some corrections.

>But I'm not quite sure that we have the code annotations exactly right. I looked 3 times but could not find the root cause of inconsistencies.

>When the specfun job that is in the cover YAML uploads the results, there are line/annotation inconsistencies. I repaired some of them and got more coverage and consistency, but we are still probably better than we thought.

I haven't really touched on anything in /detail/ yet and I know there's a lot to look at in there.

What I'm trying to do is only annotate when:

1) It's confirmed as covered by a test (lets say a multiprecision one) that isn't run with coverage analysis because it's too slow.
2) When it's a GCOV SNAFU: these are mostly multi-line statements which GCOV just can't get right for some reason, the classic is a function call:

```
f(a,
  b,  // not covered!!
  c);
```

Which I change to

```
f(a, b, c);
```

even though that often makes things harder to read, or else a table:

```
T values[] = {
... ,   // non of these lines show up as covered.
};
```

Where I bracket the whole thing in LCOV_EXCL_START/LCOV_EXCL_STOP annotations

Other than that I'm trying to actually find test cases, and/or check to see if it's actually dead code which can be removed and/or converted to an assert.

ckormanyos:
Yes John (@jzmaddock), I agree with all that.

> It's confirmed as covered by a test (lets say a multiprecision one) that isn't run with coverage analysis because it's too slow.

Regarding this particular phrase pertaining to Multiprecision, ...

I was looking at a few of these. I was wondering what you thought of the idea of making a few selected multiprecision _spot_ _values_? These could be designed to run quickly, having just a few values each, and still cover some of those Multiprecision-only sequences in _specfun_.

I could start looking into this if you think it's the way to go.

Cc: @mborland

PR #1262, opened 2025-04-28
Improve Coverage Part 4

jzmaddock:


ckormanyos:
Hi John (@jzmaddock) I see you are in `lambert_w`. I made a note a few days ago that if anyone gets in there, I'd like to see if the inclusions of `<typeinfo>`, `<exception>` and `<iostream>` can be removed. This is one of the few remaining files with those heavyweights in it.

Also `<limits>` is in twice.

Can they be removed?

jzmaddock:
>Can they be removed?

Looks like it, lets see what happens...

jzmaddock:
@mborland , @NAThompson , @ckormanyos , another relatively large batch of coverage changes ready to go.  As before I'll give it a few days at least before merging in case there are any comments.

mborland:
Maybe for now it would be best to drop the SYCL and CUDA runners? CUDA shows green but if you open the run log there's a bunch of warnings that will be errors at runtime. Once you are done with your bulk changes I can come behind and fix the GPU stuff since, unfortunately, I have to do it out of tree which is a recipe for merge conflicts. SciPy went back to pinning themselves to versions instead of tracking develop, so there's little danger in these things breaking mid release cycle.

mborland:
> Would `-Werror cross-execution-space-call` do it?

I can't say I've ever tried that. I can re-engage about getting the GPU runners moved into the Boost CI system. They sit in the CppAlliance account now because of funding, but maybe once this FSC thing is settled the CppAlliance is paying either way?

jzmaddock:
OK, just a few warnings like these left:

```
[ 85%] Building CUDA object libs/math/test/CMakeFiles/boost_math-test_ellint_d_float.dir/test_ellint_d_float.cu.o
ptxas warning : Stack size for entry function '_Z9cuda_testPKdS0_Pdi' cannot be statically determined
```

I don't *think* those functions have recursions, and certainly I haven't changed them, are these issues?

jzmaddock:
Ah, take that back, they are recursive... not a new issue though.

PR #1265, opened 2025-05-15
Adjust parameter order in Differential Evolution algorithm

Tomato-in:
### Problem Description
The optimization module implements four algorithms: Differential Evolution, Algorithm jSO, Random Search, and Evolution Strategy with Covariance Matrix Adaptation. All these algorithms accept seven parameters: `cost_function`, `params`, `gen`, `value_to_reach`, `cancellation`, `current_minimum_cost`, and `queries`. Unfortunately, the Differential Evolution algorithm doesn't maintain parameter consistency with the other three algorithms: the last two parameters (`current_minimum_cost` and `queries`) are in reverse order compared to the other algorithms. This appears to be an unintentional oversight.

### Solution
- Adjust the parameter order in the Differential Evolution algorithm to ensure consistency with other optimization algorithms;
- Update the algorithm documentation accordingly;
- No changes to example code were needed as the examples don't use these two parameters.


ckormanyos:
Hi @Tomato-in for info, ... If I understand correctly, the `sycl` failures in CI are related to known problems and not within the scope of your changes.

PR #1271, opened 2025-06-10
Fix GPU support

mborland:
Fixes up support for all our GPU devices. I have no idea why the changes to ibeta and ibeta_inv fail with SYCL, they still run totally fine on NVCC and NVRTC so it's not a lack of compatibility. They are pretty far off in some cases too:

```
/home/mborland/Documents/boost/libs/math/test/../include_private/boost/math/tools/test.hpp(248): error: in "test_main": Gross error
ibeta<float> Max = 5.105e+06 RMS Mean=5.487e+05
    worst case at row: 1039
    { 1e+16, 1.0043945e+16, 0.49890375, 0, 0, 2.1688104e-19, 1 }
Peak error greater than expected value of 2

```


mborland:
I forgot that Windows-2019 and Ubuntu 22.04 are deprecated now and need removed. I'll open a second PR for those that'll get merged into this one.

mborland:
Closes: https://github.com/boostorg/multiprecision/issues/695

PR #1272, opened 2025-06-10
Fix CI deprecations

mborland:
Windows-2019 and Ubuntu-22.04 are deprecated in Github Actions, pending removal at the end of the month. Move runs that we can to Drone.


mborland:
Rolled into #1271

PR #1274, opened 2025-06-16
docs: add Windows install instructions

SasiKiran4328:
Adds Windows build instructions using Chocolatey and CMake for Boost.Math.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,14 @@ Currently a compiler that is fully compliant with C++14 is required to use Boost
 * clang++ 5 or later
 * Visual Studio 2015 (14.0) or later
 
+### 🪟 Windows Installation
+
+To build Boost.Math on Windows using **PowerShell** or **Command Prompt** with **VS Code + CMake**, first install dependencies via Chocolatey:
+
+```powershell
+choco install cmake visualstudio2019buildtools
+
+
 ## Support, bugs and feature requests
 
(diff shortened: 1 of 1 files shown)


ckormanyos:
Hi @SasiKiran4328 our repos require approval to run CI for first-time contributors, so I've just approved your workflow run on GHA.

Cc: @jzmaddock and @mborland

SasiKiran4328:
Hi Sir! @ckormanyos Thank you so much for approving the workflow run 
I wasn't aware of that CI policy for first-time contributors. Really appreciate the support! Excited to see the checks run now

mborland (changes requested):
Did you try and run this? There's already a build folder with files in it so you can't `mkdir` a new one. Then when you "build" outside of the boost tree the library assumes you're in standalone mode. This builds nothing since it's header only in the mode.

mborland:
> Hi Sir @mborland, Thank you for pointing that out and for the helpful explanation! 🙏🏼 You're absolutely right—I overlooked the fact that the repo already has a build folder and didn't consider the implications of standalone mode for a header-only library.
> 
> I'll correct the instructions to build within the Boost tree using a custom build folder like cmake-build. Appreciate your time and guidance. I’ll push the fix shortly.

Those instructions already exist in the docs for the libraries, and on the Boost.Cmake page. Our directions are specifically for development and already outline how to do this. I am going to close this since I think competing information is confusing at best since we don't run the CMake Infrastructure, or CML is generated automatically.

Issue #1275, opened 2025-06-16
Compilation errors on MSVC 14.0


Lastique:
Recently, Boost.Math was modified in a way that fails to compile with MSVC 14.0. This affects downstream libraries, specifically Boost.Parameter tests (which depend on Boost.Graph, which depends on Boost.Math):

https://ci.appveyor.com/project/Lastique/parameter/builds/52243732/job/r4qrqpaklphyncrb?fullLog=true#L845

I know the warning says MSVC 14.0 is no longer supported, but I think the `Jamfile` in Boost.Math should list the requirements that exclude the build on this compiler, which it currently doesn't. At least, `cxx14_constexpr` is missing (if the warning is to be believed), and some other requirement (currently non-existing in Boost.Config) that checks the presence of `*_t` and `*_v` versions of type traits in `<type_traits>`.

PR #1276, opened 2025-06-16
Add cxx14_constexpr requirement

mborland:
Partial for: #1275

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -21,7 +21,7 @@ project
       #<toolset>sun:<cxxflags>-xldscope=hidden
       [ check-target-builds ../config//has_gcc_visibility "gcc visibility" : <toolset>gcc:<cxxflags>-fvisibility=hidden : ]
       [ requires cxx11_noexcept cxx11_rvalue_references sfinae_expr cxx11_auto_declarations cxx11_lambdas cxx11_unified_initialization_syntax cxx11_hdr_tuple cxx11_hdr_initializer_list cxx11_hdr_chrono cxx11_thread_local cxx11_constexpr cxx11_nullptr cxx11_numeric_limits cxx11_decltype cxx11_hdr_array cxx11_hdr_atomic cxx11_hdr_type_traits cxx11_allocator cxx11_explicit_conversion_operators ]
-      [ requires cxx14_decltype_auto cxx14_generic_lambdas cxx14_return_type_deduction cxx14_variable_templates cxx14_decltype_auto cxx14_generic_lambdas cxx14_return_type_deduction ]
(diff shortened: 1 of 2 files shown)


jzmaddock:
While completely harmless, I'm not at all convinced that this fixes the original issue (or that there is an easy fix), the point being that only a vanishingly small number of dependents will be using the Math binaries as opposed to the headers, and if you're just using the headers, then there would appear to be no appropriate target to annotate with our requirements.  Unless I'm missing something - @Lastique.

Lastique:
Given the Boost.Build modularization efforts, header-only libraries now also have targets in b2. Those targets can also have the C++ requirements, as well as dependencies, attached. Those requirements and dependencies propagate to downstream users.

Specifically, if Boost.Math ever becomes header-only, it will still have a b2 target with all C++ requirements attached. Downstream users will still "link" with that target and thus inherit the C++ requirements.

Now, to this PR. I'm not sure `cxx14_constexpr` alone will be enough to fix the problem. The compiler error messages in Boost.Parameter tests suggest that it is the `<type_traits>` lacking C++14 stuff that is the problem. And for that there currently is no Boost.Config check. Either Boost.Math will have to add its own check, or (preferably) one needs to be added to Boost.Config.

mborland:
> Given the Boost.Build modularization efforts, header-only libraries now also have targets in b2. Those targets can also have the C++ requirements, as well as dependencies, attached. Those requirements and dependencies propagate to downstream users.
> 
> Specifically, if Boost.Math ever becomes header-only, it will still have a b2 target with all C++ requirements attached. Downstream users will still "link" with that target and thus inherit the C++ requirements.
> 
> Now, to this PR. I'm not sure `cxx14_constexpr` alone will be enough to fix the problem. The compiler error messages in Boost.Parameter tests suggest that it is the `<type_traits>` lacking C++14 stuff that is the problem. And for that there currently is no Boost.Config check. Either Boost.Math will have to add its own check, or (preferably) one needs to be added to Boost.Config.

#1277 Should cover the type traits issue. For that one I had to ask Rene how to globally apply a library specific configuration since `requires` in the Jamfile only applies to Boost.Config.

jzmaddock:
> Specifically, if Boost.Math ever becomes header-only, it will still have a b2 target with all C++ requirements attached. Downstream users will still "link" with that target and thus inherit the C++ requirements.

In 99.999% of cases it *is* header only, if users are linking to the compiled library just to express a dependency then that's flat out wrong.

Lastique:
Well, the dependency is necessary at least to get include paths to Boost.Math headers, as the monolithic include directory may no longer exist.

If you want to make linking with Boost.Math binary optional, you need to introduce a new target that corresponds to the header-only Boost.Math. The existing target should still link, since otherwise you would break that 0.001% of users.

Lastique:
BTW, the same is true with CMake, only it was the case from the start. In CMake, you also have targets to express dependencies, whether those targets are built or not.

PR #1284, opened 2025-07-03
Prevent std::ldexp underflowing/overflowing because of hard-coded flo…

ak-ambi:
Prevent `std::ldexp` underflowing/overflowing because of hard-coded float type.
With exceptions disabled, I would expect `errno` to be set only on real issues. However I was hit by errno set to `ERANGE` for following example:

```
T crossover = ldexp(1.0f, iround(99 / -0.654f));
```

This means that `float` is not able to fit `crossover` even for `df` as small as 99.

### DIFF
--- a/include/boost/math/special_functions/detail/t_distribution_inv.hpp
+++ b/include/boost/math/special_functions/detail/t_distribution_inv.hpp
@@ -404,8 +404,10 @@ calculate_real:
          // where we use Shaw's tail series.
          // The crossover point is roughly exponential in -df:
          //
-         T crossover = ldexp(1.0f, iround(T(df / -0.654f), typename policies::normalise<Policy, policies::rounding_error<policies::ignore_error> >::type()));
-         if(u > crossover)
+         int u_exp;
+         T m_exp = frexp(u, &u_exp);
+         // The following is equivalent to: u > 2^df/-0.654
+         if(m_exp > 0 && u_exp < df / 0.654f)
          {
             result = boost::math::detail::inverse_students_t_hill(df, u, pol);
          }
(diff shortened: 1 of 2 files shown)


jzmaddock:
I think we need a test case, because to my eyes this doesn't actually change anything?

Note that locally I'm unable to reproduce, but that may be because whether ERANGE is set on ldexp underflow is implementation defined.  Off the top of my head, I think we need something more like:

```
T twos_power = iround(T(df / -0.654f), typename policies::normalise<Policy, policies::rounding_error<policies::ignore_error> >::type());
T crossover = twos_power < std::numeric_limits<T>::min_exponent ? 0 : ldexp(1.0f, twos_power);
```

ak-ambi:
> I think we need a test case, because to my eyes this doesn't actually change anything?

I'll try to add test case tomorrow. The notable change is that new code does not cast T to float anymore. Double precision was enough to avoid ERANGE.

> Note that locally I'm unable to reproduce, but that may be because whether ERANGE is set on ldexp underflow is implementation defined. Off the top of my head, I think we need something more like:
> 
> ```
> T twos_power = iround(T(df / -0.654f), typename policies::normalise<Policy, policies::rounding_error<policies::ignore_error> >::type());
> T crossover = twos_power < std::numeric_limits<T>::min_exponent ? 0 : ldexp(1.0f, twos_power);
> ```

it safe to use min_exponent here? I can see some boost support for FLT_RADIX higher than 2, and min_exponent is min exp for `radix` as a base.

Other issue is that limit for T is used, but `float` types are used for `iround` and `ldexp`. Main change from my pull request is that T type is used in all calculations.

jzmaddock:
Ah yes, or just replacing 1.0f with static_cast<T>(1) would do it.  And you're right, my suggestion would fall foul of decimal types, though technically this change just pushes the problem further into the long grass as df > 1021 would still underflow with double.

PR #1290, opened 2025-08-01
Avoid using long double in fourier integrals when not supported

andrjohns:
When initialising a fourier integral class with double precision (e.g., `quadrature::ooura_fourier_sin<double>`) the internals always call the `add_level` member function with `long double`, even if the platform does not support it


mborland (commented):
This looks fine to me. Any comments @NAThompson? 

Unrelated to this PR it seems like the if-else logic being modified here should be `if constexpr`/`BOOST_MATH_IF_CONSTEXPR` since we are branching on the type of `Real`

jzmaddock:
+1.

Issue #1296, opened 2025-08-10
Proposed update for the quantile function of the logistic distribution.


WarrenWeckesser:
A while back I noticed that SciPy's `logit(p)` function lost precision near p=0.5.  I experimented with different formulations and found that `log1p(2*(p - 0.5)) - log1p(-2*(p - 0.5))` maintained high precision around p=0.5.  At the time, I didn't notice the formulation `2*atanh(2*p-1)`. (I'll probably propose we update the formula in SciPy soon.)

I see that the quantile function of the logistic distribution in boost/math also used the "naive" formula `log(p/(1-p))` (and there is even a comment about trying different formulations), so it also loses precision around p=0.5.  The expression `2*atanh(2*p-1)` appears to maintain high precision except when `p` is small, where precision is lost in the expression `2*p-1`.  (Note: so far, I've been testing with double precision only.)

I propose that the logistic quantile switch to the `atanh` formula for p > 0.25.

I computed `logit(p)` on a grid on (0, 1) with the current quantile, with `atanh` on the entire interval, and with the proposed version (i.e. `atanh` for `p > 0.25`, the usual formula otherwise).  Here's a plot of the relative errors (reference values were computed with `mpmath`).

<img width="764" height="381" alt="Image" src="https://github.com/user-attachments/assets/32e294ff-acd9-49b5-b4e1-6f10f24d47f7" />

What do you think?

mborland:
> A more substantial change would be to add [`logit`](https://en.wikipedia.org/wiki/Logit) and the [logistic sigmoid function](https://en.wikipedia.org/wiki/Logistic_function) (sometimes called `expit`) to the special functions, and use them in the logistic distribution implementation. This would reduce some existing code duplication. Giving them a `Policy` parameter might be a step towards fixing [gh-1294](https://github.com/boostorg/math/issues/1294).

Does `Scipy` have a `expit` implementation that could easily be ported into Boost.Math like the `logit` discussed above? This approach sounds good to me. I could put together the logic this week to have manual promotions based on policy. Something like:

```
Using std::atanh;
Using policy_real_type = boost::math::tools::policy_promote_args_t<RealType, Policy>;
Return static_cast<RealType>(2 * atanh(2 * static_cast<policy_real_type>(p) - 1));
```

It's a little ugly having all the casts, but then it actually does what it's supposed to do.

jzmaddock:
[...]

Issue #1294, opened 2025-08-08
Double promotion for the logistic quantile doesn't work?


WarrenWeckesser:
I'm looking into improving the precision of the logistic quantile function, and along the way I've found that I can't seem to get double promotion to work with that function.  Here's my test code:

```
#include <iostream>
#include <boost/math/distributions/logistic.hpp>

using namespace std;
using namespace boost::math::policies;
using boost::math::logistic_distribution;

typedef policy<
    promote_float<false>,
    promote_double<false>
> no_promotion;

typedef policy<
    promote_float<true>,
    promote_double<true>
> with_promotion;

int main()
{
    logistic_distribution<double, no_promotion> dist;
    logistic_distribution<double, with_promotion> dist_promote;
    logistic_distribution<long double, no_promotion> dist_longdouble;
    double p = 2049.0/4096;
    double ref = 0.0009765625776102258;  // Verified with both Wolfram Alpha and mpmath
    double x = quantile(dist, p);
    double xpromote = quantile(dist_promote,  p);
    double xld = static_cast<double>(quantile(dist_longdouble, static_cast<long double>(p)));
    cout << fixed << setw(25) << setprecision(17) << p << " p" << endl;
    cout << scientific << setw(25) << setprecision(17) << x << " double"<< endl;
    cout << scientific << setw(25) << setprecision(17) << xpromote << " double with promotion" << endl;
    cout << scientific << setw(25) << setprecision(17) << xld << " long double"<< endl;
    cout << scientific << setw(25) << setprecision(17) << ref << " reference"<< endl;

    return 0;
}
```

I'm on a Linux machine using gcc, where `long double` is extended (i.e. 80 bit) precision.

Output:
```
      0.50024414062500000 p
  9.76562577610170027e-04 double
  9.76562577610170027e-04 double with promotion
  9.76562577610225755e-04 long double
  9.76562577610225755e-04 reference
```

For any value of `p` that I use, the values `x` (computed with no promotion) and `xpromote` (computed with promotion) are the same.  The value computed with the distribution explicitly declared `long double` is correct (checked with Wolfram Alpha and `mpmath`).

mborland:
It looks like all the computations are being done with STL functions rather than `boost::math::` functions where the policy is passed along with the parameters. I think we'd need to do some `std::conditional` logic to figure out what `RealType` should be promoted to based of the policy, and then cast all the arguments into STL functions.

PR #1297, opened 2025-08-12
Add `logit`, `logistic_sigmoid`, and use in `logistic` dist to support promotion policies

mborland:
Closes: #1294 
Closes: #1296

### DIFF
--- a/include/boost/math/policies/policy.hpp
+++ b/include/boost/math/policies/policy.hpp
@@ -1009,8 +1009,6 @@ struct is_noexcept_error_policy
 template <typename Policy>
 struct make_forwarding_policy
 {
-   static_assert(detail::is_valid_policy<Policy>::value, "Policy must be valid to make a forwarding policy");
-
    using type =
           typename policies::normalise<
              Policy,


include/boost/math/special_functions/logistic_sigmoid.hpp
```diff
@@ -0,0 +1,41 @@
...
+namespace math {
+
+template <typename RealType, typename Policy>
+RealType logistic_sigmoid(RealType x, const Policy&)
+{
+    BOOST_MATH_STD_USING
+
+    using promoted_real_type = typename policies::evaluation<RealType, Policy>::type;
+
+    std::fexcept_t flags;
```
jzmaddock: Rather than messing with the exception flags, I would tend to add two special cases:

```
if(-x >= tools::log_max_value<T>()) return 0;
if(-x <= -tools::log_max_value<T>()) return 1;
```

Which has the advantage of working for types T with no infinity.

include/boost/math/special_functions/logit.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+template <typename RealType, typename Policy>
+RealType logit(RealType p, const Policy&)
+{
+    BOOST_MATH_STD_USING
+    using std::atanh;
+
+    using promoted_real_type = typename policies::evaluation<RealType, Policy>::type;
+
+    std::fexcept_t flags;
```
jzmaddock: I *think* the exception checks are unnecessary here: things can only go bad if p is so small that `(1-p)/p` overflows which implies that p is denormal.  I'm not sure what the std says about `log(+INF)` but I suspect we should filter out that case anyway?

So I guess I'm suggesting that we return the result of policies::raise_overflow_error when `p < tools::min_value<T>()`.  All other p values should return finite results.

include/boost/math/distributions/logistic.hpp
```diff
@@ -145,12 +147,8 @@ namespace boost { namespace math {
           return result;
        }
        BOOST_MATH_STD_USING
-       RealType power = (location - x) / scale;
-       if(power > tools::log_max_value<RealType>())
-          return 0;
-       if(power < -tools::log_max_value<RealType>())
-          return 1;
-       return 1 / (1 + exp(power)); 
+       RealType power = -(location - x) / scale;
```
WarrenWeckesser: Here and below: shouldn't this calculation also use the promoted type?  E.g. suppose `location` is 0, `scale` is 3, and `x` is 1.  If `RealType` is double precision, then `power` is `0.3333333333333333`, which becomes `0.33333333333333331483` when promoted to extended precision.  If the translation and scaling uses the promoted type, then `power` should be `0.33333333333333333334`, and that is the value that I would expect to be passed to `logistic_sigmoid`.

WarrenWeckesser: Very minor nitpick: The extra `-` seems unnecessary. `(x - location)/scale` would be the usual translation from the `(location, scale)` family to the standard value.

include/boost/math/distributions/logistic.hpp
```diff
@@ -221,15 +219,8 @@ namespace boost { namespace math {
...
-       //Expressions to try
-       //return location+scale*log(p/(1-p));
-       //return location+scale*log1p((2*p-1)/(1-p));
-
-       //return location - scale*log( (1-p)/p);
-       //return location - scale*log1p((1-2*p)/p);
 
-       //return -scale*log(1/p-1) + location;
-       return location - scale * log((1 - p) / p);
+       return location - scale * -logit(p, Policy());
```
WarrenWeckesser: Another very minor nitpick: remove the double negative, so the expression is `location + scale * logit(p, Policy())`.  That's the usual expression for going from a 'standard' value to a value in the `(location, scale)` family.  I assume the original `-` in front of `scale` was there because `log((1 - p)/p)` was used instead of `log(p/(1 - p))`.

WarrenWeckesser (commented):
That was fast!

I have a question below about using the promoted type with the location/scale calculations that occur outside of the  calls of `logit` or `logistic_sigmoid`.

Also a couple minor nitpicks involving minus signs.

include/boost/math/distributions/logistic.hpp
```diff
@@ -145,12 +147,8 @@ namespace boost { namespace math {
           return result;
        }
        BOOST_MATH_STD_USING
```
WarrenWeckesser: It looks like this is no longer necessary (likewise in `quantile`, and in the complemented versions).

include/boost/math/distributions/logistic.hpp
mborland: Having `power` as a promoted type makes sense. I then added the ability to disable further promotion to `logistic_sigmoid`. I was thinking about this case with the policy from your original issue[1]: `RealType` is `float` which means that `power` would be promoted to `double`. Without further disabling promotion `logistic_sigmoid` would compute internally using `long double` rather than staying at the already promoted type of `double`. 

[1]
```
typedef policy<
    promote_float<true>,
    promote_double<true>
> with_promotion;
```

WarrenWeckesser: Yeah, I was wondering how disabling a second promotion would work out.

Adding that boolean template parameter to the function seems a bit awkward.  Now `logistic_sigmoid` has a regular Policy parameter, *and* a parameter to disable promotion in that Policy.  Most users of `logistic_sigmoid` would never need it--they could just use an appropriately defined Policy.  Also, I think this boolean parameter would have to be added to any other special functions that get called from within distribution functions (assuming an audit of the other distributions finds more cases where promotion isn't fully implemented).

Instead, in the `cdf` function, could `policies::normalise` be used to define a new policy from the existing one, with promotion disabled? This appears to be the suggestion given at the bottom of https://www.boost.org/doc/libs/1_87_0/libs/math/doc/html/math_toolkit/pol_tutorial/policy_tut_defaults.html.

So something like

     normalise<Policy, promote_float<false>, promote_double<false>>::type

would be the policy used when, say, `cdf` calls `logistic_sigmoid`.

This worked for a trivial example that I created with `lambert_w0` (just to test the idea), but I haven't tried updating this PR with such a change, so I don't know if there are obstacles to making it work.

jzmaddock: This is how all the special functions avoid double-promotion when calling each other, there's a bit of boilerplate in the function prologue that looks like:

```
   typedef typename policies::normalise<
      Policy,
      policies::promote_float<false>,
      policies::promote_double<false>,
      policies::discrete_quantile<>,
      policies::assert_undefined<> >::type forwarding_policy;
```

And then `forwarding_policy` is passed down to any called functions.  If I'd had a bit more foresight, I might have added that as a member typedef of class `policy`.

mborland: Thanks John. I've added a `struct make_forwarding_policy` and `using make_forwarding_policy_t` so that we don't have to keep creating the policy. The call then looks like:

```
return logistic_sigmoid(power, policies::make_forwarding_policy_t<Policy>());
```

PR #1298, opened 2025-08-13
Add new constant `log_pi`

andrjohns:
Adds the definition and tests of a new constant for `log_pi`, used across multiple distribution functions.

Let me know if I've missed anything!


ckormanyos:
It looks good to me. Thanks for contributing a nice, useful constant number.

John, Matt should we merge?

PR #1299, opened 2025-08-14
Add manual promotion to arcsine, beta, and Bernoulli distributions

mborland:
This is an extension to #1294 since it looks like most of the distributions have functions where manual promotion needs to be applied to follow what the policy says. I tried to reduce the ugliness as much as possible, but the result is that the code is generally less readable than it is without this PR. Thoughts @jzmaddock and @WarrenWeckesser. If there's no real opposition I can continue plugging along with this.

### DIFF
--- a/include/boost/math/distributions/arcsine.hpp
+++ b/include/boost/math/distributions/arcsine.hpp
@@ -239,7 +239,12 @@ namespace boost
       {
         return result;
       }
-      return  (x_min + x_max) / 2;
+
+      using policy_promoted_type = policies::evaluation_t<RealType, Policy>;
+      const auto promoted_x_max = static_cast<policy_promoted_type>(x_max);
+      const auto promoted_x_min = static_cast<policy_promoted_type>(x_min);
+
+      return static_cast<RealType>((promoted_x_min + promoted_x_max) / 2);
     } // mean
 
     template <class RealType, class Policy>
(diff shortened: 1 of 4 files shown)


JacobHass8:
> I tried to reduce the ugliness as much as possible, but the result is that the code is generally less readable than it is without this PR. Thoughts @jzmaddock and @WarrenWeckesser. If there's no real opposition I can continue plugging along with this.

Would it be a little prettier if all the computations were done in a separate function. This is seems to be how type promotion is done with special functions. It seems a little more human readable to me since all the promotions are handled in one function and the computations in another.  For example, the mean of the arcsin function would be something like this

```C++
BOOST_MATH_GPU_ENABLED inline RealType mean(const arcsine_distribution<RealType, Policy>& dist)
{ // Mean of arcsine distribution .
  RealType result;
  RealType x_min = dist.x_min();
  RealType x_max = dist.x_max();

  if (false == arcsine_detail::check_dist(
    "boost::math::mean(arcsine_distribution<%1%> const&, %1% )",
    x_min,
    x_max,
    &result, Policy())
    )
  {
    return result;
  }

  using policy_promoted_type = policies::evaluation_t<RealType, Policy>;

  return static_cast<RealType>(arcsine_mean(static_cast<policy_promoted_type>(x_min), static_cast<policy_promoted_type>(x_max)));
}

template <class RealType>
inline RealType arcsine_mean(const RealType x_min, const RealType x_max)
{
  return (x_max + x_min) / 2;
}
```
For the arcsine distribution, this isn't that big of a change. However, it might make things prettier for more complicated functions.

PR #1301, opened 2025-08-16
Repair yet again issue 1292 tests

ckormanyos:


ckormanyos:
Hi John (@jzmaddock) and Matt (@mborland) while I'm struggling with some ad-hoc tests in Math/Multiprecision mixture (on Git issue 1292), I see what might be an unrelated problem.

On drone there is an actual [failure in interpolators](https://drone.cpp.al/boostorg/math/2330/13/2) that i can't explain. Is it real or bogus?

jzmaddock:
I know nothing about that test case, but superficially, it looks to me as if we're requiring a zero result, and are instead getting back 1eps.  Assuming the test can only produce a low absolute error, then this should probably be allowed.  Probably ;)

ckormanyos:
> I know nothing about that test case, but superficially, it looks to me as if we're requiring a zero result, and are instead getting back 1eps.

Me neither. I merged my changes anyway, figuring this interpolator issue is unrelated.

> Assuming the test can only produce a low absolute error, then this should probably be allowed. Probably

Maybe Matt know something next week? Otherwise, I can take a deeper look into it.

Issue #1281, opened 2025-06-23
reverese mode autodiff


demroz:
hi all,
I've been working on a bit of a side project ml library and noticed that boost doesn't have a reverse mode autodiff implemented. I have an implementation that works (see https://github.com/demroz/mznn/tree/main/aad). I'd be interested in porting it to boost. Would there be any interest in this?

demroz:
Thanks for your comments, I'll add your suggestions.

as far as boost functions, I tried to keep the user interface somewhat similar, but in short no. With reverse mode autodiff you differentiate with respect to your entire computational graph not just locally, so you get first order derivatives in constant time for R^n -> R^1 functions. Higher order derivatives aren't necessarily practical for reverse mode, and for high orders forward mode is generally preferable. So the interface I have is something like this:

```cpp
T f(const T& x, const T& y)
{
    auto y1 = log(1 + abs(x)) * exp(y); 
    auto y2 = pow(x + y, 2.5);
    auto y3 = sqrt(1 + x * y);
    return y1 + y2 - y3;
}

rvar<T,1> x = 1.0;
rvar<T,1> y = 2.0;
rvar<T,1> z = f(x,y);
z.backward(); // computes all gradients
x.adjoint(); // dz/dx
y.adjoint(); // dz/dy
/* or for higher order derivatives */
rvar<T,4> x = 1.0;
rvar<T,4> y = 2.0;
rvar<T,4> z = f(x,y);
auto grad_nd<1>(z, &x, &y); // vector -> dz/dx, dz/y
auto grad_nd<2>(z,&x,&y); // vector<vector> -> {{d^2 z /dx^2, d^2z/dxdy},{d^2 z /dydx, d^2 z/ dy^2}}
```
this interface is probably more similar to something like pytorch or tensorflow, but its fairly standard for reverse mode autodiff. I'm open to input though.

The issue I'm having right now with boost implemented functions is with the fact that I heavily rely on expression templates to define my math overloads. If we dont want to have two versions of these functions in different namespaces, is it possible to overload them inside the boost::math namespace for my expression type? If thats not desired then I think I'll have to remove the expression template gymnastics and take the performance hit.

demroz:
Hello @mborland 
[...]

PR #1302, opened 2025-08-16
New library: Reverse-mode automatic differentiation

demroz:
This pull request introduces a new library for reverse-mode automatic differentiation. Its a tape based reverse mode autodiff, so the idea is to build a computational graph, and then call backward on the entire graph to compute all the gradients at once. 

Currently it supports all the basic operations (+,-,/,/), everything listed in conceptual requirements for real number types, and boost calls to `erf`, `erc`, `erf_inv` and `erfc_inv`. Everything is tested up to the 4th derivative. The list of tests:
`test_reverse_mode_autodiff_basic_math_ops.cpp`
`test_reverse_mode_autodiff_comparison_operators.cpp`
`test_reverse_mode_autodiff_constructors.cpp`
`test_reverse_mode_autodiff_error_functions.cpp`
`test_reverse_mode_autodiff_flat_linear_allocator.cpp`
`test_reverse_mode_autodiff_stl_support.cpp`

There are also two examples in the example directory:
`reverse_mode_linear_regression_example.cpp` -> simple linear regression that demonstrates how this library can be used for optimization

`autodiff_reverse_black_scholes.cpp` -> a rewrite of the forward mode equivalent.

Important notes
1. The intent of the library is to be an engine for gradient based optimization. (In the future I'd be interested in adding some gradient based optimizers to boost if there is interest.) I wrote it with first derivatives in mind. Although everything is tested up to the 4th derivative, the library is intended to be optimally used for calculating first order derivatives. Its best to use forward mode autodiff if very high orders are needed.
2. The library is based on expression templates. This means that although my autodiff variable rvar does satisfy all the requirements of a real-type, it doesn't necessarily play nicely with boost special functions. For example:
```cpp
rvar<double,1> x = make_rvar<T,1>(1.0);
rvar<double,1> y = make_rvar<T,1>(1.0);
rvar<double,1> z = make_rvar<T,1>(1.0);
auto f = x+y*z;
```
f in this case is not actually type `rvar`, but `add_expr<rvar,mult_expr<rvar,rvar>>`

3. reverse_mode_autodiff_memory_management.hpp has a flat_linear_allocator class thats essentially a specialized memory arena. It uses placement `new` for memory allocations. This is a deliberate design choice. The flat_linear_allocator destructor explicitly calls the destructors of individual elements. Explicit calls to `delete` shouldn't be needed here

Thank you, and looking forward to your feedback


demroz:
looks like msvc errors are gone. Not sure why CI fails but looking inside it doesn't seem to be related to any of my code

ckormanyos:
> Not sure why CI fails but looking inside it doesn't seem to be related to any of my code

The failure(s) in CI seem harmless, as you mentioned.

The job that failed CI is on Drone. This, in fact, very many jobs running on various older runner images on Drone. Most of the tests for older compilers such as GCC 7, 8 or old clang. And non-x86_64 architectures (insofar as these are available) run on Drone.

But sometimes, Drone has a hard time acquiring a VM or setting it up properly. So your failures failed in about 3 minutes - too short for a normal run, as these are mostly 7 - 30 minutes. It looks like all the Drone failures were the failure to properly stup LINUX. It gets annoying, but Drone has to be manually judged sometimes to see if failures were real or bogus. This gets doubly annoying when there is actually mixture of real and bogus errors. Which is what happens when we write new code. Sigh, that's just how it goes.

On the upside, GHA is super-reliable nowdays.

example/autodiff_reverse_black_scholes.cpp
```diff
@@ -0,0 +1,153 @@
...
+// Distributed under the Boost Software License, Version 1.0.
+//      (See accompanying file LICENSE_1_0.txt or copy at
+//           https://www.boost.org/LICENSE_1_0.txt)
+#include <boost/math/differentiation/autodiff_reverse.hpp>
+
+using namespace boost::math::differentiation::reverse_mode;
+using namespace boost::math::constants;
+
+template<typename X>
+X phi(X const& x)
```
mborland: I would use something more descriptive like `typename Real` to indicate what people should be using for `x`

example/autodiff_reverse_black_scholes.cpp
```diff
@@ -0,0 +1,153 @@
...
+{
+    return one_div_root_two_pi<double>() * exp(-0.5 * x * x);
```
mborland: You can then do: `one_div_root_two_pi<Real>()` so we're not losing precision on say `Real = long double`

example/autodiff_reverse_black_scholes.cpp
```diff
@@ -0,0 +1,153 @@
...
+    using namespace std;
+    auto const d1 = (log(S / K) + (r + sigma * sigma / 2) * tau) / (sigma * sqrt(tau));
+    auto const d2 = (log(S / K) + (r - sigma * sigma / 2) * tau) / (sigma * sqrt(tau));
+    switch (cp) {
+    case CP::call:
+        return S * Phi<T>(d1) - exp(-r * tau) * K * Phi<T>(d2);
+    case CP::put:
+        return exp(-r * tau) * K * Phi<T>(-d2) - S * Phi<T>(-d1);
+    default:
+        throw std::runtime_error("Invalid CP value.");
```
mborland: You don't need the `std::` since you already are `using namespace std`

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
```diff
@@ -0,0 +1,346 @@
...
+/****************************************************************************************************************/
+template<typename T, size_t order, typename LHS, typename RHS>
+struct add_expr
+    : public abstract_binary_expression<T, order, LHS, RHS, add_expr<T, order, LHS, RHS>>
+{
+    /* @brief addition
+   * rvar+rvar
+   * */
+    using lhs_type   = LHS;
+    using rhs_type   = RHS;
```
mborland: It doesn't look like you ever use these typedefs. I would just name LHS and RHS appropriately so that you don't need to immediately create typedefs

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
```diff
@@ -0,0 +1,346 @@
...
+}; // namespace reverse_modetemplate<typenameT,size_torder,typenameLHS,typenameRHS>structadd_expr:publicabstract_binary_expression<T,order,LHS,RHS,add_expr<T,order,LHS,RHS> >
+template<typename T, size_t order, typename ARG>
+struct add_const_expr
+    : public abstract_unary_expression<T, order, ARG, add_const_expr<T, order, ARG>>
+{
+    /* @brief
+   * rvar+float or float+rvar
+   * */
+    using arg_type   = ARG;
+    using value_type = T;
```
mborland: Same thing down the line. Name your template types so that you don't immediately make typedefs at the top of each struct

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
```diff
@@ -0,0 +1,346 @@
...
+    using inner_t    = rvar_t<T, order - 1>;
+    explicit add_const_expr(const expression<T, order, ARG> &arg_expr, const T v)
+        : abstract_unary_expression<T, order, ARG, add_const_expr<T, order, ARG>>(arg_expr, v){};
+    inner_t              evaluate() const { return this->arg.evaluate() + inner_t(this->constant); }
+    static const inner_t derivative(const inner_t &argv, const inner_t &v, const T &constant)
```
mborland: I think something like:

```
static const inner_t derivative(const inner_t& /*argv*/, const inner_t& /*v*/, const T& /*constant*/)
```

Or 

```
static const inner_t derivative(const inner_t&, const inner_t &, const T&)
```

Is preferable otherwise the users will get a ton of unused parameter warnings.

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
```diff
@@ -0,0 +1,346 @@
...
+    explicit div_expr(const expression<T, order, LHS> &left_hand_expr,
+                      const expression<T, order, RHS> &right_hand_expr)
+        : abstract_binary_expression<T, order, LHS, RHS, div_expr<T, order, LHS, RHS>>(
+              left_hand_expr, right_hand_expr)
+    {}
+
+    inner_t              evaluate() const { return this->lhs.evaluate() / this->rhs.evaluate(); };
+    static const inner_t left_derivative(const inner_t &l, const inner_t &r, const inner_t &v)
+    {
+        return 1.0 / r;
```
mborland:
```suggestion
        return static_cast<inner_t>(1.0) / r;
```

Anywhere that you have a floating point constant like this you'll want to wrap with a proper cast. Reasoning is some types will throw/warn/error because they are of different precision. E.g. types from `<stdfloat>`

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
```diff
@@ -0,0 +1,346 @@
...
+ *  subtraction float - rvar */
+template<typename U,
+         typename T,
+         size_t order,
+         typename ARG,
+         typename = typename std::enable_if<!detail::is_expression<U>::value>::type>
+add_const_expr<T, order, ARG> operator-(const expression<T, order, ARG> &arg, const U &v)
+{
+    /* rvar - float = rvar + (-float) */
+    return add_const_expr<T, order, ARG>(arg, static_cast<T>(-1.0 * v));
```
mborland: Why not just -v since it's not an expression template?

include/boost/math/differentiation/detail/reverse_mode_autodiff_erf_overloads.hpp
```diff
@@ -0,0 +1,206 @@
...
+        return detail::if_functional_dispatch<(order > 1)>(
+            [this](auto &&x) { return reverse_mode::erf_inv(std::forward<decltype(x)>(x)); },
+            [this](auto &&x) { return boost::math::erf_inv(std::forward<decltype(x)>(x)); },
+            this->arg.evaluate());
+    }
+    static const inner_t derivative(const inner_t &argv, const inner_t &v, const T &constant)
+    {
+        using multiprecision::exp;
+        using multiprecision::pow;
+        using multiprecision::sqrt;
```
mborland: Do we need to explicitly include these? I believe not

include/boost/math/differentiation/detail/reverse_mode_autodiff_erf_overloads.hpp
```diff
@@ -0,0 +1,206 @@
...
+        using std::exp;
+        using std::pow;
+        using std::sqrt;
```
mborland: All of the `using std::` math functions can be replaced with `BOOST_MATH_STD_USING`

include/boost/math/differentiation/detail/reverse_mode_autodiff_expression_template_base.hpp
```diff
@@ -0,0 +1,242 @@
+//           Copyright Maksym Zhelyenzyakov 2025-2026.
+// Distributed under the Boost Software License, Version 1.0.
+//      (See accompanying file LICENSE_1_0.txt or copy at
+//           https://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef REVERSE_MODE_AUTODIFF_EXPRESSION_TEMPLATE_BASE_HPP
+#define REVERSE_MODE_AUTODIFF_EXPRESSION_TEMPLATE_BASE_HPP
+#include <stddef.h>
```
mborland: `<cstddef>`

include/boost/math/differentiation/detail/reverse_mode_autodiff_memory_management.hpp
```diff
@@ -0,0 +1,433 @@
...
+
+#include <algorithm>
+#include <array>
+#include <cassert>
+#include <iterator>
+#include <memory>
+#include <stddef.h>
+#include <type_traits>
+#include <vector>
+#include <boost/assert.hpp>
```
mborland: `<boost/math/tools/assert.hpp>`

include/boost/math/differentiation/detail/reverse_mode_autodiff_memory_management.hpp
```diff
@@ -0,0 +1,433 @@
...
+                                            size_t                begin,
+                                            size_t                end)
+        : storage_(storage)
+        , index_(index)
+        , begin_(begin)
+        , end_(end)
+    {}
+    reference operator*()
+    {
+        BOOST_ASSERT(index_ >= begin_ && index_ < end_);
```
mborland: `BOOST_MATH_ASSERT`

include/boost/math/differentiation/detail/reverse_mode_autodiff_memory_management.hpp
```diff
@@ -0,0 +1,433 @@
...
+    template<size_t n>
+    iterator emplace_back_n()
+    {
+        size_t bid = buffer_id();
+        size_t iid = item_id();
+        if (iid + n < buffer_size) {
+            T *ptr = &(*data_[bid])[iid];
+            for (size_t i = 0; i < n; ++i) {
+                new (ptr + i) T();
+            }
```
mborland: I would add a null check with your placement new
 
T *ptr = &(*data_[bid])[iid];
if (ptr != nullptr) {
    new (ptr + i) T();
} else {
    // Handle error
}

include/boost/math/differentiation/autodiff_reverse.hpp
```diff
@@ -0,0 +1,769 @@
...
+#include <boost/math/special_functions/round.hpp>
+#include <boost/math/special_functions/trunc.hpp>
+#include <boost/math/tools/config.hpp>
+#include <boost/math/tools/promotion.hpp>
+#include <cassert>
+#include <cstddef>
+#include <iostream>
+#include <type_traits>
+#include <vector>
+#define BOOST_MATH_BUFFER_SIZE 65536
```
mborland, line 30: I would do `static constexpr std::size_t buffer_size = 65536U` inside your namespace

include/boost/math/differentiation/autodiff_reverse.hpp
```diff
@@ -0,0 +1,769 @@
+//           Copyright Maksym Zhelyenzyakov 2025-2026.
+// Distributed under the Boost Software License, Version 1.0.
+//      (See accompanying file LICENSE_1_0.txt or copy at
+//           https://www.boost.org/LICENSE_1_0.txt)
+#ifndef BOOST_MATH_DIFFERENTIATION_AUTODIFF_HPP
+#define BOOST_MATH_DIFFERENTIATION_AUTODIFF_HPP
+
+#include <boost/cstdfloat.hpp>
```
mborland: You shouldn't need this header for anything

include/boost/math/differentiation/autodiff_reverse.hpp
```diff
@@ -0,0 +1,769 @@
...
+#include <boost/math/special_functions/atanh.hpp>
+#include <boost/math/special_functions/digamma.hpp>
+#include <boost/math/special_functions/erf.hpp>
+#include <boost/math/special_functions/lambert_w.hpp>
+#include <boost/math/special_functions/polygamma.hpp>
+#include <boost/math/special_functions/round.hpp>
+#include <boost/math/special_functions/trunc.hpp>
+#include <boost/math/tools/config.hpp>
+#include <boost/math/tools/promotion.hpp>
+#include <cassert>
```
mborland: Instead of naked `assert` prefer using `BOOST_MATH_ASSERT`

include/boost/math/differentiation/autodiff_reverse.hpp
```diff
@@ -0,0 +1,769 @@
...
+    std::vector<rvar<T, order - 1>> operator()(rvar<T, order> &f, std::vector<rvar<T, order> *> &x)
+    {
+        auto &tape = get_active_tape<T, order>();
+        tape.zero_grad();
+        f.backward();
+
+        std::vector<rvar<T, order - 1>> gradient_vector;
+        gradient_vector.reserve(x.size());
+
+        for (auto xi : x) {
```
mborland: Is it expensive to make copies of each `xi`?

.gitignore
```diff
@@ -28,3 +28,6 @@ cmake-build-debug/*
 build.ninja
 .ninja*
 a.out
+include/boost/math/differentiation/build/
```
mborland: Is anything actually being built?

CMakeLists.txt.user
```diff
@@ -0,0 +1,455 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mborland: This file should be deleted from the PR

example/autodiff_reverse_black_scholes.cpp
demroz: should be fixed

example/autodiff_reverse_black_scholes.cpp
demroz: fixed

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
demroz: fixed

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
demroz: should be fixed now

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
demroz: fixed

include/boost/math/differentiation/detail/reverse_mode_autodiff_erf_overloads.hpp
demroz: removed

include/boost/math/differentiation/detail/reverse_mode_autodiff_erf_overloads.hpp
demroz: removed

include/boost/math/differentiation/detail/reverse_mode_autodiff_memory_management.hpp
demroz: I think the nullpointer check is redundant here

`buffer_ptr `is a unique_ptr
`make_unique` never returns null
and i always check or create a buffer before writing to it with 

```cpp
if (item_id() == 0 && total_size_ != 0) allocate_buffer();
```

so `data_[bid]` is never null
`&(*data_[bid])[iid]` is then never null

The only thing that could possibly throw here is the constructor for T

include/boost/math/differentiation/autodiff_reverse.hpp
demroz: I like having a macro here so the user can do -DBOOST_MATH_BUFFER_SIZE 32768 and benchmark the library for their hardware if they want to. I admit that I kind of just pulled the 65536 number from thin air, we could always do something like

```cpp
#ifndef BOOST_MATH_BUFFER_SIZE
#define BOOST_MATH_BUFFER_SIZE 65536
#endif

namespace boost::math::differentiation::reverse_mode{
static constexpr std::size_t buffer_size = BOOST_MATH_BUFFER_SIZE
}
```

what do you think?

include/boost/math/differentiation/autodiff_reverse.hpp
demroz: removed

include/boost/math/differentiation/autodiff_reverse.hpp
demroz: typo, changed to auto& xi

CMakeLists.txt.user
demroz: removed

.gitignore
demroz: I use cmake/qtcreator. This is legacy from building my tests directly in math/differentation. I removed it

include/boost/math/differentiation/autodiff_reverse.hpp
demroz: changed

include/boost/math/differentiation/detail/reverse_mode_autodiff_memory_management.hpp
demroz: changed

example/autodiff_reverse_black_scholes.cpp
demroz: removed

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp
demroz: fixed

include/boost/math/differentiation/autodiff_reverse.hpp
mborland: > I like having a macro here so the user can do -DBOOST_MATH_BUFFER_SIZE 32768 and benchmark the library for their hardware if they want to. I admit that I kind of just pulled the 65536 number from thin air, we could always do something like
> 
> ```c++
> #ifndef BOOST_MATH_BUFFER_SIZE
> #define BOOST_MATH_BUFFER_SIZE 65536
> #endif
> 
> namespace boost::math::differentiation::reverse_mode{
> static constexpr std::size_t buffer_size = BOOST_MATH_BUFFER_SIZE
> }
> ```
> 
> what do you think?

Having the user configuration option makes sense to me.

include/boost/math/differentiation/detail/reverse_mode_autodiff_memory_management.hpp
mborland: You're correct; no change needed then.

ckormanyos:
> how about some of the test cases under ...

Yeah, now we're getting into it. Thx @demroz.

OK I pulled that code out into a standalone (non-boost) test and its working. See below.

I have my first findings:

- The class `rvar` is referred to as both a `class` as well as a `struct`. I think it should be a class everywhere.
- The boost guys like `TemplateParamsLikeThis`. So when your parameter is `order` you might wand `Order` or a more specific word like `DerivativeOrder`.

I'll work through more and more code and report findings on the go.

```
#include <boost/math/differentiation/autodiff_reverse.hpp>

#include <iomanip>
#include <iostream>
#include <limits>
#include <sstream>
#include <vector>

namespace local
{
  template<typename NumericType>
  auto is_close_fraction(const NumericType& a,
                         const NumericType& b,
                         const NumericType& tol) noexcept -> bool
  {
    using std::fabs;

    auto result_is_ok = bool { };

    NumericType delta { };

    if(b == static_cast<NumericType>(0))
    {
      delta = fabs(a - b); // LCOV_EXCL_LINE

      result_is_ok = (delta <= tol); // LCOV_EXCL_LINE
    }
    else
    {
      delta = fabs(1 - (a / b));

      result_is_ok = (delta <= tol);
    }

    return result_is_ok;
  }
} // namespace local

#define MY_REQUIRE_CLOSE(a, b, tol) local::is_close_fraction((a), (b), (tol))

template<typename T>
T f(T x, T y)
{
    return (x * y) / ((x + y) * (x - y));
}

template<typename T>
std::vector<T> grad_f_a(T x, T y)
{
    T f_x = static_cast<T>(y * (-pow(x, 2) - pow(y, 2))
                           / (pow(x, 4) - 2 * pow(x, 2) * pow(y, 2) + pow(y, 4)));
    T f_y = static_cast<T>(x * (pow(x, 2) + pow(y, 2))
                           / (pow(x, 4) - 2 * pow(x, 2) * pow(y, 2) + pow(y, 4)));
    return {f_x, f_y};
}

template<typename T>
constexpr T boost_close_tol(T scale_factor = 1e5)
{
    return std::numeric_limits<T>::epsilon() * scale_factor;
}

template<typename T, size_t order = 1>
using local_rvar_type = boost::math::differentiation::reverse_mode::rvar<T, order>;

template<typename T>
auto do_something() -> bool
{
    bool result_is_ok { true };

    //RandomSample<T> rng{-1, 1};
    T x = 1.1224;   //rng.next();
    T y = 5.213414; //rng.next();

    local_rvar_type<T, 1>      x_ad              = x;
    local_rvar_type<T, 1>      y_ad              = y;

    //T               fv                = f(x, y);
    local_rvar_type<T, 1> f_ad              = f(x_ad, y_ad);
    auto                  grad_f_analytical = ::grad_f_a(x, y);
    /* intended use case */
    f_ad.backward(); // <- this computes the derivatives in place

    {
        const bool result_inplace_d0_is_ok { MY_REQUIRE_CLOSE(x_ad.adjoint(), grad_f_analytical[0], boost_close_tol<T>()) };
        const bool result_inplace_d1_is_ok { MY_REQUIRE_CLOSE(y_ad.adjoint(), grad_f_analytical[1], boost_close_tol<T>()) };

        result_is_ok = (result_inplace_d0_is_ok && result_is_ok);
        result_is_ok = (result_inplace_d1_is_ok && result_is_ok);
    }

    boost::math::differentiation::reverse_mode::gradient_tape<T, 1, BOOST_MATH_BUFFER_SIZE>& tape = boost::math::differentiation::reverse_mode::get_active_tape<T, 1>();
    tape.zero_grad();

    /* grad test */
    auto grad_func_test_grad    = grad(f_ad, &x_ad, &y_ad); // <- this returns a copy of the gradient vector w.r.t x and t
    /* grad_nd test */
    auto grad_nd_func_test_grad = grad_nd<1>(f_ad, &x_ad, &y_ad); // <- same thing, but use this function for higher order derivatives

    for (int i = 0; i < 2; i++)
    {
        const bool result_inplace_d0_is_ok { MY_REQUIRE_CLOSE(grad_func_test_grad[i], grad_f_analytical[i], boost_close_tol<T>()) };
        const bool result_inplace_dn_is_ok { MY_REQUIRE_CLOSE(grad_nd_func_test_grad[i], grad_f_analytical[i], boost_close_tol<T>()) };

        result_is_ok = (result_inplace_d0_is_ok && result_is_ok);
        result_is_ok = (result_inplace_dn_is_ok && result_is_ok);
    }

    return result_is_ok;
}

auto main() -> int
{
  const bool result_is_ok { do_something<double>() };

  std::stringstream strm { };

  strm << "result_is_ok: " << std::boolalpha << result_is_ok;

  std::cout << strm.str() << std::endl;
}
```

ckormanyos:
Also when you get into Code, then your first step is a local small-case type.

I like thinkgs like

```
template<typename RealType>
auto do_something(RealType x) -> void
{
  using local_real_type = RealType;

  static_cast<void>(std::numeric_limits<local_real_type>::is_specialized);
}
```

demroz:
just to be clear something like this? Because I think I got conflicting advice from @mborland
```cpp
template<typename RealType, size_t DerivativeOrder, typename LHS, typename RHS>
struct add_expr
    : public abstract_binary_expression<RealType, DerivativeOrder, LHS, RHS, add_expr<RealType, DerivativeOrder, LHS, RHS>>
{
    /* @brief addition
   * rvar+rvar
   * */
   using real_type = RealType;
   using order = DerivatveOrder;
   using lhs_type = LHS;
   using rhs_type = RHS:
    using inner_t    = rvar_t<real_type, order - 1>;
    explicit add_expr(const expression<real_type, order, lhs_type> &left_hand_expr,
                      const expression<real_type, order, rhs_type> &right_hand_expr)
        : abstract_binary_expression<real_type, order, lhs_type, rhs_type, add_expr<real_type, order, lhs_type, rhs_type>>(
              left_hand_expr, right_hand_expr)
    {}

    inner_t              evaluate() const { return this->lhs.evaluate() + this->rhs.evaluate(); }
    static const inner_t left_derivative(const inner_t & /*l*/,
                                         const inner_t & /*r*/,
                                         const inner_t & /*v*/)
    {
        return inner_t(1.0);
    }
    static const inner_t right_derivative(const inner_t & /*l*/,
                                          const inner_t & /*r*/,
                                          const inner_t & /*v*/)
    {
        return inner_t(1.0);
    }
};
```

ckormanyos:
Hi @demroz I ran your code in a first pass through MSVC's Intellisense and also GCC/clang. There were some very trivial findings with warnings, as desctibed [here](https://github.com/demroz/math/pull/5). I ended PR-ing it befors I say, do put a cast here, don't have a semi-colon there and whatnot.

There is no real content there.

I have also worked much more with taking derivatives in one or more variables.

I have encountered the clash between Multiprecision types and the `rval` types. I think down the road for this we might consider _bindings_ for all of the basic operations and `<cmath>` functions. We made numerous bindings over the years to make Multiprecision types _play_ _nice_ with Boost.Math. In fact, we had to since Multiprecision types are also expression template enabled (when parameterized to do so).

But if we go down the bindings-road, we will need a big long bindings file. That would be something I would kick down the road for a while. First get the gradients working, then we could talk about how they could/should evolve.

Cc: @mborland

mborland:
@ckormanyos since this is his first PR all the GitHub actions runs need to be approved manually on every run. 

@demroz I don't intend to ask for any other changes at this point either. I look through the docs once you say they're good to go.

ckormanyos:
Hi @demroz I have another query, and it might be worth looking into.

The program below computes the gradient (in this case first derivative) of a real-valued function in one variable, $x$.

I'm playing around with a very beautiful identity which is:

$
\frac{ \frac{d}{dx}\Gamma}(x)}{\Gamma}(x)}={\psi}(x)
$

ckormanyos:
Another query on syntax...

Why don't lines like [this](https://github.com/demroz/math/blob/602f227398a02646dab1243ba692a7ce3381cd30/include/boost/math/differentiation/autodiff_reverse.hpp#L580) or [this other](https://github.com/demroz/math/blob/602f227398a02646dab1243ba692a7ce3381cd30/include/boost/math/differentiation/autodiff_reverse.hpp#L613) use `(auto& xi : x)` in other words with a reference?

I am not sure, but I think a reference _might_ be OK here and better for UDTs like Multiprecision types.

ckormanyos:
> if you check the [add_special_function_overloads](https://github.com/demroz/math/blob/add_special_function_overloads/include/boost/math/differentiation/detail/autogenerated_gamma.hpp) branch on my fork, you can see a working header file for gamma functions. This solution is minimally invasive to the rest of the library

Hmmm... are we thinking through this right? I was wondering if adding boilerpülace code that binds global add/sub/mul/div, comparison operatory, all `<cmath>` functions to expression template-versions of `rval` would be sufficient to handle all of specfun at once, without dealing with each individual function. Now these binding template-functions might end up in `namespace boost::math::differentiation` or whatever the right place is, and they would be found via ADL. That would be (in my best first-guess) something like 2,000-4000 lines, but one-time only. If something like this works or would work, I'd rather do it once than for each individual function.

To be fair though, I have not thought this through in its entirety.

Your thoughts Matt (@mborland)?

ckormanyos:
> I'm surprised that the calls to tgamma work at all and produce correct derivatives anywhere

I cheated. `tgamma()` is called from within `test_function()`.

mborland:
> > if you check the [add_special_function_overloads](https://github.com/demroz/math/blob/add_special_function_overloads/include/boost/math/differentiation/detail/autogenerated_gamma.hpp) branch on my fork, you can see a working header file for gamma functions. This solution is minimally invasive to the rest of the library
> 
> Hmmm... are we thinking through this right? I was wondering if adding boilerplate code that binds global add/sub/mul/div, comparison operators, all `<cmath>` functions, etc. to expression template-versions of `rval` would be sufficient to handle all of specfun at once, without dealing with each individual function. Now these binding template-functions might end up in `namespace boost::math::differentiation` or whatever the right place is, and they would be found via ADL. That would be (in my best first-guess) something like 2,000-4000 lines, but one-time only. If something like this works or would work, I'd rather do it once than for each individual function.
> 
> To be fair though, I have not thought this through in its entirety.
> 
> Your thoughts Matt (@mborland)?

To me it sounded like the wrappers you are describing aren't feasible, which is why he has the boilerplate versions of these functions. Same thing happened in the forward autodiff library. I told him the current functions he supports is a good start and there's room for expansion should people request it.

ckormanyos:
> To me it sounded like the wrappers you are describing aren't feasible,

Uggghhh yuck. This is a hard question, I wouldn't know how to answer it, but do you know why it's not feasible? I think I took a local try at it and ran into trouble right away and was wondering if this is the non-feasibility? I could not even write the constructors i needed. And global assignment operators aren't allowed. So I gave up.

i just thought somewhere out there might be a generic solution instead of a function-by-function dedicated effort...

ckormanyos:
I suppose this is going to sound literally crazy. But in some kind of world, you could make a `number` multiprecision backend for `rval`. But again you'd need a few thousand lines for the `eval_whatever()` functions. And I do not know if this is feasible either?

demroz:
> Well I can get it compiling for `cyl_bessel_j`, but it gives me the 0.0 answer for derivatives. I wonder if it's possible to see where the call graph is breaking? And why?

could you send me the code? I'll look at it. The way I tend to try to find where the graph breaks is step through with a debugger and see if there is an unexpected call. So for tgamma, what I stepped through the call until i saw i was inside the boost version `itrunc(z, pol)` which was unexpected to me because I overload `itrunc`. Sorry that I don't have a better answer for you. I don't really know of good tools to visualize these sorts of graphs



> Also @demroz I see lines like [this](https://github.com/demroz/math/blob/b53cfc25c9e841c41fa7000ea9fa5361eb211427/include/boost/math/differentiation/autodiff_reverse.hpp#L703) at various places.
> 
> I think you might know the size of the vector up-front, although I'm not sure. If the size of the vector is known, you might want to `reserve()` enough place in the vector to avoid reallocation costs on `push_back()`?

this is a good catch. I'll update

ckormanyos:
Here is the failing `cyl_bessel_j` code. I used theinner-type `value_type` to call the special function and it still fails unexpectedly.

I am trying to take the derivative with respect to $x$ of $J_{0.25}(x){\vert}_{1.25}$

```
```

demroz:
> Here is the failing `cyl_bessel_j` code. I used the inner-type `value_type` to call the special function and it still fails unexpectedly.
> 
> I am trying to take the derivative with respect to x of

That code breaks the autodiff chain before you even call `boost::math::cyl_bessel_j` with `static_cast<local_value_type>(x)`. You're casting the `rvar` type to a double

demroz:
> > My opinion: There are very legitimate reasons to want to autodiff through special functions. I've done this in the past for physical simulations that I want to optimize over. These situations aren't particularly common though. It isn't difficult per say to write specialized 'autodiff compatible' bindings to special functions, there is just a lot of them.
> 
> I think it might be a rather clear case for even stronger motivation.
> 
> Look at it from this perspective.
> 
> * I take the `autodiff` code off the rack
> * Ok great I want to make some gradients.
> * So I put it onto some random special-func in Boost and it fails.
> 
> So there's nothing, let's say, _magic_ about Boost.Math's special functions. So who's to say when and where the differentiation code will simply fail unexpectedly again. Maybe some clients write some perfeclty reasonable numerical code and `autodiff` just fails? I suspect this is the time to work out this kind of stuff, not later when the potential library is out in the wild.

Ok i can work on this, but it make a little while. Here is how I see this being implemented. We have a `.json` file (i dont actually have a filetype preference) with a list of "special functions." This file contains basically the name of the function, boost namespace function call and its derivative information, and maybe some sort of info about the test cases needed. We have some python scripts that generate this code: check [here](https://github.com/demroz/math/blob/add_special_function_overloads/include/boost/math/differentiation/detail/generate_expression_templated_headers.py) for a rough example that doesn't work, but kind of highlights the idea of what I'm talking about.

ckormanyos:
> Ok i can work on this, but it may take a little while. Here is how I see this being implemented. We have a .json file (i dont actually have a filetype preference) with a list of "special functions." This file contains basically the name of the function, boost namespace function call and its derivative information, and maybe some sort of info about the test cases needed. We have some python scripts that generate this code: 

I need to be rather direct. This does not seem like a good solution moving forward. It is overly complicated and non-permanent in the sense of difficult to maintain.

As a potential client of this new library, I simply want it to work for my code. I mention again, there is absolutely nothing special about Boost's special functions except the work special. It is just ordinary code.

I see two possible ways out of this pickle.

- Get the `autodiff` code working on any code
- Or tell us where and why it breaks so we can fix this together.

A lot of our clients write detailed code. It's going to be difficult when you get code out in the wild and have to say, well, I guess your client cod3 was just too complex for `autodiff`. Remember our clients are extremely demanding.

The first step is to find out where, for instance, `cyl_bessel_j` breaks the code. Why? Can it be fixed?

demroz:
> > Ok i can work on this, but it may take a little while. Here is how I see this being implemented. We have a .json file (i dont actually have a filetype preference) with a list of "special functions." This file contains basically the name of the function, boost namespace function call and its derivative information, and maybe some sort of info about the test cases needed. We have some python scripts that generate this code:
> 
> I need to be rather direct. This does not seem like a good solution moving forward. It is overly complicated and non-permanent in the sense of difficult to maintain.
> 
> As a potential client of this new library, I simply want it to work for my code. I mention again, there is absolutely nothing special about Boost's special functions except the word special. Other than that, it is just ordinary, vanilla template code. This kind of code could come from literally _anywhere_.
> 
> I see two possible ways out of this pickle.
> 
> * Get the `autodiff` code working on any code
> * Or tell us where and why it breaks so we can fix this together.
> 
> A lot of our clients write detailed code. It's going to be difficult when you get code out in the wild and have to say, well, I guess your client cod3 was just too complex for `autodiff`. Remember our clients are extremely demanding.
> 
> The first step is to find out where, for instance, `cyl_bessel_j` breaks the code. Why? Can it be fixed?

So I raised this issue to @mborland in #1281.  I just looked at the implementation of `cyl_bessel_j`. The first point of failure is: 
```cpp
template <class T, class Policy>
BOOST_MATH_GPU_ENABLED inline T asymptotic_bessel_j_large_x_2(T v, T x, const Policy& pol)
{
   // See A&S 9.2.19.
   BOOST_MATH_STD_USING
   // Get the phase and amplitude:
   T ampl = asymptotic_bessel_amplitude(v, x);
   if (0 == ampl) 
      return ampl;  // shortcut.
   T phase = asymptotic_bessel_phase_mx(v, x);
   BOOST_MATH_INSTRUMENT_VARIABLE(ampl);
   BOOST_MATH_INSTRUMENT_VARIABLE(phase);
   //
   // Calculate the sine of the phase, using
   // sine/cosine addition rules to factor in
   // the x - PI(v/2 + 1/4) term not added to the
   // phase when we calculated it.
   //
   BOOST_MATH_INSTRUMENT_CODE(cos(phase));
   BOOST_MATH_INSTRUMENT_CODE(cos(x));
   BOOST_MATH_INSTRUMENT_CODE(sin(phase));
   BOOST_MATH_INSTRUMENT_CODE(sin(x));
   T cx = cos(x);
   T sx = sin(x);
   T ci = boost::math::cos_pi(v / 2 + 0.25f, pol); // <- right here 
   T si = boost::math::sin_pi(v / 2 + 0.25f, pol);
   T sin_phase = cos(phase) * (cx * ci + sx * si) - sin(phase) * (sx * ci - cx * si);
   BOOST_MATH_INSTRUMENT_VARIABLE(sin_phase);
   return sin_phase * ampl;
}
```

`boost::math::cos_pi` is no longer called on `rvar` but on `add_expr<div_expr<rvar,float>,float>`. `T` in this case is `rvar`, inside `boost::math:cos_pi` there are a few places that don't work with expression templates the way I'm using them. Looking at the code for `cos_pi`
```cpp
BOOST_MATH_GPU_ENABLED T cos_pi_imp(T x, const Policy&)
{
   BOOST_MATH_STD_USING // ADL of std names
   // cos of pi*x:
   bool invert = false;
   if(fabs(x) < T(0.25)) // <- non rvar expressions are not numeric type constructible.
      return cos(constants::pi<T>() * x); // <- this may be ok if it calls boost::math::differentiation::cos

   if(x < 0)
   {
      x = -x; //<- cannot work with an expression template because the type of x mutates
   }
   T rem = floor(x);
   if(abs(floor(rem/2)*2 - rem) > boost::math::numeric_limits<T>::epsilon())
   {
      invert = !invert;
   }
   rem = x - rem;
   if(rem > 0.5f)
   {
      rem = 1 - rem;
      invert = !invert;
   }
   if(rem == 0.5f)
      return 0;
   
   if(rem > 0.25f)
   {
      rem = 0.5f - rem;
      rem = sin(constants::pi<T>() * rem);
   }
   else
      rem = cos(constants::pi<T>() * rem);
   return invert ? T(-rem) : rem;
}
``` 
There are basically two (maybe 3 now that I'm thinking about it) ways out:
1. What I proposed above. provide overloads to special functions for the "expression" base type.
2. Take the performance hit. Ditch expression templates. Make `rvar` non-mutating. This results in longer tapes. I haven't bench-marked ET vs no-et personally. I've read from some sources its 2-3x for typical use cases.
3. I dont know if this is possible, but i can't think of a reason it shouldn't. Manually turn off expression templates for special functions. This makes the library more complicated to use though

mborland:
> But for the moment, I'd recommend you add an active template parameter for `et_on`/`et_off` and we document that with great clarity. We can evolve beyond that point in the future. I always say, evolution, not revolution.
> 
> 
> 
> Cc: @mborland 

Do you think that should wait for a separate PR? We're so early in the boost cycle that I think that can be added in a next go. I'd stay with what we have for now.

ckormanyos:
>> But for the moment, I'd recommend you add an active template parameter for `et_on`/`et_off`

> Do you think that should wait for a separate PR? 

Matt (@mborland) as so often you read my mind.

Yes, in the sense of evolution. Go with what you/we have tested now. Document the state of affairs. Then we can see if we get the next iteration into 1.90 or not.

I like how you think.

jzmaddock:
First off, my apologies for being AWOL, I'll try to look at this in due course... just a couple of quick observations:

* specializing/binding every special function seems like a fools errand to me, there may be a few that have nice identities for the derivative where it's more efficient to take advantage of that, but we're almost moving into symbolic math territory, and that's a whole lot of work!
* In the comment: "another good example is whenever there is an internal call to a special function doing something like this: `sin_pi(2.0*x);` ", that should never happen, we do test again multiprecision using expression templates, so we *should* be expression template safe.  It's possible that since you're operating at say `double` precision that you're taking a code branch we haven't tested against expression-templated UDT's yet, but that would be a bug at our end.  Being able to use a tool like this against arbitrary special functions seems like a good win to me.  But:
* Mention was made of `tgamma(5)` resulting in zero derivative: as pointed out, this is down to integer arguments being treated as constant results.  It's the correct (and fast) thing to do in general, but obviously breaks a tool like this.  The right thing to do here is to specialize the lowest level function that blows up: `unchecked_factorial` in this case.  That function is used all over (not just in tgamma) so fixing that would fix a whole bunch of things in one go.  But:
* There are many other special cases where we return a constant result: usually this is at the end point of a domain, or else say at `x = 0`, where the result is known, and might not be easily calculated by the normal methods anyway.  Again for normal scalar types it's important to filter those special cases off early, but they will potentially break an autodiff tool.  These are scattered all over, and I'm not sure how we fix them.  Possibly a trait for detecting types like this, and then some constexpr if's around problematic code.
* And finally, running any new number type through our special functions is a *really* good test of the correctness of the implementation: the type will really get hammered into all kind of corner cases that you wouldn't find otherwise, and has been invaluable in testing our multiprecision types.

demroz:
@ckormanyos Some good news on the expression template on/off macro. I have a working version under the [add_no_expression_template_support](https://github.com/demroz/math/tree/add_no_expression_template_support/) branch. With the change 
```cpp
#define BOOST_MATH_ET_OFF
#include <boost/math/differentiation/autodiff_reverse.hpp>
#include <boost/math/special_functions/bessel.hpp>
#include <boost/math/special_functions/gamma.hpp>
#include <iostream>

namespace rdiff = boost::math::differentiation::reverse_mode;

int main()
{
    constexpr std::size_t N = 1;
    using rvar              = rdiff::rvar<double, N>;
    double x                = 2.1;
    rvar   x_ad             = x;

    auto g = boost::math::tgamma(x_ad + x_ad);
    auto j = boost::math::cyl_bessel_j(0.25, x_ad + x_ad / 2);

    std::cout << "tgamma(x + x) = " << g << "\n";
    std::cout << "J_0.25(x) = " << j << "\n";

    auto &tape = rdiff::get_active_tape<double, 1>();

    g.backward();
    std::cout << "d/dx tgamma(x+x), autodiff = " << x_ad.adjoint()
              << " expected = " << 2 * boost::math::tgamma(2 * x) * boost::math::digamma(2 * x)
              << std::endl;

    tape.zero_grad();
    j.backward();

    std::cout << "d/dx J_0.25(x+x/2), autodiff = " << x_ad.adjoint() << " expected = "
              << 3.0 / 4.0
                     * (boost::math::cyl_bessel_j(-0.75, 3. * x / 2.)
                        - boost::math::cyl_bessel_j(1.25, 3. * x / 2.))
              << std::endl;

    return 0;
}

```

produces the expected gradients. `tgamma` on integer values is still problematic however

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -0,0 +1,644 @@
...
+    
+    }  // namespace reverse_mode
+    }  // namespace differentiation
+    }  // namespace math
+    }  // namespace boost
+
+[h1:description Description]
+
+Reverse mode autodiff is a header-only C++ library the [@https://en.wikipedia.org/wiki/Automatic_differentiation
+automatic differentiation] (reverse mode) of mathematical functions. This implementation builds a computational graph known as a tape, which sotres all operations and their corresponding derivatives. The total gradients are then computed by reverse accumulation via the chain rule.
```
mborland:
```suggestion
automatic differentiation] (reverse mode) of mathematical functions. This implementation builds a computational graph known as a tape, which stores all operations and their corresponding derivatives. The total gradients are then computed by reverse accumulation via the chain rule.
```

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -0,0 +1,644 @@
...
+    #include <boost/math/differentiation/autodiff_reverse.hpp>
+    
+    namespace boost {
+    namespace math {
+    namespace differentiation {
+    namespace reverse_mode {
+
+    /* autodiff variable of type RealType (numeric types), stores derivatives up to DerivativeOrder*/
+    template<typename RealType, size_t DerivativeOrder = 1>
+    class rvar : public expression<RealType, DerivativeOrder, rvar<RealType, DerivativeOrder>> {
```
mborland, line 26: I would add a brief explanation of what `expression` is since it's in the documented signature.

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -0,0 +1,644 @@
...
+      // inner_t of rvar<RealType, N> = var<RealType, N-1>
+      // used to store graphs of N-1 order derivatives
+      // rvar_t<RealType, 0> decays to RealType
+      using inner_t = rvar_t<RealType, DerivativeOrder - 1>;
+      /* constructors */
+      // default
+      rvar();
+      // construct from numeric type
+      rar(const RealType value);
```
mborland:
```suggestion
      rvar(const RealType value);
```

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -0,0 +1,644 @@
...
+      rvar<RealType, DerivativeOrder> &operator+=(const expression<RealType, DerivativeOrder, E> &expr);
+
+      template<class E>
+      rvar<RealType, DerivativeOrder> &operator*=(const expression<RealType, DerivativeOrder, E> &expr)
+
+
+    }
+
+    // gradient tape holds the computational graph
+    template<typename RealType, size_t DerivativeOrder, size_t buffer_size = BOOST_MATH_BUFFER_SIZE>
```
mborland, line 74: Add a definition of what `BOOST_MATH_BUFFER_SIZE` is prior to use.

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -0,0 +1,644 @@
...
+      // rewinds tape to last checkpoint set
+      void rewind_to_last_checkpoint();
+      // index is "checkpoint" index. so                                              // order which checkpoint was set
+      void rewind_to_checkpoint_at(size_t index);
+      // rewind to beginning of graph without clearing the tape
+      void rewind();
+    }
+
+    // tape access
+    template<typenametemplate<typename RealType, size_t DerivativeOrder>
```
mborland:
```suggestion
    template<typename template<typename RealType, size_t DerivativeOrder>>
```

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -0,0 +1,644 @@
...
+[:[:[autodiff_graph autodiff_reverse_graph.svg]]]
+
+Using reverse mode autodiff, the gradients are computed by traversing the graph backward:
+[:[:[autodiff_equation reverse_mode_autodiff_ex_eq.svg]]]
+
+Some key points about reverse mode automatic differentiation:
+
+1. Reverse mode auto-diff is exceptionally efficient for computing the gradient of functions mapping from high to low dimensional space, f : [real][super n][rarr][real]. Unlike finite differences or forward mode autodiff which scale with the number of input variables, reverse mode autodiff calculates the entire gradient vector in time proportional to the original function's evaluation time.
+
+2. It is possible to compute higher order derivatives by applying reverse mode over reverse mode differentiation. The backward function builds a new computational tape over the gradient computation. This approach is conceptually sound, but it can become computationally expensive very quickly. Forward mode autodiff is generally preferrable when computing higher order derivatives.
```
mborland: Over forward mode?

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -0,0 +1,644 @@
...
+
+      // 4. Initial forward pass and loss calculation
+      auto y_fit = model(a, b, noisy_data_x);
+      rvar<double, 1> loss_v = loss(noisy_data_y, y_fit);
+
+The model and loss functions are called. This is just to initialize y_fit and loss_v to be used inside the while loop.
+      // 5. Gradient Descent Loop
+      double learning_rate = 1e-3;
+
+The learning rate is a tunable parameter determining the "velocity" with which we descend to a solution.
```
mborland: I would add in an explanation of why we wouldn't want a higher learning rate, and maybe some recommended values or range of values.

doc/differentiation/autodiff_reverse.qbk
demroz: No. I apply reverse mode over reverse mode for higher order derivatives.

mborland: I would add that explanation. As written "applying reverse mode over reverse mode differentiation" reads awkwardly.

demroz: updated

mborland: Much better. Thanks

demroz:
Is there anything else needed from my end to get this merged?

CMakeLists.txt
mborland, line 1: This needs to be put back. You can copy and paste the one that's currently on develop

ckormanyos:
> I'll merge this in and then you can do more targeted feature branches.

Thanks @demroz nice work. I'm excited to see how this area of Math evolves.

Thanks also Matt and John.

PR #1304, opened 2025-08-17
Try repair or remove two JSO tests

ckormanyos:
Based on discussion [here]8https://github.com/boostorg/math/pull/1301#issuecomment-3193815058).

### DIFF
--- a/include/boost/math/optimization/jso.hpp
+++ b/include/boost/math/optimization/jso.hpp
@@ -389,7 +389,7 @@ jso(const Func cost_function, jso_parameters<ArgumentContainer> &jso_params,
         throw std::logic_error(oss.str());
       }
       for (size_t i = 0; i < weights.size(); ++i) {
-        weights[i] = delta_f[i] / delta_sum;
+        weights[i] = static_cast<DimensionlessReal>(delta_f[i] / delta_sum);
       }
 
       M_CR[k] = detail::weighted_lehmer_mean(S_CR, weights);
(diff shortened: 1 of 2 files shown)


ckormanyos:
As a first step, explicitly activate precise floating-point for the runners.

ckormanyos:
Older GCC compilers had a problem with one particular test-optimizer that did not explicitly name the trailing return type. This has been fixed.

Let's see if all the runners now work?

ckormanyos:
Intermediate checks [at Godbolt](https://godbolt.org/z/xTrcKfcs3) on the branch.

ckormanyos:
Ok it's coming down to the wire with jSO. In the last run, curiously enough, the only failure was on GCC8 (having C++17, as required for `test_jso.cpp`). This on our CI, even though GCC 8 passed at Godbolt.

So I'll disable _ONLY_ that compiler and see if there is any other random unexpected stuff on Drone.

ckormanyos:
So this one runs green now across the board.

- Initially, `jso_test` had limitations on GCC and only rean in its entirety on MSVC and clang.
- It needed a little syntax clarification for old GCC compilers.
- Now the only exclusion for `jso_test.cpp` is GCC8 on Drone for the original two lines in question.

So we have a lot more test coverage of `jso_test.cpp`.

I'll push this one now and we should keep a watchful eye for a bit to ensure that jSO remains well-behaved in CI/CD.

Cc: @NAThompson and @jzmaddock and @mborland

PR #1306, opened 2025-08-22
Handle more conversion warnings

ckormanyos:
In this PR we handle a few more trivial conversion warnings in new and old code

### DIFF
--- a/test/cma_es_test.cpp
+++ b/test/cma_es_test.cpp
@@ -146,18 +146,6 @@ void test_beale() {
   CHECK_ABSOLUTE_ERROR(Real(1)/Real(2), local_minima[1], Real(0.1));
 }
 
-#if BOOST_MATH_TEST_UNITS_COMPATIBILITY
-void test_dimensioned_sphere() {
-  std::cout << "Testing CMA-ES on dimensioned sphere . . .\n";
-  using ArgType = std::vector<quantity<length>>;
-  auto params = cma_es_parameters<ArgType>();
-  params.lower_bounds.resize(4, -1.0*meter);
-  params.upper_bounds.resize(4, 1*meter);
-  std::mt19937_64 gen(56789);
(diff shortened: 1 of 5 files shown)


ckormanyos:
Hi folks, I would like an independent advice on this particular PR.

I started out with some trivial syntax changes to remove warnings. The areas in the scope of changes included Bessel-J and Lambert-W implementation(s) and test(s).

While running CI, however, some [seemingly unrelated compilation issues](https://github.com/boostorg/math/actions/runs/17149309401/job/48651710121) arose on `__APPLE__` when executing dustribution tests with `Boost.Units`. So I was drawn to [this comment](https://github.com/boostorg/math/blob/e4f1d82abe7bd70dc0f1c298448ff44d471099e5/test/test_functions_for_optimization.hpp#L13). I empirically found that the tests activated with `#define BOOST_MATH_TEST_UNITS_COMPATIBILITY 1` have (as of today-ish) also begun to fail on `__APPLE__`. Three lines later, you can see my patch/hack of `#if 0`, which deactivates these tests everywhere. CI is now running again.

I was wondering if this discouraging discovery could be verified or investigated by another person? It is always good to keep maximum scope of testing. But I've come to the conclusion that minute changes on the GHA runner now cause these tests to fail also on `__APPLE__`. Can any alternate tests (that are maybe more friendly to compilers) be made with modest effort?

Cc: @mborland and @NAThompson and @jzmaddock

mborland:
I am able to replicate the error locally on my Mac. Looks like there's a static assert inside the STL now to check for a standard type being used for `uniform_real_distribution`. I think the workaround here would be to use `boost::random::uniform_real_distribution` when not in standalone mode. That should capture this use case anyway because Boost.Units does not support it's own standalone.

ckormanyos:
Thanks @NAThompson and @mborland.

Merged

PR #1307, opened 2025-08-22
Update mingw failures to end at GCC-15

mborland:
Got a report on the mailing list that a user hit this configuration error. Checked our links again, and it looks like the bug has finally been fixed in 10/24

### DIFF
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -655,7 +655,8 @@ namespace boost{ namespace math{
 //
 // When running windows-2022 or 2025 we see this issue again with GCC 12 and 14
 //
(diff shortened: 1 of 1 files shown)


ckormanyos:
I hopefully just merged the relevant changes in and fired up CI. Let's see if it goes green

ckormanyos:
> it looks like the bug has finally been fixed in 10/24

Now I see what you mean. That fix took a while. Right on, it's getting better.

mborland:
@jzmaddock Any chance you can locally test this? There's no runners in CI that have a high enough version of GCC yet.

ckormanyos:
image

It looks like pacman for UCRT64 offers GCC 13 at the moment. That's too old right?

I wonder if we could set up a run on Godbolt?

mborland:
Well despite what we thought this still fails John's threading sanity check. Maybe next time.

PR #1313, opened 2025-08-27
Feature: add enable/disable macro for expression templates

demroz:
This feature adds a `BOOST_MATH_REVERSE_MODE_ET_ON` and `BOOST_MATH_REVERSE_MODE_ET_OFF` macro to enable/disable expression templates for the reverse mode autodiff library, and adds some documentation regarding this feature.

This is also probably a good place to start discussions about the future of the library. Adding special function support is most likely the next step, and I'd love some help and advice with that. I'm guessing `tgamma` and `cyl_bessel_j` are good places to start.

The other place that I'd like to take this library is adding some optimization routines like gradient descent, SGD, Adam, LBFGS, etc...


mborland:
I will take a look tomorrow morning and leave comments

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -572,7 +577,17 @@ When in doubt, its always preferrable to compute higher order derivatives with:
...
-Reverse mode autodiff is expression template based. This means that some care has to be taken into considerations when writing code that uses this library. For example consider the code below:
+Reverse mode autodiff is expression template based by default. If you'd like to disable expression templates (for debugging, benchmarking, or simplify integration with other systems), you must define the following macro at compile time:
+
+    #define BOOST_MATH_REVERSE_MODE_ET_OFF
+
+or pass it to the compiler via flag
+
+    g++ -DBOOST_MATH_REVERSE_MODE_ET_OFF
+
+The expression templated version of this code doesn't currently interact nicely with the Boost.Math special function implementations, and will throw compile time errors. Disabling expression templates will often fix this, however various special functions are implemented in a way that breaks the automatic differentiation chain for certain values. Complete special function support may be added in the future.
```
mborland: I think it would be useful to have a table here that lists the special functions that expression templates supports, rather than the user having to experiment on seeing what compiles.

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_ops_et.hpp
```diff
@@ -0,0 +1,160 @@
...
+         typename = typename std::enable_if<!detail::is_expression<RealType2>::value>::type>
+add_const_expr<RealType1, DerivativeOrder, ARG> operator-(
+    const expression<RealType1, DerivativeOrder, ARG> &arg, const RealType2 &v)
+{
+    /* rvar - float = rvar + (-float) */
+    return add_const_expr<RealType1, DerivativeOrder, ARG>(arg, static_cast<RealType1>(-v));
+}
+
+/** @brief
+ *   subtraction float - rvar
```
mborland, line 111: Does the current testing include C++23's `<stdfloat>` types? If not I think they should be added. It has its own special rules like narrowing is a compilation error rather than a warning.

include/boost/math/differentiation/detail/reverse_mode_autodiff_stl_et.hpp
```diff
@@ -0,0 +1,275 @@
...
+    BOOST_MATH_STD_USING
+    frexp(arg.evaluate(), i);
+    return arg / pow(static_cast<RealType>(2.0), *i);
+}
+
+template<typename RealType, size_t DerivativeOrder, typename ARG>
+auto ldexp(const expression<RealType, DerivativeOrder, ARG> &arg, const int &i)
+{
+    BOOST_MATH_STD_USING
+    return arg * pow(static_cast<RealType>(2.0), i);
```
mborland, line 92: Does the case of `HUGE_VAL` need to be handled here or does the expression template value naturally handle that?

include/boost/math/differentiation/autodiff_reverse.hpp
```diff
@@ -6,12 +6,23 @@
 #define BOOST_MATH_DIFFERENTIATION_AUTODIFF_HPP
 
 #include <boost/math/constants/constants.hpp>
-#include <boost/math/differentiation/detail/reverse_mode_autodiff_basic_operator_overloads.hpp>
+
+#ifndef BOOST_MATH_REVERSE_MODE_ET_OFF
+#define BOOST_MATH_REVERSE_MODE_ET_ON
+#endif
+
```
mborland, line 17: It might be worth adding a check here that there's no crazy configuration problems:

```
#if defined(BOOST_MATH_REVERSE_MODE_ET_OFF) && defined(BOOST_MATH_REVERSE_MODE_ET_ON)
#  error "BOOST_MATH_REVERSE_MODE_ET_OFF and BOOST_MATH_REVERSE_MODE_ET_ON are mutually exclusive"
#endif
```

doc/differentiation/autodiff_reverse.qbk
demroz: That's fair. Is there any way to add tests that aren't expected to compile without breaking the entire process? If not I can test locally

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_ops_et.hpp
demroz: The `test_autodiff_reverse.hpp`  is essentially a slightly modified copy of the forward mode version. It has a macro that checks for `__STDCPP_FLOAT_*_T__`
[here](https://github.com/boostorg/math/blob/ab265d474f2f1996610df283e71a63c7a7eb1cad/test/test_autodiff_reverse.hpp#L36). I believe this is what youre asking about, I've never really worked with C++23's `<stdfloat>` types

doc/differentiation/autodiff_reverse.qbk
mborland: Yes. In addition to `run` in the Jamfile you have the options of `run-fail`, `compile`, and `compile-fail`. Example is here: https://github.com/boostorg/math/blob/develop/test/Jamfile.v2#L1629.

In the build log it'll show `failed-as-expected`

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_ops_et.hpp
mborland: Yes, but you'll need to add this to the top:

```
#if __has_include(<stdfloat>)
#  include <stdfloat>
#endif
```

Since the macros are defined inside that header instead of globally.

include/boost/math/differentiation/detail/reverse_mode_autodiff_stl_et.hpp
demroz: It naturally handles huge values correctly. I added an explicit test case for this though. Works as expected

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_ops_et.hpp
demroz: added these tests

include/boost/math/differentiation/autodiff_reverse.hpp
demroz: added

mborland: So this commit also changed the default behavior. Instead of if not explicitly ET_OFF then ET_ON, to if not explicitly ET_ON then ET_OFF (via fall through). Should ET_ON be the default for performance reasons? Or should ET_OFF be the default since it has broader compatibility. I would say ET_ON if we tell the user what works with it, and how to turn it off (which are both already covered in docs or previous comments)

@ckormanyos and @jzmaddock in case you guys have an opinion on the matter.

demroz: My fault. I didn't think too hard about about your comment and just copy and pasted that piece of code. I think ET should be enabled ON by default. At least from personal experience its what makes the most sense to me. If there is no objection from @ckormanyos and @jzmaddock I'll fix it in my next commit.

ckormanyos: >  I think ET should be enabled ON by default. At least from personal experience its what makes the most sense to me.

It also makes sense to me.

include/boost/math/differentiation/detail/reverse_mode_autodiff_basic_ops_et.hpp
ckormanyos: > es, but you'll need to add this to the top:

Do we still need to query `__has_include` itself?

```
#if defined(__has_include)
#if __has_include(<stdfloat>)
#  include <stdfloat>
#endif
#endif
```

mborland: No; we don't support any compiler that lacks `__has_include`.

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -572,7 +705,19 @@ When in doubt, its always preferrable to compute higher order derivatives with:
   auto g4 = grad_nd<4>(f, &x,&y,&z)
 
 [h2:expression-templates-and-auto Expression Templates and auto]
-Reverse mode autodiff is expression template based. This means that some care has to be taken into considerations when writing code that uses this library. For example consider the code below:
+Reverse mode autodiff is expression template based by default. If you'd like to disable expression templates (for debugging, benchmarking, or simplify integration with other systems), you must define the following macro at compile time:
```
mborland:
```suggestion
Reverse mode autodiff is expression template (`BOOST_MATH_REVERSE_MODE_ET_ON`) based by default. If you'd like to disable expression templates (for debugging, benchmarking, or simplify integration with other systems), you must define the following macro at compile time:
```

doc/differentiation/autodiff_reverse.qbk
```diff
@@ -572,7 +705,19 @@ When in doubt, its always preferrable to compute higher order derivatives with:
...
+
+    #define BOOST_MATH_REVERSE_MODE_ET_OFF
```
mborland, line 711: Maybe add an include of the library after this line? The important point is this has to come first

test/test_reverse_mode_autodiff_special_functions_compile/airy/test_reverse_mode_autodiff_airy_ai_compile_test_ET_OFF_cpp_types.cpp
```diff
@@ -0,0 +1,21 @@
+#define BOOST_MATH_REVERSE_MODE_ET_OFF
```
mborland, line 9: The doc runner in CI is failing with the message that all of these auto generated files needs a license and copyright at the top. I think it would be worth adding a line at the top as well that says these are auto-generated do no manually edit them.

mborland:
I actually think right now none of your stuff is actually cycling in the CI. I see `test-suite autodiff_reverse_specific` and `test-suite "test_reverse_mode_autodiff"`. I would consolidate these into 1 (and remove the quotes) for what you already have, and at the bottom of the file add `test_reverse_mode_autodiff` next to `autodiff` on the `github_ci_block_2` line. Then create a `test-suite autodiff-long-running-tests` with everything from this PR. We can add a few choice CI runners that only run that test-suite since we know it's lengthy. We do the same thing in Multiprecision with a test suite named long running tests.

doc/differentiation/autodiff_reverse.qbk
demroz: added

doc/differentiation/autodiff_reverse.qbk
demroz: added

test/test_reverse_mode_autodiff_special_functions_compile/airy/test_reverse_mode_autodiff_airy_ai_compile_test_ET_OFF_cpp_types.cpp
demroz: added

demroz:
@mborland so I added the fixes you asked for and fixed the Jamfile. Seems like everything is running at this point.

mborland:
Ok. I added a commit to run the long tests on our Drone CI. I tried to push it onto your branch, but since it is named develop it accidentally all got merged into boostorg:develop at the same time.

mborland:
> Ok I'll stop submitting PRs from the develop branch to avoid this in the future.

Thanks. Definitely a case of operator error though. Looks like you have tight tolerances that may be prone to random failures: https://github.com/boostorg/math/actions/runs/17607913028/job/50022812424#step:17:877

mborland:
The `long-running` tests are cycling in under an hour which is good, but they're showing a number of failures: https://drone.cpp.al/boostorg/math/2396/669/2

demroz:
> The `long-running` tests are cycling in under an hour which is good, but they're showing a number of failures: https://drone.cpp.al/boostorg/math/2396/669/2

looks like some of the function arguments didn't get parsed properly. I'll fix this evening

PR #1315, opened 2025-08-29
Repair neg-zero and tiny-arg erf and erfc

ckormanyos:


include/boost/math/special_functions/erf.hpp
```diff
@@ -153,51 +153,70 @@ T erf_imp(T z, bool invert, const Policy& pol, const Tag& t)
...
 
-   if(z < 0)
+   if(fabs(z) < tools::root_epsilon<T>())
+   {
+      // Series[Erf[x], {x, 0, 4}]
+      // Series[Erfc[x], {x, 0, 4}]
+
+      const T term2 { 2 * z / constants::root_pi<T>() };
+
+      return invert ? 1 - term2 : term2;
```
jzmaddock, line 163: Well done, I realized when I was lying in bed last night that `1` on it's own didn't quite cut it!

ckormanyos: It was a great review john. All you had to do was get me thinking. Thx.

ckormanyos: Actually John (@jzmaddock) in my new code, I think the final path for identically zero is not needed. That will be picked up by the first path, right?

ckormanyos: But then for the case of identically zoro, we would compute term, which is a waste of time.

Is this a case for `(boost::math::tools::fpclassify)(z)` since you check for NaN also?

ckormanyos: I am moving toward this implementation locally. I will also throw in a few more edge-case Multiprecision tests locally and wrap this one up shortly.

```
template <class T, class Policy, class Tag>
T erf_imp(T z, bool invert, const Policy& pol, const Tag& t)
{
   BOOST_MATH_STD_USING

   BOOST_MATH_INSTRUMENT_CODE("Generic erf_imp called");

   if ((boost::math::isnan)(z))
      return policies::raise_domain_error("boost::math::erf<%1%>(%1%)", "Expected a finite argument but got %1%", z, pol);

   const T abs_z { fabs(z) };

   const bool signbit_result = ((boost::math::signbit)(z) != 0);

   if (abs_z < tools::root_epsilon<T>())
   {
      if (abs_z < tools::epsilon<T>())
      {
         return invert ? T(1) : (!signbit_result) ? T(0) : -T(0);
      }

      // Series[Erf[x], {x, 0, 4}]
      // Series[Erfc[x], {x, 0, 4}]

      const T term2 { z * 2 / constants::root_pi<T>() };

      return invert ? 1 - term2 : term2;
   }

   if (signbit_result)
   {
      if(!invert)
         return -erf_imp(T(-z), invert, pol, t);
      else
         return 1 + erf_imp(T(-z), false, pol, t);
   }

   T result;

   if(!invert && (z > detail::erf_asymptotic_limit<T, Policy>()))
   {
      detail::erf_asympt_series_t<T> s(z);
      std::uintmax_t max_iter = policies::get_max_series_iterations<Policy>();
      result = boost::math::tools::sum_series(s, policies::get_epsilon<T, Policy>(), max_iter, 1);
      policies::check_series_iterations<T>("boost::math::erf<%1%>(%1%, %1%)", max_iter, pol);
   }
   else
   {
      T x = z * z;
      if(z < 1.3f)
      {
         // Compute P:
         // This is actually good for z p to 2 or so, but the cutoff given seems
         // to be the best compromise.  Performance wise, this is way quicker than anything else...
         result = erf_series_near_zero_sum(z, pol);
      }
      else if(x > 1 / tools::epsilon<T>())
      {
         // http://functions.wolfram.com/06.27.06.0006.02
         invert = !invert;
         result = exp(-x) / (constants::root_pi<T>() * z);
      }
      else
      {
         // Compute Q:
         invert = !invert;
         result = z * exp(-x);
         result /= boost::math::constants::root_pi<T>();
         result *= upper_gamma_fraction(T(0.5f), x, policies::get_epsilon<T, Policy>());
      }
   }

   return ((!invert) ? result : 1 - result);
}
```

ckormanyos: This is the _generic_ implementation of `erf`/`erfc` in `erf_imp`. So it is used for MP types.

PR #1328, opened 2025-10-17
Gradient optimizers

demroz:
adds a few gradient optimizers
- gradient descent
- nesterov
- and L-BFGS

everything is policy based, so these should be quite extensible. I may add a few more optimziers in the future, but these are the main ones i think. If there are any specific ones that you guys think should be added, I'd be happy to do so. Also although everything is reverse-mode autodiff centric, as long as you provide the objective function, a way to evaluate it, and a way to evaluate the gradient, everything should work correctly.

For some examples on how to use the optimziers: `test_gradient_descent_optimizer.cpp`, `test_nesterov_optimizer.cpp`, and `test_lbfgs.cpp` should be good starting points.

I'm working on the documentation currently. I wanted to hold off to see if any major revisions are necessary

### DIFF
--- a/doc/optimization/gradient_optimizers.qbk
+++ b/doc/optimization/gradient_optimizers.qbk
@@ -13,6 +13,70 @@ LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 [section:introduction Introduction]
 Gradient based optimizers are algorithms that use the gradient of a function to iteratively find locally extreme points of functions over a set of parameters. This sections provides a description of a set of gradient optimizers. The optimizers are written with `boost::math::differentiation::reverse_mode::rvar` in mind, however if a way to evaluate the funciton and its gradient is provided, the optimizers should work in exactly the same way.
 
+Below is a table that summarizes the intended usage patterns of the provided optimizers and policies, and is meant as a practical guide rather than a strict prescription:
+
+[h1:table-optimizers List of Optimizers]
+[table
+     [[Optimizer] [Order] [Uses Curvature] [Memory Cost] [Intended Problem Class] [When to Use]]
+     [[gradient descent] [first] [no] [low] [Smooth, well-scaled objectives] [Baseline method; debugging; when behavior transparency matters]]
+     [[nesterov accelerated gradient] [first] [no] [low] [Ill-conditioned or narrow-valley problems] [When plain gradient descent converges slowly or oscillates]]
+     [[L-BFGS] [quasi second order] [approximate] [medium] [Smooth, deterministic objectives] [When gradients are reliable and faster convergence is needed]]
+]
+
+[h1:table-optimizer-policies Optimizer Policies]
+[heading Initialization Policies]
+[table
+     [[Policy] [Use case] [Responsibilities]]
+     [[tape_initializer_rvar] [User initialzes all varibles manually] [initializes tape]]
+     [[random_uniform_initializer_rvar] [Initializes all variables with a random number between a min and max value] [Initializes variables. Initializes tape.]]
+     [[costant_initializer_rvar] [Initializes all variables with a constant] [Initializes variables. Initializes tape.]]
+]
+
+[heading Evaluation Policies]
+[table
+     [[Policy] [Use case] [Responsibilities]]
+     [[reverse_mode_function_eval_policy] [Default. User with boost reverse mode autodiff] [tells the optimizer how to evaluate the objective]]
+     [[reverse_mode_gradient_evaluation_policy] [Default. User with boost reverse mode autodiff] [tells the optimizer how to evaluate the gradients of an objective]]
+]
+
+These policies are intended to use with boost reverse mode autodiff. If you need to use the optimizers with a custom AD variable, or by providing the gradient of an objective manually, check the docs for policies to see how the policies are implemented.
+
+[h1:line-search-policies LBFGS line search policies]
+the table below summarizes the two line search policies provided for use with LBFGS.
+[table
+     [[Policy] [Enforced Conditions] [Per iteration cost] [Convergence] [Use case]]
+     [[Strong Wolfe] [function decrease. curvature condition] [higher] [faster] [most of the time]]
+     [[Armijo] [function decrease only] [lower] [slower] [you know what you're doing]]
+]
+[h1:minimizer-policies Minimizer Policies]
+[heading Convergence Policies]
+[table
+      [[Policy] [Criterion] [When to Use]]
+      [[gradient_norm_convergence_policy] [gradient norm < tol] [Default. Stationarity based condition]]
+      [[objective_tol_convergence_policy] [absolute difference between objective steps is small] [Well-scaled objectives]]
+      [[relative_objective_tol_policy] [relative difference between objective steps is small] [Scale-invariant convergence]]
+      [[combined_convergence_policy] [logical combination OR] [you need a combination of convergence conditions]]
+]
+
+[heading Termination Policies]
+[table
+      [[Policy] [Controls] [When to Use]]
+      [[max_iter_termination_policy] [iteration count] [Hard safety bound (almost always recommended)]]
+      [[wallclock_termination_policy] [wall clock time] [benchmarking, real-time constraints]]
+]
+
+[heading Constraint and Projection Policies]
+[table
+      [[Policy] [Constraint Type]]
+      [[unconstrained_policy] [No constraint]] 
+      [[box_constraints] [upper/lower bound clip]]
+      [[nonnegativity_constraint] [set everything below 0, to 0]]
+      [[l2_ball_constraint] [ 2-norm(x) < r]]
+      [[l1_ball_constraint] [ 1-norm(x) < r]]
+      [[simplex_constraint] [Probability simplex]]
+      [[function_constraint] [custom user provided function wrapper]]
+      [[unit_sphere_constraint] [2-norm(x) = 1]]
+]
 [endsect] [/section:introduction]
 
 [section:gradient_descent Gradient Descent]
--- a/include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
+++ b/include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
@@ -5,7 +5,6 @@
 #ifndef BOOST_MATH_OPTIMIZATION_DETAIL_DIFFERENTIABLE_OPT_UTILITIES_HPP
 #define BOOST_MATH_OPTIMIZATION_DETAIL_DIFFERENTIABLE_OPT_UTILITIES_HPP
 #include <boost/math/differentiation/autodiff_reverse.hpp>
-#include <boost/random.hpp>
 #include <cmath>
 #include <random>
 #include <type_traits>
@@ -144,11 +143,11 @@ random_vector(size_t n)
   /** @brief> generates a random std::vector<RealType> of size n
    * using mt19937 algorithm
    */
-  static boost::random::mt19937 rng{ std::random_device{}() };
-  static boost::random::uniform_real_distribution<RealType> dist(0.0, 1.0);
+  static std::mt19937 rng{ std::random_device{}() };
+  static std::uniform_real_distribution<double> dist(0.0, 1.0);
 
   std::vector<RealType> result(n);
-  std::generate(result.begin(), result.end(), [&] { return dist(rng); });
+  std::generate(result.begin(), result.end(), [&] { return static_cast<RealType>(dist(rng)); });
   return result;
 }
 
--- a/include/boost/math/optimization/detail/rdiff_optimization_policies.hpp
+++ b/include/boost/math/optimization/detail/rdiff_optimization_policies.hpp
@@ -8,7 +8,6 @@
 #include <boost/math/differentiation/autodiff_reverse.hpp>
 #include <random>
 #include <type_traits>
-#include <boost/random.hpp>
 
 namespace boost {
 namespace math {
@@ -102,10 +101,10 @@ struct random_uniform_initializer_rvar
   template<class ArgumentContainer>
   void operator()(ArgumentContainer& x) const
   {
-    static boost::random::mt19937 gen{ std::random_device{}() };
-    static boost::random::uniform_real_distribution<RealType> dist(0.0, 1.0);
+    static std::mt19937 gen{ std::random_device{}() };
+    static std::uniform_real_distribution<double> dist(0.0, 1.0);
     for (auto& xi : x) {
-      xi = rdiff::rvar<RealType, 1>(dist(gen));
+      xi = rdiff::rvar<RealType, 1>(static_cast<RealType>(dist(gen)));
     }
     auto& tape = rdiff::get_active_tape<RealType, 1>();
     tape.add_checkpoint();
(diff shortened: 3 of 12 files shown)


demroz:
At this point the code is something I'm relatively happy with. I also updated the docs. Would be great if this could get reviewed cc: @mborland @NAThompson

include/boost/math/optimization/gradient_descent.hpp
```diff
@@ -0,0 +1,207 @@
+//           Copyright Maksym Zhelyenzyakov 2025-2026.
+// Distributed under the Boost Software License, Version 1.0.
+//      (See accompanying file LICENSE_1_0.txt or copy at
+//           https://www.boost.org/LICENSE_1_0.txt)
+#ifndef GRADIENT_DESCENT_HPP
+#define GRADIENT_DESCENT_HPP
+#include <boost/math/optimization/detail/differentiable_opt_utilties.hpp>
+#include <boost/math/optimization/detail/gradient_opt_base.hpp>
+#include <boost/math/optimization/detail/rdiff_optimization_policies.hpp>
+namespace rdiff = boost::math::differentiation::reverse_mode;
```
mborland:
```suggestion
```

This alias is global, so it will pollute the consumer's environment. Fortunately, you never use it.

include/boost/math/optimization/gradient_descent.hpp
```diff
@@ -0,0 +1,207 @@
...
+    : lr_(lr) {};
+
+  template<typename ArgumentType,
+           typename = typename std::enable_if<
+             boost::math::differentiation::reverse_mode::detail::is_expression<
+               ArgumentType>::value>::type>
+  void operator()(ArgumentType& x, RealType& g)
+  {
+    // this update effectively "mutes" the tape
+    // TODO: add a tape scope guard method so that
```
mborland: Is this something that needs to be done before merge?

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
```diff
@@ -0,0 +1,166 @@
...
+         typename RealType,
+         std::size_t N,
+         typename... Args>
+struct argument_container_t<Container<rdiff::rvar<RealType, N>, Args...>>
+{
+    using type = RealType;
+};
+/******************************************************************************/
+/** @brief simple blas helpers
+ * may optimize later if benchmarks show its needed, or just switch to Eigen
```
mborland: I think it would be better to decide now whether this will need Eigen support or not. I think adding an external dependency after initial release is not ideal. If need be you could put the Eigen version behind a macro like:

`BOOST_MATH_OPTIMIZATION_USE_EIGEN`

Which will allow users to decide if they want to use Eigen, or just rely on the STL. I'm not sure how much extra work that would be.

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
```diff
@@ -0,0 +1,166 @@
...
+   * using mt19937 algorithm
+   */
+
+  /** TODO: these may need to be marked thread local
+   * in the future
+   *
+   * TODO: benchmark.
+   */
+  static boost::random::mt19937 rng{ std::random_device{}() };
+  static boost::random::uniform_real_distribution<RealType> dist(0.0, 1.0);
```
mborland: Is Boost.Math required here instead of using `<random>`? I ask because this would break the ability to be used as part of our standalone package.

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
```diff
@@ -0,0 +1,166 @@
...
+/******************************************************************************/
+template<typename RealType>
+std::vector<RealType>
+random_vector(size_t n)
+{
+  /** @brief> generates a random std::vector<RealType> of size n
+   * using mt19937 algorithm
+   */
+
+  /** TODO: these may need to be marked thread local
```
mborland: Does this need to be completed pre-merge? It will may cause you issues with MinGW if you make this change. See: https://github.com/boostorg/math/blob/develop/include/boost/math/tools/config.hpp#L657

include/boost/math/optimization/detail/gradient_opt_base.hpp
```diff
@@ -0,0 +1,92 @@
+//           Copyright Maksym Zhelyenzyakov 2025-2026.
+// Distributed under the Boost Software License, Version 1.0.
+//      (See accompanying file LICENSE_1_0.txt or copy at
+//           https://www.boost.org/LICENSE_1_0.txt)
+#ifndef GRADIENT_OPT_BASE_HPP
```
mborland:
```suggestion
#ifndef BOOST_MATH_OPTIMIZATION_DETAIL_GRADIENT_OPT_BASE_HPP
```

Similar suggestion applies to the other files for consistency

include/boost/math/optimization/detail/rdiff_optimization_policies.hpp
```diff
@@ -0,0 +1,134 @@
...
+                                  size_t seed = std::random_device{}())
+    : low_(low)
+    , high_(high)
+    , seed_(seed) {};
+
+  template<class ArgumentContainer>
+  void operator()(ArgumentContainer& x) const
+  {
+    boost::random::mt19937 gen(seed_);
+    boost::random::uniform_real_distribution<RealType> dist(low_, high_);
```
mborland: Same question about the use of boost random

include/boost/math/optimization/lbfgs.hpp
```diff
@@ -0,0 +1,384 @@
...
+//      (See accompanying file LICENSE_1_0.txt or copy at
+//           https://www.boost.org/LICENSE_1_0.txt)
+#ifndef LBFGS_HPP
+#define LBFGS_HPP
+#include <boost/math/optimization/detail/differentiable_opt_utilties.hpp>
+#include <boost/math/optimization/detail/gradient_opt_base.hpp>
+#include <boost/math/optimization/detail/rdiff_optimization_policies.hpp>
+#include <vector>
+
+#include "boost/math/optimization/detail/line_search_policies.hpp"
```
mborland:
```suggestion
#include <boost/math/optimization/detail/line_search_policies.hpp>
```

include/boost/math/optimization/lbfgs.hpp
```diff
@@ -0,0 +1,384 @@
...
+  void update_state(ArgumentContainer& x,
+                    std::vector<RealType>& g_k,
+                    RealType fk)
+  {
+
+    // iteration 0
+    if (g_prev.empty()) {
+      g_prev.assign(g_k.begin(), g_k.end());
+      x_prev.resize(x.size());
+      std::transform(x.begin(), x.end(), x_prev.begin(), [](const auto& xi) {
```
mborland, line 58: Would it be worth adding support for C++17 execution policies? This would allow the user to pass say `std::execution::par` and then you could call `std::transform(user_execution_policy, x.begin(), ...)`.

include/boost/math/optimization/lbfgs.hpp
```diff
@@ -0,0 +1,384 @@
...
+    const RealType yTy = dot(state_.Y.back(), state_.Y.back());
+    const RealType gamma = (yTy > RealType(0)) ? (sTy / yTy) : RealType(1);
+
+    std::vector<RealType> r = q;
+    scale(r, gamma);
+
+    for (std::size_t i = 0; i < L; ++i) {
+      const RealType yTr = dot(state_.Y[i], r);
+      const RealType beta = state_.rho[i] * yTr;
+      axpy(alpha[i] - beta, state_.S[i], r);
```
mborland, line 214: On the execution policy question I believe loops like this could be replaced with `std::for_each` since the iterations all seem separable.

include/boost/math/optimization/minimizer.hpp
```diff
@@ -0,0 +1,335 @@
+//           Copyright Maksym Zhelyenzyakov 2025-2026.
+// Distributed under the Boost Software License, Version 1.0.
+//      (See accompanying file LICENSE_1_0.txt or copy at
+//           https://www.boost.org/LICENSE_1_0.txt)
+#ifndef MINIMIZER_HPP
+#define MINIMIZER_HPP
+#include <boost/math/optimization/detail/differentiable_opt_utilties.hpp>
+#include <boost/math/optimization/gradient_optimizers.hpp>
+#include <vector>
```
mborland, line 9:
```suggestion
#include <vector>
#include <chrono>
```

doc/optimization/gradient_optimizers.qbk
```diff
@@ -0,0 +1,1255 @@
...
+
+[heading Parameters]
+
+* `Objective&& obj` : objective function to minimize.
+* `ArgumentContainer& x` : variables to optimize over. Updated in-place.
+* `std::size_t m` : history size. Typical values are 5–20. Default is 10. Larger m can improve directions but increases memory and per-step cost.
+* `InitializationPolicy&& ip` : initialization policy for ArgumentContainer and optimizer state. For reverse-mode AD, the default typically initializes/attaches the tape and uses the user-provided initial values in x.
+* `ObjectiveEvalPolicy&& oep` : policy for evaluating the objective function value at a given x. By default this is a reverse-mode AD evaluation policy when using `rvar`.
+* `GradEvalPolicy&& gep` : policy for evaluating the gradient of the objective. By default this is a reverse-mode AD gradient evaluation policy when using `rvar`.
+* `LineSearchPolicy&& lsp` : policy for selecting the step length alpha. Default is Strong Wolfe, but Armijo is an option.
```
mborland: Maybe explain what the differences are between these policies

doc/optimization/gradient_optimizers.qbk
```diff
@@ -0,0 +1,1255 @@
...
+mu is the momentum coefficient (typically close to 1)
+
+Setting mu = 0 reduces NAG to standard gradient descent.
+[heading Parameters]
+
+* `Objective&& obj` : objective function to minimize.
+* `ArgumentContainer& x` : variables to optimize over. Updated in place.
+* `RealType lr` : learning rate. Larger values take larger steps (faster but potentially unsable). Smaller values are more stable but converge more slowly.
+* `RealType mu` : momentum coefficient in `[0,1)`. Higher values, e.g. 0.9 to 0.99, typically accelerate convergence but may require a smaller `lr`
+* `InitializationPolicy&& ip` : initialization policy for the optimizer state and variables. For NAG, this also initializes the internal momentum/velocity state. By default the optimizer uses the same initializer as gradient descent and initializes velocity to zero.
```
mborland: Can the user adjust the initialization policy? If so how, and why would they want to?

doc/optimization/gradient_optimizers.qbk
```diff
@@ -0,0 +1,1255 @@
...
+{
+explicit costant_initializer_rvar(RealType v = 0);
+
+template<class ArgumentContainer>
+void operator()(ArgumentContainer& x) const;
+};
+```
+
+This policy initializes each element of x to the constant value v
+
```
mborland, line 1315: It may be worth adding a table at the end summarizing the different optimizers, policies, and when people should reach for each of the different options.

doc/optimization/gradient_optimizers.qbk
demroz: Added a short explanation

doc/optimization/gradient_optimizers.qbk
demroz: Yes. There are a few initialization policies I provided in `optimization/detal/rdiff_optimization_policies.hpp` for `rvar` specifically. By default the user just provides a starting point by initializing the values of `rvar`, and the initialization policy just controls how the `tape` is initialized. There are a few policies like `random_uniform_initializer_rvar` which initializes the `rvar` container with random values. A bit more complicated of a usecase is if the user doesnt want to use the boost autodiff library and provides their own AD variable, or supplies the gradient, in which case they'd need a custom init policy.

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
demroz: I think it would be better to keep this as a standalone library. After thinking about this a bit more, `Eigen` would really only be useful if we ever decide to add newton-based optimizers with explicit Hessian calculations, but even then I think rolling some small LA routines is probably sufficient

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
demroz: no, this is fine. I'll remove the comment

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
demroz: Its not required. I'll make the change

include/boost/math/optimization/gradient_descent.hpp
demroz: I have since changed my mid. This solution is much cleaner

include/boost/math/optimization/lbfgs.hpp
demroz: It’s possible this could provide a small speedup, but in practice the impact here is likely to be modest since the affected operations are simple, memory-bound transforms and typically not the dominant cost of an L-BFGS iteration.

Supporting C++17 execution policies would also require additional feature-detection macros and conditional code paths to preserve the C++14 baseline that Boost.Math requires.

I think this would be better handled as a separate PR with targeted benchmarks to justify the added complexity.

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
demroz: It could be changed to use `<random>` instead, but for `multiprecision` types `std::uniform_real_distribution<RealType>` throws compiler errors. In practice, we could cast to it `double` and cast it back to `multiprecision`. I'm not sure what's best here.

demroz (commented):
@mborland 

I updated the PR according to your suggestions. A few points : 

1.  I think adding support for `C++17` execution policies should be done in a separate PR with dedicated benchmarks. 
2.  using `<random>` instead of `<boost/random.hpp>`.  I'm not sure what should be done here. I agree with your point that this breaks the ability of this being a standalone package. However, `<random>` doesn't support multiprecision types. We could `static_cast` to `double` for multiprecision types, and `static_cast` back to `multiprecision` but thats not behavior I would expect as a user. Practically, i doubt it makes much difference as this is just an initialization policy. Any thoughts?

include/boost/math/optimization/detail/differentiable_opt_utilties.hpp
mborland: I think it's best to generate a double and then cast it to a multiprecision type. I don't think the lost precision would really matter in this case.

include/boost/math/optimization/lbfgs.hpp
mborland: Sounds good. C++17 execution policies can be a pain to support since some require linking to TBB, some provide only the header or half-baked impl, etc. etc.

mborland:
> @mborland
> 
> I updated the PR according to your suggestions. A few points :
> 
> 1. I think adding support for `C++17` execution policies should be done in a separate PR with dedicated benchmarks.
> 2. using `<random>` instead of `<boost/random.hpp>`.  I'm not sure what should be done here. I agree with your point that this breaks the ability of this being a standalone package. However, `<random>` doesn't support multiprecision types. We could `static_cast` to `double` for multiprecision types, and `static_cast` back to `multiprecision` but thats not behavior I would expect as a user. Practically, i doubt it makes much difference as this is just an initialization policy. Any thoughts?

I mentioned it above but I think generating a random double and then casting it to a multiprecision type should be fine. I would not expect the lost digits to create much of a difference.

doc/optimization/gradient_optimizers.qbk
demroz: added

demroz:
@mborland 

I updated the docs and removed the `boost::random` dependency. Let me know if anything else is needed from my end to merge

mborland (approved):
Looks good to me. Thanks for this!

PR #1334, opened 2025-10-30
Fix circular dependency on Master

mborland:


ckormanyos:
Hi Matt. Do you know if anything is needed in Multiprecision for this adaption?

PR #1337, opened 2025-11-29
Refactor safe_max to improve stability and remove overload

lighting9999:
Refactor safe_max function to prevent division by zero and overflow. Removed overloaded long double version.

### DIFF
--- a/include/boost/math/complex/details.hpp
+++ b/include/boost/math/complex/details.hpp
@@ -14,6 +14,7 @@
 #include <cmath>
 #include <complex>
 #include <limits>
+#include <algorithm>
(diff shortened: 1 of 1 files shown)


NAThompson:
Thanks for the PR. `safe_max` is only called with the constants 8 and 2 (from `acos`, `asin`, `atanh`), so the division-by-zero guard can't trigger. Removing the `long double` overload would bring back the problem its comment describes, where `sqrt(LDBL_MAX)` overflows on some platforms, and would leave it out of step with `safe_min`. Closing, but if you have a platform and input where the current code misbehaves, please open an issue with a reproducer.

Issue #1305, opened 2025-08-21
ENH: more inverses for distribution CDFs


dschmitz89:
Over at https://github.com/scipy/scipy/pull/23308 I implemented the inverse of the noncentral F distibution CDF with respect to the noncentrality parameter (the legacy cdflib library called this `ncfdtrinc`) using boost. It is a straight forward exercise for `bracket_and_solve_root` and works well.

Now my question is if such inverses are of interest for boost.math itself? For the noncentral [chi squared distribution](https://www.boost.org/doc/libs/1_42_0/libs/math/doc/sf_and_dist/html/math_toolkit/dist/dist_ref/dists/nc_chi_squared_dist.html), these inverses are for example implemented. Such functions have statistical applications (see for example here in [statsmodels](https://www.statsmodels.org/dev/_modules/statsmodels/stats/oneway.html)).

Tracker:

- [x] `noncentral_f:find_noncentrality` #1345 
- [ ] `noncentral_f:find_v1`
- [ ] `noncentral_f:find_v2`
 
- [x] `noncentral_t:find_degrees_of_freedom` #1355 
- [x] `noncentral_t:find_noncentrality` #1355 

- [ ] `fisher_f:find_v1`
- [ ] `fisher_f:find_v2`

- [ ] `student_t:invert_cdf_with_respect_to_degrees_of_freedom` (`find_degrees_of_freedom` is already taken, so we need another name ;) ) #1385

mborland:
We will certainly take it if you are offering. I can review your code here or there if need be, just let me know.

dschmitz89:
I had a look at the noncentral chi squared code, that looks almost reasonable to deal with for a non veteran C++ developer. What is more intimidating is how to setup the environment and create test cases. If I understand correctly, you use boost.multiprecision for that purpose. Do you have detailed docs somewhere?

I will try to get to that in future but for the time being, if you have time to help out with the review in scipy that would be highly appreciated.

mborland:
The distributions are in general tested against `float`, `double`, and `long double` with spot values calculated by Mathematica (and others). There's also a type called `real_concept` which emulates a multi-precision type, but is in practice a wrapper around a `long double`. I don't think there's any real documentation on the testing harness for the distributions. In your case the only types of concern are `float` and `double` right?

dschmitz89:
To be more specific, the list of distributions for which we would like to implement the inverses of the CDF with respect to its other parameters (quantiles already exist) are:

- noncentral F
[...]

PR #1338, opened 2025-12-12
Adding logarithm of incomplete gamma function

### LINKED ISSUE #1173: New special functions: logarithm of incomplete gamma and beta functions
The incomplete gamma and beta functions arise in a LOT of cumulative density functions. Very accurate implementations of them are quite common and also in Boost. What I have not seen yet are implementations of their logarithms. These would be very helpful to compute accurate logarithms of CDF and SFs. One example is an issue in SciPy where a user wants to compute the logcdf of the Poisson distribution for not even very extreme values: https://github.com/scipy/scipy/issues/8424. Here the logarithm of the incomplete regularized gamma function would come very much in handy.

SciPy includes the logarithm of the normal distribution CDF [already](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.log_ndtr.html) but incomplete beta and gamma functions are of course a larger beast to kill.

Is this something you would find worthwile for Boost?

Edit: CC ing the special function wizards @steppi @WarrenWeckesser

JacobHass8:
First draft of adding logarithm of incomplete gamma function. This still has a number of issues. Closes #1173.

### DIFF
--- a/include/boost/math/special_functions/logaddexp.hpp
+++ b/include/boost/math/special_functions/logaddexp.hpp
@@ -7,7 +7,6 @@
 #include <limits>
 #include <boost/math/special_functions/fpclassify.hpp>
 #include <boost/math/constants/constants.hpp>
-#include <complex>
 
 namespace boost { namespace math {
 
@@ -15,7 +14,7 @@ namespace boost { namespace math {
 template <typename Real>
 Real logaddexp(Real x1, Real x2) noexcept
 {
-    using std::log;
+    using std::log1p;
     using std::exp;
     using std::abs;
     
@@ -31,6 +30,33 @@ Real logaddexp(Real x1, Real x2) noexcept
 
     const Real temp = x1 - x2;
 
+    if (temp > 0)
+    {
+        return x1 + log1p(exp(-temp));
+    }
+
+    return x2 + log1p(exp(temp));
+}
+
+template <typename T>
+T logaddexpcomplex(T x1, T x2) noexcept
+{
+    using std::log;
+    using std::exp;
+    using std::abs;
+    
+    // Validate inputs first
+    if (!(boost::math::isfinite)(x1))
+    {
+        return x1;
+    }
+    else if (!(boost::math::isfinite)(x2))
+    {
+        return x2;
+    }
+
+    const T temp = x1 - x2;
+
     if (std::real(temp) > 0)
     {
         return x1 + log(1.0 + exp(-temp));
@@ -38,5 +64,5 @@ Real logaddexp(Real x1, Real x2) noexcept
 
     return x2 + log(1.0 + exp(temp));
 }
-
-}} // Namespace boost::math
+}
+}// Namespace boost::math
--- a/include/boost/math/tools/fraction.hpp
+++ b/include/boost/math/tools/fraction.hpp
@@ -20,6 +20,7 @@
 #include <boost/math/tools/cstdint.hpp>
 #include <boost/math/special_functions/logaddexp.hpp>
 #include <limits>
+#include <iostream>
 
 namespace boost{ namespace math{ namespace tools{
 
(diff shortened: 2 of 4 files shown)


include/boost/math/special_functions/logaddexp.hpp
```diff
@@ -30,12 +31,12 @@ Real logaddexp(Real x1, Real x2) noexcept
 
     const Real temp = x1 - x2;
 
-    if (temp > 0)
+    if (std::real(temp) > 0)
     {
-        return x1 + log1p(exp(-temp));
+        return x1 + log(1.0 + exp(-temp));
```
jzmaddock, line 62: This is a really bad change that will kill precision when temp is small.

include/boost/math/tools/fraction.hpp
```diff
@@ -110,7 +112,7 @@ namespace detail {
 //
 
 template <typename Gen, typename U>
-BOOST_MATH_GPU_ENABLED inline typename detail::fraction_traits<Gen>::result_type continued_fraction_b_impl(Gen& g, const U& factor, boost::math::uintmax_t& max_terms)
+BOOST_MATH_GPU_ENABLED inline typename detail::fraction_traits<Gen>::result_type continued_fraction_b_impl(Gen& g, const U& factor, boost::math::uintmax_t& max_terms, bool logArithmetic)
```
jzmaddock: Since the log and non-log versions basically have completely different code, I see this as two different functions?

include/boost/math/tools/fraction.hpp
```diff
@@ -236,7 +259,7 @@ namespace detail {
 // Note that the first a1 and b1 returned by generator Gen are both used.
 //
 template <typename Gen, typename U>
-BOOST_MATH_GPU_ENABLED inline typename detail::fraction_traits<Gen>::result_type continued_fraction_a_impl(Gen& g, const U& factor, boost::math::uintmax_t& max_terms)
+BOOST_MATH_GPU_ENABLED inline typename detail::fraction_traits<Gen>::result_type continued_fraction_a_impl(Gen& g, const U& factor, boost::math::uintmax_t& max_terms, bool logArithmetic=false)
```
jzmaddock: Again this looks like two different functions to me.

include/boost/math/special_functions/logaddexp.hpp
```diff
@@ -30,12 +31,12 @@ Real logaddexp(Real x1, Real x2) noexcept
 
     const Real temp = x1 - x2;
 
-    if (temp > 0)
+    if (std::real(temp) > 0)
```
jzmaddock, line 60: The trick here, is to add `using std::real` above, and then call `real(temp)`, that way ADL can find `boost::multiprecision::real` in the multiprecision case (likewise for any other type that has their own namespace and is forbidden from putting stuff in `std::`).

include/boost/math/special_functions/logaddexp.hpp
JacobHass8: I've reverted this for the `logaddexp` function and created a new function for adding complex numbers. I haven't been able to find an equivalent of `log1p` that works with complex numbers. Could I just make a new function `clog1p` which computes the Taylor series of $log(1+x)$ for $|x| < 1$? This seems easy enough that I'm surprised I haven't found an implementation.

include/boost/math/tools/fraction.hpp
JacobHass8: I've separated this into a new function in `gamma.hpp`. I wasn't sure what the correct implementation should have been.

jzmaddock:
> I didn't realize the CF converges so slowly. I guess that makes sense since it's a sort of iterative method.

I *love* continued fractions, but as with series approximations, they always have zones of good/bad convergence: which is only to be expected as for every CF there is a 1:1 mapping onto an equivalent series.

> That makes a lot more sense. I should have thought of this sooner. Should I delete this PR and start over with your code above? I'm not sure if there's any applications for log arithmetic continued fractions.

In that case yes, let's start with a fresh PR as it keeps the version history a bit cleaner?

JacobHass8:
> I had a very quick experiment and came up with this, which needs a lot more testing, but seems to work well:

I've done some quick tests comparing your function to Mathematica and they match up very well! It was accurate for almost all regimes I tried like $x \gg a$ and $x \ll a$.  

> //TODO add non-policy version and all the promotion boilerplate, plus some actual error checking...

The non-policy version seemed pretty easy. I took a look at other functions in `gamma.hpp` and they just passed a default policy, `policies::policy<>()`. Is that the correct approach? 

I'm not sure what you mean by "promotion boilerplate" - this is my first contribution to boost. Is this in regards to the return argument type `tools::promote_args<T>::type` which I'm seeing in most functions in `gamma.hpp`?

dschmitz89:
@JacobHass8 Thanks for your help in enhancing boost on behalf of issues raised in SciPy, much appreciated!

I thought I would ask if you would be interested in helping out with #1305. Finishing that off will enable SciPy to drop one of the ancient libraries that are unfortunately still in use, cdflib. It could also likely be easier than this issue since we can basically copy paste the approach from other distributions (it is basically a straight forward usage of a derivative-free rootfinder).

PR #1342, opened 2025-12-18
Improve check for incomplete <execution> header

mbs-c:
All versions of libc++ currently include a useless <execution> header by default, not just the version shipped with Apple Clang. For example, IBM Open XL C/C++ 17.1.3 on AIX is affected as well.

### DIFF
--- a/include/boost/math/tools/config.hpp
+++ b/include/boost/math/tools/config.hpp
@@ -84,7 +84,9 @@
 
 #else // Things from boost/config that are required, and easy to replicate
 
+#if __has_include(<version>)
 #include <version>
+#endif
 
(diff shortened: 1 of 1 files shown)


include/boost/math/tools/config.hpp
```diff
@@ -84,6 +84,8 @@
 
 #else // Things from boost/config that are required, and easy to replicate
 
+#include <version>
```
mborland, line 88:
```suggestion
#if __has_include(<version>)
#include <version>
#endif
```

This wasn't standardized until C++20. I think it's unofficially available earlier, but looks like at least some of the older clangs don't have it.

PR #1345, opened 2025-12-28
Initial implementation of `find_non_centrality` for `non_central_f_distribution`

JacobHass8:
See #1305 for for details.

### DIFF
--- a/include/boost/math/distributions/non_central_f.hpp
+++ b/include/boost/math/distributions/non_central_f.hpp
@@ -18,8 +18,6 @@
 #include <boost/math/special_functions/pow.hpp>
 #include <boost/math/policies/policy.hpp>
 #include <boost/math/distributions/complement.hpp> // complements
-#include <boost/math/distributions/fisher_f.hpp>
-#include <boost/math/special_functions/relative_difference.hpp> 
 
 namespace boost
 {


JacobHass8:
@dschmitz89 how does this first commit look in regards to https://github.com/boostorg/math/pull/1338#issuecomment-3675187807

dschmitz89:
Thanks @JacobHass8  for looking into this so quickly. This is a good start. I am not a boost maintainer (barely a boost developer), so can only comment from a high level.

The most important missing bit here is tests for the new functionality. I could not find the current noncentral F distribution tests after 2 minutes though, so we will have to rely on the boost devs to guide us a little here.

jzmaddock:
Yup, tests needed.

Non central F is treated as a special case of the non-central beta (which is where most of the tests are).  Non-central F tests are fairly lightweight, just there to make sure we have the transformation to non-central beta correct really: https://github.com/boostorg/math/blob/develop/test/test_nc_f.cpp.  A similar level of sanity checking would probably suffice for this function I would guess?

test/test_nc_f.cpp
```diff
@@ -141,6 +141,8 @@ void test_spot(
             quantile(dist, P), x, tol * 10);
       BOOST_CHECK_CLOSE(
             quantile(complement(dist, Q)), x, tol * 10);
+      BOOST_CHECK_CLOSE(
+            dist.find_non_centrality(a, b, P, x), ncp, tol * 10);
```
JacobHass8: @jzmaddock is this sufficient for tests? This matches what is done for `non_central_chi_squared_distribution`.

dschmitz89:
> This looks pretty good to me. I made a couple minor changes to get it running with CUDA. Since scipy already has an implementation, it's probably worth copying over any tests they have as well.

SciPy does not have tests for these functions at the moment yet. The functions were simply incorporated a long time ago with the cdflib library but actually never unit tested. One more plus point for the migration we are planning.

mborland:
> I know what is going wrong now. When calculating the `pdf` of `dist` for floats:
> 
> ```c++
> fisher_f_distribution<RealType, Policy> dist(v1, v2); 
> if (abs(boost::math::float_distance(pdf(dist, x), p)) <= 1){
>     return 0;
> }
> ```
> 
> the template `RealType` is upcast to a `long double` (maybe a `double`, I'm not 100% sure). If I change the `dist` variable above to `non_central_f_distribution<float, policies::policy<> > dist(v1, v2, 0);`, and take the relative distance I get 0 as expected. I also think we should be comparing to `non_central_f_distribution` with `nc=0` rather than `fisher_f_distribution` because they calculate the pdf differently.
> 
> I'm not sure if there is an elegant way to avoid this upcasting based on type because this happens before `find_non_centrality_f` is called.

Instead of using `policies::policy<>` as your template parameter to `non_central_f_distribution` you should be able to use `policies::policy<policies::promote_float<false>>` in order to avoid this.

include/boost/math/distributions/non_central_f.hpp
```diff
@@ -17,11 +17,71 @@
...
+               return policies::raise_domain_error<RealType>(function, "Can't find non centrality parameter when the probability is <=0 or >=1, only possible answer is %1%", // LCOV_EXCL_LINE
+                  RealType(boost::math::numeric_limits<RealType>::quiet_NaN()), Policy()); // LCOV_EXCL_LINE
+            }
+
+            // Check if nc = 0 (which is just the F-distribution)
+            // I first tried comparing to boost::math::tools::epsilon<RealType>(),
+            // but this was never satisfied for floats or doubles. Only 
+            // long doubles would correctly return 0.
+            fisher_f_distribution<RealType, Policy> dist(v1, v2); 
+            if (boost::math::relative_difference(pdf(dist, x), p) < 1e-7){
```
jzmaddock: This looks suspicious to me, some possibly naive questions:

1) Should we not be checking CDF and not PDF here?
2) Do we want to find non-centralities with very small values, ie below 1e-7?

One possible alternative would be to call `f(tools::min_value<T>())` and if the answer lies to the left of that, then return 0.  Hopefully bracket_and_solve_root will handle all answers to the right of `min_value()`.  Hope that makes sense!

jzmaddock:
Update: the original tests (checking for zero result) work for me with gcc cygwin and mingw.  What was the combination that was failing for you?

jzmaddock:
I had a chance to look into this: the function we are finding the root for, is basically zero for any non-centrality below about 10^-8 and bobbles about a few epsilon either side of zero "all the way down" as it were.  I think @JacobHass8 had it right the first time (sorry!) and we should check for low absolute error when non-centrality is at or near zero.  Sorry, I've not had a chance to report back before now.

PR #1346, opened 2025-12-29
Implement special function log incomplete gamma function

JacobHass8:
Implementation of log incomplete gamma function using asymptotic approximations where the incomplete gamma function underflows. See #1173 and #1338.

### DIFF
--- a/test/test_igamma.hpp
+++ b/test/test_igamma.hpp
@@ -269,7 +269,7 @@ void test_spots(T, const char* name = nullptr)
    T gcc_win_mul = 1;
 #endif
    BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(50), static_cast<T>(2)), static_cast<T>(-5.214301903317168085381693412994550732094621576607843973832e-51L), tolerance * gcc_win_mul);
-   BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(500), static_cast<T>(10)), static_cast<T>(-3.79666711621207197039397438773960431648625558027046365463e-639L), tolerance * 3);
+   BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(500), static_cast<T>(10)), static_cast<T>(-3.79666711621207197039397438773960431648625558027046365463e-639L), tolerance * 3 * gcc_win_mul);
    BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(5), static_cast<T>(1000)), static_cast<T>(-975.5430287171020511929200293377669175923128826278957569928895945L), tolerance);
    // Pairs of tests that bisect the crossover condition in our code at double and then quad precision:
    BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(10), static_cast<T>(698.75)), static_cast<T>(-652.5952453102824132865663191324423994628428404928732148525545721L), tolerance);


JacobHass8:
@jzmaddock I've tried to add error checking and the promotion boilerplate based on your comment https://github.com/boostorg/math/pull/1338#issuecomment-3655757703. Does this look okay for a start?

test/test_igamma.hpp
```diff
@@ -255,6 +255,13 @@ void test_spots(T)
    //
    BOOST_CHECK_EQUAL(::boost::math::gamma_q(static_cast<T>(1770), static_cast<T>(1e-12)), 1);
    BOOST_CHECK_EQUAL(::boost::math::gamma_p(static_cast<T>(1770), static_cast<T>(1e-12)), 0);
+
+   //
+   // Check that lgamma_q returns correct values
+   //
+   BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(5), static_cast<T>(100)), static_cast<T>(log(1.6139305336977304790405739225035685228527400976549e-37L)), tolerance);
+   BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(22.5), static_cast<T>(2000)), static_cast<T>(-1883.4897732037716195918619632721L), tolerance * 10); // calculated via mpmath
+   BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(501.2), static_cast<T>(2000)), static_cast<T>(-810.31461624182202285737730562687L), tolerance * 10); // calculated via mpmath
```
JacobHass8: I used one test case that was previously checked for `gamma_q`. I also added two more examples that I calculated using `mpmath` in python which supports arbitrary precision. Are there any other tests that I should implement?

JacobHass8:
> With regard to further tests, the area most likely to fail, would be where the implementation first switches to the log asymptotic expansion, and at higher precision (we should support 128-bit long doubles at least, I'm less concerned about full arbitrary precision just yet).

I've added some documentation but don't know how to generate the html. I tried running b2 in the doc folder but received an error from a different file. 

I also added a test for $x=999$ which is right below when the asymptotic expansion is used. Is there anything else that needs to be done before this can be merged?

test/test_igamma.hpp
```diff
@@ -18,7 +18,7 @@
 #include <boost/type_traits/is_floating_point.hpp>
 #include <boost/array.hpp>
 #include "functor.hpp"
-
+#include <iostream>
```
JacobHass8: This can be deleted, right?

mborland:
> > @JacobHass8 I added the CUDA markers and testing. Everything runs without changes needed. It does looks like you have a tolerance issue on CI machines with 128-bit long doubles (s390x and ARM64):
> 
> Would simply increasing the tolerances in this case be acceptable? This is for very large arguments and the difference seems quite small.
> 

I think that would be fine.

> In the future, how can I test with s390x and ARM64 CI machines?

They are automatically included in the check titled "continuous-integration/drone/pr". You can click on that runner's hyperlink, and it will show you the several hundred configurations that Drone runs for us.

JacobHass8:
> They are automatically included in the check titled "continuous-integration/drone/pr". You can click on that runner's hyperlink, and it will show you the several hundred configurations that Drone runs for us.

Oh, I see where that is now. Thanks! Is there a way to see if these tests will pass with my revisions before pushing?

JacobHass8:
> I use multi-arch docker containers integrated into CLion to test platforms like these I don't have. Outside of that generally no, you'll have to wait and see.
 
I had been trying to use various boost multiprecision types but to no success. I'm going to try to set up a docker container. Thanks for the help!

jzmaddock:
Yup, I've pushed some tolerance changes.

dschmitz89:
Awesome, thanks everyone involved :). scipy/xsf contains a [`log1mexp` function](https://github.com/scipy/scipy/pull/22439/files#diff-0d6c07786556f6350d4dc2392dcce30e0caaa57adba6d5df1809dbd4c508de84R72) to accurately compute `ln(1-exp(x)`. Would you consider this a viable approach to compute the log of the incomplete upper gamma `lgamma_p` using the `lgamma_q` from this PR?

PR #1349, opened 2026-01-13
Implementation of the log of the lower incomplete gamma function

JacobHass8:
See #1346 and #1173.

### DIFF
--- a/include/boost/math/special_functions/gamma.hpp
+++ b/include/boost/math/special_functions/gamma.hpp
@@ -1842,9 +1842,7 @@ BOOST_MATH_GPU_ENABLED T lgamma_incomplete_lower_imp(T a, T x, const Policy& pol
    // we will lose precision in the result, selection logic here is extracted from gamma_incomplete_imp_final:
    //
    bool need_p = false;
-   if ((x < 0.5) && (T(-0.4) / log(x) < a))
-      need_p = true;
-   else if ((x < 1.1) && (x >= 0.5) && (x * 0.75f < a))
+   if ((x < 1.1) && (x >= 0.5) && (x * 0.75f < a))
       need_p = true;
    else if ((x < a) && (x >= 1.1))
       need_p = true;
--- a/doc/sf/igamma.qbk
+++ b/doc/sf/igamma.qbk
@@ -79,10 +79,10 @@ This function changes rapidly from 0 to 1 around the point z == a:
 [graph gamma_p]
 
    template <class T1, class T2>
-   BOOST_MATH_GPU_ENABLED ``__sf_result`` lgamma_q(T1 a, T2 z);
+   BOOST_MATH_GPU_ENABLED ``__sf_result`` lgamma_p(T1 a, T2 z);
    
    template <class T1, class T2, class ``__Policy``>
-   BOOST_MATH_GPU_ENABLED ``__sf_result`` lgamma_q(T1 a, T2 z, const ``__Policy``&);
+   BOOST_MATH_GPU_ENABLED ``__sf_result`` lgamma_p(T1 a, T2 z, const ``__Policy``&);
 
 Returns the natural log of the normalized lower incomplete gamma function 
 of a and z.  


include/boost/math/special_functions/gamma.hpp
```diff
@@ -1813,6 +1843,45 @@ BOOST_MATH_GPU_ENABLED T lgamma_incomplete_imp(T a, T x, const Policy& pol)
...
+
+   // Check for invalid inputs (a < 0 or x < 0)
+   constexpr auto function = "boost::math::lgamma_p<%1%>(%1%, %1%)";
+   if(a <= 0)
+      return policies::raise_domain_error<T>(function, "Argument a to the incomplete gamma function must be greater than zero (got a=%1%).", a, pol);
+   if(x < 0)
+      return policies::raise_domain_error<T>(function, "Argument x to the incomplete gamma function must be >= 0 (got x=%1%).", x, pol);
+
+   // This still needs work...
+   if (a > x + 50){
```
JacobHass8: @jzmaddock I implemented the series formula in https://dlmf.nist.gov/8.11#E4 to calculate the log of the lower incomplete gamma function. I'm not sure what the criteria should be to use that asymptotic formula though. Do you have any thoughts?

JacobHass8:
I also changed the criteria for when the asymptotic series is used based on code paths in previous functions that used the asymptotic series. With my limited spot testing, this gives accurate results and is ready for testing.

JacobHass8:
Never mind, I got Mathematica to output higher precision. It matches the values that I've entered. So the implementation only has precision up to 10^{-18}. Here's the Mathematica code if anyone is interested. 
<p>

```Mathematica
N[Log[GammaRegularized[100,0,1/10]],64]
```

</p>
</details>

jzmaddock:
>This is strange though because in previous tests for lgamma_q, we used

>   BOOST_CHECK_CLOSE(::boost::math::lgamma_q(static_cast<T>(20), static_cast<T>(0.25)), static_cast<T>(-2.946458104491857816330873290969917497748067639461638294404e-31L), tolerance);

>and this passed for long doubles.

Yes, because all the input values have exact binary representations, so you could use 0.25f as the input argument and it would still calculate to whatever precision the result_type is.

>Is there an easy way to convert 0.6 to a binary literal? I've seen calculators online, but haven't been able to correctly convert them to a long double. I might just go with the horrible static_cast for now to see if it works.

No because 0.6 has no exact decimal representation: it's an infinitely recurring number when expressed as base 2 (at least I think it is, 0.1 is the classic example).  So you would need a way to store an infinite number of binary digits.

And that's not all... even if you use an exact fraction, say 6/10 calculated at the precision of T, then the input values for float precision are notably different (truncated) compared to long double, or indeed the arbitrary precision of mathematica.  So if the function is ill-conditioned, the 0.5ulp error in the float input values can be enough to make a significant difference to the result.

So for the sake of easiness, I always choose exact binary values, or where we use randomly generated test data, the inputs are truncated to fit inside a float without precision loss first, and then the expected result is calculated.  This can sometimes miss obscure issues where the last few bits of the input are somehow getting lost in the calculation, but for the sake of everyone's sanity, it sure does save a lot of grief :)

JacobHass8:
Everything looks like it is passing now except g++-14 C++23 autodiff. The error code is below and looks unrelated to the changes that I've made. 

<p>

```bash

testing.capture-output ../../../bin.v2/libs/math/test/test_autodiff_8.test/gcc-14/release/x86_64/link-static/threading-multi/visibility-hidden/test_autodiff_8.run
====== BEGIN OUTPUT ======
Running 53 test cases...
unknown location(0): fatal error: in "test_autodiff_8/heuman_lambda_hpp<_Float32>": Throw location unknown (consider using BOOST_THROW_EXCEPTION)
Dynamic exception type: boost::wrapexcept<std::domain_error>
std::exception::what: Error in function boost::math::heuman_lambda<N5boost4math15differentiation11autodiff_v16detail4fvarIDF32_Lm5EEE>(N5boost4math15differentiation11autodiff_v16detail4fvarIDF32_Lm5EEE, N5boost4math15differentiation11autodiff_v16detail4fvarIDF32_Lm5EEE): When 1-k^2 == 1 then phi must be < Pi/2, but got phi = depth(1)(4.20396852,1,0,0,0,0)

test_autodiff_8.cpp(38): last checkpoint

*** 1 failure is detected in the test module "test_autodiff"
```

</p>
</details>

PR #1350, opened 2026-01-13
Update fptraits.hpp for gcc/windows/aarch64.

### LINKED ISSUE #1348: Fix for buffer overflow in fp_traits.hpp on aarch64 / mingw
Hi,

I'm targeting aarch64-w64-mingw32 (probably I'm the first to try) and get this error:

```
/home/joostn/devel/wxwprojects_ptgui13/build/windows_mingw/arm64/RelWithDebInfo/thirdparty/boost-boost-1.90.0-prefix/include/boost/math/special_functions/detail/fp_traits.hpp:452:9: error: 'memcpy' will always overflow; destination buffer has size 0, but size argument is 4 [-Werror,-Wfortify-source]
  452 |         std::memcpy(reinterpret_cast<unsigned char*>(&x) + offset_, &a, 4);
      |         ^
1 error generated.
```

In include/boost/math/special_functions/detail/fp_traits.hpp you have this:

```
#if defined(BOOST_NO_INT64_T) || defined(BOOST_NO_INCLASS_MEMBER_INITIALIZATION)\
   || defined(BOOST_BORLANDC) || defined(__CODEGEAR__) || (defined(__APPLE__) && defined(__aarch64__)) || defined(_MSC_VER)
```

but AFAIK aarch64-w64-mingw32 also has 64 bit long double, so it should also take that path. This would fix it:

```#if defined(BOOST_NO_INT64_T) || defined(BOOST_NO_INCLASS_MEMBER_INITIALIZATION)\
   || defined(BOOST_BORLANDC) || defined(__CODEGEAR__) || (defined(__APPLE__) && defined(__aarch64__)) || defined(_MSC_VER) || (defined(__MINGW32__) && defined(_WIN32))
```

I'm not sure if there are other places where this should be fixed, or if there are side effects, so I don't dare to submit a pull request.

jzmaddock:
Also adds better error checking.
Fixes #1348

### DIFF
--- a/include/boost/math/special_functions/detail/fp_traits.hpp
+++ b/include/boost/math/special_functions/detail/fp_traits.hpp
@@ -272,7 +272,7 @@ template<> struct fp_traits_non_native<double, double_precision>
 
 #if defined(BOOST_NO_INT64_T) || defined(BOOST_NO_INCLASS_MEMBER_INITIALIZATION)\
    || defined(BOOST_BORLANDC) || defined(__CODEGEAR__) || (defined(__APPLE__) && defined(__aarch64__)) || defined(_MSC_VER)\
-    || (defined(__GNUC__) && defined(__aarch64__) && defined(_WIN32))
+    || (defined(__GNUC__) && defined(__aarch64__) && defined(_WIN32)) || defined(__SYCL_DEVICE_ONLY__)
 
 static_assert(LDBL_MANT_DIG == 53, "Oops, assumption that long double is a 64-bit quantity is incorrect!!");
 


include/boost/math/special_functions/detail/fp_traits.hpp
```diff
@@ -356,6 +362,8 @@ struct fp_traits_non_native<long double, extended_double_precision>
 
 // PowerPC extended double precision format (128 bits)
 
+static_assert(LDBL_MANT_DIG == 113, "Oops, assumption that long double is a 128-bit quantity is incorrect!!");
+
```
mborland, line 366: In the PowerPC case should we check both 106 and 113 for `LDBL_MANT_DIG`? I assume the original code is old enough that 106 was the only option at the time.

jzmaddock: Not in the same branch, everything has to go to the correct #if branch otherwise we don't get the correct bits.

jzmaddock:
@mborland I can't seem to get CUDA/SYCL correct: am I correct in thinking that long double is 64 bit on the device and 80-bit on the host?  If so is there a good way to configure that?

mborland:
Actually, even easier is removing the check for GPU platforms. I was wrong and `long double` is [entirely unsupported and warns if used.](https://forums.developer.nvidia.com/t/support-for-long-double-data-type-in-gpu-code/313337/4). There are also no GPU markers on any of the long double functions so they will error if called anyway.

PR #1354, opened 2026-01-27
Update fptraits.hpp with better error handling.

jzmaddock:


mborland:
Sorry for the delay, but locally this seems to help out the QEMU case. I am going to merge it in since it seems like it'll help the unknown cases.

PR #1359, opened 2026-02-10
Implement the log of the incomplete beta function

JacobHass8:
See #1173.

### DIFF
--- a/include/boost/math/special_functions/beta.hpp
+++ b/include/boost/math/special_functions/beta.hpp
@@ -1674,7 +1674,7 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
    }
    else
    {
-      return invert ? (normalised ? std::log1p(-fract) : log(boost::math::beta(a, b, pol) - fract)) : log(fract);
+      return invert ? (normalised ? boost::math::log1p(-fract) : log(boost::math::beta(a, b, pol) - fract)) : log(fract);
    }
    
 } // template <class T, class Lanczos>T ibeta_imp(T a, T b, T x, const Lanczos& l, bool inv, bool normalised)
--- a/include/boost/math/special_functions/beta.hpp
+++ b/include/boost/math/special_functions/beta.hpp
@@ -1195,7 +1195,7 @@ BOOST_MATH_GPU_ENABLED T ibeta_large_ab(T a, T b, T x, T y, bool invert, bool no
 //
 
 template <class T, class Policy>
-BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, bool normalised, T* p_derivative)
+BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, bool normalised, T* p_derivative, bool logarithm=false)
 {
    constexpr auto function = "boost::math::ibeta<%1%>(%1%, %1%, %1%)";
    typedef typename lanczos::lanczos<T, Policy>::type lanczos_type;
@@ -1235,12 +1235,32 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
          if(b == 0)
             return policies::raise_domain_error<T>(function, "The arguments a and b to the incomplete beta function cannot both be zero, with x=%1%.", x, pol);
          if(b > 0)
-            return static_cast<T>(inv ? 0 : 1);
+         {
+            if (!logarithm)
+            {
+               return static_cast<T>(inv ? 0 : 1);
+            }
+            else
+            {
+               return inv ? -std::numeric_limits<T>::infinity() : T(0);
+            }
+            
+         }
+            
       }
       else if(b == 0)
       {
          if(a > 0)
-            return static_cast<T>(inv ? 1 : 0);
+         {
+            if (!logarithm)
+            {
+               return static_cast<T>(inv ? 1 : 0);
+            }
+            else
+            {
+               return inv ? T(0) : -std::numeric_limits<T>::infinity();
+            }
+         }
       }
    }
    else
@@ -1257,7 +1277,14 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
       {
          *p_derivative = (a == 1) ? (T)1 : (a < 1) ? T(tools::max_value<T>() / 2) : T(tools::min_value<T>() * 2);
       }
-      return (invert ? (normalised ? T(1) : boost::math::beta(a, b, pol)) : T(0));
+      if (!logarithm)
+      {
+         return (invert ? (normalised ? T(1) : boost::math::beta(a, b, pol)) : T(0));
+      }
+      else
+      {
+         return (invert ? (normalised ? 0 : log(boost::math::beta(a, b, pol))) : -std::numeric_limits<T>::infinity());
+      }
    }
    if(x == 1)
    {
@@ -1265,7 +1292,15 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
       {
          *p_derivative = (b == 1) ? T(1) : (b < 1) ? T(tools::max_value<T>() / 2) : T(tools::min_value<T>() * 2);
       }
-      return (invert == 0 ? (normalised ? 1 : boost::math::beta(a, b, pol)) : 0);
+      if (!logarithm)
+      {
+         return (invert == 0 ? (normalised ? 1 : boost::math::beta(a, b, pol)) : 0);
+      }
+      else
+      {
+         return (invert == 0 ? (normalised ? 0 : log(boost::math::beta(a, b, pol))) : -std::numeric_limits<T>::infinity());
+      }
+      
    }
    if((a == 0.5f) && (b == 0.5f))
    {
@@ -1277,7 +1312,7 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
       T p = invert ? asin(sqrt(y)) / constants::half_pi<T>() : asin(sqrt(x)) / constants::half_pi<T>();
       if(!normalised)
          p *= constants::pi<T>();
-      return p;
+      return logarithm ? log(p) : p;
    }
    if(a == 1)
    {
@@ -1294,7 +1329,15 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
       {
          if(p_derivative)
             *p_derivative = 1;
-         return invert ? y : x;
+         if (!logarithm)
+         {
+            return invert ? y : x;
+         }
+         else
+         {
+            return invert ? log(y) : log(x);
+         }
+         
       }
 
       if(p_derivative)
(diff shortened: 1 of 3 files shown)


include/boost/math/special_functions/beta.hpp
```diff
@@ -1235,12 +1235,32 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
...
          if(b > 0)
-            return static_cast<T>(inv ? 0 : 1);
+         {
+            if (!logarithm)
+            {
+               return static_cast<T>(inv ? 0 : 1);
+            }
+            else
+            {
+               return inv ? -std::numeric_limits<T>::infinity() : T(0);
```
jzmaddock: For consistency with our other code, this should return

`-policies::raise_overlow_error<T>(function, nullptr, pol)`

include/boost/math/special_functions/beta.hpp
```diff
@@ -1235,12 +1235,32 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
...
          if(a > 0)
-            return static_cast<T>(inv ? 1 : 0);
+         {
+            if (!logarithm)
+            {
+               return static_cast<T>(inv ? 1 : 0);
+            }
+            else
+            {
+               return inv ? T(0) : -std::numeric_limits<T>::infinity();
```
jzmaddock: As above, raise an overflow.

include/boost/math/special_functions/beta.hpp
```diff
@@ -1257,15 +1277,30 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
...
          *p_derivative = (a == 1) ? (T)1 : (a < 1) ? T(tools::max_value<T>() / 2) : T(tools::min_value<T>() * 2);
       }
-      return (invert ? (normalised ? T(1) : boost::math::beta(a, b, pol)) : T(0));
+      if (!logarithm)
+      {
+         return (invert ? (normalised ? T(1) : boost::math::beta(a, b, pol)) : T(0));
+      }
+      else
+      {
+         return (invert ? (normalised ? 0 : log(boost::math::beta(a, b, pol))) : -std::numeric_limits<T>::infinity());
```
jzmaddock: overflow_error again, and this opens a can of worms because the regular beta can suffer underflow too, so that probably needs to be logarithmed too.

include/boost/math/special_functions/beta.hpp
```diff
@@ -1257,15 +1277,30 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
...
          *p_derivative = (b == 1) ? T(1) : (b < 1) ? T(tools::max_value<T>() / 2) : T(tools::min_value<T>() * 2);
       }
-      return (invert == 0 ? (normalised ? 1 : boost::math::beta(a, b, pol)) : 0);
+      if (!logarithm)
+      {
+         return (invert == 0 ? (normalised ? 1 : boost::math::beta(a, b, pol)) : 0);
+      }
+      else
+      {
+         return (invert == 0 ? (normalised ? 0 : log(boost::math::beta(a, b, pol))) : -std::numeric_limits<T>::infinity());
```
jzmaddock: as above.

include/boost/math/special_functions/beta.hpp
```diff
@@ -1294,7 +1329,15 @@ BOOST_MATH_GPU_ENABLED T ibeta_imp(T a, T b, T x, const Policy& pol, bool inv, b
...
          if(p_derivative)
             *p_derivative = 1;
-         return invert ? y : x;
+         if (!logarithm)
+         {
+            return invert ? y : x;
+         }
+         else
+         {
+            return invert ? log(y) : log(x);
```
jzmaddock, line 1338: Just a heads up that these conditionals can trip up expression-template-enabled multiprecision types.  You might be safer with `log(invert ? y : x)` but we can fix this up later.

test/Jamfile.v2
```diff
@@ -263,6 +263,7 @@ test-suite special_fun :
    [ run test_gamma_mp.cpp /boost/test//boost_unit_test_framework : : : release <define>TEST=3 [ check-target-builds ../config//is_ci_sanitizer_run "Sanitizer CI run" : <build>no ] : test_gamma_mp_3 ]
    [ run test_hankel.cpp /boost/test//boost_unit_test_framework  ]
    [ run test_hermite.cpp test_instances//test_instances pch_light /boost/test//boost_unit_test_framework  ]
+   [ run test_ibeta.cpp test_instances//test_instances pch_light /boost/test//boost_unit_test_framework  ]
```
jzmaddock, line 266: This should be covered by the following lines - we've deliberately split testing ibeta up into smaller chunks because if we test too many types in one go, we get massive object files which fail on cygwin/mingw and run the system out of memory on other CI machines.  The "test everything at once" behaviour that is the default is really just there for the convenience of local testing.

include/boost/math/special_functions/beta.hpp
```diff
@@ -1794,6 +1851,29 @@ BOOST_MATH_GPU_ENABLED inline typename tools::promote_args<RT1, RT2, RT3>::type
    return boost::math::ibeta(a, b, x, policies::policy<>());
 }
 
+template <class RT1, class RT2, class RT3, class Policy>
+BOOST_MATH_GPU_ENABLED inline typename tools::promote_args<RT1, RT2, RT3>::type
+   log_ibeta(RT1 a, RT2 b, RT3 x, const Policy&)
```
jzmaddock, line 1856: We need to think about consistency here, based on gamma_p/gamma_q/gamma we should just add a single "l" prefix to make `libeta`, though I admit it doesn't read too well!  Thoughts anyone?

jzmaddock:
>I'd like to ask to reconsider exposing two separate public functions, once ibeta and its log libeta. This would be more consistent with the API style you used so far as you have lgamma, lgamma_p and lgamma_q.

That's what we have here, but the log functions are called `log_ibeta`: see my comment on consistency above.

include/boost/math/special_functions/beta.hpp
JacobHass8: Great! I was going to ask how to handle overflow errors. I'll make those changes here and below.

include/boost/math/special_functions/beta.hpp
JacobHass8: I noticed there wasn't a `lbeta` function too. For now, it seems like taking `log(beta(a, b))` works pretty well. I can open another PR after to implement `lbeta` though.

test/Jamfile.v2
JacobHass8: Okay! I'll go ahead and delete this line then. I added it because when I ran `../../../b2 test_ibeta` from the `test` folder, `test_ibeta` wasn't found. It output a ton of linking errors.

include/boost/math/special_functions/beta.hpp
mborland: I'd agree that precedent says we should go with libeta even though it's kind of ugly.

JacobHass8: I could go either way with `libeta` or `log_ibeta`. The reason I didn't go with `libeta` is it looks like a library named `eta`. Using `libeta` would be consistent with `lgamma` which might be preferable. 

In regards to `gamma_p/gamma_q`, the incomplete beta function already has the complement implemented as `ibetac`. To be consistent with this, I would suggest using `libeta` and `libetac`.

include/boost/math/special_functions/beta.hpp
JacobHass8: Hmmm, I took another look at implementing `lbeta` and I might be missing something but it doesn't seem that bad. It was pretty easy to extend the implementation of [`beta_imp`](https://github.com/JacobHass8/math/blob/f1c2e6fbb991e324307fe4b885f880d288471615/include/boost/math/special_functions/beta.hpp#L45) with the Lanczos approximation to logs. Should I open a seperate PR for that change first? Sorry if I'm spamming this repository with too many PRs.

dschmitz89:
Gentle bump here now that #1363 was merged. Is a high precision `libeta` feasible?

jzmaddock:
Yes, but it's more work than first anticipated.  Basically we need a log-ified version of the series approximation when out in the tails.

@JacobHass8 : would it solve things if we used `log(ibeta(a, b, x))` when away from the tail, and a log version of ibeta_series when in the tail?  Ah hold on, we need to handle ibeta(a, b, x) near one as a special case too... maybe via reflection?

JacobHass8:
> Yes, but it's more work than first anticipated. Basically we need a log-ified version of the series approximation when out in the tails.
> 
> @JacobHass8 : would it solve things if we used `log(ibeta(a, b, x))` when away from the tail, and a log version of ibeta_series when in the tail? Ah hold on, we need to handle ibeta(a, b, x) near one as a special case too... maybe via reflection?

That seems like a reasonable fix to me. Would this entail implementing a log version of [this function](https://github.com/boostorg/math/blob/625d76783b963de14fcd4f2b86070d4d02eb6c71/include/boost/math/special_functions/beta.hpp#L753) or the [Lanczos approximation](https://github.com/boostorg/math/blob/625d76783b963de14fcd4f2b86070d4d02eb6c71/include/boost/math/special_functions/beta.hpp#L666)?

I haven't looked into the case of x close to 1 yet.

jzmaddock:
> That seems like a reasonable fix to me. Would this entail implementing a log version of [this function](https://github.com/boostorg/math/blob/625d76783b963de14fcd4f2b86070d4d02eb6c71/include/boost/math/special_functions/beta.hpp#L753) or the [Lanczos approximation](https://github.com/boostorg/math/blob/625d76783b963de14fcd4f2b86070d4d02eb6c71/include/boost/math/special_functions/beta.hpp#L666)?

Either: if we simplify things in the log world where the log of beta(a, b) is calculated using lgamma calls then they should both simplify to the same thing.  I'm assuming that we're far enough out in the tail that the lgamma calls won't suffer cancellation error (or if they do they get swamped by the rest of the terms anyway, ie that the result is large and negative).

On reflection for the near 1 case we have no problem: if we express that as `log(ibeta(a, b, x)) = log1p(-ibetac(a, b, x))` then since $log1p(x) \approx x$ there should be no cases where ibetac underflows but the log does not.

So hopefully the only question we need to answer, is are we far enough in the left tail (x small enough) that we should take the log branch?  Looking at https://functions.wolfram.com/GammaBetaErf/Beta3/06/01/03/01/01/ that should be the case whenever `bz` is small.  We might need to just check that large bz but $z^a / a$ underflowing isn't also an issue.

emilyhunt:
Coming from [this thread](https://discuss.scientific-python.org/t/new-special-functions-for-logarithm-of-incomplete-gamma-functions/2314/5?u=emilyhunt) on the scientific Python forums to mention that it would be great to see this!

Although computing in the left tail would require log-ifying the series, at least when $x$ is small it ought to converge very quickly, so it should be eminently computable. In fact, for many cases, I think only the first term of the series is even needed. Right now, as a workaround, I'm using the first term of the continued fraction approximation (works within the domain I need high precision for), and I don't think it'd be *too* horrid to formally log-ify the full series.

dschmitz89:
Would it help to implement the logarithm of the beta function first? That exists in many libraries such as SciPy as well. From what I understand, we need it for the incomplete beta anyway.

JacobHass8:
> Would it help to implement the logarithm of the beta function first? That exists in many libraries such as SciPy as well. From what I understand, we need it for the incomplete beta anyway.

I don't totally understand the internals of the incomplete beta, but I think the log of the beta would be helpful! I can review if you open a PR.

PR #1360, opened 2026-02-12
Fix conditions that disable real_concept test in test_next.cpp

mbs-c:
This test did not fail due to an MSVC bug, but due to a buggy ucrtbase in Windows Server 2022.

See my comments in #720.


mbs-c:
I have not been able to verify yet, but according to my knowledge, the underlying problem should have been fixed with an update for Windows Server 2022 this year. Therefore, the workaround could potentially be removed entirely if the build environment is up-to-date.

PR #1363, opened 2026-02-17
Fixes ibeta for large arguments

### LINKED ISSUE #1361: `ibeta` series fails to converge for large a,b and x close to a/(a+b)
When $a, b \gg 1$ and $x \approx a / (a+b)$, the incomplete beta function (`ibeta`) will throw an error because the series expansion doesn't converge. This was first found in https://github.com/scipy/scipy/issues/24566

```C++
typedef double T;

T a = 3.1622776601699636e+16; 
T b = 3.130654883566682e+18;
T x = 0.010000000000005001;

std::cout << boost::math::ibeta(a, b, x + 1e-12) << std::endl; // 0.50713
std::cout << boost::math::ibeta(a, b, x - 1e-12) << std::endl; // 0.49287

// Error in function boost::math::ibeta: Series evaluation exceeded 1000000 iterations, giving up now.
std::cout << std::setprecision(64) << boost::math::ibeta(a, b, x + 1e-13)<< std::endl; 
return 0;
```
The issue is with continued fraction in [here](https://github.com/boostorg/math/blob/c42193db07aef6f69a439815a9310729282587c7/include/boost/math/special_functions/beta.hpp#L860C1-L860C118). I've tried using the `erf` asymptotic expansion [here](https://github.com/boostorg/math/blob/c42193db07aef6f69a439815a9310729282587c7/include/boost/math/special_functions/beta.hpp#L1136C1-L1136C109), but it gave horrible results. 

I came up with a tentative solution in Python that is somewhat accurate, but it feels quite hacky. In the asymptotic expansion above ([here](https://github.com/boostorg/math/blob/c42193db07aef6f69a439815a9310729282587c7/include/boost/math/special_functions/beta.hpp#L1136C1-L1136C109)), it solves for 
```math
\eta^2 = -2 \left( x_0 \ln\left(\frac{x}{x_0}\right) + (1-x_0) \ln\left(\frac{1-x}{1-x_0}\right) \right)
```
where $x_0 = a/(a+b)$. Some care needs to be taken to see if $\eta$ is positive or negative, but this is already handled in the implementation. When $x \approx x_0$, we can Taylor expand about $x_0$ to find the approximate solution 
```math
\eta \approx \frac{x - x_0}{\sqrt{x_0(1-x_0)}}.
```
This taylor approximation actually works fairly well for the example above. When more terms are include in the continued fraction, it seems to be converging to the result in the function `ibeta_large_ab` with the taylor approximation for $\eta$. 

I only see two issues in regards to using the taylor expansion for $\eta$. First, why does the taylor expansion of $\eta$ give a more accurate result for `ibeta` than using the entire equation for $\eta$? Is it some sort of precision issue with the logs? Second, if we _should_ taylor expand $\eta$, in what regime is this accurate? I found that for $x_0 \pm \delta$ where $\delta < 10^{-13}$ the series fails to converge and the expansion works well. Should $10^{-13}$ just be hard coded into the function [here](https://github.com/boostorg/math/blob/c42193db07aef6f69a439815a9310729282587c7/include/boost/math/special_functions/beta.hpp#L1136C1-L1136C109)?

JacobHass8:
Closes #1361 and https://github.com/scipy/scipy/issues/24566.

Only the expansion for $\eta$ has been added. When to use the erf expansion and relevant tests need to be added.

### DIFF
--- a/include/boost/math/special_functions/beta.hpp
+++ b/include/boost/math/special_functions/beta.hpp
@@ -816,25 +816,25 @@ struct ibeta_fraction2_t
 {
    typedef boost::math::pair<T, T> result_type;
 
-   BOOST_MATH_GPU_ENABLED ibeta_fraction2_t(T a_, T b_, T x_, T y_) : a(a_), b(b_), x(x_), y(y_), m(static_cast<T>(0)) {}
+   BOOST_MATH_GPU_ENABLED ibeta_fraction2_t(T a_, T b_, T x_, T y_) : a(a_), b(b_), x(x_), y(y_), m(0) {}
 
    BOOST_MATH_GPU_ENABLED result_type operator()()
    {
       T denom = (a + 2 * m - 1);
       T aN = (m * (a + m - 1) / denom) * ((a + b + m - 1) / denom) * (b - m) * x * x;
 
-      T bN = m;
+      T bN = static_cast<T>(m);
       bN += (m * (b - m) * x) / (a + 2*m - 1);
       bN += ((a + m) * (a * y - b * x + 1 + m *(2 - x))) / (a + 2*m + 1);
 
-      m += 1;
+      ++m;
 
       return boost::math::make_pair(aN, bN);
    }
 
 private:
    T a, b, x, y;
-   T m;
+   int m;
 };
 //
 // Evaluate the incomplete beta via the continued fraction representation:


JacobHass8:
> In this case non-convergence isn't exceptional, it's somewhat expected right? Passing a modified policy to `ibeta_fraction2` like `ignore_error` would keep you from throwing at that check. Then this works:

I think that should work! I'm not very familiar with policies though. How would I modify the policy to `ignore_error` on within `ibeta_imp`?

JacobHass8:
@jzmaddock it looks like everything is squared away _except_ macos. I'm not really sure what to do about this because I'm running linux. The tests aren't even close to correct. The monotonicty test is failing with the error
```bash
Tests run with Clang version 17.0.0 (clang-1700.0.13.5), libc++ version 190102, Mac OS
Testing spot values with type double
./test_ibeta.hpp:508: error: in "test_main": ibeta not monotonically increasing above a/(a+b) for ibeta(10000000272564224, 9965820922822656, 0.50085594650495679) and delta=1e-10: [0.72326017914593033 >= 0.80645954095361849]
./test_ibeta.hpp:508: error: in "test_main": ibeta not monotonically increasing above a/(a+b) for ibeta(31622776601699636, 3.1306548835666821e+18, 0.010000000000005001) and delta=9.9999999999999998e-13: [0.63702741474523428 >= 0.9052874276409012]
./test_ibeta.hpp:510: error: in "test_main": ibeta not monotonically increasing below a/(a+b) for ibeta(31622776601699636, 3.1306548835666821e+18, 0.010000000000005001) and delta=9.9999999999999998e-13: [0.069925125891538167 >= 0.75500647219182093]
```
For reference, here are the tests
<p>

```C++
T a_values[2] = {10000000272564224, 3.1622776601699636e+16};
T b_values[2] = {9965820922822656, 3.130654883566682e+18};
T delta[8] = {0, 1e-15, 1e-14, 1e-13, 1e-12, 1e-11, 1e-10, 1e-9};
T a, b, x;
for (unsigned int i=0; i<2; i++){
  a = a_values[i];
  b = b_values[i];
  x = a / (a+b); // roughly the median of ibeta

  for (unsigned int j=0; j < 7; j++)
  {
     BOOST_CHECK_MESSAGE(boost::math::ibeta(a, b, x + delta[j+1]) > boost::math::ibeta(a, b, x + delta[j]), 
        "ibeta not monotonically increasing above a/(a+b) for ibeta(" << a << ", " << b << ", " << x << ") and delta=" << delta[j] << ": [" << boost::math::ibeta(a, b, x + delta[j+1]) << " >= " << boost::math::ibeta(a, b, x + delta[j]) << "]");
     BOOST_CHECK_MESSAGE(boost::math::ibeta(a, b, x - delta[j]) > boost::math::ibeta(a, b, x - delta[j+1]), 
        "ibeta not monotonically increasing below a/(a+b) for ibeta(" << a << ", " << b << ", " << x << ") and delta=" << delta[j] << ": [" << boost::math::ibeta(a, b, x - delta[j]) << " >= " << boost::math::ibeta(a, b, x - delta[j+1]) << "]");  
  }
}
```
</p>
</details> 
The only thing I can think of is printin more debug information in the log files and seeing what is going wrong. I just want to see if we're using the "erf" approximation or not. I don't want to clog up the github CI pipeline though. Do you have any thoughts?

JacobHass8:
Well, I think I figure out what is going wrong on `macos`. The value calculated [here](https://github.com/boostorg/math/pull/1363/changes#diff-d8163512afb1a16df99e5f6e7d1592401e653ab218055705745d3c37c6cbe8a7R1609) `T local_result = ibeta_power_terms(a, b, x, y, lanczos_type(), normalised, pol);` is very wrong. That's what leads to the overall answer being incorrect. I also checked that `ibeta_power_terms` is using the lanczos approximation on `macos`. 

I'm not sure what to make of this since this implementation works on every other distribution. @jzmaddock do you think it is worth hunting down what the issue is or should I just skip the tests on `macos`? I don't like the idea of skipping the tests for `macos`, but I'm using Ubuntu and don't have a good way of compiling my code with the `macos` distribution.

JacobHass8:
> No indeed, I'm not sure what to suggest.
> 
> I think the changes are a worthy improvement, and the code is in better shape than it was. This is a good thing, and worth including. However, I am wary of including tests which may or may not pass depending on the whim of the platform getting at least one bit correct! Do we have enough tests if we exclude the problematic ones? Maybe back off the size of a and b in the tests a bit - as long as they cover the new code, there's no need to be right out at the bleeding edge. Add some comments to the tests and code as well so that if this comes up again in 10 years we'll have some clue as to what's happening.

Well, I added a ton of comments to the tests and removed the failing tests. Everything should run properly on all architectures now.

jzmaddock:
I don't understand why that would suddenly start failing but it is a conceptual error that needs fixing: hopefully a fix has been pushed now.

JacobHass8:
> I don't understand why that would suddenly start failing but it is a conceptual error that needs fixing: hopefully a fix has been pushed now.

Looks like it's been fixed now. There's only one failing test but it doesn't look like it's associated with this PR.

include/boost/math/special_functions/beta.hpp
```diff
@@ -818,25 +816,25 @@ struct ibeta_fraction2_t
...
-      ++m;
+      m += 1;
 
       return boost::math::make_pair(aN, bN);
    }
 
 private:
    T a, b, x, y;
-   int m;
+   T m;
```
jzmaddock: @JacobHass8 was there a reason to change m from int to T?  It's the reason for the previous failure, but I only just spotted it :(  Generally it makes sense to keep counters as integers I guess.

include/boost/math/special_functions/beta.hpp
```diff
@@ -1181,7 +1192,7 @@ BOOST_MATH_GPU_ENABLED T ibeta_large_ab(T a, T b, T x, T y, bool invert, bool no
    T mul = 1;
    if (!normalised)
       mul = boost::math::beta(a, b, pol);
-
+   // T log_erf_remainder = -0.5 * log(2 * constants::pi<T>() * (a+b)) + a * log(x / x0) + b * log((1-x) / (1-x0)) + log(abs(1/nu - sqrt(x0 * (1-x0)) / (x-x0)));
```
jzmaddock, line 1195: Looks like we could remove this stray line too?

include/boost/math/special_functions/beta.hpp
JacobHass8: I had previously thought there was a precision issue converting the `int` to type `T`. I'll change the code back.

include/boost/math/special_functions/beta.hpp
JacobHass8: Yes, this isn't used anywhere. I left it in, but commented it out, in case anyone in the future wants to go back and look at how accurate the erf approximation is. I'll leave the decision up to you.

JacobHass8:
@jzmaddock I reverted any changes I made when I changed `int m` to `T m`. It looks like this fixed all the conceptual errors as you said. Besides the stray comment above, I think this is ready to merge. I don't care either way about deleting or keeping the line above.

PR #1368, opened 2026-02-21
Find degrees of freedom for `non_central_f` distribution

JacobHass8:
Towards #1305. I've added functions to find parameters `df1` and `df2` because they are nearly identical. 

All the tests are passing, but changing the initial guess for the degrees of freedom that is passed to `bracket_and_solve` causes most tests to fail. Thus, I'm not super confident in the current implementation. I will add more spot checks for a wider range of `df1`/`df2` to make sure things don't break. Perhaps the initial guess should change based on if the different of `p - cdf` is positive or negative? For now, I'm going to leave it though. 

I also had some difficulty with seeing if the function passed to `bracket_and_solve` was rising or falling. I ultimately found that the following worked 
```C++
RealType guess = 1; // Changing this to 20, 200 causes errors! 
std::pair<RealType, RealType> ir = tools::bracket_and_solve_root(
   f, guess, RealType(2), f(guess) < 0 ? true : false, tol, max_iter, pol);
```
where 
```C++
f_degrees_of_freedom_finder<RealType, Policy> f(x, v, nc, find_v1, p < q ? p : q, p < q ? false : true);
```
Essentially, if the difference between `cdf-p` (or is negative then `f` is increasing, and if it is positive then `f` is decreasing. I'm not entirely sure why this works though. 

Could all of this rising/falling be avoided by minimizing the squared difference `(cdf-p)^2`? Maybe this will open up a whole new can of worms though.


JacobHass8:
> So we have two roots, and it's not clear to me that we can reliably determine which way the function slopes?
> 
> One thing I thought of was 1) determine if there is a root in [1, INF), if yes, then find it, otherwise 2) determine if there is a root in (0, 1], if yes then find it, otherwise throw.
> 
> But it's not clear that this will always work, or if 1 is the correct pivot point in all cases.

I'm not entirely sure, but this might be whats going on. If you look at your plot, there is an inflection point at `x=2` where `x` is the first degree of freedom. When `x<2` the function is always decreasing and when `x>2` the function is always increasing. Thus, `x=2` is a special point. For example here's a plot of the cdf for the case you were looking at (nc=1, df2=2)

Figure_1

I think the behavior of the CDF changes when _either_ v1 or v2 = 2. Maybe this makes sense because the mean and variance aren't defined when `v2<2`. 

Then again, I'm not sure why other test cases passed if this is the case. I'll try to dig down and figure out what is going on. As you said, maybe we should check for roots on [0,2] and [2, inf]. I'm surprised that this never came up as an issue in scipy. @dschmitz89 do you know how cdflib handles multiple degrees of freedom being possible?

JacobHass8:
From sparse testing, it seems like there are two roots when $x < \nu_2, \nu_1$. See the plots below. Sorry it's a bit of a data dump. 
<p>

Figure_1
Figure_2
Figure_3
Figure_4
Figure_5
Figure_6

</p>
</details>

JacobHass8:
> > From sparse testing, it seems like there are two roots when x < ν 2 , ν 1 .
> 
> Close, but:
> 
> double v1 = 5; double v2 = 3.5; double nc = 0; double x = 3.5000000001;
> 
> Has two roots and x > v2.

You're right! Even going to x=3.51 there are two roots.

steppi:
`ncfdtridfd` can be useful for sample size calculations for ANOVA. `dfn` would be the number of treatments. `p` the desired power `nc` calculated based on the desired effect size. `x` the critical value from the central F distribution for the desired type 1 error rate. Then `dfd` would be the number of replicates to choose per treatment to achieve the desired power to detect the given effect size at the given type 1 error rate. I suspect that in most (all?) practical examples arising from real sample size calculations, there will only be one root.

I'm struggling to think of a use for `ncfdtridfn` though. The effect size of interest is mixed into `dfn` and `nc`, (`nc` is essentially the expected value of the numerator sum of squares, which depends on both the number of treatments and the treatment effect). So `ncfdtridfn` isn't what one would need to do power calculations for how many groups one can hope to investigate given a fixed sample size or things like that. I'd be reluctant to get rid of it in SciPy just because I can't think of an application offhand though.

JacobHass8:
This also appears to be an issue with the F-distribution when solving for `dfn`. See the below example and plot. Essentially, trying to invert to find `dfn` can yield 2 roots rather than 1. 

Figure_1

```python
dfn = 5
x = 3.51
dfd = 3.5
p = f.cdf(x, dfn, dfd)

dfn = np.linspace(0, 10, num=1000)
cdf1 = f.cdf(x, dfn, dfd)

fig, ax = plt.subplots()
ax.plot(dfn, -(p-cdf1))
ax.hlines(0, 0, 10, ls='--', color='k')
ax.set_xlabel('dfn')
ax.set_ylabel("cdf-p")
plt.show()
```
I'll try to investigate if there are any regimes where we can get 2 roots when inverting for `dfd`.

JacobHass8:
Yup, unfortunately this also happens when inverting for `dfd`.
Figure_1
```python
dfn = 6
x = 0.5
dfd = 4
p = f.cdf(x, dfn, dfd)

dfd = np.linspace(0, 10, num=1000)
cdf1 = f.cdf(x, dfn, dfd)

fig, ax = plt.subplots()
ax.plot(dfd, (p-cdf1))
ax.hlines(0, 0, 10, ls='--', color='k')
ax.set_xlabel('dfd')
ax.set_ylabel("p-cdf")
plt.show()
```

jzmaddock:
>In any case, I'm not sure how the boost maintainers feel about only including one root. CDFLIB in scipy returns the upper root, whereas there are several test cases already implemented in boost that want the lower root. I could simply ignore these test cases, but that feels like cheating.

Apologies for the radio silence - I think we left a lot of these out because we couldn't decide what people would actually _want_.  If there are already tests looking for the lower root, those are probably a mistake, simply because we might not have realised there were two available.

If there's a way to guarantee to find the upper root (when available), then I'm happy if we always return that.  I guess the question is whether we can always bracket the upper root?  There may also be arguments where we just graze the origin and the upper and lower roots are impossibly close together?  I'm not sure that our (or any) root finding algorithms are really designed for that.

dschmitz89:
@JacobHass8 @jzmaddock  I had a look at this for the regular F distribution with LLM assistance. It boils down to finding out if the CDF is monotonic with respect to the degrees of freedom parameter. I could not come up with a way to decide this using some rule but we can do so numerically by simply searching the maximum CDF value via optimization and comparing it to the asymptotic value.

Scipy Notation: dfn -> v1, dfd -> v2

Asymptotically, for $dfd\to\infty F(x|dfn, dfd)=\chi^2(dfn * x | dfn)$ the F distribution becomes a chi squared distribution. If the maximal value for the CDF is larger than this, we have a non monotonic case. See the figure below.

<p>

```python
from scipy.stats import f, chi2
from scipy.optimize import minimize_scalar
import numpy as np
import matplotlib.pyplot as plt

dfn = 4
x = 2
dfd = 8
p = f.cdf(x, dfn, dfd)
print(f"CDF at x={x}, dfn={dfn}, dfd={dfd} is {p:.6g}")
dfd = np.linspace(0, 100, num=10000)
cdf1 = f.cdf(x, dfn, dfd)

# asymptotic value as dfd -> inf is chi2.cdf(x*dfn, dfn)
p_inf_dfd = chi2.cdf(x=x*dfn, df=dfn)

def neg_cdf(dfd_):
	return -f.cdf(x, dfn, dfd_)

res = minimize_scalar(neg_cdf, bounds=(1e-20, 1e2), method='bounded')
max_cdf = f.cdf(x, dfn, res.x)

print(f"CDF asymptotic  = {p_inf_dfd:.6g}")
print(f"CDF maximum   = {max_cdf:.6g} at dfd = {res.x:.4g}")

# Optional: plot CDF, asymptote, and maximum
fig2, ax2 = plt.subplots()
ax2.plot(dfd, cdf1, label='CDF(x; dfn, dfd)')
ax2.axhline(p_inf_dfd, color='r', ls='--', label='Infinite dfd limit')
ax2.axhline(max_cdf, color='g', ls=':', label='Max CDF')
ax2.axvline(res.x, color='g', ls=':')
ax2.set_xlabel('dfd')
ax2.set_ylabel('CDF')
ax2.set_title(f'CDF of F-distribution at x={x}, dfn={dfn}')
ax2.legend()
plt.show()
``` 

</p>
</details> 

image

image

Suppose we have solved this problem then and we know if the problem has one or two roots. What to do with the two roots case? I think there are two possibilities:

* return `NaN`
* return a helpful error message

What do you think?

jzmaddock:
Thanks @dschmitz89 for looking into this, just thinking out loud here, I wonder if we've missed the obvious: lets call the function we're root finding `F(df)`, then based on the graphs above we have three cases:

1) `F(small) > 0` and `F(large) > 0` : this is the case with two roots.
2) `F(small) < 0` and `F(large) > 0` : one root and the "regular case". 
3) `F(small) > 0` and `F(large) < 0` : this is the case with one root very close to zero.

Cases 1 and 3 are simple enough, and now we have the direction of the slope for the bracketing routine as well.  Something like `large = sqrt(tools::max_value<T>())` and `small = 1 / large` might do the trick?

So now we just need to deal with (1), options are:

1) Return the result of `policies::evaluation_error(args...)` : this gives us either a NaN or an exception depending on what the user has requested.
2) Start at an arbitrary value (say 1) and use doubling/halving to bracket one particular root (the upper one?) and return that.  Unless there are cases with no root at all?

BTW I'm not too bothered about preserving current behaviour, as that's wrong (50% of the time) anyway.

I'm also assuming that the forth option for F(small) and F(large) doesn't happen in this case ;)

dschmitz89:
> Thanks @dschmitz89 for looking into this, just thinking out loud here, I wonder if we've missed the obvious: lets call the function we're root finding `F(df)`, then based on the graphs above we have three cases:
> 
> 1. `F(small) > 0` and `F(large) > 0` : this is the case with two roots.
> 2. `F(small) < 0` and `F(large) > 0` : one root and the "regular case".
> 3. `F(small) > 0` and `F(large) < 0` : this is the case with one root very close to zero.
> 
> Cases 1 and 3 are simple enough, and now we have the direction of the slope for the bracketing routine as well. Something like `large = sqrt(tools::max_value<T>())` and `small = 1 / large` might do the trick?
> 
> So now we just need to deal with (1), options are:
> 
> 1. Return the result of `policies::evaluation_error(args...)` : this gives us either a NaN or an exception depending on what the user has requested.
> 2. Start at an arbitrary value (say 1) and use doubling/halving to bracket one particular root (the upper one?) and return that.  Unless there are cases with no root at all?
> 
> BTW I'm not too bothered about preserving current behaviour, as that's wrong (50% of the time) anyway.
> 
> I'm also assuming that the forth option for F(small) and F(large) doesn't happen in this case ;)

I would go with the evaluation error. This will otherwise quickly lead to silent fails in actual statistical workflows.

You mentioned current behavior: which functions in boost are currently affected by a similar problem? Do we have other parameter finders with ambigous results?

jzmaddock:
>You mentioned current behavior: which functions in boost are currently affected by a similar problem? Do we have other parameter finders with ambigous results?

I'm not sure, I dimly remember @JacobHass8 mentioning one but I could be wrong.  It's one of those things that could lurk unnoticed for years until someone finds the one test case that breaks.

JacobHass8:
> > You mentioned current behavior: which functions in boost are currently affected by a similar problem? Do we have other parameter finders with ambigous results?
> 
> I'm not sure, I dimly remember @JacobHass8 mentioning one but I could be wrong. It's one of those things that could lurk unnoticed for years until someone finds the one test case that breaks.

I think that the F-distribution has similar issues (https://github.com/boostorg/math/pull/1368#issuecomment-3948523635, https://github.com/boostorg/math/pull/1368#issuecomment-3948613793)  but the "find degrees of freedom" functions aren't defined for the F-distribution in boost. Maybe the noncentral chi-squared distribution would have this issue as well?

JacobHass8:
> Thanks @dschmitz89 for looking into this, just thinking out loud here, I wonder if we've missed the obvious: lets call the function we're root finding `F(df)`, then based on the graphs above we have three cases:
> 
>     1. `F(small) > 0` and `F(large) > 0` : this is the case with two roots.
> 
>     2. `F(small) < 0` and `F(large) > 0` : one root and the "regular case".
> 
>     3. `F(small) > 0` and `F(large) < 0` : this is the case with one root very close to zero.
> 
> 
> Cases 1 and 3 are simple enough, and now we have the direction of the slope for the bracketing routine as well. Something like `large = sqrt(tools::max_value<T>())` and `small = 1 / large` might do the trick?
> 
> So now we just need to deal with (1), options are:
> 
>     1. Return the result of `policies::evaluation_error(args...)` : this gives us either a NaN or an exception depending on what the user has requested.
> 
>     2. Start at an arbitrary value (say 1) and use doubling/halving to bracket one particular root (the upper one?) and return that.  Unless there are cases with no root at all?
> 
> 
> BTW I'm not too bothered about preserving current behaviour, as that's wrong (50% of the time) anyway.
> 
> I'm also assuming that the forth option for F(small) and F(large) doesn't happen in this case ;)

I implemented a check to see if F(small) and F(large) have different signs. If this is the case, then there should be only one root. If they have the same sign, then there is either no root or two roots. In this case, the function throws an evaluation error. 

I only wrote one spot check because that was the one I found that had one root. The previously implemented spot checks sometimes have two roots.

include/boost/math/distributions/non_central_f.hpp
```diff
@@ -80,6 +81,83 @@ namespace boost
...
+
+            // There are times when f has two roots - thus, two degrees of freedom can
+            // be found. We find this case by checking if the sign of f for large and 
+            // small values of v have the same sign. If the sign is the same, then there 
+            // are an even number of roots. If the signs differ, there is only one root
+            // and we can safely find the root.
+            RealType vLarge = sqrt(boost::math::tools::max_value<RealType>());
+            RealType vSmall = 1 / vLarge;
+
+            if ((f(vLarge) < 0) == (f(vSmall) < 0)){
```
dschmitz89: Can we be confident that `f(vLarge)` is well-behaved? For many distributions computations fail for such extreme values. An alternative would be to use an asymptomatic expression similar to https://github.com/boostorg/math/pull/1368#issuecomment-4366384872.

JacobHass8: For the sparse testing I did here, `f(vLarge)` only breaks for `boost::math::real_concept` types. That's not to say it's perfect otherwise. 

Does the asymptotic chi-squared distribution not involve `vLarge` which makes it more stable? That seems reasonable to me. I wonder if there is a similar expansion for dfn -> 0.

dschmitz89: I just asked the LLM of my choice to extend my script for the F distribution and it confirmed that the noncentral F asymptotically becomes a noncentral chi squared distribution. For the `v2` case we get

image

<p>

```python
from scipy.stats import ncf, ncx2
from scipy.optimize import minimize_scalar
import numpy as np
import matplotlib.pyplot as plt

dfn = 4
nc = 2.0  # noncentrality parameter
x = 0.3
dfd_vals = np.linspace(1e-5, 200, num=1000)
cdf_vals = ncf.cdf(x, dfn, dfd_vals, nc)

# asymptotic value as dfd -> inf is ncx2.cdf(x*dfn, dfn, nc)
p_inf_dfd = ncx2.cdf(x=x*dfn, df=dfn, nc=nc)

def neg_cdf(dfd_):
    # Using a small lower bound as cdf is often not defined for dfd <= 0
    return -ncf.cdf(x, dfn, dfd_, nc)

res = minimize_scalar(neg_cdf, bounds=(1e-5, 1e4), method='bounded')
max_cdf = ncf.cdf(x, dfn, res.x, nc)

print(f"Noncentral F CDF at x={x}, dfn={dfn}, nc={nc}")
print(f"CDF asymptotic (dfd -> inf) = {p_inf_dfd:.6g}")
print(f"CDF maximum                = {max_cdf:.6g} at dfd = {res.x:.4g}")

# Optional: plot CDF, asymptote, and maximum
fig, ax = plt.subplots()
ax.plot(dfd_vals, cdf_vals, label='ncf.cdf(x; dfn, dfd, nc)')
ax.axhline(p_inf_dfd, color='r', ls='--', label='Infinite dfd limit (ncx2)')
ax.axhline(max_cdf, color='g', ls=':', label='Max CDF')
ax.axvline(res.x, color='g', ls=':')
ax.set_xlabel('dfd')
ax.set_ylabel('CDF')
ax.set_title(f'CDF of Noncentral F-distribution at x={x}, dfn={dfn}, nc={nc}')
ax.legend()
plt.show()
``` 

</p>
</details> 

This should be more stable than choosing some arbitrary large value in my opinion.

JacobHass8: > This should be more stable than choosing some arbitrary large value in my opinion.

Yeah, that should be _way_ more stable. Good catch! I'll implement that and hopefully it will fix the `real_concept` type issues.

JacobHass8: I added this approximation in so I'm going to close this.

JacobHass8:
@jzmaddock I think this is ready to merge if you have the chance to review it!

include/boost/math/distributions/non_central_f.hpp
```diff
@@ -107,6 +108,16 @@ namespace boost
             bool comp;
          };
 
+         template <class RealType, class Policy>
+         RealType large_v2_approximation(RealType x, RealType v1, RealType p, RealType q)
+         { // For v2 -> inf approximate f_degreese_of_freedome_finder with chi squared distribution 
+           // with degrees of freedom v1 at the cdf at x * v1
+               bool comp = p < q ? false : true;
+               RealType pval =  p < q ? p : q;
+               chi_squared_distribution<RealType, Policy> d(v1);
```
dschmitz89: Should this be the noncentral chi squared or the regular chi squared distribution? I have not formally checked the math.

JacobHass8: You're right. Here's a thread that shows it https://stats.stackexchange.com/questions/571682/what-is-f-k-infty-i-e-f-distribution-when-the-second-degree-of-freedom

include/boost/math/distributions/non_central_f.hpp
```diff
@@ -80,6 +82,103 @@ namespace boost
...
+            }
+            f_degrees_of_freedom_finder<RealType, Policy> f(x, v, nc, find_v1, p < q ? p : q, p < q ? false : true);
+
+            // There are times when f has two roots - thus, two degrees of freedom can
+            // be found. We find this case by checking if the sign of f for large and 
+            // small values of v have the same sign. If the sign is the same, then there 
+            // are an even number of roots. If the signs differ, there is only one root
+            // and we can safely find the root.
+            RealType vLarge = sqrt(boost::math::tools::max_value<RealType>());
+            RealType vSmall = 1 / vLarge;
```
dschmitz89, line 149: Are we using an asymptotic expression for `v1` but not for `v2`? 

It might be easier to split this over two PRs instead of one to make review easier.

JacobHass8: Yes, we're only using the asymptotic expression when solving for `v1`. This is because I can't find an expression for the distribution as `v1` -> infinity. 

We could split it up, but the asymptotic approximation is the only place where these functions differ. Otherwise, you just need to swap the order of `v1` and `v2`.

test/test_nc_f.cpp
```diff
@@ -369,6 +369,48 @@ void test_spots(RealType, const char* name = nullptr)
...
+   // Check find_v1/v2 edge cases
+   // Case when P=1 or P=0
+   nc = 2;
+   BOOST_MATH_CHECK_THROW(dist.find_v1(x, b, nc, 1), std::domain_error);
+   BOOST_MATH_CHECK_THROW(dist.find_v1(x, b, nc, 0), std::domain_error);
+   // Case when Q=1 or Q=0
+   BOOST_MATH_CHECK_THROW(dist.find_v1(boost::math::complement(x, b, nc, 1)), std::domain_error);
+   BOOST_MATH_CHECK_THROW(dist.find_v1(boost::math::complement(x, b, nc, 0)), std::domain_error);
+   // Check very small values of x an evaluation error is thrown
+   x = boost::math::tools::epsilon<long double>() / 10;
```
jzmaddock: Use of `epsilon<long double>` is tripping up the SYCL tests since long double is not supported on the device, maybe just hard code a small value like 1e-35 here?

jzmaddock:
Apart from my one comment on `long double` this all looks good to go to me.  Thanks @JacobHass8 !

dschmitz89:
Thanks a lot again @JacobHass8 for pushing this close to the finish line. Could you tell how to catch the ambiguous cases for the future SciPy wrapper?

test/test_nc_f.cpp
JacobHass8: I'm not sure why this was `long double` in the first place. I changed it to `RealType`. Let's see if that changes things.

jzmaddock:
Many thanks @JacobHass8 for pushing this one to completion, merging...

PR #1382, opened 2026-03-18
[FIX] Legacy libraries with shared linkage

philippremy:
Hi!

Thank you again for adding the legacy modules to CMake. I noticed a small issue when trying it out: Compiling the legacy modules as shared libraries failed on my macOS machine, because the `BOOST_SYMBOL_EXPORT` macro was not defined properly. The fix is as simple as (just like with MSVC) explicitly setting the export visibility to `"default"`. It seems that Boost and CMake default to hidden visibility, so in order for the shared library to expose the symbols correctly this must be adapted accordingly. I also changed the `if(MSVC)` to `if(WIN32)`, which makes it compatible with MinGW based toolchains as well.

I also added a CI workflow for the shared libraries (it just mirrors the one for the static linkage).

And theoretically it should also be possible to compile the legacy libraries in standalone mode, so I guarded any `Boost::config` logic behind the appropriate flag.

If any of these changes (or all of them) are not wanted / not fitting the project, feel free to cherry-pick or close the PR altogether.


mborland (approved):
These all seem like reasonable changes to me. Thanks for giving it a go and working out the bugs!

PR #1385, opened 2026-04-12
ENH: add parameter finder for degrees or freedom for students_t distribution

dschmitz89:
Towards #1305

This adds a parameter finder for the student t distribution with respect to the degrees or freedom.

**Disclaimer**: this PR is heavily LLM supported as C++ is not (yet?) my forte.

I verified the math with a simple Python program before. I also ran a simple sanity script to see that the function actually executes. I am not so confident about the tests though as I cannot really decipher the output of `b2`.

**Approach**: We use `bracket_and_solve_root` with an initial guess from a second order Edgeworth expansion of the CDF. The initial guess is very good for $df > 1$ where the t distribution behaves similar to the normal distribution but useless for low degrees of freedom (see the plot below). In this case or if it gives a relative error of the CDF worse than 10%, we fall back to an initial guess of 0.01.

t_cdf_edgeworth

<details><summary>Detailed approach for finding the initial guess</summary>
<p>

Given a quantile $x$ and a CDF value $p = P(T \leq x)$, we want to recover the degrees of freedom $\nu$.

**Step 1 — Edgeworth warm start.**

We use the 2nd-order Edgeworth expansion of the $t$ CDF in powers of $1/\nu$:

$$F(x;\nu) \approx \Phi(x) - \frac{\varphi(x)(x + x^3)}{4\nu} + \frac{\varphi(x)(3x + 5x^3 + 7x^5 - 3x^7)}{96\nu^2}$$
where $\Phi(x)$ is the standard normal CDF and $\phi(x)$ the standard normal PDF.
Setting $u = 1/\nu$ and $F(x;\nu) = p$ yields the quadratic

$$B u^2 - A u + (\Phi(x) - p) = 0$$

where

$$A = \frac{\varphi(x)(x + x^3)}{4}, \qquad B = \frac{\varphi(x)(3x + 5x^3 + 7x^5 - 3x^7)}{96}$$

The physically meaningful root (smallest positive $u$, i.e. largest $\nu$) gives a closed-form starting estimate $\hat\nu = 1/u$.

**Step 2 — Validation.**

We plug $\hat\nu$ into the exact $t$ CDF. If the relative residual $|F(x;\hat\nu) - p|/|p|$ exceeds 10%, we fall back to a safe low starting value $\nu_0 = 10^{-2}$.

</p>
</details>


dschmitz89:
CC @JacobHass8 would you have time to help out with a cursory initial review?

test/test_students_t.cpp
```diff
@@ -548,6 +548,103 @@ void test_spots(RealType)
...
+       // we use root finding
+       RealType tol_inv = boost::math::tools::epsilon<RealType>() * 500000;
+       if (boost::math::tools::digits<RealType>() > 100)
+          tol_inv *= 50000;
+       BOOST_CHECK_CLOSE(
+          students_t_distribution<RealType>::invert_probability_with_respect_to_degrees_of_freedom(
+             static_cast<RealType>(-6.96455673428326),
+             static_cast<RealType>(0.01)),
+          static_cast<RealType>(2),
+          tol_inv);
```
JacobHass8: I see these test values are used in quite a few places. It would be nice to combine all these into a single function which checks every method. This ensures we only have to define the values for the spot check in one place. It also cuts down on the amount of code. Maybe this would be for a separate PR though?

include/boost/math/distributions/students_t.hpp
```diff
@@ -344,6 +365,111 @@ BOOST_MATH_GPU_ENABLED RealType students_t_distribution<RealType, Policy>::find_
...
+         students_t_distribution<RealType, Policy> t_hat(nu_hat);
+         RealType exact_cdf = cdf(t_hat, x_abs);
+         RealType residual = (exact_cdf > p_adj) ? (exact_cdf - p_adj) : (p_adj - exact_cdf);
+         RealType relative_residual = (p_adj != 0) ? residual / p_adj : residual;
+         if (relative_residual <= tools::epsilon<RealType>() * 4)
+             return nu_hat; // Edgeworth estimate is already exact to machine precision.
+         if (relative_residual <= static_cast<RealType>(0.1))
+            hint = nu_hat;
+      }
+   }
```
JacobHass8: With all this in the function, it was initially hard for me to parse what was going on. Could this be put in a separate function called something like `boost::math::detail::edgeworth_approximation`? Then you could replace all this  with `hint = boost::math::detail::edgeworth_approximation(x_abs, p)`.

include/boost/math/distributions/students_t.hpp
```diff
@@ -344,6 +365,111 @@ BOOST_MATH_GPU_ENABLED RealType students_t_distribution<RealType, Policy>::find_
    return result;
 }
 
+template <class RealType, class Policy>
+BOOST_MATH_GPU_ENABLED RealType students_t_distribution<RealType, Policy>::invert_probability_with_respect_to_degrees_of_freedom(
```
jzmaddock, line 435: We already have a function for this here: https://github.com/boostorg/math/blob/625d76783b963de14fcd4f2b86070d4d02eb6c71/include/boost/math/distributions/students_t.hpp#L314

However, you have a better initial guess to the root finder which would be a very welcome addition!

include/boost/math/distributions/students_t.hpp
```diff
@@ -344,6 +365,111 @@ BOOST_MATH_GPU_ENABLED RealType students_t_distribution<RealType, Policy>::find_
...
+
+   // Use symmetry: CDF(-x; df) = 1 - CDF(x; df), so always work with x >= 0.
+   RealType x_abs = (x < 0) ? -x : x;
+   RealType p_adj = (x < 0) ? (1 - p) : p;
+
+   if (x_abs == 0)
+   {
+      // CDF(0; df) = 0.5 for all df; only solvable when p == 0.5.
+      if (p_adj == static_cast<RealType>(0.5))
+         return boost::math::numeric_limits<RealType>::infinity();
```
jzmaddock: Small point, this should return policies::raise_overflow_error for consistency.

include/boost/math/distributions/students_t.hpp
```diff
@@ -344,6 +365,111 @@ BOOST_MATH_GPU_ENABLED RealType students_t_distribution<RealType, Policy>::find_
...
+   normal_distribution<RealType, Policy> std_normal(0, 1);
+   RealType phi = pdf(std_normal, x_abs);
+   RealType Phi = cdf(std_normal, x_abs);
+
+   RealType x2 = x_abs * x_abs;
+   RealType x3 = x2 * x_abs;
+   RealType x5 = x3 * x2;
+   RealType x7 = x5 * x2;
+
+   RealType A = phi * (x_abs + x3) / 4;
```
jzmaddock: If an LLM wrote most of this then colour me quite impressed ;)

A small point of C++ etiquette: variables are normally all_lower_case and upper case is reserved for macros.

include/boost/math/distributions/students_t.hpp
jzmaddock: Maybe refactor out, and use in the existing function?  That would get automatically tested in the existing tests as well.

include/boost/math/distributions/students_t.hpp
dschmitz89: I found this function but was not sure how to map the inversion problem we have here (finding degrees of freedom given a probability and a quantile) to this one arising from sample size calculation.

The signatures are:
`find_degrees_of_freedom(RealType difference_from_mean, RealType alpha, RealType beta, RealType sd, RealType hint)`
`invert_probability_with_respect_to_degrees_of_freedom(RealType x, RealType p)`

include/boost/math/distributions/students_t.hpp
dschmitz89: My workflow was basically:

* work out the edgeworth expansion with claude opus iteratively (it got the signs wrong in the beginning, I simply plotted to check the correctness
* asked claude sonnet to invert the expression
* worked out a simple python implementation to see that it works 
* asked sonnet again to translate the python implementation to something that follows boost idioms
* iterate on the PR code with manual interventions to get what we have now

include/boost/math/distributions/students_t.hpp
jzmaddock: Ah OK, slightly different problems, I would still overload the existing function though for consistency and call it:

find_degrees_of_freedom(RealType t_stat, RealType p)

dschmitz89:
I am satisfied with the changes here now and would be happy to apply the new initial guess to the existing `find_degrees_of_freedom` function in a follow up.

I am not sure what to make of the failing tests as they pass locally for me on ARM/macOS. They cover two edge cases:
* one where the Edgeworth approximation is not computable
* one where the Edgeworth approximation yields a very bad result

From the b2 output I do not know if they only fail for float or also double and long double precision. Also, what is the difference between "CI /macos (clang, 14, github_ci_block_1" and "CI / macos (clang, 14, github_ci_block_2" for example? One fails, the other succeeds. Are they different architectures?

JacobHass8:
> From the b2 output I do not know if they only fail for float or also double and long double precision. Also, what is the difference between "CI /macos (clang, 14, github_ci_block_1" and "CI / macos (clang, 14, github_ci_block_2" for example? One fails, the other succeeds. Are they different architectures?

I think the difference between "github_ci_block_1" and "github_ci_block_2" is what tests are run. If you search for "test_student_t" in the raw logs of a "github_ci_block_2" run you'll see nothing comes up. The student_t distribution is only tested in "github_ci_block_1" which is why these are failing and "github_ci_block_2" passes. 

As far as the precision goes, it looks like the issue is with the float type. Here is the relevant output (for example on the MSVC2022 (msvc-14.3, latest, github_ci_block_1) architecture)
```
2026-04-18T11:24:46.9267469Z ====== BEGIN OUTPUT ======
2026-04-18T11:24:46.9276117Z Running 1 test case...
2026-04-18T11:24:46.9276769Z Tolerance for type f is 0.0001 %
2026-04-18T11:24:46.9299431Z test_students_t.cpp(637): [1;31;49merror: in "test_main": difference{1.70141e+37} between students_t_distribution<RealType>::invert_probability_with_respect_to_degrees_of_freedom( static_cast<RealType>(1e20), static_cast<RealType>(0.5244796002843015)){5.87746615e-41} and static_cast<RealType>(1e-3){0.00100000005} exceeds 0.1%[0;39;49m
2026-04-18T11:24:46.9306566Z test_students_t.cpp(644): [1;31;49merror: in "test_main": difference{22.8419} between students_t_distribution<RealType>::invert_probability_with_respect_to_degrees_of_freedom( static_cast<RealType>(2.0), static_cast<RealType>(0.500000000644961)){2.38418574e-09} and static_cast<RealType>(1e-10){1.00000001e-10} exceeds 0.1%[0;39;49m
2026-04-18T11:24:46.9309056Z Tolerance for type d is 0.0001 %
2026-04-18T11:24:46.9309600Z Tolerance for type e is 0.0001 %
2026-04-18T11:24:46.9310272Z Tolerance for type N5boost4math8concepts12real_conceptE is 0.0001 %
2026-04-18T11:24:46.9310883Z 
2026-04-18T11:24:46.9311413Z [1;31;49m*** 2 failures are detected in the test module "Master Test Suite"
2026-04-18T11:24:46.9313867Z [0;39;49m
2026-04-18T11:24:46.9324296Z EXIT STATUS: 201
2026-04-18T11:24:46.9324809Z ====== END OUTPUT ======
```
The output message isn't very helpful in identifying what type you're using because it's simply printing `typeid(RealType).name()`. Something is definitely going wrong in the first failing test because I think the returned value, `5.87746615e-41`, is close to the minimum, positive number a float can represent.

dschmitz89:
Thanks @JacobHass8 for the hint about the test precision. Looks like everything is passing now.

I did not add GPU tests as I do not have a GPU available and no CUDA experience.

dschmitz89:
@jzmaddock Gentle ping, would you have time to review? I am quite happy with the math here but the technicalities likely need some polishing.

include/boost/math/distributions/students_t.hpp
```diff
@@ -19,10 +19,13 @@
...
 #include <boost/math/special_functions/beta.hpp> // for ibeta(a, b, x).
 #include <boost/math/special_functions/digamma.hpp>
+#include <boost/math/special_functions/relative_difference.hpp>
 #include <boost/math/distributions/complement.hpp>
 #include <boost/math/distributions/detail/common_error_handling.hpp>
-#include <boost/math/distributions/normal.hpp> 
+#include <boost/math/distributions/normal.hpp>
+#include <boost/math/distributions/cauchy.hpp>
 #include <boost/math/policies/policy.hpp>
+#include <iostream>
```
dschmitz89:
```suggestion
```

include/boost/math/distributions/students_t.hpp
```diff
@@ -332,16 +447,62 @@ BOOST_MATH_GPU_ENABLED RealType students_t_distribution<RealType, Policy>::find_
...
+         function,
+         "No degrees of freedom can satisfy CDF(0; df) == %1% (must be 0.5).",
+         p, Policy());
    }
-   return result;
+
+   // Analytical cases: df = infinity (normal), df = 1 (cauchy)
+   RealType analytical_df;
+   boost::math::normal_distribution<RealType, Policy> norm(0, 1);
+   if (detail::analytical_df_if_cdf_matches(norm, x, p, std::numeric_limits<RealType>::infinity(), analytical_df))
```
jzmaddock: I'm really nit picking here, but there's no need for the last two parameters to analytical_df_if_cdf_matches, instead I would tend to write this as:

```
if (detail::analytical_df_if_cdf_matches(norm, x, p)
   return std::numeric_limits<RealType>::infinity();
```

Which also removes the need for the `analytical_df` variable?

Except, in this particular case, we have one more wrinkle in that strictly speaking type RealType may not have an infinity, and in any case, this is an overflow error so:

```
if (detail::analytical_df_if_cdf_matches(norm, x, p)
   return policies::raise_overflow_error<RealType>(function, nullptr, Policy());
```

And the next pair of lines become:

```
if (detail::analytical_df_if_cdf_matches(cauchy, x, p, static_cast<RealType>(1), analytical_df))
      return RealType(1);
```

and the function simplifies to:

```
template <class Distribution, class RealType>
BOOST_MATH_GPU_ENABLED inline bool analytical_df_if_cdf_matches(const Distribution& dist, RealType x, RealType p)
{
   RealType cdf_val = cdf(dist, x);
   return epsilon_difference(cdf_val, p) < 4;
}
```

doc/distributions/students_t.qbk
```diff
@@ -29,6 +29,11 @@
          RealType beta,
          RealType sd,
          RealType hint = 100);
+
+      // degrees of freedom inversion from a quantile and probability:
+      BOOST_MATH_GPU_ENABLED static RealType invert_probability_with_respect_to_degrees_of_freedom(
```
jzmaddock: Can we please rename this to find_degrees_of_freedom for consistency?  IMO It's fine to overload the existing function.

dschmitz89: That makes sense. I think you mean depending on the number of arguments one of the methods is chosen. Do you have a precedent for this in boost.math, especially regarding documentation?

dschmitz89: Should be done with the last commit.

dschmitz89:
Gentle bump here. Everything is passing and comments were addressed. I will look into using the Edgeworth approximation based guess for the other parameter finder in a followup, it should not be too difficult.

jzmaddock:
Apologies for not responding sooner @dschmitz89 many thanks for this, merging now...

PR #1390, opened 2026-04-28
Allow fallthrough of Android to 128-bit float

### LINKED ISSUE #1389: 64bit Android builds fail due incorrect long double assumption
Hello,

while compiling Boost 1.91.0 for Android the following error occurs in Boost Math:

```
[2026-04-25T11:36:24.906Z] clang-linux.compile.c++ /home/opc/iwra973m9n7wtpfh/_output/Applications_Android_x86_64_cxx20/boost/bin.v2/libs/math/build/clang-linux-android/debug/x86_64/link-static/pch-off/target-os-android/threading-multi/visibility-global/atanh.o
[2026-04-25T11:36:24.906Z] In file included from /home/opc/iwra973m9n7wtpfh/libs/math/build/../src/tr1/atanh.cpp:11:
[2026-04-25T11:36:24.906Z] In file included from ./boost/math/special_functions/atanh.hpp:22:
[2026-04-25T11:36:24.906Z] In file included from ./boost/math/special_functions/log1p.hpp:34:
[2026-04-25T11:36:24.906Z] In file included from ./boost/math/special_functions/fpclassify.hpp:21:
[2026-04-25T11:36:24.906Z] ./boost/math/special_functions/detail/fp_traits.hpp:314:1: error: static_assert failed due to requirement '113 == 64' "Oops, assumption that long double is an 80-bit quantity is incorrect!!"
[2026-04-25T11:36:24.906Z] static_assert(LDBL_MANT_DIG == 64, "Oops, assumption that long double is an 80-bit quantity is incorrect!!");
[2026-04-25T11:36:24.906Z] ^             ~~~~~~~~~~~~~~~~~~~
[2026-04-25T11:36:24.906Z] 1 error generated.
```

Using Android NDK 25 I dumped the value of `LDBL_MANT_DIG` for all architectures:
```
$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android21-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 113
$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android33-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 113

$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android19-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 53
$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android33-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 53

$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android21-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 113
$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android33-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 113

$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi19-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 53
$ $ANDROID_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi33-clang -dM -E - < /dev/null |grep LDBL_MANT_DIG
#define __LDBL_MANT_DIG__ 53
```

For each architecture I used the oldest and latest ABI version. As you can see the 64 bit Intel and ARM architectures use `113` instead of the expected `64`. If you need something to differentiate the Android platform, the `__ANDROID__` preprocessor symbol is set to 1 within the toolchain defaults.

Thanks,
Gregor

mborland:
Closes: #1389

### DIFF
--- a/.drone/drone.sh
+++ b/.drone/drone.sh
@@ -51,8 +51,8 @@ echo '==================================> BEFORE_SCRIPT'
 echo '==================================> SCRIPT'
 
 echo "using $TOOLSET : : $COMPILER : <cxxflags>-std=$CXXSTD $OPTIONS ;" > ~/user-config.jam
-(cd libs/config/test && ../../../b2 config_info_travis_install toolset=$TOOLSET && ./config_info_travis)
-(cd libs/math/test && ../../../b2 toolset=$TOOLSET $TEST_SUITE define=$DEFINE)
+(cd libs/config/test && ../../../b2 print_config_info print_math_info toolset=$TOOLSET)
+(cd libs/math/test && ../../../b2 -j3 toolset=$TOOLSET $TEST_SUITE)
 
 echo '==================================> AFTER_SUCCESS'
 
--- a/include/boost/math/special_functions/detail/fp_traits.hpp
+++ b/include/boost/math/special_functions/detail/fp_traits.hpp
@@ -270,13 +270,7 @@ template<> struct fp_traits_non_native<double, double_precision>
 
 // long double (64 bits) -------------------------------------------------------
 
-#if defined(BOOST_NO_INT64_T) || defined(BOOST_NO_INCLASS_MEMBER_INITIALIZATION)\
-   || defined(BOOST_BORLANDC) || defined(__CODEGEAR__) || (defined(__APPLE__) && defined(__aarch64__)) || defined(_MSC_VER)\
-    || (defined(__GNUC__) && defined(__aarch64__) && defined(_WIN32))\
-     || (defined(__GNUC__) && (defined(__arm__) || defined(__thumb__)))\
-      || defined(__SYCL_DEVICE_ONLY__) || (LDBL_MANT_DIG == 53)
-
-static_assert(LDBL_MANT_DIG == 53, "Oops, assumption that long double is a 64-bit quantity is incorrect!!");
+#if LDBL_MANT_DIG == 53 || defined(__SYCL_DEVICE_ONLY__)
 
 template<> struct fp_traits_non_native<long double, double_precision>
 {
@@ -305,16 +299,9 @@ private:
 
 // long double (>64 bits), x86 and x64 -----------------------------------------
 
-// Andorid is a known exception because on x64 it reports LDBL_MANT_DIG == 113
-// See: https://github.com/boostorg/math/issues/1389
-
-#elif (defined(__i386) || defined(__i386__) || defined(_M_IX86) \
-    || defined(__amd64) || defined(__amd64__)  || defined(_M_AMD64) \
-    || defined(__x86_64) || defined(__x86_64__) || defined(_M_X64)) && !defined(__ANDROID__)
+#elif LDBL_MANT_DIG == 64
 
 // Intel extended double precision format (80 bits)
-
-static_assert(LDBL_MANT_DIG == 64, "Oops, assumption that long double is an 80-bit quantity is incorrect!!");
 
 template<>
 struct fp_traits_non_native<long double, extended_double_precision>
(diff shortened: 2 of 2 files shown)


Martchus:
I noticed the same problem in my Android builds and came up with basically the same minimal fix.

However, I'm wondering whether it would make more sense to determine the floating point format *using* `LDBL_MANT_DIG` instead of making assertions on what that value is for each and every platform. This is actually already done further down in that file (`#elif (LDBL_MANT_DIG == 113)`) where the assertion (`static_assert(LDBL_MANT_DIG == 113`) is then completely redundant. There's also `#elif defined(__GNUC__) && (LDBL_MANT_DIG == 106)` further down which also *uses* `LDBL_MANT_DIG` and even already avoids redundant assertions.

mborland:
> I noticed the same problem in my Android builds and came up with basically the same minimal fix.
> 
> However, I'm wondering whether it would make more sense to determine the floating point format _using_ `LDBL_MANT_DIG` instead of making assertions on what that value is for each and every platform. This is actually already done further down in that file (`#elif (LDBL_MANT_DIG == 113)`) where the assertion (`static_assert(LDBL_MANT_DIG == 113`) is then completely redundant. There's also `#elif defined(__GNUC__) && (LDBL_MANT_DIG == 106)` further down which also _uses_ `LDBL_MANT_DIG` and even already avoids redundant assertions.

I was thinking the same thing. Now that we assert the value of `LDBL_MANT_DIG` at each point we might as well use that information exclusively so we don't end up like the current situation on different platforms. If the value is missing or an unexpected value, that should be our only point of assertion.

Martchus (approved):
Nice, that's what I had in mind.

mborland:
@jzmaddock do you have any opinions on this reorganization? I left in the special cases that we have no way of testing (e.g. Motorola 68000k)

jzmaddock:
I confess I was thinking of making much the same changes - it looks logical to me - however, I guess the question is do we have say 64-bit (long) doubles which are not in IEEE format, and will therefore cause the code logic to fail?  This is old code that we inherited, and was reviewed in it's own right as I recall, the intention was to select for platforms with known bit-layouts, the newly added static_asserts were supposed to just be a quick sanity double check that all was well.  So that worked well then!

mborland:
> I confess I was thinking of making much the same changes - it looks logical to me - however, I guess the question is do we have say 64-bit (long) doubles which are not in IEEE format, and will therefore cause the code logic to fail? This is old code that we inherited, and was reviewed in it's own right as I recall, the intention was to select for platforms with known bit-layouts, the newly added static_asserts were supposed to just be a quick sanity double check that all was well. So that worked well then!

The only mainstream non-IEEE long double I am aware of that's still floating around is `ibm128` on PPC64, and the newer operating systems are migrating those to the proper float128. There could be some complete odd ball system out there, but I think it's best to wait for a bug report from them to explain what's going on with their system.

mborland:
Maybe in a future iteration we could check `is_iec559` to alleviate your concerns John? I'm not quite sure how well that will go either. I'm going to merge this in, since it solves the issue for now and we seem to be blocking vcpkg.

PR #1391, opened 2026-04-30
constants: Protect F which is defined as a macro by some frameworks

cbicari:
For instance, ESP32 platform globally defines:

    esp32/WString.h: #define F(string_literal)  (FPSTR(PSTR(string_literal)))


mborland (approved):
Looks good to me. I approved the CI run, and will merge once it's green. Thank you!

PR #1406, opened 2026-06-11
pFq: prevent spurious overflow.

### LINKED ISSUE #1404: Unexpected exception from `hypergeometric_pFq({a}, {b}, x)` when `x` is large.
I'm exercising `hypergeometric_pFq` with this code:

```
#include <string>
#include <iostream>
#include <iomanip>
#include <boost/math/special_functions/hypergeometric_pFq.hpp>

using namespace boost::math;


int main(int argc, char *argv[])
{
    if (argc != 4 ) {
        printf("use: hyppfq a b x\n");
        return -1;
    }

    double a = std::stod(argv[1]);
    double b = std::stod(argv[2]);
    double x = std::stod(argv[3]);

    double y = hypergeometric_pFq({a}, {b}, x);
    std::cout << "pFq = " << std::scientific << std::setw(25)
        << std::setprecision(17) << y << std::endl;

    return 0;
}
```

Normal case:

```
$ ./hyppfq 2 3 10.0
pFq =   3.96478384306520911e+03
```

Overflow exception (working as expected):

```
$ ./hyppfq 2 3 10000.0
terminate called after throwing an instance of 'std::overflow_error'
  what():  Error in function boost::math::hypergeometric_pFq<double>(double,double,double): numeric overflow
Aborted (core dumped)
```

When the third argument is much larger, the exception `boost::math::rounding_error` is thrown when the code attempts to convert something to an integer:

```
$ ./hyppfq 2 3 10000000000.0
terminate called after throwing an instance of 'boost::math::rounding_error'
  what():  Error in function boost::math::itrunc<long double>(long double): Value 9999999999 can not be represented in the target integer type.
```

I would expect this case to throw `std::overflow_error` as in the previous example.

jzmaddock:
Make sure we throw the correct exceptions when the result is infinite. Fixes #1404.

### DIFF
--- a/include/boost/math/special_functions/detail/hypergeometric_pFq_checked_series.hpp
+++ b/include/boost/math/special_functions/detail/hypergeometric_pFq_checked_series.hpp
@@ -243,6 +243,9 @@
               log_scale += log_scaling_factor;
               local_scaling += log_scaling_factor;
            }
+           /*
+           * rescaling to avoid underflow is pointless here, given that the first term is 1.
+           * 
            else if ((fabs(abs_result) < lower_limit) && (fabs(abs_result) * scaling_factor < upper_limit))
            {
               abs_result *= scaling_factor;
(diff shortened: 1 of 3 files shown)


jzmaddock:
Missing coverage lines appear to be spurious (they're hit in the debugger), merging...

WarrenWeckesser:
Thanks @jzmaddock!

PR #1407, opened 2026-06-16
Initial implementation of fisher_f degrees of freedom finder

### LINKED ISSUE #1305: ENH: more inverses for distribution CDFs

JacobHass8:
This is a simple copy and paste from [the previous PR ](https://github.com/boostorg/math/pull/1368). I still need to implement a couple of tests. One set of tests to check if the function works as intended and the other to see if we properly catch the case when there are multiple degrees of freedom. 

As noted above, most of this is copy and pasted from the `non_central_f` distribution implementation of these same functions. I wonder if it would be more concise to program a special case of the `non_central_f` functions when `nc=0`. When `nc=0` we can just use the `fisher_f` distribution? Now that I type it out, I think I might go with that. 

Closes #1305.

### DIFF
--- a/include/boost/math/distributions/fisher_f.hpp
+++ b/include/boost/math/distributions/fisher_f.hpp
@@ -14,6 +14,7 @@
 #include <boost/math/distributions/fwd.hpp>
 #include <boost/math/special_functions/beta.hpp> // for incomplete beta.
 #include <boost/math/distributions/complement.hpp> // complements
+#include <boost/math/distributions/chi_squared.hpp>
 #include <boost/math/distributions/detail/common_error_handling.hpp> // error checks
 #include <boost/math/special_functions/fpclassify.hpp>
 
@@ -42,6 +43,26 @@ namespace boost{ namespace math{
          RealType p;
          bool comp;
       };
+
+      template <class RealType, class Policy>
+      RealType fisher_large_v1_approximation(RealType x, RealType v, RealType p, RealType q)
+      {     // For v1 -> inf approximate f_degreese_of_freedom_finder with chi squared distribution 
+            // with degrees of freedom v2 at the cdf at v2 / x
+            bool comp = p < q ? false : true;
+            RealType pval =  p < q ? p : q;
+            chi_squared_distribution<RealType, Policy> d(v);
+            return comp ? pval - (1 - cdf(complement(d, v / x))) : (1-cdf(d, v / x)) - pval;
+      }
+
+      template <class RealType, class Policy>
+      RealType fisher_large_v2_approximation(RealType x, RealType v, RealType p, RealType q)
+      {     // For v2 -> inf approximate f_degrees_of_freedom_finder with chi squared distribution 
+            // with degrees of freedom v1 at the cdf at x * v1
+            bool comp = p < q ? false : true;
+            RealType pval =  p < q ? p : q;
+            chi_squared_distribution<RealType, Policy> d(v);
+            return comp ? pval - cdf(complement(d, v * x)) : cdf(d, v * x) - pval;
+      }
 
       template <class RealType, class Policy>
       inline RealType find_degrees_of_freedom_fisher_f(
@@ -73,7 +94,15 @@ namespace boost{ namespace math{
          RealType vLarge = sqrt(boost::math::tools::max_value<RealType>());
          RealType vSmall = 1 / vLarge;
 
-         if ((f(vLarge) < 0) == (f(vSmall) < 0)){
+         RealType large_difference;
+         if (find_v1)
+         {
+            large_difference = fisher_large_v1_approximation<RealType, Policy>(x, v, p, q);
+         }
+         else
+            large_difference = fisher_large_v2_approximation<RealType, Policy>(x, v, p, q);
+
+         if ((large_difference < 0) == (f(vSmall) < 0)){
             return policies::raise_evaluation_error<RealType>(function, "Can't find degrees of freedom because two degrees of freedom can be found using the given parameters",
                RealType(std::numeric_limits<RealType>::quiet_NaN()), Policy()); // LCOV_EXCL_LINE 
          }
--- a/include/boost/math/distributions/fisher_f.hpp
+++ b/include/boost/math/distributions/fisher_f.hpp
@@ -51,7 +51,7 @@ namespace boost{ namespace math{
             bool comp = p < q ? false : true;
             RealType pval =  p < q ? p : q;
             chi_squared_distribution<RealType, Policy> d(v);
-            return comp ? pval - (1 - cdf(complement(d, v / x))) : (1-cdf(d, v / x)) - pval;
+            return comp ? pval - cdf(d, v / x) : cdf(complement(d, v / x)) - pval;
       }
 
       template <class RealType, class Policy>
--- a/include/boost/math/distributions/fisher_f.hpp
+++ b/include/boost/math/distributions/fisher_f.hpp
@@ -16,9 +16,85 @@
 #include <boost/math/distributions/complement.hpp> // complements
 #include <boost/math/distributions/detail/common_error_handling.hpp> // error checks
 #include <boost/math/special_functions/fpclassify.hpp>
-#include <boost/math/distributions/non_central_f.hpp> // for find_degrees_of_freedom_f
 
 namespace boost{ namespace math{ 
+   namespace detail{
+      template <class RealType, class Policy>
+      struct fisher_degrees_of_freedom_finder
+      {
+         fisher_degrees_of_freedom_finder(
+            RealType x_, RealType v_, bool find_v1_, RealType p_, bool c)
+            : x(x_), v(v_), find_v1(find_v1_), p(p_), comp(c) {}
+
+         RealType operator()(const RealType& input_v)
+         {
+            RealType v1 = find_v1 ? input_v : v; 
+            RealType v2 = find_v1 ? v : input_v; 
+            fisher_f_distribution<RealType, Policy> d(v1, v2);
+            return comp ?
+               p - cdf(complement(d, x))
+               : cdf(d, x) - p;
+         }
+      private:
+         RealType x;
+         RealType v;
+         bool find_v1;
+         RealType p;
+         bool comp;
+      };
+
+      template <class RealType, class Policy>
+      inline RealType find_degrees_of_freedom_fisher_f(
+         const RealType x, const RealType v, const RealType nc, const bool find_v1, const RealType p, const RealType q, const Policy& pol)
+      {
+         BOOST_MATH_STD_USING
+         using std::fabs;
+         const char* function = "fisher_f_distribution<%1%>::find_degrees_of_freedom_f";
+         if((p == 0) || (q == 0))
+         {
+            //
+            // Can't find a thing if one of p and q is zero:
+            //
+            return policies::raise_domain_error<RealType>(function, "Can't find degrees of freedom when the probability is 0 or 1, only possible answer is %1%",
+               RealType(std::numeric_limits<RealType>::quiet_NaN()), Policy()); // LCOV_EXCL_LINE
+         }
+         if (x < tools::epsilon<RealType>())
+         {
+            return policies::raise_evaluation_error<RealType>(function, "Can't find degrees of freedom when the abscissa value is very close to zero as all degrees of freedom generate the same CDF at x=0: try again further out in the tails!!",
+               RealType(std::numeric_limits<RealType>::quiet_NaN()), Policy()); // LCOV_EXCL_LINE
+         }
+         fisher_degrees_of_freedom_finder<RealType, Policy> f(x, v, find_v1, p < q ? p : q, p < q ? false : true);
+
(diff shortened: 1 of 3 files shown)


JacobHass8:
Hmmm, I'm not sure why this is failing on some of the platforms. The error message is 
```c++ 
../../../boost/math/distributions/fisher_f.hpp: In static member function ‘static RealType boost::math::fisher_f_distribution<RealType, Policy>::find_v1(RealType, RealType, RealType)’:
../../../boost/math/distributions/fisher_f.hpp:58:34: error: ‘find_degrees_of_freedom_f’ is not a member of ‘boost::math::detail’
   58 |       eval_type result = detail::find_degrees_of_freedom_f(
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~
../../../boost/math/distributions/fisher_f.hpp: In static member function ‘static RealType boost::math::fisher_f_distribution<RealType, Policy>::find_v1(const boost::math::complemented3_type<A, B, C>&)’:
../../../boost/math/distributions/fisher_f.hpp:81:34: error: ‘find_degrees_of_freedom_f’ is not a member of ‘boost::math::detail’
   81 |       eval_type result = detail::find_degrees_of_freedom_f(
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~
../../../boost/math/distributions/fisher_f.hpp: In static member function ‘static RealType boost::math::fisher_f_distribution<RealType, Policy>::find_v2(RealType, RealType, RealType)’:
../../../boost/math/distributions/fisher_f.hpp:103:34: error: ‘find_degrees_of_freedom_f’ is not a member of ‘boost::math::detail’
  103 |       eval_type result = detail::find_degrees_of_freedom_f(
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~
../../../boost/math/distributions/fisher_f.hpp: In static member function ‘static RealType boost::math::fisher_f_distribution<RealType, Policy>::find_v2(const boost::math::complemented3_type<A, B, C>&)’:
../../../boost/math/distributions/fisher_f.hpp:126:34: error: ‘find_degrees_of_freedom_f’ is not a member of ‘boost::math::detail’
  126 |       eval_type result = detail::find_degrees_of_freedom_f(
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~
```
but `find_degrees_of_freedom_f` is defined in `non_central_f.hpp` which I included in `fisher_f.hpp`. @jzmaddock do you have any ideas?

jzmaddock:
I suspect you've created a cyclic include - a really bad thing and all kinds of badness can ensue!  If I'm right and the non-central-f depends on the regular f (which I suspect it should), then either moving the common code into fisher_f.hpp or a common shared header of it's own would do it.

JacobHass8:
> I suspect you've created a cyclic include - a really bad thing and all kinds of badness can ensue! If I'm right and the non-central-f depends on the regular f (which I suspect it should), then either moving the common code into fisher_f.hpp or a common shared header of it's own would do it.

Hmmm, unfortunately I think I'm going to have to duplicate the code. I need the fisher and non-central fisher distributions to be able to access the function but both distributions need to used to create the function. Even putting it in a separate file would require cyclic including.

JacobHass8:
If all the tests pass, then this is probably ready to merge.

jzmaddock:
Looks like the tolerances are just slightly too tight on a couple of tests?

JacobHass8:
> Looks like the tolerances are just slightly too tight on a couple of tests?

The most recent CI run doesn't show any failures due to this PR unless I'm mistaken.

jzmaddock:
MacOS failure is spurious, syscl shows:

```
 8/63 Test  #8: run-boost_math-test_fisher_f ...........................***Failed    0.02 sec
Running 1 test case...
Tolerance = 0.2%.
Tolerance = 5.96046e-05%.
Tolerance = 5.96046e-05%.
Tolerance = 0.2%.
Tolerance = 1.11022e-13%.
Tolerance = 1.11022e-13%.
/home/runner/work/math/boost-root/libs/math/test/test_fisher_f.cpp(424): error: in "test_main": difference{1.49214e-14} between fisher_f_distribution<RealType>::find_v1(x, df2, P){10.000000000000149} and df1{10} exceeds 1.11022e-12%
/home/runner/work/math/boost-root/libs/math/test/test_fisher_f.cpp(425): error: in "test_main": difference{1.49214e-14} between fisher_f_distribution<RealType>::find_v1(boost::math::complement(x, df2, 1-P)){10.000000000000149} and df1{10} exceeds 1.11022e-12%

*** 2 failures are detected in the test module "Master Test Suite"
```

Drone failures are similar tolerance issues in test_fisher_f.

JacobHass8:
> MacOS failure is spurious, syscl shows:
> Drone failures are similar tolerance issues in test_fisher_f.

Oops, I didn't even realize syscl or Drone ran these tests. That's good to know for the future so I don't ignore these.

dschmitz89:
For the non-central f distribution we had one asymptomatic case where we could use the non-central chi squared distribution instead of some arbitrary large value. Let's do the same here with the regular chi squared distribution.

JacobHass8:
> For the non-central f distribution we had one asymptomatic case where we could use the non-central chi squared distribution instead of some arbitrary large value. Let's do the same here with the regular chi squared distribution.

That's a good idea. I tried to implement it this morning but have been having a lot of issues. I'm struggling to come up with test cases for when the inversion problem doesn't contain 2 roots.  

I think that there might be a bug somewhere. I've been having a lot of trouble with identifying 2 roots. I'm going to actually just plot the cdf for all of the tests that I've written.

JacobHass8:
As a sanity check, I just wanted to make sure that the cases that I chose only have one root when finding for df1 and df2. It appears that's the case based on the plots that I made. 

When solving for df1:
Figure_1

When solving for df2:
Figure_2

JacobHass8:
When testing for two roots I get the following plots

Solving for df1:
Figure_1

Solving for df2
Figure_2

I don't understand. It appears that the chi-squared approximation doesn't hold when there are multiple roots. I even went out to 1e100 and the cdf didn't converge. Here is my code if anyone is interested in reproducing the results. Maybe I have a bug? @dschmitz89 do you have any thoughts on this? I'm not familiar with this approximation. 

<p>

```python
import numpy as np
from matplotlib import pyplot as plt
from scipy.stats import f, chi2

if __name__ == '__main__':
    df1 = 6
    df2 = 4
    x  = 0.5

    p = f.cdf(x, df1, df2)

    df1_plot = np.geomspace(1e-1, 1e5, num=5000)
    fig, ax = plt.subplots()
    ax.set_xscale("log")
    ax.set_xlabel("df1")
    ax.set_ylabel("p-cdf(df1, df2, x)")
    ax.set_title(f"df1={df1}, df2={df2}, x={x}")
    ax.plot(df1_plot, p - f.cdf(x, df1_plot, df2))
    ax.hlines(0, min(df1_plot), max(df1_plot), ls='--', color='k')
    plt.show()

    df2_plot = np.geomspace(1e-1, 1e300, num=5000)
    fig, ax = plt.subplots()
    ax.set_xscale("log")
    ax.set_xlabel("df2")
    ax.set_ylabel("p-cdf(df1, df2, x)")
    ax.set_title(f"df1={df1}, df2={df2}, x={x}")
    ax.plot(df2_plot, p - f.cdf(x, df1, df2_plot))
    ax.hlines(0, min(df2_plot), max(df2_plot), ls='--', color='k')
    ax.hlines(p - chi2.cdf(x * df2, df2), min(df2_plot), max(df2_plot), ls='--', color='r', label='Chi-squared Approx')
    ax.legend()
    plt.show()
```

</p>
</details>

dschmitz89:
@JacobHass8 Copilot and I played around with your script. From what I see, the asymptotic formulas are correct. There is even one for the other degrees of freedom parameter.

F_Dist_Analysis


<details><summary>Script</summary>
<p>

```python
import numpy as np
from matplotlib import pyplot as plt
from scipy.stats import f, chi2

def analyze_all(df1_fixed=6, df2_fixed=4, x_vals=None):
    if x_vals is None:
        x_vals = [0.05, 0.5, 5.0]
    
    n_rows = len(x_vals)
    fig, axes = plt.subplots(n_rows, 2, figsize=(12, 4 * n_rows))
    
    # Ensure axes is always 2D
    if n_rows == 1:
        axes = axes.reshape(1, 2)
    
    df_lower_bound = 1e-1
    df_upper_bound = 1e5

    for i, x in enumerate(x_vals):
        # --- Row i, Col 0: df2 -> inf (Fixed df1) ---
        ax = axes[i, 0]
        df2_plot = np.geomspace(df_lower_bound, df_upper_bound, num=1000)
        limit_p_df2_inf = chi2.cdf(x * df1_fixed, df1_fixed)
        
        ax.set_xscale("log")
        ax.plot(df2_plot, f.cdf(x, df1_fixed, df2_plot), label="F-CDF")
        ax.hlines(limit_p_df2_inf, min(df2_plot), max(df2_plot), ls='--', color='r', label='Chi-sq limit')
        ax.set_title(f"df2 -> inf (x={x}, df1={df1_fixed})")
        ax.set_xlabel("df2")
        ax.set_ylabel("CDF")
        ax.legend()

        # --- Row i, Col 1: df1 -> inf (Fixed df2) ---
        ax = axes[i, 1]
        df1_plot = np.geomspace(df_lower_bound, df_upper_bound, num=1000)
        limit_p_df1_inf = chi2.sf(df2_fixed / x, df2_fixed)
        
        ax.set_xscale("log")
        ax.plot(df1_plot, f.cdf(x, df1_plot, df2_fixed), label="F-CDF")
        ax.hlines(limit_p_df1_inf, min(df1_plot), max(df1_plot), ls='--', color='g', label='Chi-sq limit')
        ax.set_title(f"df1 -> inf (x={x}, df2={df2_fixed})")
        ax.set_xlabel("df1")
        ax.set_ylabel("CDF")
        ax.legend()

    plt.tight_layout()
    plt.show()

if __name__ == '__main__':
    analyze_all(x_vals=[0.05, 0.5, 5])

``` 

</p>
</details> 

Let me know if there is anything else I can do to help with this PR.

JacobHass8:
> @JacobHass8 Copilot and I played around with your script. From what I see, the asymptotic formulas are correct. There is even one for the other degrees of freedom parameter.

Thanks for the help! I was getting confused on what the two different limits are when df1 or df2 -> inf. Everything looks correct now. I _think_ there might be a similar bug in `non_central_f_distribution` since I wrote that as well.

include/boost/math/distributions/fisher_f.hpp
```diff
@@ -14,10 +14,116 @@
...
+      };
+
+      template <class RealType, class Policy>
+      RealType fisher_large_v1_approximation(RealType x, RealType v, RealType p, RealType q)
+      {     // For v1 -> inf approximate f_degreese_of_freedom_finder with chi squared distribution 
+            // with degrees of freedom v2 at the cdf at v2 / x
+            bool comp = p < q ? false : true;
+            RealType pval =  p < q ? p : q;
+            chi_squared_distribution<RealType, Policy> d(v);
+            return comp ? pval - (1 - cdf(complement(d, v / x))) : (1-cdf(d, v / x)) - pval;
```
dschmitz89: Why not use the complement instead of 1-cdf(...) ?

JacobHass8: That was being lazy because I had copy and pasted the code. Good catch!

dschmitz89 (approved):

dschmitz89:
@jzmaddock If this PR looks good to you we could propagate the same asymptotic formulas to the noncentral F as well.

jzmaddock:
Looks good to me, are we all ready to merge?

dschmitz89:
> Looks good to me, are we all ready to merge?

From my side yes.

dschmitz89:
Can we press the green button here? Looks like everyone is happy with the PR. :)

PR #1411, opened 2026-06-21
Adding Symplectic Integrators

JacobHass8:
Adds symplectic solvers for ODE systems with a conserved quantity (i.e. energy). This was a requested scipy feature (see https://github.com/boostorg/math/issues/303 and https://github.com/scipy/scipy/issues/12690). I'd ultimately like to merge this into scipy using cython. I'm still working on this and have a couple of features I'd like to add:

- [x] 11th order method from [here](https://www.sciencedirect.com/science/article/pii/S0377042701004927?fr=RR-2&ref=pdf_download&rr=a0ef60fffb26b1ae)
- [x] Tests 
    - Implement Hénon–Heiles potential 
    - Add floating point tests
- [x] Add multiprecision support 
- [x] Update documentation for singleton types
- [x] Driver function which allows users to run a ode to a specific time with a specific method
- [x] Helper function for 1d integrator
- [x] Examples
- [x] Make sure the function signature is similar to other methods in boost

Does this seem like a good plan? I'd appreciate any input.


NAThompson:
@JacobHass8 : I think a good test is to identify the conserved quantity you wish to be conserved, and show that quantity is much _better_ preserved with a symplectic integrator than with (say) RK4.

I would also recommend an API for event detection and return a solution skeleton that can be interpolated as a Hermite spline, i.e., return ${t_k, y_k, dot{y}_k}_{k=0}^{n-1}$ rather than the typical solution skeleton {t_k, y_k, }_{k=0}^{n-1}.

JacobHass8:
> @JacobHass8 : I think a good test is to identify the conserved quantity you wish to be conserved, and show that quantity is much _better_ preserved with a symplectic integrator than with (say) RK4.

I've tried this on a harmonic oscillator and the energy fluctuations seem to be of order 1e-11 (for the 6th order method). I haven't checked RK4 (or any other method though!). 

> I would also recommend an API for event detection and return a solution skeleton that can be interpolated as a Hermite spline, i.e., return ${t_k, y_k, dot{y}_k}_{k=0}^{n-1}$ rather than the typical solution skeleton {t_k, y_k, }_{k=0}^{n-1}.

I'm not sure I totally understand what you mean. Are you saying return a third object that could be used to interpolate between different $t_k$?

NAThompson:
> Are you saying return a third object that could be used to interpolate between different t_k?

Exactly. Say you have an $\mathcal{O}(\delta t^6)$ accurate ODE stepper. That means you only need to store a very sparse set of $(t_k, y_k)$ pairs. And each ODE stepper does come with a "natural interpolant", so naively this is no loss of information. But the user gets this data back and has no clue how to interpolate it, or what the "natural interpolant" even is. So they resort to just using linear interpolation. This makes the display and subsequent use of the data _worse_ than if you used a less accurate ODE stepper.


However, you have to compute $\dot{y}_k = f(t_k, y_k)$ during the course of the computation anyway, so if you just store it you can at least give the user a way to display the data as a hermite spline. I've been trying to get this built up in H5Web-see [here](https://github.com/silx-kit/h5web/issues/1945).

Note that this is still inferior to the "natural interpolant", but given there seems to be no hope to get every ODE stepper's interpolant into the standard graphics packages, I think this is a reasonable compromise.

JacobHass8:
The methods implemented here do appear better than those in Scipy. I test a simple harmonic oscillator where the Hamiltonian is given by $H = p^2 + x^2$. Starting with the initial condition $x=1$ and $p=0$, the total energy of the system is $1$. I ran the system from time 0 to 1 with time steps of 0.05. Here's a plot of the position as a function of time. Both graphs match up pretty nicely so we know they are both performing as expected. 

ScipyComp

However, if we look $p^2 + x^2 -1$, which I call the energy loss, it's clear that the symplectic integrators implemented here outperform scipy's Runge-Kutta method. 

ScipyEnergyFluctuations

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,172 @@
...
+#include <string>
+#include <stdexcept>
+#include <boost/math/policies/policy.hpp>
+#include <boost/math/policies/error_handling.hpp>
+#include <boost/math/special_functions/fpclassify.hpp>
+
+namespace boost{ namespace math {namespace quadrature { 
+
+template <class RealType, class ReturnType, class Func>
+std::pair<ReturnType, ReturnType> second_order_yoshida(const ReturnType p0,
```
NAThompson: @JacobHass8 : It appears to me that if we wanted to return both objects of the same type, we would do a `std::array<ReturnType,2>`. But it also appears that these two objects have different units. Could it work with dimensioned types like mpunitz or boost::units? That would indeed force you into a `std::pair`.

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,172 @@
...
+                                                      const ReturnType q0, 
+                                                      RealType dt, 
+                                                      Func dHdp, 
+                                                      Func dHdq)
+{
+    ReturnType p = p0;
+    ReturnType q = q0;
+    
+    // Choosing "System A" solution
+    RealType w3 = 0.784513610477560;
```
NAThompson: If someone wanted to compute these to quad precision, would there be hope? Can we compute these quantities to many more digits or are they only known to double precision?

test/test_symplectic.cpp
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright Nick Thompson, 2017
```
NAThompson: your name and date

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+    if ((dt <= 0) || !(boost::math::isfinite)(dt))
+    {
+        RealType val = (boost::math::policies::raise_domain_error(function, "Time step must be positive and finite but got: dt = %1%.\n", dt, pol));
+        std::vector<ReturnType> nan_vec = {ReturnType(val)};
+        return std::make_pair(nan_vec, nan_vec);
+    }
+
+    // Check if method is available
+    std::vector<std::string> available_methods = {"Y6", "Y4", "Y2"};
```
NAThompson: Should we do an `enum` of available methods rather than a vector of strings?

test/Jamfile.v2
```diff
@@ -1341,6 +1341,7 @@ test-suite quadrature :
    [ run test_trapezoidal.cpp /boost/test//boost_unit_test_framework : : :
       release [ requires cxx11_lambdas cxx11_auto_declarations cxx11_decltype cxx11_unified_initialization_syntax cxx11_variadic_templates ]
       $(float128_type) ]
+   [ run test_symplectic.cpp /boost/test//boost_unit_test_framework ]
```
NAThompson: I don't think you used boost unit test framework here right? You just used the boost::math floating point comparisons.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: I wasn't quite sure how to implement the `enum` of available methods. I'm just not very familiar with `enum` types. Instead, I used a `map`. Is this okay or is there an easier way with `enum`?

test/Jamfile.v2
JacobHass8: I'm not entirely sure what the unit test framework is. I just copied it from test_trapezoidal...

include/boost/math/quadrature/symplectic.hpp
JacobHass8: I don't think the return object necessarily need to be the same type. It is just returning the input types of `p0` and `q0`.

JacobHass8: However, there is a constraint on these types. Either we need to be able to add the types of `p0` and `q0` or we need the function `dHdp` to take in a type of `p0` and return a type of `q0` (for `dHdq` we need `q0` -> `p0`). 

This is because of the lines 
```c++
    ReturnType p = p0;
    ReturnType q = q0;

    q = q + dt / 2 * dHdp(p);
    p = p - dt * dHdq(q);
```
I'm not familiar with mpunitz or boost::units. I was only thinking about singletons versus array types.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: It was quite the slog but I actually computed them using mathematica. Here is the code to replicate it to 64 digits 
```mathematica
FindRoot[{w0+2(w1+w2+w3)==1,
w0^3 + 2(w1^3 + w2^3+w3^3)==0, 
w0^5 + 2(w1^5 + w2^5+w3^5)==0, 
1/6(w0^2w1^3-w0^4*w1) - 1/6(w0^3w1^2-w0*w1^4)+1/6((w0+2w1)^2w2^3-(w0+2w1)(w0^3+2w1^3)w2) - 1/6((w0^3+2w1^3)w2^2-(w0+2w1)w2^4) +1/6((w0+2w1+2w2)^2w3^3-(w0+2w1+2w2)(w0^3+2w1^3+2w2^3)w3)-1/6((w0^3+2w1^3+2w2^3)w3^2-(w0+2w1+2w2)w3^4)==0}, {{w0,1.3151863206839063}, {w1,-1.17767998417887}, {w2,0.235573213359357}, {w3,0.784513610477560}}, WorkingPrecision->64]
```
The solution is then
```mathematica
w0=1.31518632068391121888424972823886251435195350615940796180785516777853373846773`64.
w1=-1.17767998417887100694641568096431573463926925263459848447536851379674155618156`64
w2=0.23557321335935813368479318297853460168646808210340111900349313095621471215223`64.
w3=0.78451361047755726381949763386634987577682441745149338456794779895125997479548`64.
```

include/boost/math/quadrature/symplectic.hpp
NAThompson: Actually, you're targeting a scipy wrap correct? So this question should be answered in the context of "what is the most expressive way to do this in python via nanobind?"

JacobHass8: Scipy actually uses cython to bind c++. I think strings are probably okay for now. This is the same behavior as `scipy.integrate.solve_ivp`.

JacobHass8:
@NAThompson I created a .qbk file to add documentation for the functions I added. I couldn't figure out how to build the documentation though. I kept getting errors when trying to follow the boost math guide. 

I wanted to include an equation so I created a .svg and .png using latex. However, I noticed that all other equations have a .mml file. Do you know how to convert these?

doc/quadrature/symplectic.qbk
```diff
@@ -0,0 +1,103 @@
...
+
+Note that the functional can be readily generalized to multiple coordinates by changing the 
+signature of `dHdp` to 
+
+    std::valarray<double> dHdp(std::valarray<double> p)
+    {
+        // calculate the partial derivatives with respect to each p_i
+    }
+
+The function must return a `valarray` type, as opposed to `std::vector`, because we must be
```
NAThompson: @JacobHass8 : I believe that `std::valarray` has been essentially deprecated in the C++ community. @mborland : Is this correct?

In general, we like to use `template typename <RandomAccessContainer>`.

JacobHass8: I only used `std::valarray` because it supports vectorized arithmetic. I think Eigen vectors would be okay too. 

Maybe this shouldn't be a requirement though? I would need to either convert to an `std::valarray` or just write for loops. The for loops option isn't bad but I thought it would obfuscate the algorithm.

NAThompson:
> I wanted to include an equation so I created a .svg and .png using latex. However, I noticed that all other equations have a .mml file. Do you know how to convert these?

I just use [this website](https://viereck.ch/latex-to-svg/) and drop pngs in it.

I also experienced errors while building the docs-I think @mborland may have to help based on your particular error.

doc/quadrature/symplectic.qbk
NAThompson: Yeah, I'd personally just leave it generic and do a for loop.

jzmaddock:
>I created a .qbk file to add documentation for the functions I added. I couldn't figure out how to build the documentation though. I kept getting errors when trying to follow the boost math guide.

What are the errors you are getting?

Unfortunately there is a rather long toolchain required for building the docs, we did have it all documented on the old wiki, but I have a hunch that's bitten the dust :(

>I wanted to include an equation so I created a .svg and .png using latex. However, I noticed that all other equations have a .mml file. Do you know how to convert these?

Don't worry about .mml, these are all a "historical artefact": when we started there was no easy way to convert LaTex to SVG or something web friendly (other than blocky pngs) but MML was the new kid on the block and there was a nice GUI editor and some scripts to convert the mml to SVG so that was what we used.  Nowadays those tools are all unmaintained, and LaTex is so much better... please just make sure that the original LaTex is archived somewhere (a comment in the docs next to the link including the SVG would be good)  that way if someone later spots an error it's so much easier/quicker to fix!

doc/quadrature/symplectic.qbk
JacobHass8: Hmmm, I can't figure out how to allow the function to accept both numeric types and iterators. I might just require the user to always input an iterator. I can't figure out how to use for loops for iterators and none for numeric types.

JacobHass8:
> What are the errors you are getting?
> 
> Unfortunately there is a rather long toolchain required for building the docs, we did have it all documented on the old wiki, but I have a hunch that's bitten the dust :(

Actually, no errors. I found the .xml file but don't know how to convert this to html. I just ran `b2` in the doc directory. 

> Don't worry about .mml, these are all a "historical artefact": when we started there was no easy way to convert LaTex to SVG or something web friendly (other than blocky pngs) but MML was the new kid on the block and there was a nice GUI editor and some scripts to convert the mml to SVG so that was what we used. Nowadays those tools are all unmaintained, and LaTex is so much better... please just make sure that the original LaTex is archived somewhere (a comment in the docs next to the link including the SVG would be good) that way if someone later spots an error it's so much easier/quicker to fix!

Sounds good! I added the latex command.

JacobHass8:
@jzmaddock or @NAThompson do you think Kahan summation might help the accuracy of these algorithms? Essentially, the algorithm is just summing together a bunch of small steps so I thought it might help. 

Edit: I implemented kahan summation, but it didn't do anything. I could have done it wrong, but I don't think that's the case. I'm only testing against the simple harmonic oscillator example. Thus, the error term just oscillates from a small negative to a small positive value.

doc/quadrature/symplectic.qbk
NAThompson: @JacobHass8 : Use:

```cpp
template<typename RandomAccessContainer>
auto foo(RandomAccessContainer::value_type dp, ...) {
     using RealType = RandomAccessContainer::value_type;
     ...
}
```

NAThompson:
@JacobHass8 : I personally don't see any reason Kahan summation is going to help here. I would try to compute the condition number of the ODE using the techniques in [Corless, A Graduate Introduction to Numerical Methods](https://link.springer.com/book/10.1007/978-1-4614-8453-0). This will tell you whether it's the _ODE itself_ that forces the inaccuracy, or it's the method that forces inaccuracy on the ODE.

doc/quadrature/symplectic.qbk
JacobHass8: > ```c++
> template<typename RandomAccessContainer>
> auto foo(RandomAccessContainer::value_type dp, ...) {
>      using RealType = RandomAccessContainer::value_type;
>      ...
> }
> ```
I tried to implement this in the newest commit. I don't quite understand how that allows us to use numeric types though. I don't think I've explained what I'm trying to do very well because I don't quite understand how to do it in C++. I have one function with the signature 
```c++
template <typename RandomAccessContainer, class Func>
std::pair<std::vector<RandomAccessContainer>, std::vector<RandomAccessContainer> > integrate_hamiltonian(const RandomAccessContainer p0,
                                                                                                         const RandomAccessContainer q0,
                                                                                                         const typename RandomAccessContainer::value_type dt,
                                                                                                         const unsigned steps,
                                                                                                         Func dHdp,
                                                                                                         Func dHdq)
```
I would like to define an identical function with the following signature
```c++
template <typename RealType, class Func>
std::pair<std::vector<RealType>, std::vector<RealType> > integrate_hamiltonian_singleton(const RealType p0,
                                                                                                         const RealType q0,
                                                                                                         const RealType dt,
                                                                                                         const unsigned steps,
                                                                                                         Func dHdp,
                                                                                                         Func dHdq)
```
here, RealType is something like double or float. 

I _could_ copy and paste all the code to convert my RandomAccessContainer implementation to singletons but this feels like bad practice. I tried wrapping the inputs of `integrate_hamiltonian_singleton` to a vector but then the signature of the function `dHdp` needs to change. In the singleton implementation, `dHdp` takes a RealType and returns a RealType. The signature needs to change to take in a `std::vector` and return a `std::vector`. I tried creating a lambda function and functor, but the signature wasn't right.

JacobHass8:
> @JacobHass8 : I personally don't see any reason Kahan summation is going to help here. I would try to compute the condition number of the ODE using the techniques in [Corless, A Graduate Introduction to Numerical Methods](https://link.springer.com/book/10.1007/978-1-4614-8453-0). This will tell you whether it's the _ODE itself_ that forces the inaccuracy, or it's the method that forces inaccuracy on the ODE.

Unfortunately, I haven't been able to find a copy of that book. I think I'll leave Kahan summation out for the first implementation. After this ports to `scipy` maybe someone will find need for it but I'll leave it out for now.

JacobHass8:
On another note, is there any need to do policy based up/down casting on the input/result? From my understanding, boost policy casting is only implemented numeric types and not containers.

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -49,17 +49,18 @@ std::pair<ReturnType, ReturnType> second_order_yoshida(const ReturnType p0,
...
 }
 
-template <class RealType, class ReturnType, class Func>
-std::pair<ReturnType, ReturnType> fourth_order_yoshida(const ReturnType p0, 
-                                                       const ReturnType q0, 
-                                                       const RealType dt, 
-                                                       Func dHdp, 
-                                                       Func dHdq)
+template <typename RandomAccessContainer, class Func>
+std::pair<RandomAccessContainer, RandomAccessContainer> fourth_order_yoshida(const RandomAccessContainer p0,
```
NAThompson: Is it not?

```cpp
std::pair<RandomAccessContainer::value_type, RandomAccessContainer::value_type> 
```

doc/quadrature/symplectic.qbk
NAThompson: Generally you don't want to force people to use `std::vector` because they might prefer `Eigen::VectorXd` right?

 `std::vector` as an instance of `RandomAccessContainer`, so `std::vector<RandomAccessContainer>` is a vector of vectors.

JacobHass8: I've started trying to separate the code to see where it differs for say `doubles` and `std::vectors`. The only issues are when I need to add two values together. If the input q is a vector I have to do,
```c++
for (unsigned j=0; j < q.size(); j++){
    q[j] = q[j] + b * dt * dHdp_val[j];
}
```
but for doubles there doesn't need to be a for loop. Is there any way we could create a function `add` which can take in either two vectors or two doubles and add them together? I think this is probably SFINAE but can't figure it out. 

I'd like to do something like this but I don't think the compiler will be able to resolve which function to use 
```c++
// Add two vectors
template <class RandomAccessContainer>
void add(RandomAccessContainer& vec1, RandomAccessContainer& vec2)
{
    for (unsigned i=0; i < vec1.size(); i++){
        vec1[i] = vec1[i] + vec2[i]
    }
}

// Add two doubles
template <typename RealType>
void add(RealType& x, RealType& y)
{
    x = x + y;
}
```

NAThompson: @JacobHass8 : I personally wouldn't worry too much about that and just write out the for loops explicitly. But of course that is just a matter of taste.

JacobHass8: Nvm, I think I figured it out!

JacobHass8: I just added support for adding arrays or floating point types! I'm going to close this then.

JacobHass8:
@NAThompson, @jzmaddock, @mborland if all the tests come back passing, I think this PR is ready for review. There are only two small issues that I can think of . 

First, if the timestep, `dt`, is less than 0 an error is thrown according to the policy. However, if the policy wants to silently return an `nan`, this isn't handled. The algorithm just proceeds as normally. I don't know why anyone would want to run this without throwing an error though.

JacobHass8:
@NAThompson, @jzmaddock or @mborland gentle ping to see if anyone would be interested in reviewing. I can also make plots of more systems with my algorithm vs scipy if that would be helpful.

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+add(T x, U y) 
+{
+    return x + y;
+}
+
+template <typename T, typename U>
+typename std::enable_if<!has_plus<T, U>::value, T>::type
+add(T vec1, U vec2) 
+{
+    for (unsigned i=0; i < vec1.size(); i++)
```
mborland:
```suggestion
    for (std::size_t i=0; i < vec1.size(); i++)
```

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+template <typename T, typename U>
+typename std::enable_if<has_plus<T, U>::value, T>::type
+add(T x, U y) 
+{
+    return x + y;
+}
+
+template <typename T, typename U>
+typename std::enable_if<!has_plus<T, U>::value, T>::type
+add(T vec1, U vec2)
```
mborland: Is it necessary to make copies of both? It looks like the goal is to add in place so you could make this

```c++
template <typename T, typename U>
std::enable_if_t<!has<plus<T, U>::value, T>
add_in_place(T& vec1, const U& vec2)
```

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+{
+    for (unsigned i=0; i < vec1.size(); i++)
+    {
+        vec1[i] = vec1[i] + vec2[i];
+    }
+    return vec1;
+}
+
+template<typename...>
+using void_t = void;
```
mborland: Redefinition of `void_t`?

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+template <typename T, typename U>
+typename std::enable_if<has_mult<T, U>::value, T>::type
+mult_prefactor(T x, U prefactor) 
+{
+    return x * prefactor;
+}
+
+template <typename T, typename U>
+typename std::enable_if<!has_plus<T, U>::value, T>::type
+mult_prefactor(T vec1, U prefactor)
```
mborland:
```suggestion
mult_prefactor(T& vec1, U prefactor) 
```

Same comment about needing to make copies for in place operations

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+}
+
+template <typename RandomAccessContainer, typename RealType, class Func>
+std::pair<RandomAccessContainer, RandomAccessContainer> second_order_yoshida(const RandomAccessContainer p0, 
+                                                                             const RandomAccessContainer q0, 
+                                                                             const RealType dt, 
+                                                                             Func dHdp, 
+                                                                             Func dHdq)
+{
+    RandomAccessContainer p = p0;
```
mborland: This is a copy of a copy, but p0 is never used. Same for q0.

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+                                                                             Func dHdp, 
+                                                                             Func dHdq)
+{
+    BOOST_MATH_STD_USING
+
+    RandomAccessContainer p = p0;
+    RandomAccessContainer q = q0;
+
+    // RealType x0 = -(std::pow(2, 1/3) / (2 - std::pow(2, 1/3)));
+    RealType x1 = 1. / (2. - std::cbrt(2.));
```
mborland:
```suggestion
    RealType x1 = 1. / (2. - cbrt(2.));
```

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+    RealType x0 = 1. - 2. * x1; 
+
+    std::vector<RealType> weights = { x1, x0, x1 };
```
mborland:
```suggestion
    std::array<RealType, 3> weights = { x1, x0, x1 };
```

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+    RandomAccessContainer q = q0;
+    
+    RealType b1 = static_cast<RealType>(0.0829844064174052);
+    RealType b2 = static_cast<RealType>(0.396309801498368);
+    RealType b3 = static_cast<RealType>(-0.0390563049223486);
+    RealType b4 = 1. - 2. * (b1 + b2 + b3);
+
+    RealType a1 = static_cast<RealType>(0.245298957184271);
+    RealType a2 = static_cast<RealType>(0.604872665711080);
+    RealType a3 = 0.5 - (a1 + a2);
```
mborland, line 251: Can these be competed to higher precision like you did above?

include/boost/math/quadrature/symplectic.hpp
mborland: Also highly recommend going the enum route instead of a vector of strings and then a map. 

```c++
enum class available_methods : unsigned 
{
    Y6 = 0,
    Y4 = 1, 
    Y2 = 2 
}
```

And then you update the signature:

```c++
template <typename RandomAccessContainer, typename RealType, class Func, class Policy>
std::pair<std::vector<RandomAccessContainer>, std::vector<RandomAccessContainer> > integrate_hamiltonian_imp(const RandomAccessContainer p0,
                                                                                                             const RandomAccessContainer q0,
                                                                                                             const RealType dt,
                                                                                                             const unsigned steps,
                                                                                                             Func dHdp,
                                                                                                             Func dHdq,
                                                                                                             available_methods method,
                                                                                                             const Policy& pol)
```

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+template<typename T, typename U, typename = void>
+struct has_plus : std::false_type {};
+
+template<typename T, typename U>
+struct has_plus<T, U, void_t<decltype(std::declval<T>() + std::declval<U>())> > : std::true_type {};
+
+template <typename T, typename U>
+typename std::enable_if<has_plus<T, U>::value, T>::type
+add(T x, U y)
```
mborland:
```suggestion
template<typename T, typename U>
struct has_plus<T, U, void_t<decltype(std::declval<T>() + std::declval<U>())> > : std::true_type {};

template <typename T, typename U>
BOOST_MATH_INLINE_CONSTEXPR bool has_plus_v = has_plus<T, U>::value;

template <typename T, typename U>
std::enable_if_t<has_plus_v<T, U>, T>
add(T x, U y) 
```

Leveraging C++14 `_t` and making your own C++17 style `_v` variables is in my opinion more readable.

mborland (changes requested):
Your accuracy is clearly good based on the plots in the main thread. I think reducing re-allocation or heavy objects will help significantly with performance.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: It is a redefinition but it was only defined in C++17. I duplicated it because I need it for C++14 test. Is there a better way to do this?

include/boost/math/quadrature/symplectic.hpp
JacobHass8: I'll have to reread the paper to see how they did it. Even for the above (simpler) case it was _quite_ complicated. I'll get a better idea and then report how difficult it would be.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: Good idea! That's a lot cleaner.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: I think p0 is just copied once now when `integrate_hamiltonian_imp` is called. Should I pass everything by reference when I can? Most function signatures have the parameters `const RealType dt, Func dHdp, Func dHdq`. The timestep dt is easy to pass by reference. Should I also pass the functions dHdp/dq?

include/boost/math/quadrature/symplectic.hpp
mborland: No I meant it as you have a definition here at line 48-49, but there's also another one at 21-22.

image

include/boost/math/quadrature/symplectic.hpp
mborland: Yes, to the extent possible either pass by reference either const or non-const to the extent you can when moving containers around. Like the in-place add I mentioned I think there are other calculations you can do in-place since you're not mutating the users data.

mborland:
The one other change now is to mark your public APIs with `BOOST_MATH_EXPORT`. Details should not be exported so they'll stay out of the purview of the consumer.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: Oops, that certainly shouldn't be there.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: I've changed the interface to use enums as you suggested. The enum class lives in the detail namespace which I'm not sure is correct. Users will have to access the detail namespace to change the method being used. Is that okay?

JacobHass8:
@mborland I think this is ready to merge if you don't have any further comments. Thanks for the help! 

The drone failure looks spurious. Here's the error message 
```
+ git clone -b develop --depth 1 https://github.com/boostorg/boost.git boost-root
Cloning into 'boost-root'...
error: RPC failed; curl 16 Error in the HTTP2 framing layer
fatal: error reading section header 'shallow-info'
```

include/boost/math/quadrature/symplectic.hpp
mborland: Was the answer here that these are overly complicated to calculate?

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+template<typename T, typename U>
+struct has_plus<T, U, void_t<decltype(std::declval<T>() + std::declval<U>())> > : std::true_type {};
+
+template <typename T, typename U>
+BOOST_MATH_INLINE_CONSTEXPR bool has_plus_v = has_plus<T, U>::value;
+
+template <typename T, typename U>
+typename std::enable_if_t<has_plus_v<T, U>, void>
+add(T& x, U& y)
```
mborland:
```suggestion
add(T& x, const U& y) 
```

Passing by constant reference should be preferred when we don't need to mutate the value(s)

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,309 @@
...
+template <typename T, typename U>
+typename std::enable_if_t<has_plus_v<T, U>, void>
+add(T& x, U& y) 
+{
+    x = x + y;
+}
+
+template <typename T, typename U>
+typename std::enable_if_t<!has_plus_v<T, U>, void>
+add(T& vec1, U& vec2)
```
mborland:
```suggestion
add(T& vec1, const U& vec2)
```

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,309 @@
...
+    add(q0, dq);
+}
+
+template <typename RandomAccessContainer, typename RealType, class Func>
+void fourth_order_yoshida(RandomAccessContainer& p0, RandomAccessContainer& q0, const RealType dt, 
+                          const Func dHdp, const Func dHdq)
+{
+    BOOST_MATH_STD_USING
+
+    // RealType x0 = -(std::pow(2, 1/3) / (2 - std::pow(2, 1/3)));
```
mborland: Does this comment matter anymore?

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,309 @@
...
+    RealType x1 = 1. / (2. - cbrt(2.));
```
mborland: If doubles are sufficient precision please use 1.0. Otherwise this can become:

```c++
RealType x1 = RealType(1) / (RealType(2) - cbrt(RealType(2));
```

Uglier but once be subject to implicit conversion loss of precision

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+    if ((dt <= 0) || !(boost::math::isfinite)(dt))
+    {
+        boost::math::policies::raise_domain_error(function, "Time step must be positive and finite but got: dt = %1%.\n", dt, pol);
+    }
+
+    typedef void (*stepperType)(RandomAccessContainer&, RandomAccessContainer&, RealType, Func, Func);
+
+    stepperType stepper; 
+    switch (method) {
```
mborland, line 346: Adding a `default: BOOST_MATH_UNREACHABLE;` will help make sure you don't ever miss a method, for example if you add additional options.

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,309 @@
...
+        case available_methods::Y6:       stepper = sixth_order_yoshida; break;
+        case available_methods::Y4:       stepper = fourth_order_yoshida; break;
+        case available_methods::Y2:       stepper = second_order_yoshida; break;
+        case available_methods::SRKNB6:   stepper = SRKN_b_order_6; break;
+        case available_methods::SRKNB11:  stepper = SRKN_b_order_11; break;
+    }
+
+    std::vector<RandomAccessContainer> p(steps);
+    std::vector<RandomAccessContainer> q(steps);
```
mborland: We should probably have some kind of assertion that both p0 and q0 have at least steps amount of elements inside of them otherwise we will end up with a SEGFAULT in the stepper below.

mborland (changes requested):
Looks much better. Comments are much smaller in size and scope this go around.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: No, it's just 1-2x1. I'll close then.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: I've changed this and some instances below to match.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: Yeah, it's just too complicated for higher precision.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: I'm not sure that I understand. The vectors p, q just store the values of p and q at each timestep. The values p0 and q0 are the starting conditions - they can be floating point or vectors. There currently isn't any guarantee that p0 and q0 have the same size though.

JacobHass8: I added a size check and relevant test.

include/boost/math/quadrature/symplectic.hpp
JacobHass8: BOOST_MATH_UNREACHABLE isn't a variable I can see. I searched the codebase and didn't find anything either.

JacobHass8:
> Looks much better. Comments are much smaller in size and scope this go around.

All the tests are passing now. I think this is ready for another look! Thanks for all your help with this.

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -0,0 +1,335 @@
+// Copyright Jacob Hass, 2026
+// Use, modification and distribution are subject to the
+// Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+// Disable warnings when defining size_check for numeric types
+#ifdef _MSC_VER
+#  pragma warning (disable : 4100) // 'function': unreferenced parameter
+#endif
```
mborland, line 11:
```suggestion
#ifdef _MSC_VER
#  pragma warning(push)
#  pragma warning(disable : 4100) // 'function': unreferenced parameter
#endif
```

and then at the end of your affected area you need to include

```c++
#ifdef _MSC_VER
#  pragma warning(pop)
#endif 
```

If the warning isn't popped then it will stay present in user code.

mborland (changes requested):
Added one more comment, but it looks like there are still a few unresolved from the last review. Getting close.

JacobHass8:
> Added one more comment, but it looks like there are still a few unresolved from the last review. Getting close.

How does everything look now?

mborland:
> > Added one more comment, but it looks like there are still a few unresolved from the last review. Getting close.
> 
> How does everything look now?

I still see three unresolved comments

JacobHass8:
> > > Added one more comment, but it looks like there are still a few unresolved from the last review. Getting close.
> > 
> > 
> > How does everything look now?
> 
> I still see three unresolved comments

Sorry about that. I forgot to close some of the resolved comments. Thanks for reviewing this!

mborland (approved):
This looks good to me now. Any other remarks from your end @NAThompson

doc/quadrature/symplectic.qbk
```diff
@@ -0,0 +1,122 @@
...
+        Y4,
+        Y2,
+        SRKNB6,
+        SRKNB11
+    };
+
+    template <typename RandomAccessContainer, typename RealType, class Func, class Policy>
+    std::pair<std::vector<RandomAccessContainer>, std::vector<RandomAccessContainer> > integrate_hamiltonian(RandomAccessContainer p0,
+                                                                                                             RandomAccessContainer q0,
+                                                                                                             const RealType dt,
```
NAThompson, line 27: @JacobHass8 : Why are we having the user specify the timestep instead of determining it adaptively? Generally we want the user to specify an accuracy goal, not an internal algorithm detail.

doc/equations/harmonic_oscillator.png
NAThompson, line 1: I think one of these is redundant-either the .png or .svg.

NAThompson:
@mborland , @JacobHass8 : One more substantive comment left. I believe claude or chat should be able to one-shot the fix. Then it's good.

NAThompson:
@JacobHass8 : Also, could you document to the user how to use the $\{t_i, p_i, q_i \}_{i=0}^{n-1}$ datastructure? Ostensibly the user should do a cubic Hermite spline interpolant, but I just want to be sure-and I think they will also want to be sure.

NAThompson (approved):

doc/quadrature/symplectic.qbk
JacobHass8: Honestly, I'm not 100% sure how to determine the accuracy/error of the algorithms. I can compute the error in the tests because I know the Hamiltonian. Only given these inputs I don't know how to define the error in the algorithm though. 

One option is to accept the Hamiltonian as parameter which would enable us to calculate the error on the fly. I'm not sure what scheme to use for the timestep though.

JacobHass8:
> @JacobHass8 : Also, could you document to the user how to use the t i , p i , q i i = 0 n − 1 datastructure? Ostensibly the user should do a cubic Hermite spline interpolant, but I just want to be sure-and I think they will also want to be sure.

Sure! I'll work up an example. I should also included an overload to return a Hermite spline interpolant as you suggested earlier.

JacobHass8:
I added the cubic hermite spline return rather than returning the p,q vectors. As a sanity check here is an interpolation of a simple harmonic oscillator. The blue is the calculated p,t values and red is interpolated. They line up exactly for the same times, but there are artifacts from the cubic spline as expected. I only wanted to see that what I was doing made sense. I'll add a brief test checking that the interpolation at each time returns the correct value. 

Figure_1
Figure_2

NAThompson:
@JacobHass8 : Do try to make the step size adaptive-that'll really improve things. Check [this article](https://link.springer.com/article/10.1023/A:1021160023092) or [this](https://dl.acm.org/doi/10.1145/210232.210242)

JacobHass8:
> @JacobHass8 : Do try to make the step size adaptive-that'll really improve things. Check [this article](https://link.springer.com/article/10.1023/A:1021160023092) or [this](https://dl.acm.org/doi/10.1145/210232.210242)

I implemented the simplest step size controller here pretty easily. 
image
I think next I'm going to try 
image
as it's an easy improvement. One quick questions though. What should the reference method be when calculating the error? Right now I'm just using the yoshida second order because this is the analog to Euler's method for symplectic integrators.

NAThompson:
> What should the reference method be when calculating the error?

Well, we should just find an ODE with an analytic solution right?

JacobHass8:
> > What should the reference method be when calculating the error?
> 
> Well, we should just find an ODE with an analytic solution right?

I'm not sure that I follow. Do we just want the most accurate solver to use as an error estimate or something different? Thanks for being so responsive and helpful. I'm new to this sort of error estimation.

NAThompson:
Oh I think I see what you're getting at-but if I'm right then the question might be a bit of a category error.

Could you read [Chapter 17](https://numerical.recipes/book.html)? It's a very easy intro to this topic.

JacobHass8:
> Oh I think I see what you're getting at-but if I'm right then the question might be a bit of a category error.
> 
> Could you read [Chapter 17](https://numerical.recipes/book.html)? It's a very easy intro to this topic.

Oh, this makes way more sense than what I was doing/thinking. This is pretty easy to get working. Let me give it a shot and get back to you.

JacobHass8:
> Oh I think I see what you're getting at-but if I'm right then the question might be a bit of a category error.
> 
> Could you read [Chapter 17](https://numerical.recipes/book.html)? It's a very easy intro to this topic.

I've added adaptive step sizes using Chapter 17 but only in 1 dimension (haven't committed it yet). This is because boost's `cubic_hermite` only supports one dimensional interpolation. Extending the adaptive step sizes to higher dimensions is easy but I'm not sure about the spline. I thought about using a separate spline for each dimension but thought I'd check first. If we decide to do that, should it be added to cubic_hermite.hpp?

NAThompson:
> Extending the adaptive step sizes to higher dimensions is easy but I'm not sure about the spline. I thought about using a separate spline for each dimension but thought I'd check first. If we decide to do that, should it be added to cubic_hermite.hpp?

Yeah that would be a nice feature add!

Of course, you could also just return the solution skeleton $\{ t_k, y_k, \dot{y}_k\}_{k=0}^{n-1}$ which would be perfectly acceptable.

NAThompson:
@JacobHass8 : FWIW, now that adaptivity is in you should merge when you think it's good.

JacobHass8:
> @JacobHass8 : FWIW, now that adaptivity is in you should merge when you think it's good.

Tbh, I'm not super happy with where it is at right now. There's still a number of features that I think need to be implemented before it's up to boost standards. I'm just going to drop a list here so I can keep track

- [x] Write a test making sure that vectors cannot be passed to `integrate_hamiltonian_cubic`
- [x] Add option to return error in solution to `integrate_hamiltonian_adaptive`
- [x] Write test to make sure that the error is below the requested threshold 
- [x] Update documentation to reflect new functions

After I finish these, I think it will be ready to commit. Here are a couple more features that are more long term but I think are necessary too.

- [ ] Implement PI controller to account for error in previous step (probably the easiest)
- [ ] Multi-dimensional hermite spline 
        -  Just treat every dimension as independent 
        - Maybe we can overload the current function
- [ ] Figure out a better container/method for addition/multiplication/division of floating point vs random access containers
        - I introduced a lot of duplicate functions which are used for either floating point or random access containers
        - More than likely, this could be streamlined. The easiest would be to simply convert floats to vectors to so that the actual implementation only uses a single function

NAThompson:
FWIW, I don't believe changing the cubic spline to support vector outputs should be a big change.

Lemme see . . .

include/boost/math/quadrature/symplectic.hpp
```diff
@@ -444,19 +444,19 @@ struct error_info
 
             if (rejected)
             {
-                dt *= std::min(scale, RealType(1.0));
+                dt *= (std::min)(scale, RealType(1.0));
```
NAThompson: Do a `using std::min` so you can go to multiprecision types.

Also recommend doing a unit test with multiprecision types.

JacobHass8:
@NAThompson I'm going to merge if all the tests come back passing (or unrelated). Let me know if you have any other comments.

NAThompson:
@JacobHass8 : I've just done quite a bit of maintenance work on the CI system-could you rebase? Right now you'll get spurious failures.

After the rebase, you'll probably _still_ get spurious failures, but hopefully there will be fewer of them.

JacobHass8:
> @JacobHass8 : I've just done quite a bit of maintenance work on the CI system-could you rebase? Right now you'll get spurious failures.
> 
> After the rebase, you'll probably _still_ get spurious failures, but hopefully there will be fewer of them.

Yup, only OSX errors and autodifferentiation errors. I think you can merge when ready.

NAThompson:
@JacobHass8 : Great work and good job pushing this past the finish line!

JacobHass8:
> @JacobHass8 : Great work and good job pushing this past the finish line!

Thanks for all your help!

PR #1413, opened 2026-06-30
<sstream> is not needed in exception free environment

### LINKED ISSUE #1412: error_handling.hpp requires `<sstream>`/`prec_format` even with `BOOST_MATH_NO_EXCEPTIONS`
**Version:** 1.89.0 (standalone)

**Problem**

With `BOOST_MATH_NO_EXCEPTIONS`, `policies/error_handling.hpp` still includes `<sstream>`/`<iomanip>` and defines `detail::prec_format` (uses `std::stringstream`/`std::setprecision`). These only format exception `what()` text; their sole caller `raise_error` is already under `#ifndef BOOST_MATH_NO_EXCEPTIONS`, but the includes and `prec_format` are not -- so they compile even with exceptions off. `prec_format` is a template, but its non-dependent `std::stringstream` is diagnosed at definition time, so the error fires without instantiation.

**Impact:** breaks Boost.Math as a no-exceptions back-end on a stdlib without a usable `<sstream>` (e.g. libc++ without localization). LLVM libc++ is adopting Boost.Math for the C++17 special functions and hits this.

**Reproduction**

Godbolt: https://godbolt.org/z/dn5Krja8q (gcc+libstdc++ and clang+libc++ both fail).
Flags: `-std=c++17 -DBOOST_MATH_STANDALONE -DBOOST_MATH_NO_EXCEPTIONS -fno-exceptions`

```cpp
// Pre-claim <sstream>'s guard so it expands to nothing (mirrors no-localization libc++):
#define _GLIBCXX_SSTREAM 1   // libstdc++
#define _LIBCPP_SSTREAM      // libc++
#include <boost/math/special_functions/laguerre.hpp>
int main() { return static_cast<int>(boost::math::laguerre(1u, 0u, 0.5)); }
```
-> `'stringstream'/'streamsize'/'setprecision' is not a member of 'std'` in `prec_format`.

**Fix:** move the `<iomanip>`/`<sstream>` includes into the existing `#ifndef BOOST_MATH_NO_EXCEPTIONS` block, and wrap `detail::prec_format` (+ its `BOOST_MATH_USE_CHARCONV_FOR_CONVERSION` specialization) in the same guard.

```diff
--- a/include/boost/math/policies/error_handling.hpp
+++ b/include/boost/math/policies/error_handling.hpp
@@ -26,9 +26,9 @@
 #include <cstring>
+#ifndef BOOST_MATH_NO_EXCEPTIONS
 #include <iomanip>
 #include <sstream>
-#ifndef BOOST_MATH_NO_EXCEPTIONS
 #include <stdexcept>
 #endif
@@ -95,6 +95,8 @@ namespace detail
 {
+// Only used to build exception messages; relies on <sstream>.
+#ifndef BOOST_MATH_NO_EXCEPTIONS
 template <class T>
 inline std::string prec_format(const T& val)
@@ -125,6 +127,7 @@ inline std::string prec_format<std::float128_t>(const std::float128_t& val)
 }
 #endif
+#endif // BOOST_MATH_NO_EXCEPTIONS
```

mborland:
Closes: #1412

### DIFF
--- a/include/boost/math/policies/error_handling.hpp
+++ b/include/boost/math/policies/error_handling.hpp
@@ -26,9 +26,9 @@
 #include <complex>
 #include <cstdint>
 #include <cstring>
+#ifndef BOOST_MATH_NO_EXCEPTIONS
 #include <iomanip>
 #include <sstream>
-#ifndef BOOST_MATH_NO_EXCEPTIONS
 #include <stdexcept>
 #endif
 #include <string>
@@ -95,6 +95,8 @@ T user_indeterminate_result_error(const char* function, const char* message, con
 namespace detail
 {
 
+#ifndef BOOST_MATH_NO_EXCEPTIONS
+// Only used to build exception messages; relies on <sstream>
 template <class T>
 inline std::string prec_format(const T& val)
 {
@@ -124,7 +126,9 @@ inline std::string prec_format<std::float128_t>(const std::float128_t& val)
    return std::string(buffer, r.ptr);
 }
 
-#endif
+#endif // BOOST_MATH_USE_CHARCONV_FOR_CONVERSION
+
+#endif // BOOST_MATH_NO_EXCEPTIONS
 
 inline void replace_all_in_string(std::string& result, const char* what, const char* with)
 {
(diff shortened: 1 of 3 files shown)


mborland:
Spurious failure. Merging.

PR #1414, opened 2026-07-01
Add M nearest-neighbour Chatterjee correlation (#990)

### LINKED ISSUE #990: Can Chatterjee Correlation be improved?
Related paper: https://arxiv.org/pdf/2108.06828.pdf

su-senka:
## Summary

Implements the revised (M nearest-neighbour) Chatterjee rank correlation of
Lin & Han (2021), addressing #990. Adds a new function
`chatterjee_correlation_mnn(u, v, M)` alongside the existing
`chatterjee_correlation`, with the same C++11 and C++17 overload structure.

The original coefficient has a detection boundary of n^(-1/4) for independence
testing, well short of the parametric n^(-1/2) rate. By using the M right
nearest neighbours of each point (rather than the single right neighbour) and
letting M grow with n, the revised statistic consistently estimates the same
dependence measure while approaching near-parametric efficiency. See Lin & Han,
*On boosting the power of Chatterjee's rank correlation*, Biometrika 110(2)
(2023) 283–299, [arXiv:2108.06828](https://arxiv.org/abs/2108.06828).

## Design notes

- **Separate function rather than an extended signature.** The M-NN statistic
  uses `min(R_i, R_j)` and a different normalisation, so even at M = 1 it is not
  identical to `chatterjee_correlation`. A distinct function avoids silently
  changing existing results and keeps the statistical intent explicit. 
 `M` is a required argument with no default.

- **Rank base.** The internal `rank()` returns 0-based ranks; the paper's
  formula uses 1-based ranks. The offset cancels in the existing M = 1 statistic
  (which uses `|R_i - R_{i+1}|`) but not under `min(.,.)`, so it is applied
  explicitly. This is noted in a comment where it matters.

- **Complexity.** O(n log n + nM). Near-linear for small M; tends to O(n²) as
  M → n.

- **Parallel path.** The outer index loop is partitioned across threads into
  disjoint ranges, each reading the shared rank vector read-only (indices up to
  i + M may fall in a neighbouring range; there are no writes). This differs
  from the M = 1 parallel path, which splits the data array for the
  difference-based transform.

- **Ties / degenerate input.** Like `chatterjee_correlation`, the function
  assumes distinct Y (continuous data). A constant Y returns a quiet NaN; this
  is detected on the input directly, since `rank()` collapses tied values.

- **Choice of M.** The asymptotic null variance is minimised at M ~ sqrt(n); the
  choice is documented but left to the caller.

## Tests

Added to `test_chatterjee_correlation.cpp`, covering `float`, `double`, and
`long double`:

- Exact closed-form checks against the paper's Remark 2.5 (strictly increasing
  and strictly decreasing dependence), which require no external reference.
- Small exact spot values computed independently as rationals.
- Constant-Y → NaN, and invariance under strictly increasing transforms of X
  and Y.
- Sequential/parallel agreement across several M (under the parallel build).

The sequential path was verified locally under `b2` with `cxxstd=14` and
`cxxstd=17` (clang, arm64).

### DIFF
--- a/doc/overview/roadmap.qbk
+++ b/doc/overview/roadmap.qbk
@@ -17,6 +17,7 @@ Recent issues on GitHub [@https://github.com/boostorg/math/issues?utf8=%E2%9C%93
 * Add improvement to Heuman Lambda precision.
 * Improve Skew Normal root finding, see [@https://github.com/boostorg/math/issues/1120 1120].
 * Lots of minor fixes and improved code coverage.
+* Add support for the M nearest-neighbour Chatterjee correlation of Lin and Han (2021), see [@https://github.com/boostorg/math/issues/990 990].
 
 [h4 Math-4.2.0 (Boost-1.85)]
 
--- a/doc/statistics/chatterjee_correlation.qbk
+++ b/doc/statistics/chatterjee_correlation.qbk
@@ -72,3 +72,83 @@ Of course, random numbers are not used internally, but the result is not guarant
 
 [endsect]
 [/section:chatterjee_correlation Chatterjee Correlation]
+
+[section:chatterjee_correlation_mnn Chatterjee Correlation (M nearest neighbours)]
+
+[heading Synopsis]
+
+``
+#include <boost/math/statistics/chatterjee_correlation.hpp>
+
+namespace boost::math::statistics {
+
+    C++17:
+    template <typename ExecutionPolicy, typename Container>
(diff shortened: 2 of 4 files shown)


mborland (approved):
I've approved the workflow. This all looks good to me! @NAThompson if you have time could you give this a quick look?

mborland:
macOS failure has already been fixed on develop. Merging. Thank you for this contribution!

PR #1415, opened 2026-07-01
Merge to Master for 1.92

mborland:
Deadline is today, kind of slipped my mind


mborland:
Strangely the macOS runs are starting to fail consistently with:

```
====== BEGIN OUTPUT ======
Error at cubic_roots_test.cpp:test_zero_coefficients:44
  Condition 'nan == -1' is not satisfied:
  Expected =  nan = nan = nan
  Computed =  -1.000000 = -1.000000e+00 = -0x1p+0
Error at cubic_roots_test.cpp:test_zero_coefficients:45
  The computed value should be a nan, but is instead -1.000000 = -1.000000e+00 = -0x1p+0
Error at cubic_roots_test.cpp:test_zero_coefficients:50
  Condition 'nan == 1' is not satisfied:
  Expected =  nan = nan = nan
  Computed =  1.000000 = 1.000000e+00 = 0x1p+0
Error at cubic_roots_test.cpp:test_zero_coefficients:51
  The computed value should be a nan, but is instead 1.000000 = 1.000000e+00 = 0x1p+0
libc++abi: terminating due to uncaught exception of type std::domain_error: Error in CHECK_ULP_CLOSE: Expected value cannot be a nan. Callsite: cubic_roots_test.cpp:test_zero_coefficients:57.

EXIT STATUS: 134
====== END OUTPUT ======
```

It's been >4 years since the last commit to the cubic roots code?

mborland:
Mac failures are fixed in #1417, so I will wait for that to be merged so that it can automatically be populated into master

mborland:
#1417 ran green and #1416 better tests libc++'s integration issue so I think we are going to be good now

PR #1416, opened 2026-07-01
Check all of special functions can be included under same conditions

mborland:
CC: @PaulXiCao

This seems to run fine with all of special functions included. It looks like the other naked uses of `<sstream>` are in the interpolators and gradient optimizers. Those should all be outside of what you need, and are effectively isolated modules within boost.math.

### DIFF
--- a/test/git_issue_1412_pt_2.cpp
+++ b/test/git_issue_1412_pt_2.cpp
@@ -4,20 +4,23 @@
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 //
 // Regression test for https://github.com/boostorg/math/issues/1412
+//
+// Part 2: Test all of #include <boost/math/special_functions.hpp> as that is what libc++ is going to use
 
 #define BOOST_MATH_NO_EXCEPTIONS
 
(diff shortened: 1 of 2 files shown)


test/git_issue_1412_pt_2.cpp
```diff
@@ -0,0 +1,23 @@
...
+#define _GLIBCXX_IOMANIP 1     // libstdc++
+
+#include <boost/math/policies/error_handling.hpp>
+#include <boost/math/special_functions.hpp>
+
+int main()
+{
+    // Exercise a special function so the no-exceptions header path is
+    // instantiated, not just parsed.
+    return static_cast<int>(boost::math::laguerre(1u, 0u, 0.5));
```
PaulXiCao: Is the inclusion already enough to instantiate the possible problematic locale-dependent print functions? I could imagine that we'd need to call at least one function per header in `boost/math/special_functions/*.hpp`?

mborland: It seemed sufficient to catch the same error as your original issue: https://godbolt.org/z/xe5zqj1a7 
We have an instantiation harness so I could setup a different driver and instantiate every function. That would be more comprehensive.

PR #1417, opened 2026-07-01
Fix UB in sorting results with the cubic root finder

mborland:


mborland:
Gets our build back to green as hoped

PR #1422, opened 2026-07-15
Implement and test consumption of Boost.Math as a module

mborland:
This implements module support for a large subset of the library as well as testing with associated infrastructure. Since we don't control our dependencies, consumption of the module only supports standalone mode. A modules.qbk files was added that explicitly lays out limitations that we can't circumvent such as policy configuration macros will not work. 

Please let me know what you think @jzmaddock, @NAThompson, @ckormanyos, @ChuanqiXu9


mborland:
This is green in general and with both GCC and Clang as a module

module/math.cppm
```diff
@@ -0,0 +1,223 @@
+// Copyright 2026 Matt Borland
```
ChuanqiXu9, line 1: nit: I prefer math.cppm

doc/overview/modules.qbk
```diff
@@ -0,0 +1,57 @@
...
+* The `user_error` policy is not supported through the module, because a user supplied `boost::math::policies::user_..._error` definition cannot attach to the module.
+* A program should consume Boost.Math either through the module or through textual includes, not both.
+  The only headers that are safe to include textually alongside the import are `boost/math/tools/config.hpp` and `boost/math/tools/assert.hpp`, which the test support headers rely on.
+* Excluded from the module: the TR1/C99 interfaces and `common_factor` headers, `boost/math/bindings`, `boost/math/concepts`, `boost/math/cstdfloat`, all deprecated headers, and the headers that require external non-Boost dependencies (`chebyshev_transform.hpp` and `interpolators/cardinal_trigonometric.hpp` need FFTW, `optimization/cma_es.hpp` needs Eigen).
+  These remain available as ordinary includes.
+* `__float128` support (`BOOST_MATH_USE_FLOAT128`) is not enabled in the module build.
+* On libc++ the parallel execution policy overloads in statistics compile out (no `<execution>` support), matching the ordinary header behavior on that standard library.
+
```
ChuanqiXu9, line 49: I noticed you didn't use `extern C++` in math.cxx. So the linkage will change. That is not bad. It is a choice. But I think we need to say it here.

ChuanqiXu9 (commented):
Excellent work

doc/overview/modules.qbk
mborland: Is extern C++ preferable? There's a couple of smaller boost libs with module support so I'll check and see what they're doing for consistency.

ChuanqiXu9: Not prefer or not. It is just a style. Whether or not we want to break the ABI.  I discussed it in https://clang.llvm.org/docs/StandardCPlusPlusModules.html#abi-breaking-style . The short version is, it looks like the  famous ABI breaking in libstdc++5. But it may not be so bad as now most code compile from source and boost is not libstdc++.

jzmaddock:
Wow!  Looks like good work to me, time will tell what if any consequences there are?

I notice it's not yet tested with msvc which has really rather good module support, has this been tried yet?

Also any word on compile times?

mborland:
> Wow! Looks like good work to me, time will tell what if any consequences there are?
> 

I expect so. I think the ecosystem is still feeling it out.

> I notice it's not yet tested with msvc which has really rather good module support, has this been tried yet?
> 

I have not tried with MSVC, but I can add that to the CI.

> Also any word on compile times?

Clang takes 33 seconds to build the module plus compile and run the 25 tests. I can slice the module build into a separate CI step to measure it directly.

doc/overview/modules.qbk
mborland: Other boost libs use `extern C++` so I will change what is here to match the others.

mborland:
The MSVC module is in CI and works. Build times for just the boost.math module:

MSVC - 15 seconds
Clang - 9 seconds
GCC - 6 seconds

The other Boost libraries that support modules all use `extern c++` so I have changed our approach to match that.

ChuanqiXu9:
> The MSVC module is in CI and works. Build times for just the boost.math module:
> 
> MSVC - 15 seconds Clang - 9 seconds GCC - 6 seconds
> 
> The other Boost libraries that support modules all use `extern c++` so I have changed our approach to match that.

Generally we prefer to test the time for using/import the modules. As the story modules builds on everything was built once and we can avoid duplications.

e.g., the blog (https://anarthal.github.io/cppblog/modules3) says:

> Build performance gains are higher when lots of translation units consume the same library. Building the Boost.Mp11 test suite (which has around 200 translation units) yields the following results:
> 
> Headers: 2min 10s.
> 
> Modules: 39s (this includes the time required to build the std and Boost modules).
>
> Benchmarks performed on Ubuntu 22.04 with clang-19 and libc++.

This reads very good.

mborland:
Anything else you'd like to see for this one @jzmaddock? If not I can merge this PR and close the old one.

jzmaddock:
Looking good.  However, when trying to get some concept checks working with modules I see:

```
use_module.obj : error LNK2001: unresolved external symbol "double const * const boost::math::lambert_w_detail::lambert_w_lookup::wm1es" (?wm1es@lambert_w_lookup@lambert_w_detail@math@boost@@3QBNB)
1>use_module.obj : error LNK2001: unresolved external symbol "double const * const boost::math::lambert_w_detail::lambert_w_lookup::wm1zs" (?wm1zs@lambert_w_lookup@lambert_w_detail@math@boost@@3QBNB)
1>use_module.obj : error LNK2001: unresolved external symbol "double const * const boost::math::lambert_w_detail::lambert_w_lookup::halves" (?halves@lambert_w_lookup@lambert_w_detail@math@boost@@3QBNB)
1>use_module.obj : error LNK2001: unresolved external symbol "double const * const boost::math::lambert_w_detail::lambert_w_lookup::sqrtwm1s" (?sqrtwm1s@lambert_w_lookup@lambert_w_detail@math@boost@@3QBNB)
```

These are all defined in lambert_w_lookup_table.hpp and there are some tricky ones here, because I don't think we can make the tables `inline constexpr` but ` static constexpr` won't work with modules?

mborland:
BOOST_MATH_INLINE_CONSTEXPR should fix these. It expands to static constexpr in C++14 mode and inline constexpr for 17 and up.

mborland:
@jzmaddock Try your test now; the stray linkage errors should be resolved now. If it does not work could you please post the code you're trying to run?

mborland:
This is now back to green on all platforms with significantly expanded tests.

mborland:
@jzmaddock any opposition to merging this or further comments?

jzmaddock:
No lets get it out there and see!

PR #1435, opened 2026-08-19
Multi dimensional integrator

JacobHass8:
I made a small function to integrate in higher dimensions by recursively integrating in 1D. It gets slower and less accurate as the dimension increases. I only wrote test for 2 and 3 dimensions and 3 dimensions was already getting slow.

However, this could be useful for low dimensional integrals that don't require high precision. When I researched higher dimensional integrals, most answers online suggested repeated 1D integration. From my understanding, this is actually how scipy implements their arbitrary dimensional integrator [`nquad`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.nquad.html). This could work as a quick solution before implementing sparse grids. 

There's still a few more changes that I need to make, but the rough draft is here.


NAThompson:
I'm not 100% sure of the utility of this. IMO Claude will be able to 1-shot repeated 1D integration for anyone who wants it-but it's an anti-pattern in general. 

These would be better in low d: https://nines.cs.kuleuven.be/research/ecf/, https://link.springer.com/article/10.1007/s002110050358

And I don't think sparse grids are beyond your talents! I think you could get them working.

JacobHass8:
> I'm not 100% sure of the utility of this. IMO Claude will be able to 1-shot repeated 1D integration for anyone who wants it-but it's an anti-pattern in general.

Thanks for taking a look! I'll go ahead and close this then. 

> These would be better in low d: https://nines.cs.kuleuven.be/research/ecf/, https://link.springer.com/article/10.1007/s002110050358

The first link looks promising but I can't access any of the tables. Even if I did, is this something I could use under the boost copyright? 

I was looking at implementing this first https://www.researchgate.net/publication/225149450_Padua2DM_Fast_interpolation_and_cubature_at_the_Padua_points_in_MatlabOctave

NAThompson:
@JacobHass8 : Padua points would be an awesome first step!

jzmaddock:
Just for fun (ha!), I wondered what recursion should look like and came up with this:

```
template <class Integrator, class Fun, class T>
inline auto integrateND(const Integrator& i, Fun f, T lower, T upper)
{
    // We need a functor passed to integrate that takes a single argument only
    // as if it's var-arg it can throw off function overload resolution:
    auto single_arg_f = [&](T x) { return f(x); };
    return i.integrate(single_arg_f, lower, upper);
}

template <class Integrator, class Fun, class T, typename... Args>
inline auto integrateND(const Integrator& i, Fun f, T lower, T upper, Args... args)
{
    auto recursion = [&](T x) 
        { 
            // std::bind_front works fine, but std::bind fails if we recurse more than twice!
            return integrateND(i, std::bind_front(f, x), args...); 
        };
    return i.integrate(recursion, lower, upper);
}
```

The idea is that the function to be integrated is written "naturally" as:

```
    auto gaussian = [](double x, double y) { return std::exp(-(x * x + y * y)); };
```

or whatever.

and the function call looks like:

```
    std::cout << integrateND(boost::math::quadrature::tanh_sinh<double>(), gaussian, -inf, inf, -inf, inf) << std::endl;
```

Works a treat for toy examples, including the 3D Gaussian, Watson's first integral works too, but takes quite a while:

```
    std::cout << integrateND(boost::math::quadrature::tanh_sinh<double>(), [](double x, double y, double z) { return 1 / (1 - cos(x) * cos(y) * cos(z)); }, 0.0, pi, 0.0, pi, 0.0, pi) / boost::math::pow<3>(pi) << std::endl;
```

However, there's an important class of integrals which can't be handled by this kind of boilerplate, and that's mostly geometric problems where the inner integration bounds are dependent on the outer/enclosing variable(s).  I'm not really sure what interface design would handle those to be honest?

Anyway, just throwing this out there in case it helps...

JacobHass8:
> Just for fun (ha!), I wondered what recursion should look like and came up with this:

Interesting! I had been trying to get function recursion to work but couldn't figure out how to correctly bind the function.

PR #1438, opened 2026-08-26
Fix catastrophic cancellation error in poisson logpdf when k and mu are large

rahulb0802:
This replaces the current direct-formula method in ``logpdf`` with Loader's (2000) saddle-point approx, ported from a change I've been working on in SciPy's ``scipy.stats.poisson.logpmf``. The current formula loses precision due to cancellation when k is close to mu and both are large. The new form implemented avoids this.

Two new helpers are added to ``poisson_detail``:
- ``stirlerr(n)``: Stirling's series remainder. Direct path for small `n`, asymptotic series for large `n`
- ``bd0(mean, k)``: Deviance term. Direct path and series expansion for ``|v| < 0.1`` to avoid cancellation when k and mu are close

All existing poisson tests pass (across all real types). I also added accuracy checks for other values, validated against mpmath (1000-digit) reference values, including large-magnitude cases. No regressions as per CI checks.

### DIFF
--- a/include/boost/math/distributions/poisson.hpp
+++ b/include/boost/math/distributions/poisson.hpp
@@ -142,31 +142,23 @@ namespace boost
         return true;
       } // bool check_dist_and_prob
 
-      template <class RealType>
-      BOOST_MATH_GPU_ENABLED inline RealType stirlerr(const RealType& n) {
+      template <class RealType, class Policy>
+      BOOST_MATH_GPU_ENABLED inline RealType stirlerr(const RealType& n, const Policy& pol) {
         BOOST_MATH_STD_USING // for ADL of std functions.
         using boost::math::lgamma;
 
-        // Stirling's series coefficients
-        const RealType S0 = RealType(1)/12;
-        const RealType S1 = RealType(1)/360;
-        const RealType S2 = RealType(1)/1260;
-        const RealType S3 = RealType(1)/1680;
-        const RealType S4 = RealType(1)/1188;
-
-        // Use Stirling's series if n is small; use the direct formula otherwise
-        bool is_small = n < 15;
-        if (is_small) {
-          return lgamma(n + 1) - (n * log(n) - n + 0.5 * log(2 * boost::math::constants::pi<RealType>() * n));
-        } else {
-          RealType n2 = n * n;
-          return (S0 - (S1 - (S2 - (S3 - S4/n2)/n2)/n2)/n2)/n;
+        // Use the direct formula for small n
+        if (n < boost::math::detail::minimum_argument_for_bernoulli_recursion<RealType>()) {
+          return lgamma(n + 1) - (n * log(n) - n + RealType(0.5) * log(RealType(2) * boost::math::constants::pi<RealType>() * n));
         }
+
+        // Use the Stirling series approximation for large n
+        return boost::math::detail::bernoulli_stirling_series<RealType>(n, pol);
 
       }
 
-      template <class RealType>
-      BOOST_MATH_GPU_ENABLED inline RealType bd0(const RealType& mean, const RealType& k) {
+      template <class RealType, class Policy>
+      BOOST_MATH_GPU_ENABLED inline RealType bd0(const RealType& mean, const RealType& k, const Policy& pol) {
         BOOST_MATH_STD_USING // for ADL of std functions.
 
         // Calculate v = (k - mean) / (k + mean) from Loader (2000) approximation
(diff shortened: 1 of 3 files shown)


jzmaddock:
Thanks for this @rahulb0802 , this is a good idea, but not quite right just yet.  The problem is that the new routines evaluate a fixed number of terms of the Stirling remainder, and therefore are definitely not arbitrary precision.  BTW your new test values are limited to double precision too, so the tests should NOT be passing for any long double larger than 64 bits, certainly not for the 128-bit ones, I need to investigate why that is.

But for now taking the log of the PDF as the true result, and with `old_logpdf` a copy of the previous logpdf code:

```
    using boost::math::poisson_distribution;

    using mp = boost::multiprecision::number<boost::multiprecision::debug_adaptor<boost::multiprecision::backends::cpp_bin_float<200>>>;

    std::int64_t mean = 20;
    std::int64_t k    =  18;

    std::cout << std::setprecision(18);

    mp val = log(pdf(poisson_distribution<mp>(mean), mp(k)));
    double dval = (double)val;

    std::cout << "mean =             " << mean << " k = " << k << std::endl;
    std::cout << "MP logpdf:         " << boost::math::relative_difference(logpdf(poisson_distribution<mp>(mean), mp(k)), val) << std::endl;
    std::cout << "MP old logpdf:     " << boost::math::relative_difference(old_logpdf(poisson_distribution<mp>(mean), mp(k)), val) << std::endl;
    std::cout << "double logpdf:     " << boost::math::relative_difference(logpdf(poisson_distribution<double>(mean), double(k)), dval) << std::endl;
    std::cout << "double old logpdf: " << boost::math::relative_difference(old_logpdf(poisson_distribution<double>(mean), double(k)), dval) << std::endl;
```

Gives:

```
mean =             20 k = 18
MP logpdf:         1.19456007034065095e-17
MP old logpdf:     1.03043973124271719e-198
ddouble logpdf:     1.79628534341220374e-16
double old logpdf: 8.98142671706101575e-16
```

So when evaluated at 200-decimal precision the new routine gives just 17 decimal places compared to nearly full precision before.  Evaluating at larger mean looks a little better:

```
mean =             1000000 k = 1300000
MP logpdf:         2.60446354798830633e-75
MP old logpdf:     1.30516145170170221e-197
double logpdf:     2.12532375379900552e-15
double old logpdf: 5.75608516653897354e-14
```

But we're further out into the asymptotic region for Sterlings formula so fewer terms generate more digits of precision.

So... what to do?  I can't see Sterling converging to full arbitrary precision for small means (from memory the terms become alternating and divergent eventually?)  Although we might get there if the `is_small` check is appropriately tweaked inside `stirlerr`?  We do have an arbitrary precision version of Sterling here: https://github.com/boostorg/math/blob/9b192b023bec782dd47c4aa2cbb2b9458a275487/include/boost/math/special_functions/gamma.hpp#L490 which calculates tgamma(z) / (z/e)^z, could that be pressed into action here?

rahulb0802:
Yep, that is the fix, thanks for pointing this out. The sum variable being accumulated in this function is literally the exact same thing I was trying to compute in `stirlerr`, but not capped at a fixed number of terms (five in my original implementation). So I moved that loop part to its own function and now the `scaled_tgamma_no_lanczos` function just calls that before performing final computations. 

`bd0` had a similar problem but the fix was much simpler there. Instead of using 10 terms, just put in a convergence check. But doing it this way introduces an edge case for nonconvergence when k = mean. However, since we know the series is trivially zero at that point, we can safely skip the loop.

Running your original case:
```cpp
using mp = boost::multiprecision::number<boost::multiprecision::backends::cpp_bin_float<200>>;

std::cout << std::setprecision(18);

std::int64_t mean = 20;
std::int64_t k    = 18;

mp val = log(boost::math::pdf(poisson_distribution<mp>(mean), mp(k)));
double dval = (double)val;

std::cout << "mean =             " << mean << " k = " << k << std::endl;
std::cout << "MP logpdf:         " << boost::math::relative_difference(boost::math::logpdf(poisson_distribution<mp>(mean), mp(k)), val) << std::endl;
std::cout << "double logpdf:     " << boost::math::relative_difference(boost::math::logpdf(poisson_distribution<double>(mean), double(k)), dval) << std::endl;

```

```
mean =             20 k = 18
MP logpdf:         6.86959820828478126e-200
double logpdf:     1.79628534341220374e-16
```

```
mean =             1000000 k = 1300000
MP logpdf:         8.80528085179990715e-199
double logpdf:     2.12532375379900552e-15
```
Double results remain unchanged but the mp result is substantially better in terms of precision.

rahulb0802:
Hey @jzmaddock , just wanted to follow up on this. Let me know if you notice anything else that needs fixing

jzmaddock:
I'm currently away, but I'll look at this shortly, thanks for the bump @rahulb0802

PR #1439, opened 2026-08-26
Make implicit integral to `Real` conversions explicit

mborland:
We are already properly casting the return type, so apply that to the initial conditions. Makes all of the statistics functions work without any workaround with Boost.Decimal.


mborland:
Several instances of this failure in jso_test which should have been completely unaffected by this change:

```
====== BEGIN OUTPUT ======
terminate called after throwing an instance of 'std::invalid_argument'
  what():  ../../../boost/math/optimization/jso.hpp:89:validate_jso_parameters: There must be more individuals in the population than threads.
Testing jSO on Ackley function . . .
Aborted (core dumped)

EXIT STATUS: 134
====== END OUTPUT ======
```

PR #1440, opened 2026-08-28
Van den Bos quadrature

NAThompson:


NAThompson:
@mborland , @JacobHass8 : I investigated the Padua points, but it appears there is no known literature that satisfies the properties:

- strictly positive weights
- arbitrary nesting
- optimal convergence Kexp(-c\sqrt{n})
- reasonable compute ratios between stages


[This paper](https://arxiv.org/pdf/1809.09842) seems to satisfy all these constraints, and also can extend to nD. Interested?

We probably would need to test https://dl.acm.org/doi/abs/10.5555/2837.2842.

test/test_van_den_bos.cpp
```diff
@@ -0,0 +1,100 @@
+#include "boost/math/quadrature/van_den_bos.hpp"
```
mborland:
```suggestion
/*
 * Copyright 2026 Nick Thompson
 *
 * Use, modification and distribution are subject to the
 * Boost Software License, Version 1.0.
 * 
 */

#include <boost/math/quadrature/van_den_bos.hpp>
```

test/test_van_den_bos.cpp
```diff
@@ -0,0 +1,100 @@
...
+    std::vector<double> w1 = {1.0/3.0, 4.0/3.0, 1.0/3.0};
+
+    std::vector<point> points;
+    std::vector<double> weights;
+    for (std::size_t i = 0; i < 3; ++i)
+        for (std::size_t j = 0; j < 3; ++j)
+        {
+            points.push_back({x[i], x[j]});
+            weights.push_back(w1[i] * w1[j]);
+        }
```
mborland:
```suggestion
    for (std::size_t i = 0; i < 3; ++i)
    {
        for (std::size_t j = 0; j < 3; ++j)
        {
            points.push_back({x[i], x[j]});
            weights.push_back(w1[i] * w1[j]);
        }
    }
```

Nesting without braces is evil

test/test_van_den_bos.cpp
```diff
@@ -0,0 +1,100 @@
...
+        6,   // target size
+        0);  // no protected nodes
+
+    std::cout << "reduced size = " << reduced.size() << "\n";
+    std::cout << std::setprecision(17)
+              << "moment residual = "
+              << van_den_bos_moment_residual(reduced, basis, moments)
+              << "\n";
+
+    assert(reduced.size() <= 6);
```
mborland: Math unit test instead of asserts if you actually plan to merge this

include/boost/math/quadrature/van_den_bos.hpp
```diff
@@ -0,0 +1,510 @@
...
+            throw std::invalid_argument(
+                "positive_cubature_rule: points and weights must have the same size.");
+    }
+
+    std::size_t size() const noexcept { return points_.size(); }
+
+    std::vector<Point> const& points() const noexcept { return points_; }
+    std::vector<Real> const& weights() const noexcept { return weights_; }
+
+    Point const& point(std::size_t i) const { return points_.at(i); }
```
mborland: `.at()` will hinder performance if you're using this for comparison to our new method.

include/boost/math/quadrature/van_den_bos.hpp
```diff
@@ -0,0 +1,510 @@
...
+                p = i;
+            }
+        }
+
+        if (best <= tol)
+            continue;
+
+        if (p != row)
+            for (std::size_t j = 0; j < n; ++j)
+                std::swap(A(row, j), A(p, j));
```
mborland:
```suggestion
        if (p != row)
        {
            for (std::size_t j = 0; j < n; ++j)
                std::swap(A(row, j), A(p, j));
        }
```

JacobHass8:
This looks very interesting! Do you know how runtime scales with dimension and number of points used?

mborland:
> This looks very interesting! Do you know how runtime scales with dimension and number of points used?

I believe empirical evaluation is the next step. Nick and I had a competing idea, but literature review found this method.

JacobHass8:
> > This looks very interesting! Do you know how runtime scales with dimension and number of points used?
> 
> I believe empirical evaluation is the next step. Nick and I had a competing idea, but literature review found this method.

I've been working on implementing sparse grids using the previously implemented Gauss-Legendre quadrature. Would that still be a worthwhile contribution or would this algorithm supersede?

NAThompson:
> I've been working on implementing sparse grids using the previously implemented Gauss-Legendre quadrature. Would that still be a worthwhile contribution or would this algorithm supersede?

@JacobHass8 : I think sparse grids would still be very valuable! We have many 1D quadrature methods, naively it seems like 2D is much more difficult and could be served by many 2D quadrature methods.

NAThompson:
@JacobHass8 , @mborland , @jzmaddock : The code is in a pretty good place I think. I had originally thought that maybe we could generate more nodes/weights at runtime if we hadn't achieved convergence, but it's a big dense "find a nullspace" problem that really cannot be done online.

So this method is great for (say) assembling a stiffness matrix in a finite element calculation, but can't be used for (say) David H Bailey-style high precision quadrature.

mborland:
I've pushed a bunch of very minor changes. I also had fable run through the paper and the implementation here. This is the output:

image

NAThompson:
@mborland : Build failure is unrelated and documented in #1441. Merging.

PR #1442, opened 2026-09-01
Fix JSO CI breakage

NAThompson:


NAThompson:
@mborland , @jzmaddock : Some much needed CI cleanup here. Low risk so will probably merge once things go green, or at least get . . . greener.

PR #1448, opened 2026-09-01
Fix underflow in BesselK

### LINKED ISSUE #1229: Spurious underflow in cyl_bessel_k
`cyl_bessel_k(100, 746)`  returns zero when evaluating at double precision (with no extended long double type), but the result should be 9.49147e-66.

Underflow occurs in k0 and k1 prior to iteration.  The problem is also present if v is perturbed to be a non-integer in which case CF2_ik underflows.

Reduced test case from https://github.com/boostorg/math/pull/1228.

NAThompson:
Closes #1229

### DIFF
--- a/test/test_bessel_k.cpp
+++ b/test/test_bessel_k.cpp
@@ -134,6 +134,30 @@ BOOST_AUTO_TEST_CASE( test_main )
    test_bessel(0.1F, "float");
 #endif
    test_bessel(0.1, "double");
+
+   {
+      // https://github.com/boostorg/math/issues/1229
+      // Force evaluation at double precision so extended long double
+      // evaluation does not mask underflow in the starting K values.
+      typedef boost::math::policies::policy<
+         boost::math::policies::promote_double<false> > no_promote_policy;
+      const no_promote_policy pol;
+      const double tol = 256 * std::numeric_limits<double>::epsilon();
+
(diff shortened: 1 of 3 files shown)


test/test_bessel_k.cpp
```diff
@@ -134,6 +134,26 @@ BOOST_AUTO_TEST_CASE( test_main )
...
+         boost::math::policies::promote_double<false> > no_promote_policy;
+      const no_promote_policy pol;
+      const double tol = 256 * std::numeric_limits<double>::epsilon();
+
+      BOOST_CHECK_CLOSE_FRACTION(boost::math::cyl_bessel_k(1000, 747.0, pol),
+         9.4914699277133192873957343540027064981e-66, tol);
+      BOOST_CHECK_CLOSE_FRACTION(boost::math::cyl_bessel_k(1000.25, 747.0, pol),
+         1.2500811621640357229654772264173947351e-65, tol);
+
+      if (std::numeric_limits<double>::has_denorm == std::denorm_present)
```
mborland, line 155: I think you can #if this out for the SYCL test. It's likely reporting `has_denorm` for the host and then flushing to 0 on the device.

jzmaddock:
Thanks for this, does it fix my revised test case in #1229?  Looks like it should?

NAThompson:
@jzmaddock : Added your test case-cycling through CI now.

Take a look at your convenience.

PR #1449, opened 2026-09-01
Fix skew normal by exp-sinh quadrature

### LINKED ISSUE #1190: Skewnorm CDF inaccurate in the left tail
SciPy currently falls back to quadrature integration of the PDF in the left tail of the skewnorm CDF for positive $a$ as the boost function has issues for low values. There seem to be catastrophic cancellations resulting in CDF values of 0. See the python example below. This is probably not a super serious issue but still wanted to make you aware in case you would like to fix it on your end.

```python

import numpy as np
from scipy import special
import matplotlib.pyplot as plt

a = 1
x = np.linspace(-50, 0, 100000)

plt.semilogy(x, special._ufuncs._skewnorm_cdf(x, 0.0, 1.0, a))
plt.show()

``` 

Which results in the following plot:

image

Zooming in shows 

image

For reference what the parameters of `_skewnorm_cdf` mean here the wrapper code:

```c++

template<typename Real>
Real
skewnorm_cdf_wrap(const Real x, const Real l, const Real sc, const Real sh)
{
    if (std::isfinite(x)) {
        return boost::math::cdf(
            boost::math::skew_normal_distribution<Real, StatsPolicy>(l, sc, sh), x);
    }
    // -inf => 0, inf => 1
    return 1 - std::signbit(x);
}

```

NAThompson:
Closes #1190

### DIFF
--- a/include/boost/math/distributions/skew_normal.hpp
+++ b/include/boost/math/distributions/skew_normal.hpp
@@ -17,6 +17,7 @@
 #include <boost/math/special_functions/owens_t.hpp> // Owen's T function
 #include <boost/math/distributions/complement.hpp>
 #include <boost/math/distributions/normal.hpp>
+#include <boost/math/quadrature/exp_sinh.hpp>
 #include <boost/math/distributions/detail/common_error_handling.hpp>
 #include <boost/math/constants/constants.hpp>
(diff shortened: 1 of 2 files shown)


mborland:
image

Issue #1450, opened 2026-09-01
minimax Jamfile missing GMP/MPFR/quadmath link dependencies


ecorm:
I had trouble building the `libs/math/tools/minimax` example. AI helped me find a workaround: add missing linker flags for libquadmath, GMP, and MPFR in the Jamfile.

I don't understand Jam-ese, so here is an AI-generated bug report. I hope it makes sense and is useful.

---

## Bug

`libs/math/tools/minimax` fails to link with undefined references to `ldexpq`, `mpfr_prec_round`, etc., because its Jamfile never declares these dependencies — unlike `libs/math/test/Jamfile.v2` and `libs/math/example/Jamfile.v2`, which both properly detect and link GMP/MPFR (via `modules.peek : GMP_PATH/MPFR_PATH` and `../../config/checks/config`) and `quadmath` (via a check-target-builds against `../config//has_float128`).

`tools/minimax/Jamfile.v2` has no such handling: the `gmp`/`mpfr` lib targets are commented out entirely, and there's no `has_float128`/quadmath check at all, despite `f.cpp/main.cpp` transitively depending on `boost::multiprecision::mpfr_float` and `boost::math::cstdfloat` (`__float128`) through headers.

## Steps to reproduce

```
cd boost_1_92_0/libs/math/tools/minimax
../../../../b2
```

## Result

Sample linker failures:

- Undefined reference to `ldexpq` (from `boost/math/cstdfloat/cstdfloat_cmath.hpp`), resolved by adding `-lquadmath` to the link line.
- Undefined reference to `mpfr_prec_round` (from `boost/multiprecision/mpfr.hpp`), resolved by adding `-lmpfr -lgmp` (order matters: `mpfr` before `gmp`).

## Environment

Boost 1.92.0
Linux Mint (Ubuntu-based), GCC 13
libgmp-dev, libmpfr-dev installed via apt
libquadmath0 present (ships with GCC)

## Workaround

Add linker flags to the `exe minimax` requirements in the Jamfile manually:
```
exe minimax : f.cpp main.cpp
    : <linkflags>"-lquadmath -lmpfr -lgmp"
    ;
```

## Suggested fix

Bring `tools/minimax/Jamfile.v2` in line with `libs/math/example/Jamfile.v2` — reuse its GMP/MPFR path-detection pattern and its `check-target-builds ../config//has_float128 ... : <linkflags>-lquadmath` idiom, rather than hardcoding flags.

NAThompson:
[...]

PR #1451, opened 2026-09-02
Cleanup Jamfile for minimax

NAThompson:


NAThompson:
Closes #1450

ecorm:
When I try:
```
../../../../b2 -a
```
within the `libs/math/tools/minimax` directory, I get:
```
error: Unable to find file or target named
error:     '../config//has_mpfr'
error: referred to from project at
error:     '.'
error: could not resolve project reference '../config'
```

NAThompson:
@ecorm : apologies-try once more!

ecorm:
@NAThompson , it compiles for me now with your last change. Thanks!

```
./minimax
Running 1 test case...
info
Nothing to display
Variant              = 0
range                = [0,1]
Relative Error       = 1
Pin to Origin        = 0
Order (Num/Denom)    = 3/1
Target Precision     = 64
Working Precision    = 128
Skew                 = 0
Brake                = 50
X Offset             = 0
X scale              = 1
Y Offset             = 0
```

ecorm:
@NAThompson , the `Running 1 test case...` output is a bit worrisome. The concern is that the Boost.Math test suite might try to run `minimax` and then get stuck because it's an interactive program. I don't know how the test suite is put together, sorry.

PR #1452, opened 2026-09-02
Polynomial roots

NAThompson:


NAThompson:
https://arxiv.org/abs/2302.06244

amlel-el-mahrouss (commented):
I have some questions and remarks, so, per `void check_residuals` in test/polynomial_roots_test.cpp:
- This could be `const auto&` for  `for(auto root : roots)`.

per `void fft` in polynomial_roots.hpp:
- This could benefit of some comments, and more testing as well due to the nested loops.

jeremy-murphy:
Just wondering naming, since there are so many algorithms for finding polynomial roots; you wouldn't call it `imbach_moroz_polynomial_roots`?

NAThompson:
@jeremy-murphy : I believe this algorithm finishes polynomial rootfinding as it is nlogn and stable.

jeremy-murphy:
> @jeremy-murphy : I believe this algorithm finishes polynomial rootfinding as it is nlogn and stable.

Cool, that's pretty exciting! Do you have any results on how accurate it is compared to the existing algorithms we have or against some optimal answer?

NAThompson:
> Cool, that's pretty exciting! Do you have any results on how accurate it is compared to the existing algorithms we have or against some optimal answer?

Not yet-I think Aurentz's 2015 work (which is O(n^2)) is probably the reference for stability. I do not believe this algorithm is stable in quite the same sense.

test/polynomial_roots_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+   check_residuals({1, 0, 1}, quadratic);
+   assert(quadratic.size() == 2);
+
+   std::vector<double> with_zero_roots{0, 0, -2, 1};
+   auto zero_roots = polynomial_roots(with_zero_roots);
+   check_residuals(with_zero_roots, zero_roots);
+   assert(std::count(zero_roots.begin(), zero_roots.end(), std::complex<double>(0, 0)) == 2);
+
+   std::mt19937_64 generator(0x5eed);
+   std::uniform_real_distribution<double> distribution(-2, 2);
```
JacobHass8, line 73: Have you tested how accurate this method is with multiprecision types like `boost::multiprecision::cpp_bin_float`?

include/boost/math/tools/polynomial_roots.hpp
```diff
@@ -0,0 +1,574 @@
...
+#include <vector>
+
+namespace boost { namespace math { namespace tools {
+namespace polynomial_roots_detail {
+
+template <class Real>
+inline Real log2_abs(Real x)
+{
+   using std::abs;
+   using std::log2;
```
JacobHass8, line 29: Would adding BOOST_MATH_STD_USING here and below allow multiprecision types?

test/polynomial_roots_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+   // Newton polygon has a single edge.
+   std::vector<double> roots_of_unity(257);
+   roots_of_unity.front() = -1;
+   roots_of_unity.back() = 1;
+   check_residuals(roots_of_unity, polynomial_roots(roots_of_unity), 2e-8);
+
+   bool threw = false;
+   try { polynomial_roots(std::vector<double>{0, 0}); }
+   catch(std::domain_error const&) { threw = true; }
+   assert(threw);
```
JacobHass8, line 93: Should all tests here use boost test functions like BOOST_CHECK_THROW? Additionally, is there any benefit in adding policies to `polynomial_roots`?

PR #1461, opened 2026-09-07
Gauss-Hermite Quadrature

JacobHass8:
Before continuing with this PR I wanted to get some feedback on the approach I took. For Gauss-Hermite quadrature, I need the roots of the Hermite polynomials. For small N, I found it's common to use the [Golub-Welsch algorithm](https://en.wikipedia.org/wiki/Gaussian_quadrature#The_Golub-Welsch_algorithm) which finds the roots via the eigenvalues of a matrix. To find the eigenvalues, I used the `Eigen` library guarded behind a `__has_include`. I figured this was okay since I generated and saved the same abscissa/weights as was in gauss.hpp. @NAThompson do you have any thoughts? 

If this is okay, I'll press on with writing some tests.

### DIFF
--- a/include/boost/math/quadrature/hermite.hpp
+++ b/include/boost/math/quadrature/hermite.hpp
@@ -23,6 +23,7 @@
    #include <boost/multiprecision/cpp_bin_float.hpp>
    #include <boost/math/constants/constants.hpp>
    #include <boost/math/special_functions/hermite.hpp>
+   #include <boost/math/tools/roots.hpp>
 
    #define EIGEN_SUPPORT
 #endif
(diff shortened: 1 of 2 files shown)


NAThompson:
@JacobHass8 : Yeah, we've done guarded Eigen includes in this library many times. I think it's acceptable for sure.

JacobHass8:
I added root finding after the Golub-Welsch algorithm runs to further refine the roots. It doesn't look like the parameters moved that much. If someone tries to use higher N it might help. I did a couple of spot checks against Mathematica and the current values match to 100 digits.

NAThompson:
@JacobHass8 : Let's push this one past the finish line next. Do you need my polynomial roots class or can you finish without?

JacobHass8:
> @JacobHass8 : Let's push this one past the finish line next. Do you need my polynomial roots class or can you finish without?

I can finish without. If we want a boost standalone in the future, then we can implement the polynomial roots class. 

I think the only thing left is to write tests and documentation. There will probably be unseen issues with multi-precision types.

NAThompson:
@JacobHass8 : Yeah-try to integrate a function taking complex value and vector values, and taking inputs as multiprecision types. That should cover a large set of issues.

Also: Is there a "Gauss-Hermite-Kronrod" extension of Gauss-Hermite?

NAThompson:
https://arxiv.org/html/1203.1554v2

Looks like the adaptive extension of Gauss-Hermite is called "Genz-Keister".

JacobHass8:
> @JacobHass8 : Yeah-try to integrate a function taking complex value and vector values, and taking inputs as multiprecision types. That should cover a large set of issues.
> 
> Also: Is there a "Gauss-Hermite-Kronrod" extension of Gauss-Hermite? Generally we _vastly_ prefer quadrature schemes that can provide an estimate of their own error.

Yeah, I'm not sure how we would estimate the error. There is a way to estimate but it requires a very high derivative which wouldn't make sense to compute https://math.stackexchange.com/questions/298206/hermite-gauss-quadrates-error-bound. 

I had left out the Kronrod extension just for simplicity. It shouldn't be difficult to add back in.

NAThompson:
> What do you mean vector values? As in vectors in the output? I don't know if that's possible because you would need to be able to add vectors element-wise.

We can now do quadrature of matrices-so we can also do vectors.

> Yeah, I'm not sure how we would estimate the error. There is a way to estimate but it requires a very high derivative which wouldn't make sense to compute https://math.stackexchange.com/questions/298206/hermite-gauss-quadrates-error-bound.

Just use two successive estimates of the integral and take the absolute value of their difference.

PR #1455, opened 2026-09-05
Fix Daubechies derivative scaling and complete coverage

NAThompson:
Exercise explicit refinements, absolute-error mode, interior derivatives, and the Haar wavelet. Remove obsolete interpolators and migrate the remaining example to matched_holder_aos.

Correct missing grid-spacing factors in cardinal quintic and septic derivatives, including septic node evaluation. Add cubic-polynomial regressions for both data layouts and non-unit grid spacing.

Fixes boostorg/math#1101

PR #1462, opened 2026-09-11
Guard cardinal quintic Hermite against out-of-bounds read at the right endpoint

vishwa2710:
Fixes #<issue number, to be filed>

## Problem

`cardinal_quintic_hermite` and `cardinal_quintic_hermite_aos` read one element past the end of `y_`, `dy_` and `d2y_` for some `x` that are strictly inside `domain()`. `operator()` guards `x <= xf` with `xf = x0_ + (n-1)/inv_dx_` and returns early only for `x == xf`, but the interval lookup computes `s = (x - x0_)*inv_dx_` independently, and for `x` a few ulps below `xf` this can round to `n-1 + O(1e-14)`. Then `floor(s) == n-1`, `t != 0`, and `y_[i+1]` is out of bounds. It only occurs when `dx` is not a power of two, which is why the existing tests (`dx = 1` and `1/8`) never see it. In a random sweep over `x0 in (-5,5)`, `dx in (1e-3,3)`, `n in [2,200]`, about 0.26% of grids trigger it within 4 ulps of the right endpoint, for `float`, `double` and `long double` alike.

The basis-function formulas are correct: both cardinal layouts agree with the non-cardinal `quintic_hermite` and with scipy's `BPoly.from_derivatives` to ~4e-15 (value), ~1.5e-13 (first derivative) and ~1.4e-11 (second derivative) on a 12-node grid with `dx = 0.1`, and reproduce `x^5` exactly.

Minimal reproducer (aborts with `-D_GLIBCXX_ASSERTIONS`, or reports heap-buffer-overflow at `quintic_hermite_detail.hpp:283` under `-fsanitize=address`):

```cpp
#include <boost/math/interpolators/quintic_hermite.hpp>
#include <cmath>
#include <vector>
int main()
{
    const double x0 = -2.6660221376669786, dx = 2.6599332164560212;
    const std::size_t n = 26;
    std::vector<double> y(n), dy(n), d2y(n);
    for (std::size_t i = 0; i < n; ++i) { double x = x0 + i*dx; y[i] = std::sin(x); dy[i] = std::cos(x); d2y[i] = -std::sin(x); }
    auto qh = boost::math::interpolators::cardinal_quintic_hermite(std::move(y), std::move(dy), std::move(d2y), x0, dx);
    double x = std::nextafter(qh.domain().second, -INFINITY);   // x < xf, but (x - x0)*(1/dx) == 25.000000000000004
    return qh(x) > 2;                                             // reads y_[26]
}
```

## Fix

In all six `unchecked_*` evaluators (three per data layout), after the existing `t == 0` fast path, clamp to the final interval when `i + 1 >= size()`: `i = size() - 2`, `t = s - i`. `t` is then within a few ulps of 1, so the result is continuous with the endpoint value.

## Tests

* `test_cardinal_endpoint_rounding_case`: the fuzz-found grid above, evaluated for both layouts on a `checked_vector` whose `operator[]` calls `at()`, so the out-of-range read throws `std::out_of_range` deterministically without a sanitizer. With the old header this reports 25 failures across the new tests; with the fix, none.
* `test_cardinal_endpoint_rounding<Real>`: 2000 random grids per type for `float`, `double`, `long double`, walking 8 ulps below the right endpoint, checking values against linear data.
* Existing `quintic_hermite_test` passes unchanged under `-fsanitize=address,undefined`; the header still builds as C++14 with `-Wall -Wextra -Wpedantic`.

## Note

`cardinal_cubic_hermite_detail*` and `cardinal_septic_hermite_detail*` use the same interval lookup and are very likely affected in the same way; this PR is deliberately limited to the quintic interpolator. `daubechies_scaling`/`daubechies_wavelet` construct the AoS variant with `x0 = 0`, `dx = 2^-r`, for which the arithmetic is exact, so they are not affected.

### DIFF
--- a/include/boost/math/interpolators/detail/quintic_hermite_detail.hpp
+++ b/include/boost/math/interpolators/detail/quintic_hermite_detail.hpp
@@ -279,6 +279,14 @@ public:
         {
             return y_[i];
         }
+        // Rounding in (x-x0_)*inv_dx_ can place floor(s) on the final node even though
+        // x < xf, in which case index i+1 is one past the end. Evaluate on the final
+        // interval instead; t is then within a few ulps of 1.
+        if (i + 1 >= y_.size())
+        {
+            i = y_.size() - 2;
+            t = s - static_cast<Real>(i);
+        }
         Real y0 = y_[i];
         Real y1 = y_[i+1];
         Real dy0 = dy_[i];
@@ -326,6 +334,14 @@ public:
         {
             return dy_[i]*inv_dx_;
         }
+        // Rounding in (x-x0_)*inv_dx_ can place floor(s) on the final node even though
+        // x < xf, in which case index i+1 is one past the end. Evaluate on the final
+        // interval instead; t is then within a few ulps of 1.
+        if (i + 1 >= y_.size())
+        {
+            i = y_.size() - 2;
+            t = s - static_cast<Real>(i);
+        }
         Real y0 = y_[i];
         Real y1 = y_[i+1];
         Real dy0 = dy_[i];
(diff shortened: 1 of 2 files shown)


mborland:
Have you looked at and validated any of this yourself?

vishwa2710:
Hi @mborland, I did not mean to push this MR.
I was debugging an issue with the double_prime derivative computation for the cardinal quintic hermite (I saw that the numbers weren't adding up with my tests) and an agent went down another rabbit hole and opened this MR. Will close this MR.

I see that https://github.com/boostorg/math/pull/1455 fixes the issue I was looking through.

Apologies for wasting your time.

PR #1467, opened 2026-09-17
Fix noncentral t degrees-of-freedom domain checking

NAThompson:


NAThompson:
@WarrenWeckesser : This can at best be regarded as a partial fix, but . . .

PR #1468, opened 2026-09-17
Improve precision and speed of the Kolmogorov-Smirnov distribution

evanmiller:
Measured against a 50-digit reference (`cpp_bin_float_50` through the same header), the Kolmogorov-Smirnov distribution had three problems. This PR fixes them without touching the Jacobi theta functions, which the CDF still uses.

### Quantile

The Newton-Raphson bracket was hard-coded to `[0, 1]` in `x`, so for small `n`, where the upper-tail quantile exceeds unity, it either threw (`quantile(dist(1), 0.9)`) or was clamped: for `n = 10` the complement quantile of any `q <= 1e-10` returned exactly 1. Mid-range it needed about 9 iterations.

The iteration now runs on `ln cdf` in `W = pi^2/(8 x^2 n)` for `p < 0.6`, and on `ln(1 - cdf)` in `V = x^2 n` above, where the residual is very nearly linear. The starting value inverts the first two terms of the series in each variable; that is already exact to working precision over most of the range, so a single evaluation usually confirms it. `p = 0` returns 0 and `p = 1` raises an overflow error, as in the other distributions.

### Mode, median, skewness, kurtosis

These are constants of the standardized distribution times a power of `n`. The mode was found at run-time by Brent's minimizer, which cannot locate a flat maximum better than `sqrt(epsilon)`; skewness and kurtosis were computed from closed forms that cancel away 40-260 ulp. They are now tabulated to 100 decimal digits (verified two independent ways), with run-time fallbacks for types of higher precision. A `median()` overload is added.

### PDF

Each series is evaluated from a single exponential, with the remaining terms obtained by multiplication, and the rounding of `x*x*n`, of `pi^2/8` and of the division is tracked with Dekker products so it is not amplified by the size of the exponent. Just above the smallest normal number the leading term is evaluated as a square of `exp(-W/2)`, which fixes a loss of three decimal digits to an intermediate underflow (visible as the clipped points in the old ULP graph).

### Results, `n = 10`, double

| | before | after |
|---|---|---|
| quantile worst error | ~1e6 ulp (complement: ~1e13) | 3.1 ulp, mean 0.6 |
| pdf mean / worst error | 11 / 694 ulp | 1.1 / 5.8 ulp |
| mode / skewness / kurtosis excess | 1e7 / 39 / 261 ulp | 0.6 / 0.3 / 0.4 ulp |
| quantile time | 375 ns | ~190 ns |
| mode time | 549 ns | 3 ns |

A random stress test of 640,000 quantiles (`n` from 1 to 1e6, `p` log-uniform over the full range, float and double) had no failures.

### Tests and docs

`test_kolmogorov_smirnov.cpp` now covers small `n`, both tails down to the smallest normal number (by bracketing, which stays meaningful where one ulp of `x` is many ulps of `p`), reference values computed at 120 digits, the moment constants against their closed forms, and the high-precision fallbacks with a 110-digit type. The implementation table in the docs is updated and the PDF ULP graph regenerated; its vertical axis is now ±5 ulp.

Not addressed: the complement CDF keeps its 10-120 ulp error in the far upper tail, because the rounding of `tau` is amplified inside the theta function. That is only fixable inside theta, or by evaluating the `z = 0` series here directly.


NAThompson:
@evanmiller , @claude : Chat review: https://chatgpt.com/share/6aacafea-68ec-83e8-848b-827d57fb13bc

NAThompson:
@evanmiller , @claude : I have merged a "CI greenup" MR-which though not yet _fully_ green, should give a better baseline.

Could you rebase and kick off a new CI job then we'll examine the remaining failures?

evanmiller:
Rebased onto the CI green-up and pushed. The old run also had one failure that was this PR's own: the two quantile functors took `log(p)` in a member initializer, where `BOOST_MATH_STD_USING` cannot reach, so the `std_real_concept_check` compile tests failed with "no matching function for call to 'log'". Fixed in 81bec9990 by assigning in the constructor body; all four `std_real_concept_check` variants compile locally now and `test_kolmogorov_smirnov` passes. Let's see what remains after the new run.

NAThompson:
@evanmiller , @claude : Do you have a before/after on the ulp plot? It seems like it only measures the "after"-which is fine for documentation, but even so some of the evals are falling outside the εcond(f(x)) threshold

evanmiller:
ks_pdf
ks_ccdf
ks_quantile

NAThompson:
@evanmiller , @claude : Please rebase to pick up the CI maintenance work.

evanmiller:
I think the CI issues are all fixed now

include/boost/math/distributions/kolmogorov_smirnov.hpp
```diff
@@ -111,77 +122,216 @@
...
+
+template <class RealType>
+inline const RealType& kolmogorov_smirnov_skewness_constant() {
+    static const RealType value = BOOST_MATH_BIG_CONSTANT(RealType, 1000, 0.8604261371436682558667183685173007452393495203060929924698665373211030237490355834020123201092217303);
+    return value;
+}
+
+template <class RealType>
+inline const RealType& kolmogorov_smirnov_kurtosis_excess_constant() {
+    static const RealType value = BOOST_MATH_BIG_CONSTANT(RealType, 1000, 0.8816189679105236704015538304328295408355055629289034536289160037385510119639244078254098968815665536);
```
NAThompson: @evanmiller , @claude : Can these constants not be recovered to arbitrary precision?

NAThompson:
@evanmiller , @claude : Barring the last nitpick I had are you ready for merge?

include/boost/math/distributions/kolmogorov_smirnov.hpp
evanmiller: Claude says

> The constants can be recovered to arbitrary precision, and the PR already does it. `kolmogorov_smirnov_use_constants` only uses the table when policies::digits<RealType, Policy>() <= 330

evanmiller:
> @evanmiller , @claude : Barring the last nitpick I had are you ready for merge?

Yes I am OK to merge any time

NAThompson:
Done thanks!

PR #1469, opened 2026-09-17
Form the Jacobi theta exponents without amplified rounding

evanmiller:
Follow-up to #1468, independent of it.

Every term of the theta series is `exp(-E)`, where `E` is a product or quotient of `tau`, `pi` and a small integer, or (for `tau < 1`, after the modular transformation) the square of `z` plus a multiple of `pi/2`, divided by `pi*tau`. Rounding `E` to working precision costs `|E|` ulps in the term, and the exponents run into the hundreds before the terms underflow. So the functions lost tens to hundreds of ulps wherever the exponents were large: near `tau = 1` in the direct series, and for small `tau` throughout, where the argument reduction by a rounded `pi` added an absolute error that the Gaussians then amplified by `2z/(pi*tau)`.

### What changed

- A small object (`_jacobi_theta_exponents`) holds `pi` and `a = pi*tau` (or `-ln q`) as unevaluated sums `hi + lo` and forms every exponent while tracking the rounding of each product, sum and quotient with Dekker's and Knuth's error-free transformations. These work for any binary floating-point type; where `<cmath>` advertises a fast fma (`FP_FAST_FMA` etc.) it is used instead. The error is applied to each term as a factor `(1 - dE)`.
- The argument reduction for `tau < 1` uses the same split `pi` and passes the remainder's error into the Gaussian terms.
- The `z = 0` shortcuts form the nome of `1/tau` without rounding the reciprocal.
- The `q` parameterization builds `a = -ln q` directly instead of going through `tau = -ln(q)/pi` and back, which halves its remaining error.

### Results

Against a 50-digit reference (double, `z` in [0, 5], `tau` in [0.005, 31]):

| | before (worst / mean ulp) | after |
|---|---|---|
| `tau` parameterization, `tau < 1` | 177–360 / 6–12 | 3–6 / 0.5 |
| `tau` parameterization, `tau >= 1` | 20–76 / 2–8 | 2.5 / 0.5 |
| `q` parameterization, `q > exp(-pi)` | 380–790 / 12–14 | 64–81 / 3 |

What remains in the `q` parameterization is the rounding of `ln q` itself, which is inherent to it. The "minus one" functions below `tau = 1` still cancel against 1, as documented.

For the Kolmogorov-Smirnov complement CDF, which calls `jacobi_theta4m1tau` with exponents up to 2x²n, theta's own contribution to the error drops from 70 ulp worst (10 mean) to 1.9 (0.45).

### Cost

The Gaussian-sum paths (`tau < 1`, `z != 0`) take about twice as long as before; the direct series about a quarter more (a few nanoseconds of fixed overhead per call).

### Tests and docs

The periodicity test shifts `z` by the rounded `two_pi`; with an exact reduction that is a genuinely different argument, and for large `q` the functions are steep enough that it changes them by more than the evaluation error, so the test's tolerance now accounts for it (`100 eps + 4 pi² eps / -ln q`). The Mellin transform integrals, which sum thousands of theta values, get a couple more ulps of slack. `test_jacobi_theta` and `test_kolmogorov_smirnov` pass.

The docs' accuracy notes are updated and the eight ULP graphs regenerated; the `q` sweep of the plot generator started at `q = 0`, which has been a domain error since the previous round of fixes, so it now starts at 1e-6. (The generator also traps when built at `-O2` with this clang, with the old header as well; I generated the graphs at `-O0`.)


NAThompson:
@evanmiller , @claude : Could you drop the  ULPs plot you created here?

Also, please check the committed SVG size-when I wrote the `ulps_plot` function I made it zero dependency but at the cost of those SVGs often being huge.

Also, as to cost there are google benchmarks-can we add one and then put it in `reporting/performance/jacobi_theta_performance.cpp` and report the before/after results here?

evanmiller:
Here are the benchmark numbers:

image

The committed SVG files are slightly smaller than the ones they replace but I can reduce the point count if you'd like to reduce further. Here are overlaid versions:

theta1tau_z_tau0 05
theta4_5_q
theta4tau_0_logtau
theta4tau_z_tau0 3

evanmiller:
And one more

theta3tau_z_tau0 05

NAThompson:
@evanmiller , @claude : Great improvement! Getting robustly below the condition number of function evaluation without resorting to higher intermediate precision is difficult!

I believe the cost increase is a perfectly acceptable price to pay for this accuracy improvement.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -228,12 +229,231 @@ struct _jacobi_theta_q_power {
...
+// the standard library advertises a fast fused multiply-add it is used
+// instead, since fma(a, b, -ab) is exactly this quantity.
+template <class RealType>
+inline RealType _jacobi_theta_product_error(RealType a, RealType b, RealType ab, RealType splitter) {
+    RealType ah, al, bh, bl;
+    _jacobi_theta_split(a, splitter, ah, al);
+    _jacobi_theta_split(b, splitter, bh, bl);
+    return (((ah * bh - ab) + ah * bl) + al * bh) + al * bl;
+}
+#ifdef FP_FAST_FMAF
```
NAThompson: Is this necessary? I feel like basically all modern hardware has FMAs now, and if you're unfortunate enough to not have it then degrading to whatever `std::fma` backends to is correct.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -228,12 +229,231 @@ struct _jacobi_theta_q_power {
...
+// precision costs |E| ulps in the exponential, so instead the rounding
+// errors of pi, of the products and of the division are tracked exactly with
+// Dekker's product and Knuth's sum (which work for any binary floating-point
+// type) and applied as a correction factor exp(-dE) = 1 - dE. What remains
+// is the rounding of tau itself, which is the caller's.
+// (Cached per type: for a type whose precision can change at run time the
+// cached value may split at the wrong place, in which case the corrections
+// below are merely inexact, i.e. no worse than not applying them.)
+template <class RealType>
+inline RealType _jacobi_theta_splitter() {
```
NAThompson: Should these be in anonymous namespaces?

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Sounds good, removed the ifdefs

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: I moved the underscore functions in this module into the `detail` namespace

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -220,20 +223,233 @@ _jacobi_theta_check_q(RealType q, const Policy& pol, const char* function, RealT
...
+// (Cached per type: for a type whose precision can change at run time the
+// cached value may split at the wrong place, in which case the corrections
+// below are merely inexact, i.e. no worse than not applying them.)
 template <class RealType>
-struct _jacobi_theta_tau_power {
-    RealType tau;
-    RealType operator()(RealType exponent) const {
+inline RealType jacobi_theta_splitter() {
+    BOOST_MATH_STD_USING
+    static const RealType splitter = ldexp(RealType(1), (tools::digits<RealType>() + 1) / 2) + 1;
```
NAThompson, line 249: I do wonder if there is good reason to make this static instead of making it `constexpr`?

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -322,420 +538,419 @@ _jacobi_theta4m1_series(RealType z, const NomePower& q_pow, const Policy&) {
...
         if (n%2)
             pair = -pair;
 
         result += pair;
         n++;
-    } while (!_jacobi_theta_converged(last_g, g, eps));
+    } while (!jacobi_theta_converged(last_g, g, eps));
 
-    return result * sqrt(tau);
+    return result / sqrt(x.tau);
```
NAThompson, line 679: rsqrt?

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Claude says ldexp is not constexpr before C++23

NAThompson:
@evanmiller , @claude : I believe chat found a bug in the implementation; please see [here](https://chatgpt.com/share/6aac95a2-23f4-83e8-a008-528696badaef)

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: I am told

> For the built-in types boost::math::rsqrt is 1/sqrt(x), so result * rsqrt(tau) does the same division and adds one more rounding. rsqrt.hpp also hard-errors below C++17.

evanmiller:
Added a commit addressing the blocking bug found by chat affecting large z inputs.

include/boost/math/special_functions/jacobi_theta.hpp
```diff
@@ -220,20 +224,290 @@ _jacobi_theta_check_q(RealType q, const Policy& pol, const char* function, RealT
...
+}
+inline float jacobi_theta_product_error(float a, float b, float ab, float splitter) {
+    return jacobi_theta_fma_is_fused<float>() ? std::fma(a, b, -ab) : jacobi_theta_product_error<float>(a, b, ab, splitter);
+}
+inline double jacobi_theta_product_error(double a, double b, double ab, double splitter) {
+    return jacobi_theta_fma_is_fused<double>() ? std::fma(a, b, -ab) : jacobi_theta_product_error<double>(a, b, ab, splitter);
+}
+inline long double jacobi_theta_product_error(long double a, long double b, long double ab, long double splitter) {
+    return jacobi_theta_fma_is_fused<long double>() ? std::fma(a, b, -ab) : jacobi_theta_product_error<long double>(a, b, ab, splitter);
+}
```
NAThompson: @evanmiller , @claude : These three can't be merged can't be a template? I thought the multiprecision types _did_ have an `fma` that could be found via ADL.

@mborland: Is that correct?

jzmaddock: Yes, the multiprecision types have fma.  The catch is there's no simple macro to determine when it actually offers better performance/accuracy as that varies by the specific multiprecision type.

NAThompson:
@evanmiller : I think this can be punched past the finish line with just a quick check for the fma for multiprecision types.

include/boost/math/special_functions/jacobi_theta.hpp
evanmiller: Done in 3e2e6a4e5: the three overloads are now one template. `fma` is looked up with `std::fma` in scope plus ADL (falling back to the split via SFINAE for a type with no `fma` at all), and the run-time fusedness probe now runs for every type, which answers @jzmaddock's point that no macro says when the multiprecision `fma` is the real thing. What the probe reports here (Apple clang, arm64, MPFR 4):

| type | fma fused? | path |
|---|---|---|
| `float`, `double`, `long double` | yes | fma |
| `mpfr_float_50` (et_on / et_off), variable-precision `mpfr_float` | yes (`mpfr_fma`) | fma |
| `cpp_bin_float_50` (et_on / et_off) | no (default `eval_multiply_add` is multiply, then add) | split |
| `real_concept` | has no `fma` | split |

`float128` should come out fused as well (`fmaq`), but I have no libquadmath here to confirm. For every type above the fma result is bit-identical to the Dekker product error, and over 144 (z, tau) points against a 400-bit reference the four `theta*tau` functions are no worse on the fma path (worst relative error 6.1e-50 for `mpfr_float_50` vs 1.2e-49 for a 168-bit `cpp_bin_float` on the split path). Also rebased onto the CI green-up.

NAThompson:
Build failure is:

```
++ git clone -b develop --depth 1 https://github.com/boostorg/boost.git boost-root

You have not agreed to the Xcode license agreements, please run 'sudo xcodebuild -license' from within a Terminal window to review and agree to the Xcode license agreements.
```

Merging.

PR #1474, opened 2026-09-20
Fix second variance accumulator in parallel correlation

### LINKED ISSUE #1473: Parallel correlation uses only the first chunk variance for its second input
The parallel path of `correlation_coefficient` returns `1` for the example below, while the sequential path returns about `0.70710678118654`. The exact correlation is `1/sqrt(2)`.

```cpp
#include <boost/math/statistics/bivariate_statistics.hpp>
#include <execution>
#include <iostream>
#include <vector>

int main() {
    std::vector<double> u(200000), v(u.size());
    for (std::size_t i = 0; i < u.size(); ++i) {
        u[i] = i % 4;
        v[i] = u[i] + (i / 4) % 4;
    }
    using boost::math::statistics::correlation_coefficient;
    std::cout << correlation_coefficient(std::execution::seq, u, v) << '\n';
    std::cout << correlation_coefficient(std::execution::par, u, v) << '\n';
}
```

Tested on Linux with GCC 11.5, C++17, and current develop (`7a5caf3`). The input is large enough to exercise the parallel path rather than its sequential fallback.

In `correlation_coefficient_parallel_impl`, the merge loop assigns the combined variance to `Qv_b` instead of `Qv_a`. That leaves the first chunk's variance in the running accumulator, so the final normalization uses the wrong value. I have a one-line fix and a regression test that checks both argument orders against the exact correlation.

Robertboy18:
The parallel merge writes the combined variance of the second input back to `Qv_b`, then discards it. `Qv_a` keeps only the first chunk's variance, so correlation can be inflated and clamped to 1.

This changes the assignment to `Qv_a`. The regression uses 65,536 samples with a known correlation of `1 / sqrt(2)` and checks both argument orders above the parallel dispatch threshold.

The new test fails in both orders before the fix. The full `bivariate_statistics_test.cpp` suite passes after it with GCC 11.5 in C++17 and C++14 modes, using Boost 1.92.0 dependencies.

Fixes #1473.


NAThompson:
@Robertboy18 : Thanks! Should be merged after CI cycles.

NAThompson:
@mborland : Take a look at your convenience.

PR #1476, opened 2026-09-20
Avoid overflow and underflow in correlation normalization

### LINKED ISSUE #1475: Correlation normalization overflows or underflows for finite variances
Found a range issue in `correlation_coefficient`: scaling two inputs changes their correlation from about `0.866` to either `0` or `1`, even though the covariance and both variances remain finite.

```cpp
#include <boost/math/statistics/bivariate_statistics.hpp>
#include <iomanip>
#include <iostream>
#include <vector>

int main() {
    using boost::math::statistics::correlation_coefficient;
    std::cout << std::setprecision(17);
    for (double s : {1.0, 1e100, 1e-100}) {
        std::vector<double> u{0, s, 2*s}, v{0, 0, s};
        std::cout << s << ": " << correlation_coefficient(u, v) << '\n';
    }
}
```

On current develop (`7a5caf3`), GCC 11.5, Linux:

```text
1:      0.86602540378443871
1e+100: 0
1e-100: 1
```

The expected correlation is `sqrt(3)/2` in every case. The final `cov/sqrt(Qu*Qv)` forms the variance product before taking its square root. That product overflows for the large inputs and underflows to zero for the small ones; the latter result is then clamped to 1.

Taking the square roots separately fixes these examples. I have a patch with regressions for `float`, `double`, and `long double`, including different scales for each input and negative correlation. This addresses the final normalization; it does not make the earlier variance accumulation safe for arbitrary finite inputs.

This is separate from the parallel accumulator typo in #1473. Can attach a small Lean/FloatLib example of the normalization issue if useful.

Robertboy18:
Fixes #1475.

`correlation_coefficient` computes `cov/sqrt(Qu*Qv)`. The variance product can overflow or underflow even when the covariance, both variances, and the final correlation are representable. For `u = {0, s, 2*s}` and `v = {0, 0, s}`, this returns 0 at `s = 1e100` and 1 at `s = 1e-100`, instead of `sqrt(3)/2`.

Take the square roots separately in the sequential and parallel normalization. The input loop is unchanged.

The regressions cover `float`, `double`, and `long double`, independent scaling of the two inputs, negative correlation, and large inputs that exercise parallel dispatch. They fail 20 checks before the fix. The full bivariate statistics suite passes afterward with GCC 11.5 in C++17 and C++14 modes, using Boost 1.92.0 dependencies.

This only addresses normalization; the variance accumulation can still overflow for larger inputs. The separate parallel accumulator bug is covered by #1474.

I also have a small Lean/FloatLib example with checked binary64 counterexamples and the real-valued normalization identity. Can attach it if useful; it doesn't prove the C++ implementation.


NAThompson:
@Robertboy18 : I am not convinced by this. First, it increases the number of operations necessary to compute the result, as well as induces additional rounding. Second, the _point_ of floating point numbers is so that we can essentially ignore underflow and overflow by having such astronomically huge `max` and `denorm` mins that we can safely assume that if `x` is _being used_ and representable, forming `x*x` is going to overflow or underflow with negligible probability.

If we were to accept the premise of this PR, the entire library would have to be gone over with a fine-tooth comb for these sorts of problems, leaving us less accurate and slower after we finished.

